PR #44, opened 2020-10-21
bridge: in-place debugging using dlv

leoluk:
* #54 bridge: log submission for cross-submissions to Solana
* #53 bridge: do not resubmit submitted VAAs during aggregation
* #52 bridge: do not log errors for duplicate VAA submissions
* #51 bridge: have all nodes submit VAAs to Solana
* #48 bridge/pkg/solana: retry VAA submission on transient errors
* #47 agent: return gRPC Internal error on submission failure
* #46 bridge: move initial guardian set fetching to pkg/ethereum/watcher.go
* #45 bridge: propagate panics from runnables
* **#44 bridge: in-place debugging using dlv**

Optional support for in-place debugging using an external
IDE debugger. Disabled by default.

The dlv binary is always compiled in the Dockerfile -
this is cheap due to it only depending on go.mod.


devnet/bridge.yaml
```diff
@@ -42,7 +42,16 @@ spec:
         - name: guardiand
           image: guardiand-image
           command:
+# Uncomment this for in-place debugging using dlv
```
hendrikhofstadt, line 45: Why is this commented out? You could always run it with delve and use `--continue` to not block and wait for a debugger to be attached (for anyone trying to debug right from the start, this could then be commented out)?

hendrikhofstadt (changes requested):

hendrikhofstadt (approved):

hendrikhofstadt (approved):

PR #51, opened 2020-10-21
bridge: have all nodes submit VAAs to Solana

### LINKED ISSUE #20: "Leader selection" for on-chain publishing
https://github.com/certusone/wormhole/blob/c370a94911c3b06237fd5e92837739ff2f98d191/bridge/cmd/guardiand/processor.go#L306-L308

We need at-least-once delivery of signed VAAs to the chain.

Complications:

- [ ] We need a fallback/retry mechanism if the tx fails to be published.
- [ ] The requirements state that non-guardian nodes may wish to aggregate signatures and act as fallback submitters, how will this work?

leoluk:
* #54 bridge: log submission for cross-submissions to Solana
* #53 bridge: do not resubmit submitted VAAs during aggregation
* #52 bridge: do not log errors for duplicate VAA submissions
* **#51 bridge: have all nodes submit VAAs to Solana**
* #48 bridge/pkg/solana: retry VAA submission on transient errors
* #47 agent: return gRPC Internal error on submission failure
* #46 bridge: move initial guardian set fetching to pkg/ethereum/watcher.go
* #45 bridge: propagate panics from runnables
* #44 bridge: in-place debugging using dlv

VAAs are deduplicated by the on-chain contracts. For Ethereum,
submission happens outside of the bridge anyway, and for Solana, the
first tx to be confirmed wins. Subsequent attempts to submit it
will fail in preflight, so the fee won't be spent multiple times.

This eliminates the need for leader selection and fixes #20.

### DIFF
--- a/bridge/cmd/guardiand/processor.go
+++ b/bridge/cmd/guardiand/processor.go
@@ -227,18 +227,11 @@ func vaaConsensusProcessor(lockC chan *common.ChainLock, setC chan *common.Guard
 						zap.Int("have_sigs", len(sigs)),
 					)
 
-					if *unsafeDevMode && len(sigs) >= quorum {
-						idx, err := devnet.GetDevnetIndex()
-						if err != nil {
-							return err
-						}
-
+					if len(sigs) >= quorum {
 						vaaBytes, err := signed.Marshal()
 						if err != nil {
 							panic(err)
 						}
-
-						// TODO: proper leader selection
 
 						if t, ok := v.Payload.(*vaa.BodyTransfer); ok {
 							switch {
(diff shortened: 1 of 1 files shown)


leoluk:
> Subsequent attempts to submit it
> will fail in preflight, so the fee won't be spent multiple times.

@hendrikhofstadt What happens if it's submitted to two nodes simultaneously? Would both spend a fee?

hendrikhofstadt:
> > Subsequent attempts to submit it
> > will fail in preflight, so the fee won't be spent multiple times.
> 
> @hendrikhofstadt What happens if it's submitted to two nodes simultaneously? Would both spend a fee?

Both nodes could pass preflight but on one node the transaction confirmation would time out (no fee).

hendrikhofstadt (approved):

PR #55, opened 2020-10-22
bridge: use mlockall(2) to prevent pages from being swapped out

leoluk:
* **#55 bridge: use mlockall(2) to prevent pages from being swapped out**

The extra capability is harmless and is, at worst, a DoS risk.


leoluk:
(reviewed by Lorenz)

Issue #50, opened 2020-10-21
2/3+ majority


leoluk:
The spec says 2/3+, implementation is 2/3. Fix this and align across all components.

More of a formality due to how our consensus works.

hendrikhofstadt:
Can be closed when #58 is merged

PR #69, opened 2020-10-28
bridge: fix quorum calculation

hendrikhofstadt:
Previously we had rounding errors due to the usage of integers for calculating `n * 2/3`.

This was fixed by using one decimal fix point numbers during this part of the calucation.

Overflows were mitigated using a u16 cast on Rust. No mitigations are needed in Solidity since `.length` is a u256 (https://solidity.readthedocs.io/en/v0.5.3/types.html#array-members)


leoluk (approved):

hendrikhofstadt:
Partially fixed #50

leoluk:
See #58 for the Go implementation and test.

PR #88, opened 2020-11-18
README added for the Terra contracts and several security fixes to the contracts

ysavchenko:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/88)


leoluk:
What was the security issue(s) that were fixed?

ysavchenko:
Two actually:
- Wormhole could burn tokens without allowance given by the user, so lockTokens could be called by any user, not token account holder
- Transaction which was used by the bridge to watch for token lock events could also be called by anyone, so they could fool the bridge into thinking the token lock happened

ysavchenko:
And I removed burn tests because they were testing the wrong thing: wormhole should not be able to burn tokens without allowance

leoluk (approved):
:lgtm: with the same caveat: external review required for Terra contract changes.

> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/88#-:-MMQEfKM0VVja6QftDWA:b-cm2gl2)* status: 3 of 5 files reviewed, all discussions resolved

PR #130, opened 2020-12-03
Terra contract deployment moved to a separate k8s job

ysavchenko:
Basically what was done:
- Previous Terra node moved from `terra/docker` directory to `terra/devnet`, because now we have another docker image for Terra, so we should make naming less generic
- New docker image added, which is a multi-stage build, at first we use rust-optimizer to build all smart contracts and then add them to node-based image which will include JS code for deploying contracts to the devnet
- This new image is added as a Job to the Tilt configuration (reasons being that it only needs to run once)
- Afterwards I removed separate docs.md for Terra testing because now its job is done by the new `terra-contracts` job
- And since addresses generated in localterra are deterministic we can avoid changing them in the bridge config

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/130)


leoluk (changes requested):
Awesome! Once this is merged, we can integrate it with the e2e testing framework (bridge/e2e).

> Reviewed 22 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/130#-:-MNc96wNEaFLExSbptmL:bwmeeuu)* status: 22 of 23 files reviewed, 2 unresolved discussions (waiting on @leoluk and @ysavchenko)

___
*[devnet/terra-devnet.yaml, line 31 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNc7xR83Hl5rSlA356X:-MNc7xR83Hl5rSlA356Y:bma2ymj) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/devnet/terra-devnet.yaml#L31)):*
> ```YAML
> ---
> apiVersion: batch/v1
> kind: Job
> ```

The contract deployment is not idempotent, right? We need to restart terrad every time we want to redeploy it, otherwise it fails:

    Error Error: store code failed. code: 4, codespace: sdk, raw_log: unauthorized: signature verification failed; verify correct account sequence and chain-id

Given this, it would be better to deploy this as a sidecar container with terrad such that the two are always restarted simultaneously (this is how it works with the Solana and Ethereum contracts).
___
*[terra/Dockerfile, line 1 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNc9RmD9Cy37jAoRj_i:-MNc9RmD9Cy37jAoRj_j:b-f8tcrl) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/terra/Dockerfile#L1)):*
> ```Dockerfile
> # This is a multi-stage docker file, first stage builds contracts
> ```

We use Buildkit, which means that we can use cache mounts during builds (see the Dockerfile). This avoids a full npm rebuild every time the contract changes, refer to existing Dockerfiles how this works. I have no strong feeling either way, change it if you want or leave it if you believe it's not worth the effort.
___
*[terra/tools/deploy.sh, line 3 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNcAPtXCRimdft1NFMU:-MNcAPtXCRimdft1NFMV:b4id1zd) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/terra/tools/deploy.sh#L3)):*
> ```Shell
> npm run prepare-token http://terra-lcd:1317
> npm run prepare-wormhole http://terra-lcd:1317
> npm run lock-token terra18vd8fpwxzck93qlwghaj6arh4p7c5n896xzem5 terra174kgn5rtw4kf6f938wm7kwh70h2v4vcfd26jlc 1000 http://terra-lcd:1317
> ```

Please add these to the "Devnet addresses" section in DEVELOP.md.

leoluk (commented):
> Reviewed 1 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/130)* status: all files reviewed, 2 unresolved discussions (waiting on @ysavchenko)

ysavchenko (commented):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/130)* status: 18 of 24 files reviewed, 2 unresolved discussions (waiting on @leoluk)

___
*[devnet/terra-devnet.yaml, line 31 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNc7xR83Hl5rSlA356X:-MNcYgyS6-tcevp3Ld6B:b-wqqner) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/devnet/terra-devnet.yaml#L31)):*
<details><summary><i>Previously, leoluk (Leopold Schabel) wrote…</i></summary><blockquote>

The contract deployment is not idempotent, right? We need to restart terrad every time we want to redeploy it, otherwise it fails:

    Error Error: store code failed. code: 4, codespace: sdk, raw_log: unauthorized: signature verification failed; verify correct account sequence and chain-id

Given this, it would be better to deploy this as a sidecar container with terrad such that the two are always restarted simultaneously (this is how it works with the Solana and Ethereum contracts).
</blockquote></details>

This makes sense. I've got it working as a sidecar, however it keeps restarting after its job is finished. What I ended up with was `sleep infinity` at the end of the test script. I hope it is ok. There is also a small change in the test JS part, where the first deploy error is not shown, because terrad initialization was not finished at this point. This did not cause any problems, but always shown an extra error message which I got rid of.
___
*[terra/Dockerfile, line 1 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNc9RmD9Cy37jAoRj_i:-MNcRpRq47XQMu6VzXFj:b1l42tu) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/terra/Dockerfile#L1)):*
<details><summary><i>Previously, leoluk (Leopold Schabel) wrote…</i></summary><blockquote>

We use Buildkit, which means that we can use cache mounts during builds (see the Dockerfile). This avoids a full npm rebuild every time the contract changes, refer to existing Dockerfiles how this works. I have no strong feeling either way, change it if you want or leave it if you believe it's not worth the effort.
</blockquote></details>

Yes, I saw this on other dockerfiles and tbh it is a little out my expertise as I barely made working the Dockerfile as it is now :). So if it is ok, I'll leave it as it is.
___
*[terra/tools/deploy.sh, line 3 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNcAPtXCRimdft1NFMU:-MNcRlWQ8LecS6Guuhvj:b-896fix) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/terra/tools/deploy.sh#L3)):*
<details><summary><i>Previously, leoluk (Leopold Schabel) wrote…</i></summary><blockquote>

Please add these to the "Devnet addresses" section in DEVELOP.md.
</blockquote></details>

Done.

leoluk (approved):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/130)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[devnet/terra-devnet.yaml, line 31 at r1](https://reviewable.io/reviews/certusone/wormhole/130#-MNc7xR83Hl5rSlA356X:-MNcnsDn7JTK8eBgW6Eu:bdomdem) ([raw file](https://github.com/certusone/wormhole/blob/a6d0a6aa47df0a50cdb7d2176b24fe4ae947c834/devnet/terra-devnet.yaml#L31)):*
<details><summary><i>Previously, ysavchenko (Yuriy Savchenko) wrote…</i></summary><blockquote>

This makes sense. I've got it working as a sidecar, however it keeps restarting after its job is finished. What I ended up with was `sleep infinity` at the end of the test script. I hope it is ok. There is also a small change in the test JS part, where the first deploy error is not shown, because terrad initialization was not finished at this point. This did not cause any problems, but always shown an extra error message which I got rid of.
</blockquote></details>

LGTM - `sleep infinity` is a good way to do it. For the other pods it was changed to a TCP listener for the readiness check, but that's something for the e2e test integration.

PR #155, opened 2021-01-19
Fix guardian set expiration on Solana

hendrikhofstadt:
The expiration set is set to the time when the set was changed + expiration constant. So the guardian set is expired when that time is in the past.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/155)

### DIFF
--- a/solana/bridge/src/processor.rs
+++ b/solana/bridge/src/processor.rs
@@ -699,7 +699,7 @@ impl Bridge {
         }
 
         // Check that the guardian set is still active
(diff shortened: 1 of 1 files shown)


leoluk (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/155)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #159, opened 2021-01-23
Use /usr/bin/env for shell scripts; web/.dockerignore: node_modules

drozdziak1:
# `#!/usr/bin/env bash`
Hi, I've been trying to run the project according to `DEVELOP.md` on NixOS. I
managed to get most things to work, but Tilt had trouble executing
some of the jobs because of `/bin/bash` being used. NixOS will do the right
thing for `#!/usr/bin/env bash` and `#!/bin/sh` shell scripts but not direct
interpreter paths. Take it with a grain of salt, I replaced that with a "find
and replace in project" tool.

# `web/.dockerignore`
I saw that the web interface container tried to copy the context including
node_modules. I added it to an ignorefile.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/159)

### DIFF
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 #
 # This script downloads and installs development dependencies required for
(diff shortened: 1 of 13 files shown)


leoluk (approved):
Ah yes, the magic NixOS `/usr/bin/env` :) Nice catch, thanks!

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/159#-:-MRkFZqWEzAABT7yoAZu:b-afok2a)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #172, opened 2021-02-02
don't subsidize guardian set creation

hendrikhofstadt:
This works around https://github.com/solana-labs/solana/issues/9711 which causes issues when the guardian set creation is subsidized and another CPI call is done subsequently

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/172)

### DIFF
--- a/solana/bridge/src/processor.rs
+++ b/solana/bridge/src/processor.rs
@@ -875,7 +875,7 @@ impl Bridge {
             payer_info,
             program_id,
             &guardian_seed,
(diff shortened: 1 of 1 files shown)


leoluk (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/172)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #182, opened 2021-02-10
Added VAA verification query to the Terra smart contract, lock fees, refactoring

ysavchenko:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/182)

### DIFF
--- a/terra/contracts/wormhole/src/contract.rs
+++ b/terra/contracts/wormhole/src/contract.rs
@@ -11,7 +11,7 @@ use crate::msg::{GuardianSetInfoResponse, HandleMsg, InitMsg, QueryMsg};
 use crate::state::{
     config, config_read, guardian_set_get, guardian_set_set, vaa_archive_add, vaa_archive_check,
     wrapped_asset, wrapped_asset_address, wrapped_asset_address_read, wrapped_asset_read,
-    ConfigInfo, GuardianAddress, GuardianSetInfo,
+    ConfigInfo, GuardianAddress, GuardianSetInfo, ParsedVAA,
 };
 
 use cw20_base::msg::HandleMsg as TokenMsg;
@@ -30,6 +30,8 @@ use k256::ecdsa::Signature;
 use k256::ecdsa::VerifyKey;
 use k256::EncodedPoint;
 use sha3::{Digest, Keccak256};
+
+use generic_array::GenericArray;
 
 use std::convert::TryFrom;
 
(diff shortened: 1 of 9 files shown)


ysavchenko:
Added another commit with configurable lock fees, new stuff:
- Default fee is zero, contract owner can update it
- Locks without fee attached to the transaction fail (only if fee is non-0)
- Added owner method to transfer fee out of the contract
- Added contract state query (returns if it is active and fee value)
- Added tests for all of the above (except for moving funds out of the contract)

terra/contracts/wormhole/src/contract.rs
```diff
@@ -48,6 +50,7 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
         wrapped_asset_code_id: msg.wrapped_asset_code_id,
         owner: deps.api.canonical_address(&env.message.sender)?,
         is_active: true,
+        fee: Coin::new(0, ""),  // No fee by default
```
YunSuk-Yeo, line 53: It is just idea, not related with security or performance.

You can use Optional type and can put None or Some(Coin::new())

terra/contracts/wormhole/src/contract.rs
```diff
@@ -480,6 +449,13 @@ fn handle_lock_assets<S: Storage, A: Api, Q: Querier>(
         return ContractError::ContractInactive.std_err();
     }
 
+    // Check fee
+    if state.fee.amount > Uint128::zero() {
```
YunSuk-Yeo, line 453: If you use optional data type, you can check like 

```
if let Some(fee) = state.fee {
}
```

terra/contracts/wormhole/src/contract.rs
```diff
@@ -569,8 +545,45 @@ pub fn handle_set_active<S: Storage, A: Api, Q: Querier>(
...
+) -> StdResult<HandleResponse> {
+    let state = config_read(&deps.storage).load()?;
+
+    if deps.api.canonical_address(&env.message.sender)? != state.owner {
+        return ContractError::PermissionDenied.std_err();
+    }
+
     Ok(HandleResponse {
-        messages: vec![],
+        messages: vec![CosmosMsg::Bank(BankMsg::Send{
```
YunSuk-Yeo, line 582: It is just concern. If you always left $1 balance or feet token is Luna, you don't need to care about this.

In Terra blockchain, we charge tax to all native token transfer action except Luna. 
When the contract try to withdraw fee to other account, the chain will charge a tax from the left contract balance. 

To prevent over-withdraw due to tax, you can pre-compute the tax amount will be charged to this transfer action.

Ex) https://github.com/terraswap/terraswap/blob/72c60c05c43841499f760710a03f864c5ee4db3b/packages/terraswap/src/asset.rs#L32-L56

YunSuk-Yeo (changes requested):
overall updates are looks fine, but some concerns if a fee is not Luna denom. 
If a fee denom is Luna, you can just ignore the comment :)

terra/contracts/wormhole/src/contract.rs
ysavchenko: Yes, I also thought about this, but chose 0-sum coin simply for more effective storage strategy (Option takes more space when serialized). But looking at it again I have to agree, using Option would be more elegant.

terra/contracts/wormhole/src/contract.rs
ysavchenko: This is a great point, but I think we should keep the current logic for a couple of reasons:

- Fee on the contract is for spam prevention only, so I do not think it will be collected frequently if at all
- This is the admin method called by the contract owner, so I would assume that they know what they're doing

If it was a method a regular user could call this would be a different story and we need to add checks for Tax there.

ysavchenko: And also yes, I think the fee will be in LUNA :)

YunSuk-Yeo (approved):
LGTM thanks for updates

terra/contracts/wormhole/src/contract.rs
YunSuk-Yeo: cool

ysavchenko:
The last version removes contract activity status (it is always active now) and fees made hardcoded (to 0.01 LUNA). Tests updated to the new logic, obsolete tests removed.

leoluk (approved):

PR #437, opened 2021-09-15
bridge_ui: get terra testnet token on tesnet build

skqksh:


leoluk:
Is this change still relevant? Can you provide an explanation of your change?

PR #608, opened 2021-12-13
Stub out algorand support in wormhole

jumpsiegel:
1) Introduce the algorand chain constant in all the appropriate places
  2) Deploy pyth/hernandc algorand smart contracts into devnet
  3) Fund all the correct contracts for devnet testing


node/pkg/algorand/watcher.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"encoding/hex"
+//	"fmt"
+//	"github.com/certusone/wormhole/node/pkg/p2p"
+//	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+//	"github.com/prometheus/client_golang/prometheus/promauto"
+//	"io/ioutil"
+//	"net/http"
+//	"time"
+
```
leoluk, line 13: We have a "no commented out code on the main branch" habit

node/pkg/algorand/watcher.go
```diff
@@ -0,0 +1,58 @@
...
+	ID uint64 `json:"id"`
+}
+
+// NewWatcher creates a new Algorand contract watcher
+func NewWatcher(urlRPC string, urlToken string, contract string, lockEvents chan *common.MessagePublication, setEvents chan *common.GuardianSet) *Watcher {
+	return &Watcher{urlRPC: urlToken, contract: contract, msgChan: lockEvents, setChan: setEvents}
+}
+
+func (e *Watcher) Run(ctx context.Context) error {
+        return nil
```
leoluk, line 56: This will cause the runnable to crash-loop. If you want this to block forever, use `select {}`

node/go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tendermint/tendermint v0.33.7
 	github.com/terra-project/terra.go v1.0.1-0.20210129055710-7a586e5e027a
+	github.com/algorand/go-algorand-sdk v1.11.0
```
leoluk, line 42: There's no matching import so `go mod tidy` will garbage-collect this

node/cmd/guardiand/node.go
```diff
@@ -373,6 +384,16 @@ func runNode(cmd *cobra.Command, args []string) {
...
+        if *algorandRPC == "" {
+                logger.Fatal("Please specify --algorandRPC") 
+        }
+        if *algorandToken == "" { 
+                logger.Fatal("Please specify --algorandToken") 
+        }
+        if *algorandContract == "" { 
+                logger.Fatal("Please specify --algorandContract") 
+        }
+
```
leoluk, line 396: Best to put this behind a `*unsafeDevMode` condition such that it only runs in devnet.

(also, the formatting is wrong - Go uses exclusively tabs - just have your editor run `gofmt` on save and it fixes it up automatically)

leoluk (changes requested):

node/pkg/algorand/watcher.go
leoluk: And you need to call `readiness.SetReady` here or CI will spend an hour waiting for it to come up and then fail

devnet/algorand.yaml
```diff
@@ -52,7 +52,7 @@ spec:
           command:
             - /bin/sh
             - -c
-            - ./goal kmd start -d /network/Node && ./goal account list && sleep infinity
+            - ./goal kmd start -d /network/Node && ./goal account list && /setup/setup.sh && sleep infinity
```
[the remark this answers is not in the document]
jumpsiegel: Factored it out so that it only runs the wallet/contract setup instead of the pip dependencies

node/cmd/guardiand/node.go
jumpsiegel: Can you send me your emacs settings you use for go?

node/pkg/algorand/watcher.go
jumpsiegel: teach me to cleanup before a PR... I had that in the middle of a bunch of commented out code...

node/cmd/guardiand/node.go
leoluk: Well... I don't anymore, just use Emacs for org-mode these days and IntelliJ for everything else :P

PR #617, opened 2021-12-13
sdk/js: added terra transfer/deposit integration test

kev1n-peters:


sdk/js/src/token_bridge/__tests__/integration.ts
```diff
@@ -604,4 +608,93 @@ describe("Integration Tests", () => {
...
+          try {
+            await lcd.tx.estimateFee(wallet.key.accAddress, [transfer], {
+              memo: "localhost",
+              feeDenoms: ["uluna"],
+              gasPrices,
+            });
+          } catch (e) {
+            error = e.response.data.error.includes("Overflow: Cannot Sub");
+          }
+          expect(error).toEqual(true);
```
evan-gray, line 689: can you add a `withdraw_tokens` message at the end here so the tests are repeatable / leave the state the way they found it 🙂 I think it takes the same asset format as `initiate_transfer`

evan-gray (changes requested):

sdk/js/src/token_bridge/__tests__/integration.ts
kev1n-peters: For sure, that's a good idea.

kev1n-peters: @evan-gray should be good now

evan-gray (approved):

PR #625, opened 2021-12-16
hotfix parseNFTPayload

evan-gray:


evan-gray:
Wait to confirm encoding in devnet.

PR #631, opened 2021-12-17
bridge_ui: Transfer already redeemed feature

kev1n-peters:
Notify the user when they're trying to redeem a transfer that has
already been redeemed.

Added useGetIsTransferCompleted hook

### DIFF
--- a/bridge_ui/src/components/Recovery.tsx
+++ b/bridge_ui/src/components/Recovery.tsx
@@ -335,6 +335,10 @@ export default function Recovery() {
               targetAddress: parsedPayload.targetAddress,
               originChain: parsedPayload.originChain,
               originAddress: parsedPayload.originAddress,
+              amount:
+                "amount" in parsedPayload
+                  ? parsedPayload.amount.toString()
+                  : "",
             },
           })
         );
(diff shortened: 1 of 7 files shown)


bridge_ui/src/hooks/useFetchTargetAsset.ts
```diff
@@ -114,7 +109,7 @@ function useFetchTargetAsset(nft?: boolean) {
     ]
   );
   useEffect(() => {
-    if (isRecovery || argsMatchLastSuccess) {
```
kev1n-peters, line 117: For some context - the targetAsset isn't set in the redux store when the user is navigated to the redeem step (4) on the Transfer page during the recovery workflow. This hook will now fire after they're navigated to the redeem step and set the targetAsset.

bridge_ui/src/components/Recovery.tsx
```diff
@@ -335,6 +335,10 @@ export default function Recovery() {
               targetAddress: parsedPayload.targetAddress,
               originChain: parsedPayload.originChain,
               originAddress: parsedPayload.originAddress,
+              amount:
```
kev1n-peters, line 338: For some context - this could also be a NFT payload which doesn't have an amount property, so I have to do this to get around Typescript complaining

evan-gray (approved):

PR #634, opened 2021-12-19
node: add support for backfilling via RPC

leoluk:
**Stack**:
- #641
- #640
- #635
- #634 ⮜
- #639
- #638
- #637
- #636
- #633


hendrikhofstadt (approved):
Since the findMissing function is not public this looks good, otherwise we should add a comment making clear that this does no verification at all and is fully trusted

leoluk:
✓ Commit merged in pull request [#641](https://github.com/certusone/wormhole/pull/641)

PR #657, opened 2021-12-21
bridge_ui: Show checkmark on step 4 when complete

kev1n-peters:


evan-gray:
NFT transfers probably need the same treatment

evan-gray (changes requested):
NFT transfers probably need the same treatment

evan-gray (approved):

PR #672, opened 2021-12-24
bridge_ui: Warn when transferring shuttle wrapped tokens from Polygon

kev1n-peters:


bridge_ui/src/components/Transfer/TokenWarning.tsx
```diff
@@ -84,6 +92,15 @@ function RewardsWarning() {
   );
 }
 
+function PolygonTerraWrappedWarning() {
+  const classes = useStyles();
+  return (
+    <Alert severity="warning" variant="outlined" className={classes.alert}>
+      There are no liquid markets for Shuttle Wrapped Wormhole Wrapped Tokens.
```
kev1n-peters, line 99: Not sure if this is the message we want to display

chase-45 (approved):

PR #687, opened 2021-12-29
Token Bridge Transfer With Payload

evan-gray:
@calebrate wanted to run this by you: had to restructure a bit due to "stack too deep" - ended up returning a struct and splitting out the log function.

This PR implements "payload 3" for EVM and Terra and Solana.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/687)


ethereum/contracts/bridge/Bridge.sol
```diff
@@ -410,14 +491,15 @@ contract Bridge is BridgeGovernance {
...
+        
         transfer.toChain = encoded.toUint16(index);
         index += 2;
 
         transfer.fee = encoded.toUint256(index);
         index += 32;
 
-        require(encoded.length == index, "invalid Transfer");
+        // payload 3 allows for an arbitrary additional payload
+        require(encoded.length == index || transfer.payloadID == 3, "invalid Transfer");
```
kcsongor, line 502: this doesn't set `transfer.payload` (which doesn't exist, because the function return `Transfer` and not `TransferWithPayload`)

terra/contracts/token-bridge/src/contract.rs
```diff
@@ -120,6 +123,11 @@ const CHAIN_ID: u16 = 3;
...
 const WRAPPED_ASSET_UPDATING: &str = "updating";
 
+pub enum TransferType<A> {
+    WithoutPayload,
+    WithPayload { payload: A },
+}
+
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
     let mut bucket = wrapped_asset_address(deps.storage);
```
kcsongor, line 133: TODO: it's likely that by the time we roll out this patch to mainnet, this migration will have already gone through, so it should be removed.

kcsongor:
The solana implementation is in #760

ethereum/contracts/bridge/Bridge.sol
kcsongor: Just want to add to this: right now this function returns the whole VAA binary, and not just the payload (in case there is a payload). It's workable, clients just have to index into the binary at the right offset, but that essentially means leaking out the signature protocol, and changing it in the future would be a breaking change for downstream users of payload 3.

terra/contracts/token-bridge/src/contract.rs
```diff
@@ -247,6 +283,27 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> S
...
+        } => handle_initiate_transfer(
+            deps,
+            env,
+            info,
+            asset,
+            recipient_chain,
+            recipient.as_slice().to_vec(),
+            fee,
+            TransferType::WithPayload {
+                payload: payload.as_slice().to_vec(),
```
jynnantonix, line 305: This is making an unnecessary copy.  We can just do `payload: payload.into()`, which will unwrap the `Binary` and return the inner `Vec<u8>`.

karlkempe: When you have time, check out the latest push to this branch. I changed this and others to use `into()`.

ethereum/test/bridge.js
```diff
@@ -678,11 +820,200 @@ contract("Bridge", function () {
...
+            // tokenchain
+            testBridgedAssetChain +
+            // receiver (must be self msg.sender)
+            web3.eth.abi.encodeParameter("address", accounts[0]).substr(2) +
+            // receiving chain
+            web3.eth.abi.encodeParameter("uint16", testChainId).substring(2 + (64 - 4)) +
+            // fee
+            web3.eth.abi.encodeParameter("uint256", new BigNumber(fee).div(1e10).toString()).substring(2) +
+            // additional payload
+            web3.eth.abi.encodeParameter("address", accounts[1]).substr(2);
```
claudijd, line 853: In the above examples, we have been using arbitrary payloads, is there a reason why we're using address in this context?

claudijd: There's a few examples below, but mentioning here just so I can understand.

claudijd (approved):

kcsongor (approved):

claudijd (approved):

ethereum/test/bridge.js
evan-gray: no particular reason here, just matches what we do with the mock integration which actually uses that address to send the funds to

ethereum/test/bridge.js
```diff
@@ -614,6 +615,147 @@ contract("Bridge", function () {
...
+            gasLimit: 2000000
+        });
+
+        // deposit tokens
+        const initialized = new web3.eth.Contract(BridgeImplementationFullABI, TokenBridge.address);
+
+        const accountBalanceBefore = await token.methods.balanceOf(accounts[0]).call();
+        const bridgeBalanceBefore = await token.methods.balanceOf(TokenBridge.address).call();
+
+        assert.equal(bridgeBalanceBefore.toString(10), "0");
```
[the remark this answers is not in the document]
evan-gray: not really, since it would depend on the results of the previous tests and therefore incorrectly causes failures based on the state of the account. the real test is that it differs by the amount after the transfer

PR #713, opened 2022-01-04
Initial Rust SDK

Reisen:
This PR contains several commits, each commit is individually reviewable.

The SDK is structured into two layers:

- `sdk/rust/core` contains generic Rust types for Wormhole, including parsers.
- `sdk/rust/sdk` pulls in our core modules (token/nft) and re-exposes them along with core.

This means `core` can be relied on for working with pure Wormhole data wheras SDK is meant to be used to interact with the Wormhole ecosystem and contracts.

There are two examples, one for Solana and the second for Terra with tests that can be run to get used to how Wormhole works, and how to use the libraries.


sdk/rust/core/src/vaa/nft.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(PartialEq, Debug, Clone)]
+pub struct Transfer {
+    /// Address of the token. Left-zero-padded if shorter than 32 bytes
+    pub nft_address: [u8; 32],
+
+    /// Chain ID of the token
+    pub nft_chain: Chain,
+
+    /// Symbol of the token
+    pub symbol: String,
```
kcsongor, line 39: Aren't `symbol` and `name` supposed to be [u8; 32]? The nft spec says so

sdk/rust/sdk/src/chains/terra.rs
```diff
@@ -0,0 +1,63 @@
...
+    Addr::unchecked("terra1dq03ugtd40zu9hcgdzrsq6z2z4hwhc9tqk2uy5")
+}
+
+/// Export Core Devnet Contract Address
+#[cfg(feature = "devnet")]
+pub fn id() -> Addr {
+    Addr::unchecked("terra1pd65m0q9tl3v8znnz5f5ltsfegyzah7g42cx5v")
+}
+
+pub fn post_message<T>(wormhole: Addr, nonce: u32, message: &T) -> StdResult<CosmosMsg>
```
kcsongor, line 33: Why do these functions take the wormhole address as an argument, and not just use `id()` internally?

hendrikhofstadt (changes requested):
Pls drop the binary .so files

sdk/rust/core/src/vaa/nft.rs
kcsongor: On the receiving end, `String` is nicer for sure, but when you construct one of these, I think it's better if the type tells you about the size limitation explicitly as it's an important detail in the spec

hendrikhofstadt (approved):

PR #728, opened 2022-01-09
node/pkg/ethereum: check for rpc.ErrNoResult with non-nil tx

leoluk:
<pre>
This cannot currently happen the way TransactionReceipt is implemented,
but make sure to check the tx != nil case anyway in case the API
is changed in future releases of go-ethereum.
</pre>


hendrikhofstadt (approved):

leoluk:
✓ Commit merged in pull request [#743](https://github.com/certusone/wormhole/pull/743)

PR #730, opened 2022-01-10
solana: hide wasm-bindgen dependency behind the feature flag

drozdziak1:


leoluk (approved):
LGTM (assuming CI passes)

PR #747, opened 2022-01-12
update WROSE_ADDRESS to 0x21C7

keone:


evan-gray (changes requested):
This should not be changed unless the address the contract uses changes.

evan-gray (approved):
approving in coordination with governance

PR #756, opened 2022-01-13
Use Solana's log feature instead of accounts

hendrikhofstadt:
Change-Id: Ibbc34fe7292baa82fea51410d61642c59ce77085

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/756)


hendrikhofstadt:
Dropping this due to the potential risks of using logs and log parsing.
Continuing cost reduction efforts in #1077

PR #764, opened 2022-01-19
Terra sdk fixes

sekoje:
Terra sdk fixes and fix to tests in examples messenger

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/764)


sdk/rust/sdk/src/chains/terra.rs
```diff
@@ -33,19 +33,16 @@ pub fn id() -> Addr {
 /// Export Core Devnet Contract Address
 #[cfg(feature = "devnet")]
 pub fn id() -> Addr {
-    Addr::unchecked("terra1pd65m0q9tl3v8znnz5f5ltsfegyzah7g42cx5v")
+    Addr::unchecked("terra18vd8fpwxzck93qlwghaj6arh4p7c5n896xzem5")
```
hendrikhofstadt, line 36: This is not the devnet address according to the docs

hendrikhofstadt (changes requested):

sdk/rust/sdk/src/chains/terra.rs
Reisen: `terra18vd8fpwxzck93qlwghaj6arh4p7c5n896xzem5` seems to be correct, the previous address here was testnet not devnet so this looks like the right address to me now.

examples/messenger/terra/src/lib.rs
```diff
@@ -117,18 +117,14 @@ mod testing {
...
-                    110, 99, 101, 34, 58, 48, 125, 125
+                    34, 109, 101, 115, 115, 97, 103, 101, 34, 58, 34, 65, 119, 65, 65, 65, 69,
+                    74, 118, 89, 103, 115, 65, 65, 65, 66, 73, 90, 87, 120, 115, 98, 121, 66, 66, 98,
+                    71, 108, 106, 90, 81, 61, 61, 34, 44, 34, 110, 111, 110, 99, 101, 34, 58, 48, 125, 125
                 ]),
             }))]
         );
     }
-
+/*
```
Reisen, line 127: This test fails due to the fact that cosmwasm doesn't support mocking out cross-contract calls currently. Rather than commenting it out, for now can we remove the `#[test]` attribute and leave the function intact instead of committing commented code, and then adding a comment above this with something like:

```
/// This test currently would fail until cosmwasm can mock cross contract calls, see:
/// https://github.com/CosmWasm/cosmwasm/blob/0.16/packages/std/src/mock.rs#L389
```

Reisen (changes requested):

Reisen (approved):

hendrikhofstadt (approved):

hendrikhofstadt (approved):

Reisen (approved):

PR #706, opened 2022-01-03
p2w-sdk: don't build ETH contracts, only the types

drozdziak1:
**Stack**:
- #876
- #804
- #765
- #752
- #708
- #707
- #706 ⮜
- #705
- #704
- #703

<pre>
Change-Id: I7cbd18328227700635d7688aa24a9671e8919fcd
</pre>

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/706)

PR #705, opened 2022-01-03
p2w-sdk: Expand README

drozdziak1:
**Stack**:
- #876
- #804
- #765
- #752
- #708
- #707
- #706
- #705 ⮜
- #704
- #703

<pre>
Change-Id: I17cb547d6aaddc240588923561c26d11a787df2e
</pre>

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/705)

PR #704, opened 2022-01-03
Add p2w-relay, a p2w-sdk integration test

drozdziak1:
**Stack**:
- #876
- #804
- #765
- #752
- #708
- #707
- #706
- #705
- #704 ⮜
- #703

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/704)

PR #765, opened 2022-01-20
p2w-attest: price batching

drozdziak1:
**Stack**:
- #876
- #804
- #765 ⮜
- #752
- #708
- #707
- #706
- #705
- #704
- #703

<pre>
This commit introduces support for multiple Pyth product/price pairs
per call. The initial maximum batch size is 5 and is enforced using a
`P2W_MAX_BATCH_SIZE` constant.

solana/pyth2wormhole/program:
* On-chain batching logic
* Batch message parsing logic

solana/pyth2wormhole/client:
* Off-chain batching logic - divides any number of symbols into
largest possible batches
* Use a multi-symbol config file instead of CLI arguments

third_party/pyth/p2w-sdk:
* Expose batch parsing logic

third_party/pyth/p2w-relay:
* Comment out target chain calls until ETH contract supports batching
* Test the batch parsing function

third_party/pyth/p2w_<a href="autoattest.py">autoattest.py</a>:
* Generate and use the symbol config file  with pyth2wormhole-client

third_party/pyth/pyth_<a href="publisher.py">publisher.py</a>:
* Add a configurable number of mock Pyth symbols
* Adjust HTTP endpoint for multiple symbols
</pre>

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/765)

### DIFF
--- a/solana/pyth2wormhole/client/src/main.rs
+++ b/solana/pyth2wormhole/client/src/main.rs
@@ -1,4 +1,13 @@
+pub mod attestation_cfg;
 pub mod cli;
+
+use std::{
+    fs::File,
+    path::{
+        Path,
+        PathBuf,
+    },
+};
 
 use borsh::{
     BorshDeserialize,
@@ -60,7 +69,10 @@ use bridge::{
 };
 
 use pyth2wormhole::{
-    attest::P2WEmitter,
+    attest::{
+        P2WEmitter,
+        P2W_MAX_BATCH_SIZE,
+    },
     config::P2WConfigAccount,
     initialize::InitializeAccounts,
     set_config::SetConfigAccounts,
@@ -68,6 +80,8 @@ use pyth2wormhole::{
     AttestData,
     Pyth2WormholeConfig,
 };
+
+use crate::attestation_cfg::AttestationConfig;
 
 pub type ErrBox = Box<dyn std::error::Error>;
 
@@ -84,25 +98,25 @@ fn main() -> Result<(), ErrBox> {
 
     let (recent_blockhash, _) = rpc_client.get_recent_blockhash()?;
 
-    let tx = match cli.action {
+    let txs = match cli.action {
         Action::Init {
             owner_addr,
             pyth_owner_addr,
             wh_prog,
-        } => handle_init(
+        } => vec![handle_init(
             payer,
             p2w_addr,
             owner_addr,
             wh_prog,
             pyth_owner_addr,
             recent_blockhash,
-        )?,
+        )?],
         Action::SetConfig {
             ref owner,
             new_owner_addr,
             new_wh_prog,
             new_pyth_owner_addr,
-        } => handle_set_config(
+        } => vec![handle_set_config(
             payer,
             p2w_addr,
             read_keypair_file(&*shellexpand::tilde(&owner))?,
@@ -110,37 +124,35 @@ fn main() -> Result<(), ErrBox> {
             new_wh_prog,
             new_pyth_owner_addr,
             recent_blockhash,
-        )?,
+        )?],
         Action::Attest {
-            product_addr,
-            price_addr,
-            nonce,
+            ref attestation_cfg,
         } => handle_attest(
             &rpc_client,
             payer,
             p2w_addr,
-            product_addr,
-            price_addr,
-            nonce,
+            attestation_cfg.as_path(),
             recent_blockhash,
         )?,
     };
 
-    let sig = rpc_client.send_and_confirm_transaction_with_spinner(&tx)?;
+    for tx in txs {
+        let sig = rpc_client.send_and_confirm_transaction_with_spinner(&tx)?;
 
-    // To complete attestation, retrieve sequence number from transaction logs
-    if let Action::Attest { .. } = cli.action {
-        let this_tx = rpc_client.get_transaction(&sig, UiTransactionEncoding::Json)?;
+        // To complete attestation, retrieve sequence number from transaction logs
+        if let Action::Attest { .. } = cli.action {
+            let this_tx = rpc_client.get_transaction(&sig, UiTransactionEncoding::Json)?;
 
-        if let Some(logs) = this_tx.transaction.meta.and_then(|meta| meta.log_messages) {
-	    for log in logs {
-		if log.starts_with(SEQNO_PREFIX) {
-		    let seqno = log.replace(SEQNO_PREFIX, "");
-		    println!("Sequence number: {}", seqno);
-		}
-	    }
-        } else {
-            warn!("Could not get program logs for attestation");
+            if let Some(logs) = this_tx.transaction.meta.and_then(|meta| meta.log_messages) {
+                for log in logs {
+                    if log.starts_with(SEQNO_PREFIX) {
+                        let seqno = log.replace(SEQNO_PREFIX, "");
+                        println!("Sequence number: {}", seqno);
+                    }
+                }
+            } else {
+                warn!("Could not get program logs for attestation");
+            }
         }
     }
 
@@ -165,6 +177,7 @@ fn handle_init(
     };
 
     let config = Pyth2WormholeConfig {
+        max_batch_size: P2W_MAX_BATCH_SIZE,
         owner: new_owner_addr,
         wh_prog: wh_prog,
         pyth_owner: pyth_owner_addr,
@@ -202,6 +215,7 @@ fn handle_set_config(
     };
 
     let config = Pyth2WormholeConfig {
+        max_batch_size: P2W_MAX_BATCH_SIZE,
         owner: new_owner_addr,
         wh_prog: new_wh_prog,
         pyth_owner: new_pyth_owner_addr,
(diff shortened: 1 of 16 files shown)


solana/pyth2wormhole/client/src/main.rs
```diff
@@ -239,66 +256,116 @@ fn handle_attest(
...
-        },
-    );
-
-    let ix = Instruction::new_with_bytes(p2w_addr, ix_data.try_to_vec()?.as_slice(), acc_metas);
+    // Read the current max batch size from the contract's settings
+    let max_batch_size = config.max_batch_size;
+
+    let mut nonce = start_nonce;
+
+    let mut txs = Vec::with_capacity(attestation_cfg.symbols.len() / max_batch_size as usize);
```
jayantk, line 271: I think this needs to round up in case `attestation_cfg.symbols.len()` is not evenly divisible by `max_batch_size` ?

solana/pyth2wormhole/client/src/main.rs
```diff
@@ -84,63 +98,63 @@ fn main() -> Result<(), ErrBox> {
...
+            start_nonce,
         } => handle_attest(
             &rpc_client,
             payer,
             p2w_addr,
-            product_addr,
-            price_addr,
-            nonce,
+            attestation_cfg.as_path(),
+            start_nonce,
```
jayantk, line 136: how do these nonces work? do we lose the state of the nonce when this program quits? Who's responsible for maintaining what the last used nonce is?

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -49,14 +49,52 @@ use solitaire::{
...
 pub type P2WEmitter<'b> = Derive<Info<'b>, "p2w-emitter">;
 
+/// Important: must be manually maintained until native Solitaire
+/// variable len vector support. The number must correspond to maximum
+/// number of symbols and reflect how many pyth product/price pairs
+/// are expected in the Attest struct below.
+pub const P2W_MAX_BATCH_SIZE: u16 = 5;
+
 #[derive(FromAccounts, ToInstruction)]
 pub struct Attest<'b> {
```
jayantk, line 59: let me make sure I understand how this works. The limitation in this struct is purely a limitation of the on-chain program that we invoke to construct the attestation message, right? That is, this is something we can improve later once we figure out the stack size thing.

solana/pyth2wormhole/program/src/config.rs
```diff
@@ -10,6 +10,8 @@ pub struct Pyth2WormholeConfig {
     pub wh_prog: Pubkey,
     /// Authority owning Pyth price data
     pub pyth_owner: Pubkey,
+    /// How many product/price pairs can be sent and attested at once
+    pub max_batch_size: u16,
```
jayantk, line 14: is there a check at config loading time that this number is < P2W_MAX_BATCH_SIZE ?

third_party/pyth/p2w_autoattest.py
```diff
@@ -148,9 +156,27 @@ def serve_attestations():
         logging.error("Bad Content type")
         sys.exit(1)
 
-    PYTH_PRICE_ACCOUNT = pyth_accounts["price"]
-    PYTH_PRODUCT_ACCOUNT = pyth_accounts["product"]
-    logging.info(f"Retrieved Pyth accounts from endpoint: {pyth_accounts}")
+    cfg_yaml = f"""
+---
+symbols:"""
```
jayantk, line 161: FYI python has a `yaml` library that will generate the yaml text from python arrays / dictionaries. I think that's better than generating the string yourself here. docs: https://pyyaml.org/wiki/PyYAMLDocumentation

third_party/pyth/p2w_autoattest.py
```diff
@@ -186,10 +211,20 @@ def serve_attestations():
 readiness_thread = threading.Thread(target=readiness, daemon=True)
 readiness_thread.start()
 
-seqno_regex = re.compile(r"^Sequence number: (\d+)")
+seqno_regex = re.compile(r"Sequence number: (\d+)")
 
-nonce = 1
 while True:
+    # Make sure we start by
```
jayantk, line 217: nitpick: comment trails off

jayantk (commented):
Seems pretty reasonable to me. I didn't follow the transaction construction or the relayer / pyth_publisher, but i think that's ok.

solana/pyth2wormhole/program/src/config.rs
drozdziak1: Currently, it effectively *is* P2W_MAX_BATCH_SIZE and is only stored as courtesy toward attesters - the number of symbols expected by the business logic is maintained manually in `attest.rs` (Solana framework limitation that can be lifted in the future) and hard-coded in pyth2wormhole-client config-related options. The idea is that attesters read it right before deciding how to divide their symbols into batches. If we upgrade the contract with a new max_batch_size (especially changing to a smaller one), all attesters will be able to use the correct value immediately after the change is reflected on-chain in this struct.

third_party/pyth/p2w_autoattest.py
drozdziak1: I did not consider it worth it. This YAML's schema is insignificant and I expect that to be the case for forseeable future.

solana/pyth2wormhole/client/src/main.rs
drozdziak1: I am not sure. Wormhole accepts a nonce in its payload struct (PostMessageData). however It does not do anything significant with it (see `solana/bridge/program/src/api/post_message.rs`) in its contract. We might be well off without it.

drozdziak1: The nonce is redundant and I removed it

solana/pyth2wormhole/client/src/main.rs
drozdziak1: This may be premature of me. The pre-allocation may not be worth the raised eyebrows. I would not expect batch counts to go higher  than a thousand or two.. Removing in favor of starting with empty vector.

solana/pyth2wormhole/program/src/attest.rs
```diff
@@ -143,9 +252,9 @@ pub fn attest(ctx: &ExecutionContext, accs: &mut Attest, data: AttestData) -> So
     let post_message_data = (
         bridge::instruction::Instruction::PostMessage,
         PostMessageData {
-            nonce: data.nonce,
-            payload: price_attestation.serialize(),
-            consistency_level: data.consistency_level,
+	    nonce: 0, // Superseded by the sequence number
```
Reisen, line 255: Is it the case that attest will be executed more than once in a slot? If so you will still need different nonces if two attests are done in the same block, the way this is done in Wormhole is to just make the caller pass `nonce` in the instruction data and forward it through. Likely not a problem here though as far as I can tell.

solana/pyth2wormhole/program/src/types/mod.rs
```diff
@@ -51,12 +53,16 @@ pub const PUBKEY_LEN: usize = 32;
 #[repr(u8)]
 pub enum PayloadId {
     PriceAttestation = 1,
```
Reisen, line 55: Is this to maintain the ability to attest single prices, or could the batch be used with an `N` of 1 for the same purpose? Curious mostly to confirm if this is going to be the standard format for P2W messages in case it needs to be documented.

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: I don't think two of them would end up in the same slot. On a different note, I believe at one point you suggested that seqno has exactly the same job and that nonce is a relic+ of the past?

Reisen: It's a relic in that sequence and nonce perform the same job but it's still hashed by guardians in the VAA payload, and `node` still de-duplicates within a block for the same reason, so we still have to work with it.

Reisen: Example for the token_bridge in the JS SDK: https://github.com/certusone/wormhole/blob/dev.v2/sdk/js/src/token_bridge/attest.ts#L44 -- we can just let the user specify it when calling P2W and let them fill it with a random number.

solana/pyth2wormhole/program/src/config.rs
Reisen: Got it, so we have to be careful to make sure to include config change transactions with upgrade to avoid downtime with this approach.

drozdziak1: Yes, batching is a special snowflake in this respect. I was not able to back the config value with logic that would respect it dynamically (Option is not enough for that, we would need Vec<>). Otherwise I think the right thing to do is have the program understand the preceding config and use the older one until the privileged set-config call is made. This is an actual hard requirement as the program must understand an older config enough to verify the authority attempting to migrate it.

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: So the guardians don't use the sequence number when hashing? Why?

solana/pyth2wormhole/program/src/types/mod.rs
drozdziak1: My instinct was to leave it in place so that there can be no confusion about what is being sent

solana/pyth2wormhole/program/src/attest.rs
drozdziak1: **Are we concerned about duplicate seqnos for two separate messages processed in the same slot?**
I just confirmed that the seqno *is* used for hashes that later are used for signatures, as seen here: https://github.com/certusone/wormhole/blob/dev.v2/node/pkg/vaa/structs.go#L324

I'm sorry, but I'm struggling to understand the importance of the seqno, because as it stands, the seqno is just as good if not better.

drozdziak1:
Taken care of in the large PR

PR #772, opened 2022-01-24
fix: parseSequenceFromLogSolana -> Throw correct error if a Solana Tx is not found

davidediak:


evan-gray (approved):

evan-gray:
Thanks for this @davidediak !

PR #774, opened 2022-01-25
node: avoid clobbering terminal with binary data

leoluk:
**Stack**:
- #775
- #744
- #774 ⮜

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #744, opened 2022-01-11
node/pkg/processor: expire late observations

leoluk:
**Stack**:
- #775
- #744 ⮜
- #774

<pre>
Fixes <a href="https://github.com/certusone/wormhole/issues/685">https://github.com/certusone/wormhole/issues/685</a>.

Example occurrence this fixes: <a href="https://i.imgur.com/gZWKf1n.png">https://i.imgur.com/gZWKf1n.png</a>

Possible future optimizations include:

- Ignore late messages in the processor (but we can only ignore
them post settlement time, so we need the cleanup logic regardless).

- Ignoring late observations from other nodes.

- Using the stored VAA to calculate misses.

- Drop incomplete local observations. However, this is not trivial
since we do not know the message ID for those.
</pre>
**Issue references**: 

 - Fixes https://github.com/certusone/wormhole/issues/685

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #775, opened 2022-01-25
node/pkg/telemetry: reduce log level to INFO

leoluk:
**Stack**:
- #775 ⮜
- #744
- #774

<pre>
It appears that GCP Cloud Logging cannot handle the volume of logs
we're throwing at it... full text search slows to a crawl (LOL)

Reduce log level until we can move to something else.

commit-id:b71c3467
</pre>

### DIFF
--- a/node/cmd/guardiand/logging.go
+++ b/node/cmd/guardiand/logging.go
@@ -1,0 +1,28 @@
+package guardiand
+
+import (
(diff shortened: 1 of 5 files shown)


SEJeff (approved):
But @leoluk, google is web scale! /s

leoluk:
It's kind of funny for Google to fail at _full text search_

PR #779, opened 2022-01-26
improve logging around GCP PubSub message path

justinschuldt:
This change increases logging around the GCP PubSub message path, in order to help with debugging.

### DIFF
--- a/node/pkg/reporter/bigtablewriter.go
+++ b/node/pkg/reporter/bigtablewriter.go
@@ -135,8 +135,7 @@ func BigTableWriter(events *AttestationEventReporter, connectionConfig *BigTable
 					publishResult := pubsubTopic.Publish(ctx, &pubsub.Message{
 						Data: []byte(b),
 					})
-					_, err = publishResult.Get(ctx)
-					if err != nil {
+					if _, err = publishResult.Get(ctx); err != nil {
 						logger.Error("Failed getting GCP PubSub publish reciept",
 							zap.String("rowKey", rowKey),
 							zap.Error(err))
--- a/node/pkg/reporter/bigtablewriter.go
+++ b/node/pkg/reporter/bigtablewriter.go
@@ -57,9 +57,14 @@ func BigTableWriter(events *AttestationEventReporter, connectionConfig *BigTable
 
 		pubsubClient, err := pubsub.NewClient(ctx, e.connectionConfig.GcpProjectID)
 		if err != nil {
-			return fmt.Errorf("failed to create PubSub client: %w", err)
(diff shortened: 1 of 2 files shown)


node/pkg/reporter/bigtablewriter.go
```diff
@@ -57,9 +57,14 @@ func BigTableWriter(events *AttestationEventReporter, connectionConfig *BigTable
 
 		pubsubClient, err := pubsub.NewClient(ctx, e.connectionConfig.GcpProjectID)
 		if err != nil {
-			return fmt.Errorf("failed to create PubSub client: %w", err)
+			logger.Error("failed to create GCP PubSub client", zap.Error(err))
+			return fmt.Errorf("failed to create GCP PubSub client: %w", err)
```
SEJeff, line 61: Logging and returning an error in the same place is generally discouraged. [This post](https://www.orsolabs.com/post/go-errors-and-logs/) gives an idea of why at the very top, but just as a general rule, it is not very idiomatic go. One of the golang core developers [has the same opinion](https://dave.cheney.net/tag/error-handling) and has discussed it multiple times.

Log where the error is handled, not inside the function returning the error.

node/pkg/reporter/bigtablewriter.go
```diff
@@ -122,15 +126,21 @@ func BigTableWriter(events *AttestationEventReporter, connectionConfig *BigTable
...
 							zap.String("columnFamily", colFam),
 							zap.Error(err))
 						errC <- err
 					}
-					pubsubTopic.Publish(ctx, &pubsub.Message{
+					publishResult := pubsubTopic.Publish(ctx, &pubsub.Message{
 						Data: []byte(b),
 					})
+					_, err = publishResult.Get(ctx)
+					if err != nil {
```
SEJeff, line 139:
```suggestion
					if _, err = publishResult.Get(ctx); err != nil {
```

SEJeff (approved):

node/pkg/reporter/bigtablewriter.go
justinschuldt: Yeah, I agree that is overkill. Wanted to be explicit because it should bubble up to here https://github.com/certusone/wormhole/blob/dev.v2/node/cmd/guardiand/node.go#L726, but after being handled a bit by the Supervisor code.
Wanted to be **VERY** sure that I will be able to find it in logs if it is happening.

Would like to leave it for the short term, then will remove all with the "init success" logs.

node/pkg/reporter/bigtablewriter.go
justinschuldt: good idea, updating now.

leoluk (approved):

PR #782, opened 2022-01-26
node/pkg/solana: fix context memory leak

leoluk:
<pre>
commit-id:69632fa8
</pre>


SEJeff (approved):
`defer cancel()` inside an infinite loop... memory go 💹

justinschuldt (approved):

PR #797, opened 2022-01-29
node/hack: add repair script

leoluk:
Fetches the list of missing Solana messages, finds the corresponding
on-chain transactions, sends an ObservationRequest to the network
and waits for consensus.

Hacky one-off script with a couple of edge cases (doesn't filter
failed txs, crashes if the boundary is not a valid tx). We can later
refactor this and add it to the find-missing-messages admin command
if we keep needing it.

---

**Stack**:
- #821
- #798
- #797 ⮜


node/hack/repair/repair.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func main() {
+	flag.Parse()
+
+	ctx := context.Background()
+	sr := rpc.New(*solanaRPC)
+
+	conn, err, admin := getAdminClient(ctx, *adminRPC)
+	defer conn.Close()
```
SEJeff, line 68: Don't you normally do defer after the err check?

node/hack/repair/repair.go
```diff
@@ -0,0 +1,306 @@
...
+						}})
+					if err != nil {
+						log.Fatalf("SendObservationRequest: %v", err)
+					}
+
+					for {
+						log.Printf("verifying %d", p.Sequence)
+						req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf(
+							"%s/v1/signed_vaa/%d/%s/%d",
+							publicRPCEndpoints[0],
```
SEJeff, line 221: It's a shame we don't retry this to a node other than the certus node here, but I get why you might not want to do that.

SEJeff (approved):

node/hack/repair/repair.go
leoluk: Actually... we could, but I wanted to avoid picking a random node on the first try that wasn't available

node/hack/repair/repair.go
leoluk: Indeed

node/hack/repair/repair.go
SEJeff: Works for me.

leoluk:
✓ Commit merged in pull request [#821](https://github.com/certusone/wormhole/pull/821)

PR #807, opened 2022-02-02
node/pkg/db: add db.VaaIDFromString and vaa.StringToAddress helpers

leoluk:
**Stack**:
- #810
- #809
- #808
- #807 ⮜

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #808, opened 2022-02-02
node/hack: add repair script

leoluk:
Fetches the list of missing Solana messages, finds the corresponding
on-chain transactions, sends an ObservationRequest to the network
and waits for consensus.

Hacky one-off script with a couple of edge cases (doesn't filter
failed txs, crashes if the boundary is not a valid tx). We can later
refactor this and add it to the find-missing-messages admin command
if we keep needing it.

---

**Stack**:
- #810
- #809
- #808 ⮜
- #807


leoluk:

PR #809, opened 2022-02-02
node/pkg/solana: skip failed transactions

leoluk:
Avoid unnecessary RPC calls/retries.

Makes no difference for safety, which relies on VAA accounts rather than
any transaction metadata.

---

**Stack**:
- #810
- #809 ⮜
- #808
- #807


leoluk:

PR #810, opened 2022-02-02
staging/algorand: apply global Apache 2 license

leoluk:
As previously discussed, staging/algorand will be relicensed as
Apache 2. Replace all license headers by the global license.

Rand Labs needs to sign off on this PR to formally relicense
their contributions.


evan-gray (approved):

hernandp:
We agree with the new licensing.

Thanks.

PR #829, opened 2022-02-08
node: increase Polygon minimum confirmation to 512

leoluk:
It appears that in spite of exceeding the recommended value by 2x,
256 offers insufficient safety margin given that we saw two reorgs
of depths 178 and 187 recently: <a href="https://polygonscan.com/blocks_forked">https://polygonscan.com/blocks_forked</a>

This will increase confirmation delay from ~10 minutes to ~20 minutes.


nodimdim:
While this may be fine for Mainnet, it makes testing on Mumbai Testnet extremely difficult. In Mumbai, 512 block confirmations take about 52 minutes. This is hampering development flow.

PR #836, opened 2022-02-09
Deploy instructions

kcsongor:
Cherry-picked out a few commits from various other PRs to aid in deployments.
Includes scripts and instructions to aid terra testnet & mainnet deployments, and script fixes + truffle config updates for evm deployments.


kcsongor:
@evan-gray awesome, thanks! do you know why these dependencies were needed? Nothing in the PR requires python I think

evan-gray:
I’m guessing that particular truffle version or some dependency did not have a prebuilt binary so it builds something with node-gyp, which requires g++, make, and Python 🙂

PR #839, opened 2022-02-09
node: terra reobservation

evan-gray:


hendrikhofstadt (approved):

node/pkg/terra/watcher.go
```diff
@@ -156,6 +167,63 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case r := <-e.obsvReqC:
+				if vaa.ChainID(r.ChainId) != vaa.ChainIDTerra {
+					panic("invalid chain ID")
+				}
+
+				tx := eth_common.BytesToHash(r.TxHash)
```
leoluk, line 180: It's a bit weird to use a go-ethereum Hash type here - any specific reason we do that?

node/pkg/terra/watcher.go
```diff
@@ -156,6 +167,63 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+
+				logger.Info("received observation request for terra",
+					zap.String("tx_hash", tx.Hex()))
+
+				client := &http.Client{
+					Timeout: time.Second * 5,
+				}
+
+				// Query for tx by hash
+				resp, err := client.Get(fmt.Sprintf("%s/cosmos/tx/v1beta1/txs/%s", e.urlLCD, strings.TrimPrefix(tx.Hex(),"0x")))
```
leoluk, line 190: In a similar vein, what's the `strings.TrimPrefix` for? Because the Eth library prepends a 0x?

leoluk (changes requested):

node/pkg/terra/watcher.go
evan-gray: prior art. no specific reason, we use this same thing in eth and in StringToHash at the bottom of the file

leoluk: StringToHash uses it because it explicitly needs a common.Hash for MessageObservation. It's just a needlessly complicated way of converting plain bytes to hex otherwise.

leoluk (approved):

node/pkg/terra/watcher.go
evan-gray: Done and tested, thanks!

SEJeff (approved):

leoluk (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/839)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @evan-gray)

leoluk:
@SEJeff ?

SEJeff:
@leoluk my browser showed that it needed 1 more ack, so I clicked the request review button. Then I refreshed and it showed merged. Computers are hard.

PR #865, opened 2022-02-16
Add stronger nonce/confidence parsing

claudijd:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/865)

This change moves the int parsing logic up in the logic path.  Side benefit here is that it will now prevent negative numbers for each value.

### DIFF
--- a/clients/eth/main.go
+++ b/clients/eth/main.go
@@ -51,7 +51,7 @@ func init() {
 }
 
 func postMessage(cmd *cobra.Command, args []string) {
-	nonce, err := strconv.Atoi(args[0])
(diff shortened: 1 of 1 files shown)


leoluk (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/865)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SEJeff)

SEJeff (approved):
Nice and simple change!

This would be a good time to add cursory tests if you have cycles, but we can come back to that, especially since this is an internal thing only really currently used for development.

PR #907, opened 2022-02-25
Token/NFT Bridge Shutdown (aka: "Emergency Shutdown")

claudijd:
Opening this PR for open feedback on a security feature to allow the security guardians to enable/disable transactions for the Token/NFT bridge during existential threat scenarios.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/907)


mcf-rocks:
hi, obviously we need a bunch safeguards, probably I’m about to be that guy who doesn’t know what he is talking about, talking to the guy who knows what he is talking about :) but still it would be nice if you can alleviate my concerns.

“Wormhole security guardians (a subset of the 19)”

That sounds manual. I think we need something quicker, like automated… and maybe that thing is also being developed idk… also if it is manual, a fixed subset of 19 will react slower than a self-selected subset of 19 (first responders), and if the subset is a fixed subset who would want to have that responsibility?

“may need to take quick and decisive action (typically minutes) to enable/disable transactions”

don’t we want to fail the system pre-emptively to the fail condition… like if an issuance will cause an imbalance that issuance doesn’t happen?

kcsongor:
@mcf-rocks re your last point, you're absolutely right that pre-emptive safety guards are crucial. This device is meant to be a manual kill switch, as quick & dirty way of shutting the contract down when someone thinks there's something off, in the most low tech way possible. Balance checking and automated response requires cross chain accounting which is also on the roadmap but is a different facet of the security strategy

kcsongor:
The incremental nonce in Option 1 could be a per-signer counter, so no synchronisation is required between the guardians, and the signatures can all be included in the same block, which could matter in an emergency situation. Another option is to have the nonce be a random number, but then the replay protection code becomes more complex as it requires allocating storage for storing previous signatures, which we should avoid.

claudijd:
There's been a couple conversations where we've iterated on this approach, I really need to go back and fill in the options, I think it's possible the full Guardian set could still vote down instead of a subset, which is most safe and less overhead than a patched smart contract, but it necessitates being able to get a 2/3+ majority quickly in a time of need.  There are also other ideas and proposals for risk mitigation (some of which will be viable, others just academic), but I really need to get to task and write some of those up for public review and consideration.

claudijd:
@mcf-rocks the "first responders" concept was also discussed briefly, I really should rework this to be more about the general approach rather than the implementation primitives on the contract.

whitepapers/0007_killswitch.md
```diff
@@ -0,0 +1,87 @@
...
+* Less complexity and thus less potential for abuse when compared to option 1
+* Normal governance can be used to update signers as the security gaurdians change
+
+### Cons
+
+* Security Guardians would be limited from sending a `vote` from a single wallet.  Alternatively, if an implementation didn't bind wallets and signers 1:1, a single guardian could send `signed message` on another security guardians behalf if that guardian couldn't use their wallet.
+
+## Detailed Design (Option N)
+
+If you have other proposals, please add them here or suggest them via the PR.
```
ruuda, line 87: Here’s an alternative idea: instead of a list of signers and booleans, have a list of token accounts, that are controlled by the signers. Put some tokens in it (it doesn’t really matter which ones or how many). The contract would check that a majority of those accounts have a positive balance. To trigger the kill switch, signers temporarily move the funds to a different account.

### Pros

* Doesn’t require any special tooling, signers can just use their everyday wallet, which means it should be very quick. (I.e. no figuring out last moment that the CLI utility is broken and needs to be rebuilt from source, or figuring out what the right syntax was for referencing a particular derivation path on the Ledger, etc.)
* Can easily check the status on block explorers without custom tooling.

### Cons

* Infeasible on Solana because all contracts would need all of these accounts as input and Wormhole already needs a lot of account inputs.
* Not possible for a 2/3 majority of the guardians to re-enable the network after the kill switch is triggered, it can only be undone by the same people who triggered it. Although I suppose this could be addressed with a contract update if needed.

I don’t think this is a feasible alternative, but maybe some iteration of it can be helpful.

whitepapers/0007_killswitch.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Eliminates the burden of binding shutdown actions from a specific wallet
+* Normal governance can be used to update signers as the security gaurdians change
+
+### Cons
+
+* Will require some CLI tooling or manual steps for a signer to obtain the current nonce before signing a message to ensure speed
+* If for some reason the nonce rolls over and becomes non-unique, it could be succeptable to replay attacks that the nonce aims to solve for
+* Will require each signer to wait until the next block before signing
+
```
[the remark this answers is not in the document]
claudijd: Yes and Yes.  It would likely take shape of a CLI tool held by the Guardians to make this process easier.

hoos97:
Adding an idea for consideration.  The irony of decision making in an decentralized environment is that operating by majority sentiment institutes considerable delay and drag in the process.  We point out that circuit breaker removes decentralization, but don't forget that approval to have a circuit breaker is a strategic decision that is approved in a decentralized manner (2/3 of guardians).  But the initiation of the breaker when the situation dictates is a tactical decision that needs to made in a non-decentralized manner...time is of the essence.  
The decision to activate a circuit breaker should be done with command by negation.  We empower certain members of the wormhole team to take circuit breaker action.  If a need for activation is identified, the guardian network is notified and a short window is provided to enable rapid responses.  In the absence of any negative votes, the circuit breaker is activated and wormhole is placed in a safe condition.  
This could be built on further.  I.e., should votes start to trickle in after the action has been taken, and turns out a threshold of negative votes are received, then you could close the breaker and restore activities.  But, if votes remain supportive, than you keep wormhole in safe condition as you determine extent of the vulnerability and enable full engagement of the guardian network to determine recovery steps.

bruce-riley (dismissed):

whitepapers/0007_killswitch.md
claudijd: Reverted to quorum on this for now, keeps the security assumptions a bit cleaner, but it's a trade off on some speed.

PR #922, opened 2022-02-28
pyth2wormhole: only prepare 2 test Pyth symbols in CI

drozdziak1:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/922)


evan-gray:
This looks like it still takes 12+ minutes 😕

PR #927, opened 2022-03-01
solitaire: Infer entry point argument type in solitaire! macro

kcsongor:
The type of the `ix_data` binding withing the solitaire! macro expansion
is explicitly annotated with its type, which comes from the macro's
pattern binding `$kind`. However, this type annotation is entirely
optional, since `ix_data` is passed in as an argument to `$fn`, whose
type forecs `ix_data` anyway, and rust will happily infer that from the
application.

I propose we remove it, so the `solitaire!` entry point is easier to use. The syntax is made-up anyway and not particularly intuitive, so I think this is an improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/927)

### DIFF
--- a/solana/bridge/cpi_poster/src/lib.rs
+++ b/solana/bridge/cpi_poster/src/lib.rs
@@ -17,5 +17,5 @@ pub use api::{
 };
 
 solitaire! {
-    PostMessage(PostMessageData)                => post_message,
+    PostMessage                => post_message,
 }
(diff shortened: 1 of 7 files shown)


Reisen (dismissed):
Looks good to me, this is much more intuitive. It also makes refactoring a tiny bit easier when renaming types.

jynnantonix (dismissed):

hendrikhofstadt (dismissed):

jynnantonix (approved):

hendrikhofstadt (approved):

Reisen (approved):

PR #948, opened 2022-03-04
Refactor Symmetric Unit-tests to Testify

claudijd:
Refactors symmetric unit-test to use testify and adds more of a framework for adding additional cases.

Breaking the tests out also guard against a theoretical bug scenario where both the encrypt and decrypt methods are effected by a bug in common code, anchoring us to binary blobs from a known good ciphertext.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/948)

### DIFF
--- a/node/pkg/common/symmetric_test.go
+++ b/node/pkg/common/symmetric_test.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"encoding/hex"
 	"github.com/stretchr/testify/assert"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


node/pkg/common/symmetric_test.go
```diff
@@ -1,24 +1,36 @@
...
+		{data: []byte("AlphaNum1")},
+		{data: []byte("12345")},
 	}
 
-	if string(dec) != string(data) {
-		t.Fatalf("expected %s got %s", string(data), string(dec))
+	key := []byte("01234567890123456789012345678901")
+
+	for _, tc := range tests {
+		// Test that we can encrypt data without error.
```
SEJeff, line 25: This is a good refactor with 1 caveat. The way you did this golang treats this as one MONSTER test and if a single assertion fails, the rest never run. It is better to have go treat each of the test cases in `tests` be an individual test that can be run independently.

```suggestion

	for _, testCase := range tests {
		// Test that we can encrypt data without error.
		t.Run(string(testCase.data), funct(t *testing.T){
```

Note that if you do that you'll have an extra { and ( you need to close.

SEJeff (approved):
Approving as this is a big improvement, however consider adopting my suggestion. It makes it much nicer when you can specify a single testcase with `go test ... -run /foo/`

node/pkg/common/symmetric_test.go
claudijd: Made those changes, thank you!

node/pkg/common/symmetric_test.go
```diff
@@ -1,24 +1,62 @@
...
+			enc, err := EncryptAESGCM(testCase.data, key)
+			assert.Nil(t, err)
+			assert.NotNil(t, enc)
+
+			dec, err := DecryptAESGCM(enc, key)
+			assert.Nil(t, err)
+			assert.NotNil(t, dec)
+			assert.Equal(t, testCase.data, dec)
+
+			assert.Equal(t, dec, testCase.data)
```
claudijd, line 59: Dup text case, going to fix this.

node/pkg/common/symmetric_test.go
```diff
@@ -1,24 +1,60 @@
...
 
-	if string(dec) != string(data) {
-		t.Fatalf("expected %s got %s", string(data), string(dec))
+	for _, testCase := range tests {
+		// Verify that we can encrypt plain text
+		t.Run(string(testCase.data), func(t *testing.T) {
+			enc, err := EncryptAESGCM(testCase.data, key)
+			assert.Nil(t, err)
+			assert.NotNil(t, enc)
+			// AESGCM is non-deterministic, so we cannot expect consistent cipher text
```
claudijd, line 32: The reason this is non-deterministic, is that the nonce is generated on the fly and not exposed as an interface.

leoluk (approved):

SEJeff (approved):

claudijd (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/948#-:-Mz1hFgR7bWe15brSpUB:bnfp4nl)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @evan-gray and @SEJeff)

PR #974, opened 2022-03-08
remove dependabot

evan-gray:
Dependabot was disabled, but the yml remained

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/974)


SEJeff (approved):
I want to un-ack this so I can ack it twice.

kev1n-peters (approved):

PR #993, opened 2022-03-11
Implement deposit addresses

hendrikhofstadt:
Deposit addresses allow users to initiate cross-chain transfers by sending tokens to an EOA (e.g. to be used to send cross-chain transfers from an exchange). The transfer can then be completed by a "initial-relayer" for the specified fee [there should be an additional fee specified for this in addition to the cross-chain fee].

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/993)


hendrikhofstadt:
This was merely a demo and not meant to be merged

PR #997, opened 2022-03-14
bridge_ui: Make transfers to/from chains configurable

kev1n-peters:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/997)

### DIFF
--- a/bridge_ui/src/config.ts
+++ b/bridge_ui/src/config.ts
@@ -1,0 +1,33 @@
+import { ChainId } from "@certusone/wormhole-sdk";
+
+export type DisableTransfers = boolean | "to" | "from";
+
+export interface WarningMessage {
+  text: string;
+  link?: {
+    url: string;
+    text: string;
+  };
+}
+
+export interface ChainConfig {
+  disableTransfers?: DisableTransfers;
+  warningMessage?: WarningMessage;
+}
+
+export type ChainConfigMap = {
+  [key in ChainId]?: ChainConfig;
+};
+
+export const CHAIN_CONFIG_MAP: ChainConfigMap = {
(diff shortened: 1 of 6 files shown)


bridge_ui/src/config.ts
```diff
@@ -0,0 +1,33 @@
...
+  warningMessage?: WarningMessage;
+}
+
+export type ChainConfigMap = {
+  [key in ChainId]?: ChainConfig;
+};
+
+export const CHAIN_CONFIG_MAP: ChainConfigMap = {};
+
+CHAIN_CONFIG_MAP[CHAIN_ID_POLYGON] = {
```
kev1n-peters, line 24: type checking doesn't work on this object's properties when it's included in the initialization of CHAIN_CONFIG_MAP...

evan-gray: It will if you use `as ChainConfigMap` after :wink:

kev1n-peters: Changed

evan-gray (approved):
:lgtm:

Was wondering if the warning message should show regardless of the disabled state, but I think I agree with the way it is. The message can be written in a way to convey the status, and folks probably want to know if the opposite way is disabled ahead of time. In the future may want to consider it showing as an error if the selected direction is disabled.

> Reviewed 5 of 7 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/997#-:-MyK5XJID1v7PHbjxiIt:b-1s8ypx)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @chase-45 and @kev1n-peters)

chase-45 (approved):

PR #1001, opened 2022-03-15
sdk: Add vaa_payload crate

jynnantonix:
The vaa_payload crate implements a serde data format for VAA payloads.
This will let us replace all the hand-rolled serialization code with
auto-generated code, which is less error-prone and easier to review.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1001)

### DIFF
--- a/sdk/rust/vaa_payload/src/lib.rs
+++ b/sdk/rust/vaa_payload/src/lib.rs
@@ -1,0 +1,426 @@
+//! Serialize and deserialize rust values from the VAA payload wire format.
+//!
+//! As of this writing (June, 2022) there is no proper specification for the VAA payload wire
+//! format so this implementation has mostly been reverse engineered from the existing messages.
+//! While the rest of this document talks about how various types are represented on the wire this
+//! should be seen as an explanation of how things are implemented *in this crate* and not as
+//! official documentation. In cases where the serialization of a payload produced by this crate
+//! differs from the one use by the wormhole contracts, the serialization used by the actual
+//! contract is considered the canonical serialization.
+//!
+//! Unless you want to interact with existing wormhole VAA payloads, this crate is probably not what
+//! you are looking for. If you are simply using the wormhole bridge to send your own payloads then
+//! using a schema with auto-generated code (like protobufs or flatbuffers) is probably a better
+//! choice.
+//!
+//! ## Wire format
+//!
+//! The VAA payload wire format is not a self-describing format (unlike json and toml). Therefore it
+//! is necessary to know the type that needs to be produced before deserializing a byte stream.
+//!
+//! The wire format currently supports the following primitive types:
+//!
+//! ### `bool`
+//!
+//! Encoded as a single byte where a value of 0 indicates false and 1 indicates true. All other
+//! values are invalid.
+//!
+//! ### Integers
+//!
+//! `i8`, `i16`, `i32`, `i64`, `i128`, `u8`, `u16`, `u32`, `u64`, and `u128` are all supported and
+//! encoded as full-width big-endian integers, i.e., `i16` is 2 bytes, `u64` is 8 bytes, etc.
+//!
+//! ### `char`
+//!
+//! Encoded as a big-endian `u32`, with the additional restriction that it must be a valid [`Unicode
+//! Scalar Value`](https://www.unicode.org/glossary/#unicode_scalar_value).
+//!
+//! ### Sequences
+//!
+//! Variable length heterogeneous sequences are encoded as a single byte length followed by the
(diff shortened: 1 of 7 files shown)


kcsongor:
Why is this only for terra? Can't we use it in solana too?

terra/packages/vaa_payload/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+//! the wire format for a tuple with the same fields in the same order. The only exception is unit
+//! structs (structs with no fields), which are not represented in the wire format at all.
+//!
+//! ### `[T; N]`
+//!
+//! Arrays are treated as tuples with homogenous fields and have the same wire format.
+//!
+//! ### Enums
+//!
+//! Enums are encoded as a single byte identifying the variant followed by the serialization of the
```
kcsongor, line 73: What's the rationale for this encoding? Did you find any VAA types that encode enums in this way? The only enum-like thing that comes to mind is the choice between the different types of payloads, such as "token bridge contract upgrade" and "nft bridge transfer", but discriminating between these is not done by an encoded enum tag, or any other identifier. In fact, there's no explicit discriminator, and I'm not even sure if the VAA grammar used just by the token bridge itself is unambiguous. Even if it's unambiguous, an arbitrary VAA certainly can't be parsed into an appropriate payload type without backtracking.

kcsongor: Actually, I take that back. The token bridge and NFT bridge payloads are encoded in this way.

jynnantonix:
> Why is this only for terra? Can't we use it in solana too?

No particular reason.  I just hadn't looked at the solana code at all and so didn't think about making this more general purpose.  I'll move it into the sdk directory.

jynnantonix:
I think I'm just going to close this.  The goal was to avoid having to manually write parsing code for the various VAA messages we already have and I'm not sure about the benefit of moving it into the sdk.  People building on top of wormhole should use a proper schema for their payloads rather than the hand rolled one we're using for VAAs.

jynnantonix:
I've moved the crate into the sdk/rust directory and also added support for strings and maps.

jynnantonix:
> could we get the byte_utils.rs from terra/contracts/wormhole/src added into here as well? I use state.rs and byte_utils.rs in every single rust contract I build...

The purpose of this crate is to be a replacement for byte_utils.rs so I don't think moving it here helps us with that goal.

sdk/rust/vaa_payload/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+//! Variable length heterogeneous sequences are encoded as a single byte length followed by the
+//! concatenation of the serialized form of each element in the sequence. Note that this means that
+//! sequences cannot have more than 255 elements. Additionally, during serialization the length must
+//! be known ahead of time.
+//!
+//! ### Byte arrays - `&[u8]`, `Vec<u8>`, and `Cow<'a, [u8]>`
+//!
+//! Byte arrays are treated as a subset of variable-length sequences and are encoded as a single
+//! byte length followed by that many bytes of data. Again, since the length of the byte array has
+//! to fit in a single byte it cannot be longer than 255 bytes.
```
hendrikhofstadt, line 49: fields like the VAA payload don't have a length field and can have unlimited length (till the end of the packet)

sdk/rust/vaa_payload/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        signatures: Cow<'s, [Signature]>,
+        timestamp: u32, // Seconds since UNIX epoch
+        nonce: u32,
+        emitter_chain: u16,
+        emitter_address: [u8; 32],
+        sequence: u64,
+        consistency_level: u8,
+        #[serde(borrow)]
+        payload: GovernancePacket<'p>,
+        map: BTreeMap<u32, u32>,
```
hendrikhofstadt, line 274: not part of the spec (https://github.com/wormhole-foundation/wormhole/blob/dev.v2/whitepapers/0001_generic_message_passing.md)

hendrikhofstadt (changes requested):

sdk/rust/vaa_payload/src/lib.rs
hendrikhofstadt: That sounds good

sdk/rust/vaa_payload/src/lib.rs
hendrikhofstadt: I missed that this is in tests only.

hendrikhofstadt (approved):

jynnantonix:
Superseeded by #1919

PR #1003, opened 2022-03-15
hack/repair_solana: improve sequence guessing

leoluk:
Increases tolerance for out-of-order txs when trying to guess the
right offset to search for the transaction.

Obviously, this is truly a hack and the proper solution is to rely
on a service which indexes the logs... except we don't have that yet.


SEJeff (approved):

node/hack/repair_solana/repair.go
```diff
@@ -128,16 +128,24 @@ func main() {
...
 
-			last := txs[len(txs)-1].Signature
-
-			_, lastSeq, err := fetchTxSeq(ctx, sr, last)
-			if err != nil {
-				log.Fatalf("fetch last tx seq: %v", err)
+			var lastSeq, firstSeq uint64
+
+			var last solana.Signature
+			for i := 0; lastSeq == 0; i-- {
```
kcsongor, line 134: it's not entirely clear to me what this loop has to do with the offset change. Is this just an unrelated improvement to make sure we find the last tx that has a sequence number in it?

node/hack/repair_solana/repair.go
```diff
@@ -172,8 +183,8 @@ func main() {
 							continue
 						}
 						if nseq != p.Sequence {
-							log.Printf("%d != %d, offset +%d", nseq, p.Sequence, nseq-p.Sequence)
-							offset += nseq - p.Sequence
+							offset += 1
```
kcsongor, line 186: what does this do?

kcsongor (approved):

bruce-riley (approved):

node/hack/repair_solana/repair.go
leoluk: Yep!

node/hack/repair_solana/repair.go
leoluk: It advances the offset by one if we landed on the wrong seq no - it's just a dumb brute force way of narrowing down the set of txs to request (for lack of an index we could query)

PR #1006, opened 2022-03-16
Add more comprehensive structs unit-tests

claudijd:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1006)

### DIFF
--- a/node/pkg/vaa/structs_test.go
+++ b/node/pkg/vaa/structs_test.go
@@ -13,6 +13,26 @@ import (
 	"time"
 )
 
+func getVAA() *VAA {
+	var payload = []byte{97, 97, 97, 97, 97, 97}
+	var governanceEmitter = Address{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4}
+
+	vaa := &VAA{
+		Version:          uint8(1),
+		GuardianSetIndex: uint32(1),
+		Signatures:       nil,
+		Timestamp:        time.Unix(0, 0),
+		Nonce:            uint32(1),
+		Sequence:         uint64(1),
+		ConsistencyLevel: uint8(32),
+		EmitterChain:     ChainIDSolana,
+		EmitterAddress:   governanceEmitter,
+		Payload:          payload,
+	}
+
+	return vaa
+}
+
 func TestChainIDFromString(t *testing.T) {
 	type test struct {
 		input  string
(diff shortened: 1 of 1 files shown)


SEJeff (approved):
This is really good with 1 small caveat. You copy / paste `payload`, `governanceEmitter`, and `vaa` multiple times.

Perhaps make something vaguely like this?

```
func getVAA() *VAA {
	var payload = []byte{97, 97, 97, 97, 97, 97}
	var governanceEmitter = Address{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4}

	return &VAA{
		Version:          uint8(1),
		GuardianSetIndex: uint32(1),
		Signatures:       nil,
		Timestamp:        time.Unix(0, 0),
		Nonce:            uint32(1),
		Sequence:         uint64(1),
		ConsistencyLevel: uint8(32),
		EmitterChain:     ChainIDSolana,
		EmitterAddress:   governanceEmitter,
		Payload:          payload,
	}
```

That gives you a new copy each time and you don't need to worry about mutating the struct. Then your tests simply do:

```
vaa := getVAA()
...
```

That is also a light enough abstraction it doesn't impede readability.

node/pkg/vaa/structs_test.go
```diff
@@ -123,3 +129,179 @@ func TestChainId_String(t *testing.T) {
...
+	h := vaa.SigningMsg()
+	pubKey, err := crypto.Ecrecover(h.Bytes(), vaa.Signatures[0].Signature[:])
+	assert.Nil(t, err)
+	assert.NotNil(t, pubKey)
+
+	// Translate that public key back to an address
+	addr := common.BytesToAddress(crypto.Keccak256(pubKey[1:])[12:])
+
+	// Make sure that it verifies
+	assert.True(t, vaa.VerifySignatures([]common.Address{addr}))
```
SEJeff, line 299: Also this tests all of the positive cases, what what about the negative cases where there are errors? We probably want to cover those with a followup to this later on. Error code paths are always the least often used and most likely to break things.

claudijd: I have another PR that goes into much greater depth on this topic.  This PR is largely just to establish a base from which to build an extensive suite of tests on top.

claudijd:
@SEJeff great suggestions, used the abstraction technique and I will pile on to this pattern with a more extensive verifySignatures test suite.

bruce-riley (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1006#-:-MzL2yt0B2S6Ww9IrDD2:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @SEJeff)

SEJeff (approved):

PR #1026, opened 2022-03-23
Add unit-tests for registry

claudijd:


node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+	registry := NewRegistry()
+	assert.Equal(t, "", registry.guardianAddress)
+
+	registry.SetGuardianAddress("foo")
+	assert.Equal(t, "foo", registry.guardianAddress)
+}
+
+func TestSetNetworkStats(t *testing.T) {
+	registry := NewRegistry()
+	assert.Equal(t, 0, len(registry.networkStats))
```
SEJeff, line 30: Don't test the same code twice. You covered that in `TestNewRegistry`.

```suggestion
```

node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	pub, _, _ := ed25519.GenerateKey(rand.Reader)
+	contractAddr := base58.Encode(pub[:])
+
+	heart_beat := &gossipv1.Heartbeat_Network{
```
SEJeff, line 35: As much as the pythonista in me hates it... It is idiomatic to [use mixedCase for variables names](https://go.dev/doc/effective_go#mixed-caps).

Writing idiomatic code for the language you're using makes maintenance easier down the road and especially for new developers added to the project.

node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+		ContractAddress: contractAddr,
+	}
+
+	registry.SetNetworkStats(vaa.ChainIDEthereum, heart_beat)
+	assert.Equal(t, 1, len(registry.networkStats))
```
SEJeff, line 40: The chain id should be deterministic, right? If so, it would be better to assert the actual contents of `registry.networkStats` instead of the length. Asserting the value by design also asserts the length :)

node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func TestAddErrorCount(t *testing.T) {
+	registry := NewRegistry()
+	assert.Equal(t, 0, len(registry.errorCounters))
```
SEJeff, line 45: Already covered in `TestNewRegistry`
```suggestion
```

node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	registry.AddErrorCount(vaa.ChainIDEthereum, uint64(1))
+	assert.Equal(t, 1, len(registry.errorCounters))
```
SEJeff, line 48: Strive to assert values for correctness, not just the length. That way if the value is wrong, the tests will fail. Can you just create a variable named something like `expected` and assert it directly equal to `registry.errorCounters`?

node/pkg/p2p/registry_test.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func TestGetErrorCount(t *testing.T) {
+	registry := NewRegistry()
+	assert.Equal(t, 0, len(registry.errorCounters))
```
SEJeff, line 53: I understand you're doing this for completeness, but it isn't necessary as it is covered elsewhere. These add up over time and cause test harnesses to run slow.
```suggestion
```

SEJeff (approved):
Minor nits, great work sir!

node/pkg/p2p/registry_test.go
claudijd: 100%

node/pkg/p2p/registry_test.go
claudijd: generally, this behavior I'm exhibiting here is coming from a mindset where parallel test execution is possible and state pollution could occur.  I also like the idea of seeing the state change in a give test, as a guarantee I'm getting the same starting point.  That said, I'll pull those, just noting why my thinking behind it.

claudijd:
@SEJeff should have all the feedback sorted.  thanks again!

SEJeff (approved):

bruce-riley (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1026)* status: all files reviewed, 6 unresolved discussions (waiting on @claudijd and @SEJeff)

bruce-riley (approved):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1026)* status: all files reviewed, 6 unresolved discussions (waiting on @claudijd and @SEJeff)

PR #1027, opened 2022-03-23
Shutdown Switch Solidity Code

bruce-riley:
Change-Id: Ibe029c221b9315e3960335697d7b76b0ed9fa2e1

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1027)


ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // Called by the voter's client to update their votes.
+    function castShutdownVote(uint16 _chainId, bool _enabled) public {
+        require(_chainId == getChainId(), "invalid chain id");
+        require(isRegisteredVoter(msg.sender), "you are not a registered voter");
+
+        // Our set of voters is the set of guardians. If the guardian set has changed, rebuild our set of voters, setting all votes to enabled.
+        uint32 gsi = getWH().getCurrentGuardianSetIndex();
+        if ((guardianSetIndex != gsi) || (! voterSetInitialized)) {
```
kcsongor, line 62: the formatting here is a little funny, `(guardianSetIndex != gsi || !voterSetInitialized)` wold look more natural

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+            voterSetInitialized = true;
+            Structs.GuardianSet memory gs = getWH().getGuardianSet(gsi);
+
+            uint256 quorum = ((gs.keys.length * 10 / 3) * 2) / 10 + 1;
+            requiredNoVotes = uint16(gs.keys.length - quorum + 1);
```
kcsongor, line 67: so `requiredNoVotes = gs.keys.length / 3` ?

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+        return enabled;
+    }
+
+    event ShutdownVoteCast(address indexed voter, bool votedToEnable, uint16 numVotesToDisable, bool enableFlag);
+    event ShutdownStatusChanged(bool enabledFlag, uint16 numVotesToDisable);
+
+    function getWH() public virtual view returns (IWormhole);
+    function getChainId() public virtual view returns (uint16);
+
+    uint16 constant REQUIRED_NO_VOTES = 1;
```
kcsongor, line 33: this doesn't seem to be used anywhere

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+        // Our set of voters is the set of guardians. If the guardian set has changed, rebuild our set of voters, setting all votes to enabled.
+        uint32 gsi = getWH().getCurrentGuardianSetIndex();
+        if ((guardianSetIndex != gsi) || (! voterSetInitialized)) {
+            voterSetInitialized = true;
+            Structs.GuardianSet memory gs = getWH().getGuardianSet(gsi);
+
+            uint256 quorum = ((gs.keys.length * 10 / 3) * 2) / 10 + 1;
+            requiredNoVotes = uint16(gs.keys.length - quorum + 1);
+
+            for (uint idx = 0; (idx < voters.length); idx++) {
```
kcsongor, line 69: can we pull out the initialisation code to a separate internal function called `initializeShutdownSwitch`? it would help readability

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    event ShutdownStatusChanged(bool enabledFlag, uint16 numVotesToDisable);
+
+    function getWH() public virtual view returns (IWormhole);
+    function getChainId() public virtual view returns (uint16);
+
+    uint16 constant REQUIRED_NO_VOTES = 1;
+
+    ShutdownStructs.VoterEntry[20] voters;
+    uint32 guardianSetIndex = 0;
+    uint16 votesToDisable = 0;
```
kcsongor, line 37: what exactly is the purpose of the `votesToDisable` variable? it's not clear

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    event ShutdownVoteCast(address indexed voter, bool votedToEnable, uint16 numVotesToDisable, bool enableFlag);
+    event ShutdownStatusChanged(bool enabledFlag, uint16 numVotesToDisable);
+
+    function getWH() public virtual view returns (IWormhole);
+    function getChainId() public virtual view returns (uint16);
+
+    uint16 constant REQUIRED_NO_VOTES = 1;
+
+    ShutdownStructs.VoterEntry[20] voters;
```
kcsongor, line 35: Reading through the code, there's a lot of looping around this array, it looks a little error-prone. How about we represent this instead as
```solidity
mapping(address => bool) voters;
```
then for counting up the votes, we just iterate through the keys in the current guardian set and see what the guardian is set to.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+
+            guardianSetIndex = gsi;
+            votesToDisable = 0;
+            enabled = true;
+        }
+
+        bool somethingChanged = false;
+        uint16 numDisabled = 0;
```
kcsongor, line 85: with the mapping, this code would be much simpler:

```solidity
bool previousVote = votes[msg.sender];
votes[msg.sender] = _enabled;

if (perviousVote != _enabled) {
  emit ShutdownVoteCast ...
}
```

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+import "./interfaces/IWormhole.sol";
+
+contract ShutdownStructs {
+    struct VoterEntry {
+        address key;
+        bool enabled;
+    }
+}
+
+abstract contract ShutdownSwitch {
```
kcsongor, line 14: It would be nice to have an overview comment of what this class does. Something like https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol#L6-L21

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    function enabledFlag() public view returns (bool){
+        return enabled;
+    }
+
+    event ShutdownVoteCast(address indexed voter, bool votedToEnable, uint16 numVotesToDisable, bool enableFlag);
+    event ShutdownStatusChanged(bool enabledFlag, uint16 numVotesToDisable);
+
+    function getWH() public virtual view returns (IWormhole);
```
kcsongor, line 30: I think this leaks out too much internal details. In principle, the emergency shutdown class isn't tied to wormhole at all, so instead of asking for the whole `IWormhole`, we could just have a
```solidity
function eligibleVoters() public virtual view returns address[];
```
which could be implemented by returning the current guardian set.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+                numDisabled++;
+            }
+        }
+
+        bool newEnabledFlag = (numDisabled < requiredNoVotes);
+        if (somethingChanged) {
+            emit ShutdownVoteCast(msg.sender, _enabled, numDisabled, newEnabledFlag);
+        }
+
+        if (votesToDisable != numDisabled) {
```
kcsongor, line 108: I'm not sure what `votesToDisable` means, but this conditional is unnecessary, and can just be replaced by
```solidity
votesToDisable = numDisabled
```

ethereum/contracts/nft/NFTBridgeSetters.sol
```diff
@@ -3,9 +3,11 @@
 
 pragma solidity ^0.8.0;
 
+import "../ShutdownSwitch.sol";
+
 import "./NFTBridgeState.sol";
 
-contract NFTBridgeSetters is NFTBridgeState {
+contract NFTBridgeSetters is NFTBridgeState, ShutdownSwitch {
```
kcsongor, line 10: it's a nitpick, but semantically I think this class should be extended by the `NFTBridgeGovernance` class instead (same for `BridgeGovernance`)

kcsongor:
requiring that the guardians sign these transactions with their guardian private key is convenient, but feels dangerous (need a hot wallet with the same key they use to sign wormhole observations with) and non-portable (the guardian keys are secp256k1, so are ethereum keys, but IIRC solana uses ed25519)

ethereum/contracts/ShutdownSwitch.sol
claudijd: So it `eligibleVoters()` would be something to this effect?

```
Structs.GuardianSet memory guardianSet = getGuardianSet(getCurrentGuardianSetIndex());
return guardianSet.keys
```

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    event ShutdownVoteCast(address indexed voter, bool votedToEnable, uint16 numVotesToDisable, bool enableFlag);
+    event ShutdownStatusChanged(bool enabledFlag, uint16 numVotesToDisable);
+
+    function getWH() public virtual view returns (IWormhole);
+    function getChainId() public virtual view returns (uint16);
+
+    uint16 constant REQUIRED_NO_VOTES = 1;
+
+    ShutdownStructs.VoterEntry[20] voters;
+    uint32 guardianSetIndex = 0;
```
claudijd, line 36: We shouldn't need to hard-code this, we should be able to obtain this from a pre-existing getter method for https://github.com/certusone/wormhole/blob/dev.v2/ethereum/contracts/Getters.sol#L13-L15

claudijd:
General comment: I think we should be very explicit in the comments about what various things mean.  The most grok-able this is, the better chance we'll have at not missing something in our various stages of review.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+        require(_chainId == getChainId(), "invalid chain id");
+        require(isRegisteredVoter(msg.sender), "you are not a registered voter");
+
+        // Our set of voters is the set of guardians. If the guardian set has changed, rebuild our set of voters, setting all votes to enabled.
+        uint32 gsi = getWH().getCurrentGuardianSetIndex();
+        if ((guardianSetIndex != gsi) || (! voterSetInitialized)) {
+            voterSetInitialized = true;
+            Structs.GuardianSet memory gs = getWH().getGuardianSet(gsi);
+
+            uint256 quorum = ((gs.keys.length * 10 / 3) * 2) / 10 + 1;
```
claudijd, line 66: Could we move this out into Getters.sol?

ethereum/contracts/ShutdownSwitch.sol
claudijd: I think the rationale here is that this would be less arithmetic, but I think largely we should keep the existing logic, but move the quorum logic out into Getters.sol.  This way, if we define quorum differently this will apply globally.  If follow that path, I think the simple division here wouldn't fly, as it makes assumptions about quorum never changing.

ethereum/contracts/ShutdownSwitch.sol
kcsongor: Yes, that's what I had in mind.

bruce-riley:
There has been a lot of good feedback on this PR that will result in significant rework, so I am closing this PR and will issue a new one. Thanks.

kcsongor:
@bruce-riley I think the best thing to do would be to just add new changes to this PR, since then we can keep all the historical discussion. How do you feel about that?

ethereum/contracts/ShutdownSwitch.sol
kcsongor: My rationale other than less arithmetic is that it's easier to understand. Is it important that this threshold is a function of the quorum?

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+            requiredNoVotes = uint16(gs.keys.length - quorum + 1);
+
+            for (uint idx = 0; (idx < voters.length); idx++) {
+                if (idx < gs.keys.length) {
+                    voters[idx].key = gs.keys[idx];
+                    voters[idx].enabled = true;
+                } else {
+                    voters[idx].key = address(0);
+                    voters[idx].enabled = false;
+                }
```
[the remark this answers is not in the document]
kcsongor: Ideally the intent should be apparent from the code (or documented)

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    modifier isEnabled() {
+        require(enabledFlag(), "transfers are temporarily disabled");
+        _;
+    }
+
+    // Called by the voter's client to update their votes.
+    function castShutdownVote(uint16 _chainId, bool _enabled) public {
+        require(_chainId == getChainId(), "invalid chain id");
```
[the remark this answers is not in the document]
bruce-riley: We thought this was necessary to prevent replay attacks but now we know better. Removed it.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    function setUpShutdownSwitch() internal {
+        guardianSetIndex = 0;
+        votesToDisable = 0;
+        requiredNoVotes = 0;
+        voterSetInitialized = false;
+        enabled = true;
+    }
+
+    modifier isEnabled() {
```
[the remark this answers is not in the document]
bruce-riley: Removed them.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,129 @@
...
+    function getChainId() public virtual view returns (uint16);
+
+    uint16 constant REQUIRED_NO_VOTES = 1;
+
+    ShutdownStructs.VoterEntry[20] voters;
+    uint32 guardianSetIndex = 0;
+    uint16 votesToDisable = 0;
+    uint16 requiredNoVotes = 0;
+    bool voterSetInitialized = false;
+    bool enabled = true;
```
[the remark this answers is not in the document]
bruce-riley: Made them private.

ethereum/contracts/ShutdownSwitch.sol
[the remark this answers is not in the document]
claudijd: we went through it, it's going to be refactored.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: This has been deleted.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: After further discussions, we are now using a static required number of votes of three, so this calculation has been deleted.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Changed it to use a mapping.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Did this.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Well, now we are back to using a hard coded value of three. . .

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Code has been deleted.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Code has been deleted.

ethereum/contracts/ShutdownSwitch.sol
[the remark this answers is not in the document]
bruce-riley: The code has been completely reworked.

ethereum/contracts/nft/NFTBridgeSetters.sol
bruce-riley: The problem is that BridgeSetup::setup() needs access to the setUp method in ShutdownSwitch, and BridgeSetUp inherits from BridgeSetters, not BridgeGovernance. BridgeSetters is the common root of both Bridge and BridgeSetup.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Covered in the rework.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: It's gone now.

bruce-riley (commented):
As discussed in the sync up today, it is desirable to keep this feature as simple as possible, to increase the probability of it working when necessary, so this approach is okay+

> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1027#-:-MywQDnY3FOp-5VB6HPh:b-lttc5q)* status: 0 of 10 files reviewed, 17 unresolved discussions (waiting on @calebrate, @claudijd, @evan-gray, @hendrikhofstadt, @justinschuldt, and @kcsongor)

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,134 @@
...
+ * @dev Abstract contract module that provides a bridge contract with the ability to temporarily suspend transfers.
+ *
+ * Inheriting from `ShutdownSwitch` will allow a bridge contract to use the isEnabled modifier to prevent transactions
+ * when the shutdown switch is engaged via a guardian voting mechanism.
+ *
+ * This contract provides a public function that allows the guardians to vote to disable / enable transfers.
+ * When a predetermined number of guardians vote to disable, then transfers are blocked until the number of
+ * disable votes is reduced below that threshold.
+ *
+ * NOTE: The number of votes to enable is irrelevant. As long as there are enough votes to disable, trabsfers are blocked.
```
claudijd, line 15: nit: typo "trabsfers"

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * This contract provides public getters to query the shutdown status. It also emits events when valid votes are cast,
+ * or the shutdown status changes.
+ */
+
+import "./interfaces/IWormhole.sol";
+
+abstract contract ShutdownSwitch {
+    /// @dev A map of all guardians that have active votes to disable. If a guardian is removed from the guardian set while they
+    /// have an active vote to disable, they would get left in the map. The current assumption is that this will have minimal impact.
```
claudijd, line 35: Suggestion: "The current assumption is that this will have minimal impact because we will never reference that old guardian again in future passes"

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,471 @@
...
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
+
+        await initialized.methods.castShutdownVote(false).send({
+            value: 0,
+            from: accounts[1],
+            gasLimit: 2000000
+        });
+
+        assert.equal((await initialized.methods.numVotesToDisable().call()), 2)
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
```
claudijd, line 348: I like how you are checking each step of the way here, on the state transitions.

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,471 @@
...
+        // "no quorum", but that's okay for this test, since that check happens after the enable check.
+        try {
+            await initialized.methods.completeTransfer("0x" + vaa).send({
+                value: 0,
+                from: accounts[0],
+                gasLimit: 2000000
+            })
+        } catch (error) {
+            assert.equal(error.message, "Returned error: VM Exception while processing transaction: revert no quorum")
+        }
```
claudijd, line 329: This is mostly just a comment of opportunity/optimization... if you took a similar strategy as you did with `clearAllVotes()`, you might be able to abstract this out something like `testTransfer(bool result, err)`.  Doing so might allow you to verify the transfer status as your votes ramp up.

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,471 @@
...
+            })
+        } catch (error) {
+            assert.equal(error.message, "Returned error: VM Exception while processing transaction: revert transfers are temporarily disabled")
+        }
+    })
+
+    it("a vote changing back to enabled should allow redeems again", async function () {
+        const accounts = await web3.eth.getAccounts();
+        const initialized = new web3.eth.Contract(BridgeImplementationFullABI, TokenBridge.address);
+        // Don't clear the votes here.
```
claudijd, line 375: in the interest in having tests not depend on each other's state, I'd almost suggest a call to `clearAllVotes()` and re-add (but without all the interim asserts from above).  This is mainly to try and avoid an unexpected state change in one of the other tests having a negative impact on this test.

claudijd: I had something like this bite me in an environment where you weren't exactly sure the order in which tests would execute, so you could get emergent pollution across tests.  My sense is that we're serial in this context atm.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Fixed

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            disabledVotes[msg.sender] = true;
+        }
+
+        // Update the number of votes to disable.
+        uint16 votesToDisable = computeNumVotesDisabled(gs);
+
+        // Determine the new shutdown status and generate the appropriate events.
+        bool newEnabledFlag = (votesToDisable < computeRequiredVotesToDisable(gs.keys.length));
+        emit ShutdownSwitch.ShutdownVoteCast(msg.sender, _enabled, votesToDisable, newEnabledFlag);
```
kcsongor, line 98: So we emit an event even when the voter votes twice in the same direction. Is that intentional?

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,134 @@
...
+            if (voter == gs.keys[idx]) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /// @dev Returns the number of disable votes required to go disabled.
+    function computeRequiredVotesToDisable(uint numGuardians) private pure returns(uint16) {
```
kcsongor, line 118: can we call this `requiredVotesToShutdown` or something like that? 'disable' is a little confusing to me, because it's not immediately clear if it disables the shutdown, or transfers (I know it's the latter after reading the code)

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,134 @@
...
+        // If the number of active guardians is less than the pre-configured threshold, use that.
+        return uint16(numGuardians >= REQUIRED_NO_VOTES ? REQUIRED_NO_VOTES : numGuardians);
+    }
+
+    /// @dev Counts up the number of current guardians that have active votes to disable.
+    function computeNumVotesDisabled(Structs.GuardianSet memory gs) private view returns (uint16) {
```
kcsongor, line 124: I would just call this `getShutdownVoteCount`

ethereum/contracts/ShutdownSwitch.sol
claudijd: I agree, I think this would read cleaner.

ethereum/contracts/ShutdownSwitch.sol
claudijd: same, I think it's more direct.

ethereum/contracts/ShutdownSwitch.sol
kcsongor: It's not a concern, but it will emit repeated events, and just wanted to check if that's OK

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Added that detail

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,471 @@
...
+    })
+
+    async function clearAllVotes(initialized, accounts) {
+        for (let idx = 0; (idx < 4); ++idx) {
+            await initialized.methods.castShutdownVote(true).send({
+                value: 0,
+                from: accounts[idx],
+                gasLimit: 2000000
+            });
+        }
```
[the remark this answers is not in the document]
bruce-riley: Done.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Before the rework, I only generated the event when something changed, but as I reworked it, I decided to always generate the event when a valid voter casts a vote, even if it doesn't change anything. It might be worthwhile to know when guardians are voting. . .

bruce-riley: So if there are no objections, I am going to leave it like it is.

ethereum/test/shutdownswitch.js
bruce-riley: Yeah, the tests have to be serial, or at least not execute in parallel, or they will not work. Also, I put the checks at the start of this test to verify that the previous test worked before carrying on with this test. So unless you feel strongly about it, I would like to leave it as is.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Renamed things to use Shutdown rather than Disable.

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Rename it.

ethereum/test/shutdownswitch.js
bruce-riley: Leaving this as is for now, unless you object strongly.

ethereum/test/shutdownswitch.js
claudijd: seems fine, more just seeing this for the first time, so please forgive my noob questions.

ethereum/test/shutdownswitch.js
claudijd: no strong objection.

bruce-riley:
Code has been modified to use the auth proof rather than the message sender to validate voters. Would appreciate it if you gave it another look.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,181 @@
...
+    mapping(address => bool) private shutdownVotes;
+
+    bool private enabled = true;
+
+    /// @dev Returns the number of votes required to disable transfers.
+    function requiredVotesToShutdown() public view returns (uint16) {
+        return computeRequiredVotesToShutdown(getCurrentGuardianSet().keys.length);
+    }
+
+    /// @dev Returns the current number of votes to disable transfers.
```
claudijd, line 58: The comments between 53 and 58 feel a bit too similar.  I wonder if we could differentiate a bit more between the to to drive clarity?

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,181 @@
...
+        // The autoProof is made up as follows:
+        //    r: bytes32
+        //    s: bytes32
+        //    v: uint8
+
+        require((authProof.length == 65), "invalid auth proof");
+
+        bytes32 r = authProof.toBytes32(0);
+		bytes32 s = authProof.toBytes32(32);
+		uint8 v = authProof.toUint8(64) + 27; // Adding 27 is required, see here for details: https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739
```
claudijd, line 147: Formatting got a bit wonky here.

ethereum/contracts/nft/NFTBridge.sol
```diff
@@ -97,7 +97,7 @@ contract NFTBridge is NFTBridgeGovernance {
         }(nonce, encoded, 15);
     }
 
-    function completeTransfer(bytes memory encodedVm) public {
+    function completeTransfer(bytes memory encodedVm) public isEnabled {
```
claudijd, line 100: Odd trailing spaces here.

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,518 @@
...
+            value: 0,
+            from: accounts[0],
+            gasLimit: 2000000
+        });
+
+        assert.equal((await initialized.methods.numVotesToShutdown().call()), 1)
+        assert.equal((await initialized.methods.requiredVotesToShutdown().call()), 3)
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
+
+        // This first vote should succeed, but we should still be enabled.
```
claudijd, line 122: I think this is your second vote, though the comments suggest the first.

claudijd (dismissed):

bruce-riley:
Changes to tests did not get committed. Sorry about that!

ethereum/contracts/ShutdownSwitch.sol
bruce-riley: Added more detail.

ethereum/contracts/nft/NFTBridge.sol
bruce-riley: Removed.

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        if (newEnabledFlag != enabled) {
+            enabled = newEnabledFlag;
+            emit ShutdownSwitch.ShutdownStatusChanged(newEnabledFlag, votesToShutdown);
+        }
+    }
+    
+    /// @dev Extract the guardian key from the authProof.
+    function decodeVoter(address sender, bytes memory authProof) public pure returns (address) {
+        // The autoProof is made up as follows:
```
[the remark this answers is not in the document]
bruce-riley: Fixed.

bruce-riley:
During the sync meeting on 4/18/2022, we discussed the make up of the auth proof. A concern was raised that the auth proof never expires, so a disgruntle employee could use it. However, casting the vote requires access to a specific wallet, and the guardian firm would already have to do something to secure that wallet, independent of shutdown switch. If they didn't, the worst thing that could happen is that a single vote could be cast to shutdown, which would require the guardian to change their guardian private key. It was agreed that this was okay. (The alternative was to specify a "good until" block in the digest, and the sol code would compare that to the current block before deciding if a vote was valid. That could be added in the future if necessary, but there are advantages to the proofs not expiring, since the guardians could have them pre-generated to be used during a crisis.)

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -0,0 +1,181 @@
...
+    function numVotesToShutdown() public view returns (uint16) {
+        return computeNumVotesShutdown(getCurrentGuardianSet());
+    }
+
+    /// @dev returns the current shutdown status, where true means transfers are enabled.
+    function enabledFlag() public view returns (bool) {
+        return enabled;
+    }
+
+    /// @dev Returns the number of votes required to disable transfers.
```
kcsongor, line 68: Looks like this returns the guardian keys that have voted to shut down

kcsongor (approved):

justinschuldt (dismissed):

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,528 @@
...
+    
+    //////////////// NOTE: Every test after this assumes four guardians! ///////////////////////////////
+    
+    it("should start out with transfers enabled", async function () {
+        const initialized = new web3.eth.Contract(BridgeImplementationFullABI, TokenBridge.address);
+
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
+        assert.equal((await initialized.methods.numVotesToShutdown().call()), 0)
+
+        // Since we start out with only one guardian, the required votes should be one.
```
dsterioti, line 111: Nit-picky, but is this comment outdated?

ethereum/test/shutdownswitch.js
```diff
@@ -0,0 +1,528 @@
...
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
+
+        // This first vote should succeed, but we should still be enabled.
+        await initialized.methods.castShutdownVote(GuardianAuthProof0).send({ from: EthPublicKey });
+
+        assert.equal((await initialized.methods.numVotesToShutdown().call()), 1)
+        assert.equal((await initialized.methods.requiredVotesToShutdown().call()), 3)
+        assert.equal((await initialized.methods.enabledFlag().call()), true)
+
+        assert.equal((await initialized.methods.numVotesToShutdown().call()), 1)
```
dsterioti, line 165: Another nit-picky comment, this looks like a copy and paste error. I also see this on line 261 and 369.

dsterioti (approved):

claudijd (dismissed):

justinschuldt (approved):
> Reviewed 4 of 10 files at r1, 3 of 6 files at r2, 1 of 2 files at r6, 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1027)* status: all files reviewed, 35 unresolved discussions (waiting on @bruce-riley, @calebrate, @claudijd, @evan-gray, @hendrikhofstadt, and @kcsongor)

claudijd (dismissed):

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -3,6 +3,8 @@
 pragma solidity ^0.8.0;
 pragma experimental ABIEncoderV2;
 
+import "@openzeppelin/contracts/utils/StorageSlot.sol";
```
kcsongor, line 6: Remove this please, redundant now

ethereum/contracts/ShutdownSwitch.sol
```diff
@@ -179,11 +174,54 @@ abstract contract ShutdownSwitch {
...
                 votesToShutdown++;
             }
         }
 
         return votesToShutdown;
     }
+    
+    /// @dev Since other contracts will inherit from this one, it cannot have any of its own local state.
+    /// This is because that would affect the layout of local state in the inheriting contracts. To prevent this,
+    /// we use the storage slot library to put our state in deterministic storage.
```
kcsongor, line 187: we're not using the storage slot library here, just doing the inline assembly directly

kcsongor (approved):
🚀

bruce-riley:
This PR is being closed because a decision has been made to implement the shutdown switch in a different manner.

PR #1028, opened 2022-03-24
Run ethereum tests on CI

kcsongor:
The tests currently fail. Ideally we could fix those tests and merge this in, so CI runs these tests for each PR.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1028)

### DIFF
--- a/ethereum/Makefile
+++ b/ethereum/Makefile
@@ -26,7 +26,4 @@ test: build .env dependencies
 	pkill ganache-cli
 
 clean:
-	rm -f ganache.log
-	rm -f .env
-	rm -rf node_modules
-	rm -rf build
+	rm -rf ganache.log .env node_modules build
--- a/ethereum/Makefile
+++ b/ethereum/Makefile
@@ -1,0 +1,32 @@
+SOURCE_FILES:=$(shell find contracts -name "*.sol")
+
+.PHONY: dependencies test clean all
+
+all: build
+
+node_modules: package-lock.json
+	touch -m node_modules
+	npm ci
+
+dependencies: node_modules
+
+build: node_modules ${SOURCE_FILES}
+	mkdir -p build
+	touch -m build
+	npm run build
+
+.env: .env.test
+	cp $< $@
+
+test: build .env dependencies
+	if pgrep ganache-cli; then pkill ganache-cli; fi
+	npx ganache-cli > ganache.log &
+	sleep 5
+	npx truffle test || (pkill ganache-cli && exit 1)
+	pkill ganache-cli
+
+clean:
+	rm -f ganache.log
+	rm -f .env
+	rm -rf node_modules
+	rm -rf build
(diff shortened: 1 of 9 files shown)


ethereum/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	npx ganache-cli > ganache.log &
+	sleep 5
+	npx truffle test || (pkill ganache-cli && exit 1)
+	pkill ganache-cli
+
+clean:
+	rm -f ganache.log
+	rm -f .env
+	rm -rf node_modules
+	rm -rf build
```
SEJeff, line 32:
```suggestion
	rm -rf ganache.log .env node_modules build
```

ethereum/Makefile
```diff
@@ -0,0 +1,32 @@
+SOURCE_FILES:=$(shell find contracts -name "*.sol")
+
+.PHONY: dependencies test clean all
+
+all: build
+
+node_modules: package-lock.json
+	touch -m node_modules
```
SEJeff, line 8: Is it possible for this to run before the `node_modules` directory exists? If so, it will create a file, which I'm assuming `npm ci` will not be terribly happy about.

SEJeff (commented):
I like the idea, but have a few questions.

ethereum/Makefile
kcsongor: It is very much possible, but npm actually removes the file in that case. Not the cleanest though, should we `mkdir -p node_modules` first?

evan-gray (changes requested):
will review further, but is this worth spinning up an additional runner when it could be added to tilt?

kcsongor:
> will review further, but is this worth spinning up an additional runner when it could be added to tilt?

two things about this:
1. I don't know how to run the tests with tilt

2. I like the fact that these tests work without tilt

Reason 2. of course doesn't mean that they should not be added to tilt, but I'd like to keep them runnable outside of tilt if possible

bruce-riley (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1028)* status: all files reviewed, 3 unresolved discussions (waiting on @dsterioti, @kcsongor, and @SEJeff)

ethereum/Makefile
SEJeff: That would be better than `touch node_modules`, yes.

evan-gray:
> > will review further, but is this worth spinning up an additional runner when it could be added to tilt?
> 
> two things about this:
> 
> 1. I don't know how to run the tests with tilt
> 2. I like the fact that these tests work without tilt
> 
> Reason 2. of course doesn't mean that they should not be added to tilt, but I'd like to keep them runnable outside of tilt if possible

Sounds like we might try to keep the other unit test like things outside, so we can leave this for now

evan-gray (approved):

evan-gray (commented):
> Dismissed @SEJeff and @kcsongor from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1028)* status: 2 of 9 files reviewed, all discussions resolved (waiting on @bruce-riley and @dsterioti)

SEJeff (approved):

PR #1033, opened 2022-03-28
Never upgrade to an empty guardianSet for Terra

claudijd:
This is largely just footgun avoidance on terra to make sure we never upload an empty GuardianSet, similar how to we check on EVM.

This PR is incomplete as is, and requires that we write some unit-tests for it to ensure the behavior is as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1033)

### DIFF
--- a/terra/contracts/wormhole/src/contract.rs
+++ b/terra/contracts/wormhole/src/contract.rs
@@ -261,11 +261,11 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
         return ContractError::GuardianSetIndexIncreaseError.std_err();
     }
 
-    /* A new GuardianSet with zero addresses should never be accepted,
+    /* An GuardianSet with no addresses should not be accepted,
      * it would break assumptions of the underlying contracts
      */
-    if new_guardian_set.addresses.len() == 0
-        return ContractError::GuardianSetZeroAddresses.std_err();
+    if new_guardian_set.addresses.is_empty()
+        return ContractError::GuardianSetEmpty.std_err();
     }
 
     let old_guardian_set_index = state.guardian_set_index;
--- a/terra/contracts/wormhole/src/contract.rs
+++ b/terra/contracts/wormhole/src/contract.rs
@@ -264,7 +264,7 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
     /* An GuardianSet with no addresses should not be accepted,
      * it would break assumptions of the underlying contracts
      */
-    if new_guardian_set.addresses.is_empty()
+    if new_guardian_set.addresses.is_empty() {
         return ContractError::GuardianSetEmpty.std_err();
     }
 
--- a/terra/contracts/wormhole/src/contract.rs
+++ b/terra/contracts/wormhole/src/contract.rs
@@ -261,6 +261,13 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
         return ContractError::GuardianSetIndexIncreaseError.std_err();
     }
 
+    /* A new GuardianSet with zero addresses should never be accepted,
+     * it would break assumptions of the underlying contracts
+     */
+    if new_guardian_set.addresses.len() == 0
+        return ContractError::GuardianSetZeroAddresses.std_err();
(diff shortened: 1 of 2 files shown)


terra/contracts/wormhole/src/contract.rs
```diff
@@ -261,6 +261,13 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
         return ContractError::GuardianSetIndexIncreaseError.std_err();
     }
 
+    /* A new GuardianSet with zero addresses should never be accepted,
+     * it would break assumptions of the underlying contracts
+     */
+    if new_guardian_set.addresses.len() == 0
+        return ContractError::GuardianSetZeroAddresses.std_err();
```
kcsongor, line 268: this error message is a little confusing, because it sounds like we have a guardian whose address is the 0 address, rather than having an empty guardian set. On that note, we should have a check that ensures we don't have a guardian with zero address (at least on EVM)

terra/contracts/wormhole/src/contract.rs
```diff
@@ -261,6 +261,13 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
         return ContractError::GuardianSetIndexIncreaseError.std_err();
     }
 
+    /* A new GuardianSet with zero addresses should never be accepted,
+     * it would break assumptions of the underlying contracts
+     */
+    if new_guardian_set.addresses.len() == 0
```
jynnantonix, line 267: style nit: `if new_guardian_set.addresses.is_empty()`

If you run `cargo clippy` it will catch stuff like this though you might want to apply #983 first to fix unrelated warnings.

terra/contracts/wormhole/src/error.rs
```diff
@@ -67,6 +67,11 @@ pub enum ContractError {
     #[error("GuardianSetIndexIncreaseError")]
     GuardianSetIndexIncreaseError,
 
+
+    /// Guardian set must not be zero
+    #[error("GuardianSetZeroAddresses")]
```
jynnantonix, line 72: There's no harm in making the actual error message more detailed:

```
#[error("Guardian set must not be empty")]
GuardanSetZeroAddresses,
```

terra/contracts/wormhole/src/contract.rs
claudijd: I can rebase once we land #983, but I'll also fix this suggestion now.

I like your suggestion on `cargo clippy`, let's just make it part of the build to keep us honest going forward.

claudijd:
Suggestions fixed, please don't merge this until after https://github.com/certusone/wormhole/pull/983 lands.

terra/contracts/wormhole/src/error.rs
claudijd: fixed.

terra/contracts/wormhole/src/contract.rs
claudijd: fixed, I left the zero'd address fix out, but I think it's a good idea.  I'll create a ClickUp so I don't miss it, and make sure we get it covered for EVM as well.

terra/contracts/wormhole/src/contract.rs
jynnantonix: This line as also missing a curly brace (`{`) so it won't actually build but maybe that's intentional.

kcsongor (approved):

terra/contracts/wormhole/src/contract.rs
```diff
@@ -261,6 +261,13 @@ fn vaa_update_guardian_set(deps: DepsMut, env: Env, data: &Vec<u8>) -> StdResult
         return ContractError::GuardianSetIndexIncreaseError.std_err();
     }
 
+    /* An GuardianSet with no addresses should not be accepted,
+     * it would break assumptions of the underlying contracts
+     */
+    if new_guardian_set.addresses.is_empty()
```
kcsongor, line 267: you have a missing `{` here

kcsongor (changes requested):
Other than the syntax error, this looks good to me

jynnantonix (dismissed):
lgtm but please make sure the this PR passes the updated checks in dev.v2

jynnantonix (approved):

PR #1034, opened 2022-03-28
Add chainlock_test

claudijd:
This ended up being more than I originally planned, it does the following:

- Adds unit test for MessagePublication
- Also attempts to convey a lack of implicit uint8 safety risk (but we don't seem to have any abuse cases)

My thinking about making safety issues more explicit is that I believe they could be fixed in Go 2.x given some of the proposals mentioned, which would eliminate the need to be overly cautious on add/sub operations.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1034)


claudijd:
I realized I would like to extend this a bit, so pulling reviewers until those are done.

node/pkg/common/chainlock_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	assert.Equal(t, txHash, mp.TxHash)
+	assert.Equal(t, timeStamp, mp.Timestamp)
+	assert.Equal(t, nonce, mp.Nonce)
+	assert.Equal(t, sequence, mp.Sequence)
+	assert.Equal(t, consistencyLevel, mp.ConsistencyLevel)
+	assert.Equal(t, emitterChain, mp.EmitterChain)
+	assert.Equal(t, emitterAddress, mp.EmitterAddress)
+	assert.Equal(t, payload, mp.Payload)
+}
```
SEJeff, line 63: I don't actually see the point of what this test is testing? It is supposed to test that assigning a struct works?

SEJeff (approved):

bruce-riley (dismissed):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1034)* status: all files reviewed, 3 unresolved discussions (waiting on @claudijd and @SEJeff)

___
*[-- commits, line 5 at r2](https://reviewable.io/reviews/certusone/wormhole/1034#-MzKrAga65Cw93MbdP4p):*
So is the intent here that these tests will most likely break on go 2.x? If so, can you maybe throw a comment to that effect at the top of the files?
___
*[node/pkg/common/chainlock_test.go, line 33 at r2](https://reviewable.io/reviews/certusone/wormhole/1034#-MzKn6DJ3L0MaP0TQ3Ti:-MzKn6DJ3L0MaP0TQ3Tj:bur2tmj) ([raw file](https://github.com/certusone/wormhole/blob/914ba50aa4f364deb8b914a076251c5dcd3bb0b8/node/pkg/common/chainlock_test.go#L33)):*
> ```Go
> }
> 
> // Base test to ensure the
> ```

Extremely trivial, but "to ensure the" what?
___
*[node/pkg/common/chainlock_test.go, line 63 at r2](https://reviewable.io/reviews/certusone/wormhole/1034#-MzGVaJYhnl3PiZ654Ng-r2-63:-MzKndlX8hJK29A_VPjI:b-snt50k) ([raw file](https://github.com/certusone/wormhole/blob/914ba50aa4f364deb8b914a076251c5dcd3bb0b8/node/pkg/common/chainlock_test.go#L63)):*
<details><summary><i>Previously, SEJeff (Jeff Schroeder) wrote…</i></summary><blockquote>

I don't actually see the point of what this test is testing? It is supposed to test that assigning a struct works?
</blockquote></details>

In a language like C++ testing a user implemented constructor makes sense, but in go, since the language implements it, I'm not sure this is necessary? But it doesn't hurt anything.

node/pkg/common/chainlock_test.go
claudijd: It locks in a firm expectation of what the struct will do.  For example, if the struct changes (or behaves differently) it will fail and require the person to revisit the test.   I feel like we should go back vaa/common structs similarly, just to lock in their expected behavior, so they don't change in some unplanned/emergent way.

I think another/additional way this could be spec'd in could be to marshal the struct to a hex string, such that if the struct is iterated on it will require a unit-test change.

SEJeff: If someone adds something to the struct, this test will not catch it. You're testing that go works as expected here more than anything else. This is not the kind of thing to make assertions against IMO. If someone changes an existing thing, it will catch it, but this literally is not actually making any assertions other than go works as designed.

I'm not convinced this test has any utility, but I'm still going to approve it.

bruce-riley (dismissed):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1034)* status: all files reviewed, 1 unresolved discussion (waiting on @SEJeff)

node/pkg/common/chainlock_test.go
claudijd: Removed.

bruce-riley (dismissed):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1034)* status: all files reviewed, 1 unresolved discussion (waiting on @SEJeff)

SEJeff (approved):

bruce-riley (approved):

PR #1036, opened 2022-03-29
terra/token_bridge: fixed wrapped asset ids

kcsongor:
The wrapped assets on terra have been migrated to code id 767, but the
config wasn't updated to store this new code id, so newly created
wrapped assets all use an older code id. This patch upgrades all wrapped
assets to 767 and changes the config so future wrapped assets also use
that code id.

Also added a new function `migrate_wrapped_assets` so this task can be
done in the future in a less error-prone way.

### DIFF
--- a/terra/contracts/token-bridge/src/contract.rs
+++ b/terra/contracts/token-bridge/src/contract.rs
@@ -121,8 +121,41 @@ const CHAIN_ID: u16 = 3;
 const WRAPPED_ASSET_UPDATING: &str = "updating";
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
-    Ok(Response::new())
+pub fn migrate(deps: DepsMut, env: Env, _msg: MigrateMsg) -> StdResult<Response> {
+    let messages = migrate_wrapped_assets(deps, env, 767)?;
+    let count = messages.len();
+
+    Ok(Response::new()
+        .add_messages(messages)
+        .add_attribute("migrate", "upgrade cw20 wrappers")
+        .add_attribute("count", count.to_string()))
+}
+
+/// Migrate all wrapped assets to a new code id.
+/// This function should be called in [`migrate`].
+fn migrate_wrapped_assets(deps: DepsMut, _env: Env, new_code_id: u64) -> StdResult<Vec<CosmosMsg>> {
+    let bucket = wrapped_asset_address(deps.storage);
+
+    // Produce a migrate message for each wrapped asset.
+    let mut messages = vec![];
+    for item in bucket.range(None, None, Order::Ascending) {
+        let contract_address = item?.0;
+        messages.push(CosmosMsg::Wasm(WasmMsg::Migrate {
+            contract_addr: deps
+                .api
+                .addr_humanize(&contract_address.into())?
+                .to_string(),
+            new_code_id,
+            msg: to_binary(&MigrateMsg {})?,
+        }));
+    }
+
+    // Update config so future wrapped assets will be deployed with new code id
+    let mut c = config(deps.storage).load()?;
+    c.wrapped_asset_code_id = new_code_id;
+    config(deps.storage).save(&c)?;
+
+    Ok(messages)
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]


terra/contracts/token-bridge/src/contract.rs
```diff
@@ -121,8 +121,41 @@ const CHAIN_ID: u16 = 3;
 const WRAPPED_ASSET_UPDATING: &str = "updating";
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
-    Ok(Response::new())
+pub fn migrate(deps: DepsMut, env: Env, _msg: MigrateMsg) -> StdResult<Response> {
+    let messages = migrate_wrapped_assets(deps, env, 767)?;
```
jynnantonix, line 125: Where is this number coming from?  Is there some meaning behind it or is it just random?  What should be the next `code_id`?  When should this be removed?

Please add comments in the code explaining this.

terra/contracts/token-bridge/src/contract.rs
```diff
@@ -121,8 +121,41 @@ const CHAIN_ID: u16 = 3;
...
+
+    Ok(Response::new()
+        .add_messages(messages)
+        .add_attribute("migrate", "upgrade cw20 wrappers")
+        .add_attribute("count", count.to_string()))
+}
+
+/// Migrate all wrapped assets to a new code id.
+/// This function should be called in [`migrate`].
+fn migrate_wrapped_assets(deps: DepsMut, _env: Env, new_code_id: u64) -> StdResult<Vec<CosmosMsg>> {
```
jynnantonix, line 136: Can we add some tests for this?  It looks like we can set up a unit test environment with [`cosmwasm_std::testing::mock_dependencies`](https://docs.rs/cosmwasm-std/0.16.6/cosmwasm_std/testing/fn.mock_dependencies.html) and [`cosmwasm_std::testing::mock_env`](https://docs.rs/cosmwasm-std/0.16.6/cosmwasm_std/testing/fn.mock_env.html).

In addition to the normal case, what should be the behavior if `wrapped_asset_code_id == new_code_id`?  What happens if we deploy a new version of the contract without removing this migration code?  Or if we accidentally deploy an older version of the contract?  It feels like we need some basic check to make sure that we only run the migration when going from version X -> version Y and either skip or throw an error in any other case.

Is there a way for us to prevent this from happening again?  For example, we could define the code id as a constant somewhere and add a check in `initialize` that will error out if the stored code id doesn't match the constant (and then it's easy to add a test for that behavior).

terra/contracts/token-bridge/src/contract.rs
kcsongor: Added an explanation of this code.

terra/contracts/token-bridge/src/contract.rs
kcsongor: > Is there a way for us to prevent this from happening again?

Here's the code that was ran during the previous migration: https://github.com/certusone/wormhole/commit/c832b123fcfb017d55086cb4d71241370ed270c6#diff-4d1ad320d9f5ab48ff71ac9974eb9a9b10739c15e740a137d85f4fd0354e98f5R124-R144
this current PR is supposed to prevent this issue from happening again by setting the state too in `migrate_wrapped_assets`.

> In addition to the normal case, what should be the behavior if wrapped_asset_code_id == new_code_id?

We could do a check to reject that moving forward. If we forget, that's a no-op, because upgrading the wrapped asset contract to a given id is idempotent (since the wrapped asset contracts have an empty `migrate` handler:
https://github.com/certusone/wormhole/blob/ddd2a60ab8da54966f5383df88eaf5c5bbe73973/terra/contracts/cw20-wrapped/src/contract.rs#L223-L226)

However, the migrate code of the token bridge in general is not idempotent (since it can run arbitrary code), so we must make sure that once it's done, the code is removed. Added a note about this too.

> Or if we accidentally deploy an older version of the contract?

This would require putting the wrong code id into the VAA. When the guardians vote on the VAA, they independently verify that the checksum of the code that's being deployed (by building it locally) matches the checksum of the binary that's stored on-chain at the given code id. There's of course a human element here, but I think the real risk is upgrading to a contract that doesn't belong to the token bridge. I don't know of a good way to protect against that, but it would be worthwhile to investigate.

> Can we add some tests for this? It looks like we can set up a unit test environment with [cosmwasm_std::testing::mock_dependencies](https://docs.rs/cosmwasm-std/0.16.6/cosmwasm_std/testing/fn.mock_dependencies.html) and [cosmwasm_std::testing::mock_env](https://docs.rs/cosmwasm-std/0.16.6/cosmwasm_std/testing/fn.mock_env.html).

As part of this PR, we should investigate that testing framework.

jynnantonix: I've fixed all the terra tests in #1046 so it should be possible to add tests now (after applying that PR locally).  I think the tests in cw20-wrapped/{src/contract.rs,tests/integration.rs} provide a good template for testing the contract functions.

jynnantonix (commented):
Not sure if you're waiting on anything from me.  If you're planning on adding tests to this change I'll wait till those are in.  Otherwise, I can approve this now.

kcsongor:
Yes I want to add tests before merging this. I don't think we're planning on making a release of this very soon so it's not that high up on my list right now, but I want to get around to it.

jynnantonix (approved):
yolo

panoel (approved):

kev1n-peters (approved):

PR #1039, opened 2022-03-30
Add `cargo check` to CI

jynnantonix:
This makes sure that the code builds successfully.


kcsongor:
or could we build all of the rust code with that nightly compiler?

jynnantonix:
> or could we build all of the rust code with that nightly compiler?

The nightly compiler is not stable and can have bugs / other issues.  Also the toolchain team wants to land mutable noalias support so it's regularly turned on in nightly and then reverted when miscompilations (like [this one](https://github.com/rust-lang/rust/issues/84958)) are discovered.  `mutable noalias` is not a well exercised code path in llvm and so miscompilations are common as rust is the first real user of that code.  Since the nightly version isn't even pinned we could easily get unlucky and end up with a bad version of the toolchain that we then use to build + deploy.

If this were new code, I would *very strongly* discourage use of nightly for production.  Whatever feature you think you need from nightly is not worth the hassle of dealing with potential compiler bugs.  Unfortunately it looks like the nightly feature that solitaire uses is deeply baked into the design and cannot be easily changed.  At the very minimum we need to pin the nightly toolchain to a known good version before I would recommend running it in CI.

jynnantonix:
Superseeded by #1045

PR #1040, opened 2022-03-30
Add unit-test for attestation events and Gossip SignedObservations

claudijd:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1040)

### DIFF
--- a/node/pkg/proto/gossip/v1/gossip.pb_test.go
+++ b/node/pkg/proto/gossip/v1/gossip.pb_test.go
@@ -1,0 +1,119 @@
+package gossipv1
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"encoding/hex"
+	vaa "github.com/certusone/wormhole/node/pkg/vaa"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/assert"
+	"testing"
+
+	"time"
+)
+
+func getVAA() vaa.VAA {
+	var payload = []byte{97, 97, 97, 97, 97, 97}
+	var governanceEmitter = vaa.Address{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4}
+
+	vaa := vaa.VAA{
+		Version:          uint8(1),
+		GuardianSetIndex: uint32(1),
+		Signatures:       nil,
+		Timestamp:        time.Unix(0, 0),
+		Nonce:            uint32(1),
+		Sequence:         uint64(1),
+		ConsistencyLevel: uint8(32),
+		EmitterChain:     vaa.ChainIDSolana,
+		EmitterAddress:   governanceEmitter,
+		Payload:          payload,
+	}
+
+	return vaa
+}
+
+func TestSignedObservation(t *testing.T) {
+	vaa := getVAA()
+	digest := vaa.SigningMsg()
+
+	txhash := []byte{97, 97, 97, 97, 97, 97}
+	signature := []byte{97, 97, 97, 97, 97, 97}
+
+	privKey, _ := ecdsa.GenerateKey(crypto.S256(), rand.Reader)
+
+	obsv := SignedObservation{
+		Addr:      crypto.PubkeyToAddress(privKey.PublicKey).Bytes(),
+		Hash:      digest.Bytes(),
+		Signature: signature,
+		TxHash:    txhash,
+		MessageId: vaa.MessageID(),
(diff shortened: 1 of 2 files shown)


node/pkg/proto/gossip/v1/gossip.pb_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Addr:      crypto.PubkeyToAddress(privKey.PublicKey).Bytes(),
+		Hash:      digest.Bytes(),
+		Signature: signature,
+		TxHash:    txhash,
+		MessageId: vaa.MessageID(),
+	}
+
+	assert.Equal(t, signature, obsv.Signature)
+	assert.Equal(t, txhash, obsv.TxHash)
+	assert.Equal(t, vaa.MessageID(), obsv.MessageId)
```
SEJeff, line 54: You assert that `SignedObservation.Signature` was set to the signature which you set. You then assert that `SignedObservation.TxHash` was set to the txhash you set. Lastly you assert that `vaa.MessageID()` is equal to the obvs.MessageId that you set to `vaa.MessageID()`.

This test is testing that go sets structs to what you set them to. There is no beneficial value in this. Consider removing this test entirely. Testing that setting structs works should be in go's stdlib tests, not in application tests. Consider not doing this.

node/pkg/proto/gossip/v1/gossip.pb_test.go
```diff
@@ -0,0 +1,119 @@
...
+	obsv := getSignedObservation()
+	obsvString := obsv.String()
+	assert.Contains(t, obsvString, "addr")
+	assert.Contains(t, obsvString, "hash")
+	assert.Contains(t, obsvString, "signature")
+	assert.Contains(t, obsvString, "tx_hash")
+	assert.Contains(t, obsvString, "message_id")
+	assert.Greater(t, len(obsvString), 200)
+	assert.Less(t, len(obsvString), 300)
+}
```
SEJeff, line 85: It sucks that you're generating a new private key every time in `getSignedObservation()`. If it was quasi-static, you could simply assert the string contents, which would be a better test overall.

This is solid given the value changes eachtime, but could miss things in the future (unlikely), but less magic and more bare asserts are good, as @cjw296 taught me.

node/pkg/proto/gossip/v1/gossip.pb_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	assert.Equal(t, []int([]int{3}), num)
+	assert.Equal(t, expected, hex.EncodeToString(bytes))
```
SEJeff, line 93: perfect 👍 

No logic or real magic. Just data and assertions.

node/pkg/reporter/attestation_events_test.go
```diff
@@ -0,0 +1,34 @@
...
+	attestationEventReporter := EventListener(logger)
+	assert.Equal(t, subs_expect, attestationEventReporter.subs)
+}
+
+func TestGetUniqueClientId(t *testing.T) {
+	logger, _ := zap.NewProduction()
+	defer logger.Sync()
+
+	attestationEventReporter := EventListener(logger)
+	assert.Equal(t, 498081, attestationEventReporter.getUniqueClientId())
```
SEJeff, line 24: This seems like a solid test, but where does that number come from?

If you can swing it, [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) are better when they are set to more obvious variables describing what they do.

node/pkg/reporter/attestation_events_test.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func TestSubscribe(t *testing.T) {
+	logger, _ := zap.NewProduction()
+	defer logger.Sync()
+
+	attestationEventReporter := EventListener(logger)
+	activeSubscription := attestationEventReporter.Subscribe()
+	assert.Equal(t, 727887, activeSubscription.ClientId)
```
SEJeff, line 33: Ditto as above. Mind making that number a variable that makes it more obvious what it is or why it is that specific number?

SEJeff (commented):
Maybe wrap the defer in an anonymous function that checks the possible error in `logger.Sync()` to make errcheck (which is a fantastic linter btw!) happier?

This is causing the [checks to fail here](https://github.com/certusone/wormhole/runs/5760602597?check_suite_focus=true#step:4:63).

Sans that one test, this is really really good!

bruce-riley (approved):

node/pkg/reporter/attestation_events_test.go
claudijd: Once https://github.com/certusone/wormhole/pull/1127 merges, I'll readjust this to reflect the new behavior.

PR #1041, opened 2022-03-30
Don't require padding on addrs in bridge clients

bruce-riley:
Change-Id: Id32bfb90ee20d1bccc6b860de028c0bdecfd8786

Allow users to specify the bridge address in the usual EVM format of 0x599CEa2204B4FaECd584Ab1F2b6aCA137a0afbE8 rather than having to pad it with 24 leading zeroes.


kcsongor (approved):
This has been a little annoying, thanks for fixing it 🙂

PR #1045, opened 2022-04-01
Add `cargo check` to CI

jynnantonix:
This makes sure that the code builds successfully.

---

**Stack**:
- #1060
- #1059
- #1046
- #1045 ⬅

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1045)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,3 +77,28 @@ jobs:
           go-version: '1.17.5'
       - run: make generate && ./lint.sh
       - run: cd node && go test -v ./...
+
+  # Run rust lints and tests
+  rust-lint-and-tests:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        manifest:
+          - terra/Cargo.toml
+          - sdk/rust/Cargo.toml
+    steps:
+      - name: Check out source
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


kcsongor (dismissed):

.github/workflows/build.yml
```diff
@@ -77,3 +77,31 @@ jobs:
           go-version: '1.17.5'
       - run: make generate && ./lint.sh
       - run: cd node && go test -v ./...
+
+  # Run rust lints and tests
+  rust-lint-and-tests:
+    runs-on: tilt-kube-public
```
evan-gray, line 83: I don't think this needs to run on the tilt kube box. can we just run it on ubuntu-20.04 like the others?

evan-gray, line 83: I don't think this needs to run on the tilt kube box. can we just run it on ubuntu-20.04 like the others?

evan-gray (changes requested):

.github/workflows/build.yml
jynnantonix: Done.  I also removed the concurrency section because I didn't see it in the other places that use ubuntu-20.04.

evan-gray (approved):

kcsongor (approved):

PR #1049, opened 2022-04-02
Add scripts and readme to deploy and verify terra contracts

kcsongor:
Now a deployment can be done with
```
cd terra
export NETWORK=mainnet
make artifacts
make deploy/bridge
```


terra/Makefile
```diff
@@ -0,0 +1,55 @@
...
+VALID_testnet=1
+VALID_devnet=1
+.PHONY: check-network
+check-network:
+ifndef VALID_$(NETWORK)
+	$(error Invalid or missing NETWORK. Please call with `$(MAKE) $(MAKECMDGOALS) NETWORK=[mainnet | testnet | devnet]`)
+endif
+
+$(WASMS) artifacts/checksums.txt: $(SOURCE_FILES)
+	docker build --target artifacts -o artifacts .
```
claudijd, line 24: Can we do a --no-cache on the build action?  I realize it will take more time per run, but I'm overly cautious about shipping cached artifacts.  Similarly, I'd also suggest having this step should call clean, for similar reasons.

claudijd (dismissed):
Approving, but still suggesting we consider removing artifacts at every build and running docker in --no-cache mode for the build artifacts step.

terra/generate_governance
```diff
@@ -0,0 +1,71 @@
...
+       ;;
+  esac
+done
+shift $((OPTIND - 1))
+
+[ -z "$code_id" ] && { echo "$usage" >&2; exit 1; }
+[ -z "$module" ] && { echo "$usage" >&2; exit 1; }
+
+cd "${0%/*}/.."
+make node >/dev/null
```
kcsongor, line 40: oops, this should not happen here, instead just use `guardiand`

terra/Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+payer-$(NETWORK).json:
+	$(error Missing private key in payer-$(NETWORK).json)
+
+.PHONY: deploy/bridge
+## Deploy core bridge
+deploy/bridge: bridge-code-id-$(NETWORK).txt
+
+## Deploy token bridge
+deploy/token_bridge: token_bridge-code-id-$(NETWORK).txt
```
kcsongor, line 34: should be phony (and the next one too)

evan-gray (approved):

terra/Makefile
kcsongor: I think independent verification should eliminate any issues from this, and no-cache is too painful during development.

karlkempe (approved):

PR #1051, opened 2022-04-05
Celo testnet support

bruce-riley:
The majority of the work for this PR is in the guardian watcher, because Celo requires their own version of go-ethereum. The code in node/pkg/celo is basically a copy of node/pkg/ethereum, but using the Celo library. Most of the differences are in marshalling from the Celo Address and Hash types to the Ethereum ones (used in the guardian code) and vice versa. It might be possible to refactor all of this to share more code, but I think that's outside the scope of this PR.

Note that on the client side, it will be necessary to use the Celo provider and signer, as the standard ethers ones fail (returning "Invalid Big Number" on API calls). Here is what needs to be done:

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1051)


bruce-riley:
Got off guard by this failing to build in tilt on github since it builds in tilt for me in my local repo. The build error is due to duplicate symbols between the Celo go library and go-ethereum, and the solution is to add the following to the go build line:  --ldflags '-extldflags "-Wl,--allow-multiple-definition"'

I added that to the node Dockerfile, but apparently the github build doesn't do that?? Anyone know how to fix this?

node/pkg/celo/by_transaction.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// Bail early when the transaction receipt status is anything other than
+	// 1 (success). In theory, this check isn't strictly necessary - a failed
+	// transaction cannot emit logs and will trigger neither subscription
+	// messages nor have log messages in its receipt.
+	//
+	// However, relying on that invariant is brittle - we connect to a lot of
+	// EVM-compatible chains which might accidentally break this API contract
+	// and return logs for failed transactions. Check explicitly instead.
+	if receipt.Status != 1 {
```
SEJeff, line 51: Maybe I'm weird, but I'd almost make that 1 a constant like `const statusSuccess = 1` at the top of this file so it is exceedingly obvious, even with the helpful comment. I never like [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))

node/pkg/celo/by_transaction.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	msgs := make([]*common.MessagePublication, 0, len(receipt.Logs))
+
+	// Extract logs
+	for _, l := range receipt.Logs {
+		// SECURITY: Skip logs not produced by our contract.
+		if l.Address != contract {
+			continue
+		}
```
SEJeff, line 68: Also @claudijd ^

We probably want tests for these security bits. Tagging for later.

third_party/abigen-celo/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apk add curl git gcc libc-dev linux-headers
+
+WORKDIR /app
+
+ADD go.mod .
+ADD go.sum .
+
+#RUN go build -mod=readonly -o /bin/abigen github.com/celo-org/celo-blockchain/cmd/abigen
+RUN go build -o /bin/abigen github.com/celo-org/celo-blockchain/cmd/abigen
```
SEJeff, line 11: Try switching this to use a newer go image and also one that isn't using alpine and I think the rust problems might go away.

node/pkg/celo/by_transaction.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/certusone/wormhole/node/pkg/vaa"
+	eth_common "github.com/ethereum/go-ethereum/common"
+)
+
+var (
+	// SECURITY: Hardcoded ABI identifier for the LogMessagePublished topic. When using the watcher, we don't need this
+	// since the node will only hand us pre-filtered events. In this case, we need to manually verify it
+	// since ParseLogMessagePublished will only verify whether it parses.
+	logMessagePublishedTopic = celo_common.HexToHash("0x6eb224fb001ed210e379b335e35efe88672a8ce935d981a6896b27ffdf52a3b2")
+)
```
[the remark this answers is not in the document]
claudijd: I don't fully understand the security context here, could someone add some more context?

node/pkg/celo/by_transaction.go
claudijd: yeah, eth/by_transaction works the same.  Was chatting about this with Leo this AM, sounds like we might benefit from a refactor on this to make it more testable.  However, I don't think it should block this PR.

third_party/abigen-celo/Dockerfile
claudijd: I'd recommend 1.17.5 or 1.17.8

claudijd (dismissed):

node/pkg/celo/by_transaction.go
claudijd: This is copy pasta from ethereum.  If we want to refactor this, we should probably do it comprehensively.

node/pkg/celo/by_transaction.go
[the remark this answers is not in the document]
bruce-riley: Sorry, this is a straight copy from pkg/ethereum that was necessary to make celo work. I do not know the answer to this.

bruce-riley:
As I mentioned in the original comment, this code is really a copy and past of EthWatcher to CleoWatcher, because Celo uses their own version of go-ethereum. I'm not thrilled with this approach, but don't really know a better way to handle it. @leoluk I would really appreciate your input on this.

generate-abi-celo.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Getting this error when building the docker image:
+#10 40.92 /usr/lib/gcc/x86_64-alpine-linux-musl/9.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: /go/pkg/mod/github.com/celo-org/celo-bls-go@v0.2.4/bls/../libs/x86_64-unknown-linux-gnu/libbls_snark_sys.a(std-f14aca24435a5414.std.54pte2sm-cgu.0.rcgu.o): in function `std::sys::unix::net::on_resolver_failure':
+#10 40.92 /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39//library/std/src/sys/unix/net.rs:376: undefined reference to `__res_init'
+#10 40.92 collect2: error: ld returned 1 exit status
+
+# (
+#   cd third_party/abigen-celo
+#   docker build -t localhost/certusone/wormhole-abigen-celo:latest .
+# )
```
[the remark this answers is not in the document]
bruce-riley: Got the docker image working.

third_party/abigen-celo/Dockerfile
bruce-riley: Done.

SEJeff (dismissed):
Seems like we could merge this as-is and refactor the eth bit to work with this AND the other evm chains, but refactoring it AND adding this has a big chance to break everything. Let's try to come back and clean that up, but one thing at a time.

leoluk (changes requested):
Sorry - all this duplicate code is impossible to maintain.

bruce-riley:
This PR has been massively refactored to eliminate most of the redundant code. I would appreciate if everyone would give it another look.

node/pkg/celo/celoimpl.go
```diff
@@ -0,0 +1,263 @@
...
+	return e.caller.GetCurrentGuardianSetIndex(opts)
+}
+
+func (e *CeloImpl) GetGuardianSet(ctx context.Context, index uint32) (ethAbi.StructsGuardianSet, error) {
+	if e.caller == nil {
+		panic("caller is not initialized!")
+	}
+
+	opts := &celoBind.CallOpts{Context: ctx}
+	celoGs, err := e.caller.GetGuardianSet(opts, index)
```
SEJeff, line 69: You need to check this err btw. The [errcheck](https://github.com/kisielk/errcheck) linter will find these sorts of things for you automatically.

node/pkg/celo/celoimpl.go
```diff
@@ -0,0 +1,263 @@
...
+		Bloom:       ethTypes.BytesToBloom(ev.Bloom.Bytes()),
+		// Celo doesn't have Difficulty.
+		Number: ev.Number,
+		// Celo doesn't have GasLimit.
+		GasUsed: ev.GasUsed,
+		Time:    ev.Time,
+		Extra:   ev.Extra,
+		// Celo doesn't have MixDigest.
+		// Celo doesn't have Nonce.
+		// Celo doesn't have BaseFee.
```
SEJeff, line 261: Good comments 👍

SEJeff (commented):
I do wish there was a cleaner way to check all of the required fields for CeloImpl or EthImpl so. you don't need to check then panic in so many methods in the implementations of Ethish.

You missed one `err` that could cause a panic, but the rest looks fine.

node/pkg/celo/celoimpl.go
bruce-riley: Sticking with the panics, since these errors would be programming bugs, rather than legitimate runtime errors.

Added that error check.

SEJeff (approved):
Nice work considering you were given a 💩 by celo.

third_party/abigen-celo/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/certusone/wormhole/third_party/abigen-celo
+
+go 1.15
```
leoluk, line 3: That's a pretty old version of Go - how comes?

node/pkg/ethereum/watcher.go
```diff
@@ -124,7 +125,16 @@ func NewEthWatcher(
...
 	minConfirmations uint64,
-	obsvReqC chan *gossipv1.ObservationRequest) *Watcher {
+	obsvReqC chan *gossipv1.ObservationRequest,
+	unsafeDevMode bool) *Watcher {
+
+	var ethIntf common.Ethish
+	if chainID == vaa.ChainIDCelo && ! unsafeDevMode {
+		ethIntf = &celo.CeloImpl{NetworkName: networkName}
+	} else {
+		ethIntf = &EthImpl{NetworkName: networkName}
```
leoluk, line 135: Why is this necessary?

It would be best to avoid leaking `unsafeDevMode` into pkg/ and instead have specific flags for the functionality you want to toggle.

node/pkg/common/ethish.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"go.uber.org/zap"
+)
+
+type Ethish interface {
+	SetLogger(l *zap.Logger)
+	DialContext(ctx context.Context, rawurl string) error
+	NewAbiFilterer(address ethCommon.Address) error
+	NewAbiCaller(address ethCommon.Address) error
+	GetCurrentGuardianSetIndex(ctx context.Context) (uint32, error)
```
leoluk, line 24: What's the point of having three separate "init" methods?

leoluk (changes requested):
Looks good overall - good way to avoid Eth code duplication

.github/workflows/build.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         with:
           go-version: '1.17.5'
       - run: make generate && ./lint.sh
-      - run: cd node && go test -v ./...
+      - run: cd node && go test -v --ldflags '-extldflags "-Wl,--allow-multiple-definition" ' ./...
```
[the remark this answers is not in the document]
leoluk: Also:

> These options allow multiple definitions and the first definition will be used.

Which one gets used, celo or eth? Could this cause unintended behavior or introduce extra risks?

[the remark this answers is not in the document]
bruce-riley: Added a comment. They both use the exact same secp256k1 C header file. It is picking up the go-ethereum version, but I'm not sure that's deterministic.

third_party/abigen-celo/go.mod
bruce-riley: This was copied from the existing abigen version of this file. It is only used to build the celo abigen tool, which works fine with this version. If / when we need to regenerate the ABIs for both the go-ethereum and celo-ethereum watchers, we can upgrade both. Leaving this as is for now.

node/pkg/common/ethish.go
bruce-riley: This is to maintain the full flexibility of the go-ethereum library. For instance, parse_eth_tx uses this, but it does not need the caller, so it only calls the first two init methods. Going to leave this as is.

.github/workflows/build.yml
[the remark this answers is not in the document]
bruce-riley: During our standup today, we discussed this. All of the duplicate symbols are in two files (one *.h and one *.go) that are literally identical between the two packages. We decided that, before Celo goes to mainnet, we will implement a step in the build process that confirms that the only multiple definitions in the guardian are in these two files. This should prevent other duplicate definitions from seeping in.

generate-abi-celo.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Regenerate node/pkg/celo/abi.
+
+set -euo pipefail
+
+(
+  cd third_party/abigen-celo
+  docker build -t localhost/certusone/wormhole-abigen-celo:latest .
```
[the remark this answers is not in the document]
bruce-riley: Note going to bite this off right now.

clients/token_bridge/main.ts
```diff
@@ -339,6 +341,10 @@ yargs(hideBin(process.argv))
     }, async (argv: any) => {
         const bridge = await importCoreWasm()
 
+        // let provider= new celo.CeloProvider(argv.rpc)
+        // await provider.ready
+        // let signer = new celo.CeloWallet(argv.key, provider)
```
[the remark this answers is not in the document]
bruce-riley: This was just here as documentation. Csongar is in the process of deleting this file as part of his CLI refactor, so I just deleted these comments. (I also have it on the testnet deploy page.)

evan-gray (approved):
approving for testnet deployment

SEJeff (approved):

PR #1059, opened 2022-04-08
solana: Merge into single workspace

jynnantonix:
Merge all the separate solana crates into a single workspace.  This is
the same thing we do for terra and will make it easier to check / lint /
test all the crates together.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1059)


kcsongor (approved):
This looks like a net win to me. Are there any workflows that are not possible with this merged setup that were possible before?

kcsongor (approved):

evan-gray (approved):

PR #1046, opened 2022-04-01
sdk, terra: Fix tests and add them to CI

jynnantonix:
The terra tests had bit rotted quite terribly. It seems no one has run
them in almost a year because the API changed with the 0.14 release of
cosmwasm_vm in April, 2021.

Completely remove cosmwasm_vm from dev-dependencies.  The tests aren't
doing anything that requires interacting with the wasm code and can just
test the relevant functions directly.

Once this goes in the tests should be running in the CI so hopefully no
one will be able to push any more breaking changes.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1046)

PR #1060, opened 2022-04-08
solana: bridge: Fix tests

jynnantonix:
Fix all the tests for the solana bridge program so that we can start
running it in CI.

Use the `solana-program-test` crate as the test framework. Previously,
running the tests required spinning up a test validator and manually
deploying the program there. Now developers can just do `cargo test-bpf`
to run the tests without needing any additional setup (beyond installing
the solana tools). This also lets us split out the tests so that they
can be run separately rather than having to run everything sequentially
in a single integration test.

Unfortunately, I suspect there is a race in the test framework
somewhere.  The `duplicate_messages_fail` test sometimes fails when all
the tests are run in parallel but has not failed even once when running
it by itself or with `--test-threads 1` (i.e. running all tests
sequentially).

---

**Stack**:
- #1059
- #1046
- #1045


solana/bridge/program/tests/integration.rs
```diff
@@ -325,53 +382,91 @@ fn test_invalid_emitter(context: &mut Context) {
...
-        10_000,
+        &[payer, &emitter, &message],
+        &[
+            system_instruction::transfer(&payer.pubkey(), &fee_collector, fee),
+            instruction,
+        ],
+        CommitmentLevel::Processed,
     )
+    .await
     .is_err());
```
jynnantonix, line 463: I had to re-write this test and it's very likely I've screwed something up.  Like I mentioned in the commit message, this test fails sometimes when all tests are run in parallel but hasn't failed a single time when run by itself or with `--test-threads 1`.

The test failure is always the same.  The second execution of the instructions succeeds when it shouldn't.  The weird thing is that it should fail because the message account is already initialized, which should cause the derivation to fail and lines 438 - 449 check the contents of the message account to ensure that it does exist but somehow the second execution still succeeds.

If anyone has thoughts on what might be going on here, I'd love to hear them.

kcsongor:
This is really neat!

solana/bridge/program/src/lib.rs
```diff
@@ -8,7 +8,6 @@ use solitaire::*;
 pub const MAX_LEN_GUARDIAN_KEYS: usize = 19;
 pub const CHAIN_ID_SOLANA: u16 = 1;
 
-#[cfg(feature = "no-entrypoint")]
```
evan-gray, line 11: I don't think we should remove this, we don't want the instruction builders built into the on-chain program. Maybe we need to add a different feature flag to handle whatever this was needed for?

evan-gray, line 11: I don't think we should remove this, we don't want the instruction builders built into the on-chain program. Maybe we need to add a different feature flag to handle whatever this was needed for?

evan-gray (changes requested):

solana/bridge/program/tests/integration.rs
kcsongor: The message account is indeed already initialised... what's going on? 🤔

solana/bridge/program/src/lib.rs
jynnantonix: The tests need the builder functions.  I ran into some other issues getting the tests working with the "no-entrypoint" feature enabled so I added a separate feature for the instructions module only.

solana/bridge/program/tests/integration.rs
jynnantonix: As discussed offline this test was wrong so I've removed it.  I didn't spend more time trying to figure out why it was only failing in multi-threaded mode.

jynnantonix:
I've updated the Tiltfile and the Dockerfile so that we run the integration tests in the tilt CI and I also added a `test` target to the Makefile.

kcsongor:
Not too long ago I added a makefile to the bridge folder to run the tests outside of tilt: https://github.com/certusone/wormhole/blob/dev.v2/solana/bridge/Makefile#L15-L32.
It spins up a new test validator - a little hacky, and running in tilt is probably more robust.
Should we delete that makefile?

kcsongor (dismissed):

jynnantonix:
Once this lands, we won't use solana-test-validator for the bridge tests so we should probably drop the Makefile. Should I do that here or in another change?

solana/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update && \
     && \
     rm -rf /var/lib/apt/lists/* && \
     rustup component add rustfmt && \
-    rustup default nightly-2022-01-02
+    rustup default nightly-2021-12-03
```
evan-gray, line 15: is this dangerous to revert our solana builder to an older rust nightly? @kcsongor @hendrikhofstadt ?

evan-gray (changes requested):

solana/Dockerfile
kcsongor: It's hard to quantify the effects of switching (or even not switching) nightly versions, but what seems like the safest thing to do is to track the compiler version used by the given version of solana we're using. In this case https://github.com/solana-labs/solana/blob/v1.9.4/ci/rust-version.sh#L27 the solana 1.9.4 toolchain seems to be best tested with `2021-12-03`, so I think this is as safe as it gets.

jynnantonix: Yes, I changed it based on the slack discussion that we should match whatever version solana uses.

evan-gray:
re-requested @kcsongor 's review, also looks like there are conflicts @jynnantonix - probably just a simple rebase

jynnantonix:
> re-requested @kcsongor 's review, also looks like there are conflicts @jynnantonix - probably just a simple rebase

Conflicts resolved via github's UI

kcsongor (approved):

solana/modules/nft_bridge/program/tests/integration.rs
```diff
@@ -423,209 +316,219 @@ fn test_register_chain(context: &mut Context) -> () {
...
-}
 
-fn test_create_wrapped(context: &mut Context) -> (Pubkey) {
-    println!("CreateWrapped");
-    use token_bridge::{
-        accounts::ConfigAccount,
-        types::Config,
-    };
+    // What this actually does is initialize the spl token metadata so that we can then burn the NFT
+    // in the future but of course, we can't call it something useful like
```
kcsongor, line 512: 😬

kcsongor (approved):

claudijd (approved):

Reisen (approved):

evan-gray (approved):

PR #1070, opened 2022-04-11
Solitaire account initialisation check

kcsongor:
This changes solitaire's logic for determining when an account is initialised. Currently, any account that holds lamports is considered to be initialised. With this patch, only the accounts that hold data are.
Furthermore, solana's `create_account` primitive reverts when the account already holds lamports, so we don't use it. Instead, account creation proceeds by
1. topping up the account to be rent-exempt
2. allocating memory
3. assigning ownership to the program

Also made a quick attempt to get tests runnable in CI, but the tests currently require running a testing validator. A much more robust solution is https://github.com/certusone/wormhole/pull/1060, so we should just use that approach.


claudijd (approved):

solana/bridge/program/tests/integration.rs
```diff
@@ -335,40 +435,6 @@ fn test_invalid_emitter(context: &mut Context) {
     .is_err());
 }
 
-fn test_duplicate_messages_fail(context: &mut Context) {
```
jynnantonix, line 338: Is there any particular reason why this test was dropped?  If it's ok for it to be removed I can also drop it from my refactor, which will let us run the tests in parallel.

PR #1077, opened 2022-04-13
Allow account reuse for message posting

hendrikhofstadt:
This change introduces a new instruction that allows to post messages while reusing the message account.
Usage of this instruction is unreliable because VAAs can't be recovered when the message is overwritten.
Applications that frequently send messages can make use of this to save costs.

A new instruction is introduced to make sure existing programs can't accidentially introduce unreliability by reusing message accounts (or users reusing message accounts as this logic mostly lives in the client).


claudijd (approved):

solana/bridge/program/tests/integration.rs
```diff
@@ -322,6 +323,84 @@ fn test_bridge_messages(context: &mut Context) {
     }
 }
 
+fn test_bridge_messages_unreliable(context: &mut Context) {
```
jynnantonix, line 326: Can we also add a test to ensure that messages with different sizes return an error?  And also one for messages with different emitters.  I think it's trivially obvious from reading the source that these will fail but the tests are mostly so that those checks don't get accidentally removed in the future.

node/pkg/solana/client.go
```diff
@@ -386,7 +387,7 @@ func (s *SolanaWatcher) processInstruction(ctx context.Context, logger *zap.Logg
 		return false, nil
 	}
 
-	if inst.Data[0] != postMessageInstructionID {
```
jynnantonix, line 389: Unrelated but do we need to check that len(inst.Data) > 0 first? Looks like this is coming from untrusted input.

solana/bridge/program/src/lib.rs
```diff
@@ -56,6 +58,7 @@ pub use api::{
     InitializeData,
     PostMessage,
     PostMessageData,
+    PostMessageUnreliable,
```
jynnantonix, line 61: naming nit: If I was using a library and I saw a method called `*_unreliable` I probably wouldn't even consider using it, especially since there are no doc comments explaining what exactly is unreliable about it.  How about something like `post_message_reusable`?

hendrikhofstadt: Added a comment; we want to warn users and only have them use it if they know what they're doing.

solana/bridge/program/src/api/post_message.rs
```diff
@@ -77,71 +98,141 @@ pub fn post_message(
...
+    Ok(())
+}
+
+pub fn post_message_unreliable(
+    ctx: &ExecutionContext,
+    accs: &mut PostMessageUnreliable,
+    data: PostMessageData,
+) -> Result<()> {
+    // Accounts can't be resized so the payload sizes need to match
+    if accs.message.is_initialized() && accs.message.payload.len() != data.payload.len() {
```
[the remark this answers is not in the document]
hendrikhofstadt: I think this may lead to `serde` issues with borsh and potential inconsistencies. Wanted to just not support this for now due to complexity.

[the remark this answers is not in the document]
jynnantonix: Ack.

jynnantonix (dismissed):
> Reviewed 5 of 9 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @evan-gray, @Reisen, and @sekoje)

jynnantonix (changes requested):
> Reviewed 17 of 17 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: all files reviewed, 7 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @Reisen, and @sekoje)

*[a discussion](https://reviewable.io/reviews/certusone/wormhole/1077#-N3Iwby6EjJbFkjVKDEJ:-N3Iwby6EjJbFkjVKDEK:bh9ftbt) (no related file):*
My preference would be to keep the solana-sdk upgrade separate from the feature work so that it can be easily reverted if necessary but I don't feel that strongly about it.

___
___
*[`-- commits` line 9 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3Ion_C0UwDarIoo2IA):*
I'm surprised solana itself didn't get hit by this.
___
*[`solana/rust-toolchain.toml` line 4 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3IoREa2WmVr3q_BzxL:-N3IoREb86-ubn9ZSsaB:b-s0zytc) ([raw file](https://github.com/certusone/wormhole/blob/5b879a8c929f9c19fd158738a971915f33e78236/solana/rust-toolchain.toml#L4)):*
> ```TOML
> # the branch of the solana repo that corresponds with the version we're using.
> [toolchain]
> channel = "nightly-2022-05-26"
> ```

Please update the toolchain version in the Dockerfile as well.
___
*[`solana/bridge/program/src/api/post_message.rs` line 154 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3Iql_02K8ehEohtsU_:-N3Iql_02K8ehEohtsUa:b-yzdxej) ([raw file](https://github.com/certusone/wormhole/blob/5b879a8c929f9c19fd158738a971915f33e78236/solana/bridge/program/src/api/post_message.rs#L154)):*
> ```XML
>     if accs.message.is_initialized()
>         && data.payload.len() as isize - accs.message.payload.len() as isize
>             > solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE as isize
> ```

Do we need to do checked arithmetic here or are these values trusted?  It looks like we're casting to isize to sidestep underflow issues.  Another option is using `saturating_sub` like:



_Suggestion:_
```XML
        data.payload.len().saturating_sub(accs.message.payload.len()) > MAX_PERMITTED_DATA_INCREASE
```
___
*[`solana/bridge/program/src/api/post_message.rs` line 163 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3Iux6J7iJvFvl6WFu0:-N3Iux6J7iJvFvl6WFu1:b-4xhr0i) ([raw file](https://github.com/certusone/wormhole/blob/5b879a8c929f9c19fd158738a971915f33e78236/solana/bridge/program/src/api/post_message.rs#L163)):*
> ```XML
>         accs.message.info().realloc(
>             (accs.message.info().data_len() as isize
>                 + ((data.payload.len() as isize) - (accs.message.payload.len() as isize)))
> ```

`realloc` [takes the new length as a parameter](https://docs.rs/solana-program/latest/solana_program/account_info/struct.AccountInfo.html#method.realloc) so why would this not be `accs.message.info().data_len() + accs.message.payload.len()`?  It's unclear to me why the length of the previous payload needs to be a part of the calculation.
___
*[`solana/bridge/program/src/api/post_message.rs` line 165 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3IvrjP-szM8LujLOKA:-N3IvrjP-szM8LujLOKB:b2qwnls) ([raw file](https://github.com/certusone/wormhole/blob/5b879a8c929f9c19fd158738a971915f33e78236/solana/bridge/program/src/api/post_message.rs#L165)):*
> ```XML
>                 + ((data.payload.len() as isize) - (accs.message.payload.len() as isize)))
>                 as usize,
>             true,
> ```

Is this necessary?  The realloc docs say the memory is already zero-initialized so we only need to re-zero it if we realloc "from larger to smaller and back to larger again" in the same call.
___
*[`solana/bridge/program/tests/integration.rs` line 370 at r3](https://reviewable.io/reviews/certusone/wormhole/1077#-N3It99U9Yvvmzyap3oT:-N3It99U9Yvvmzyap3oU:b-uwn71w) ([raw file](https://github.com/certusone/wormhole/blob/5b879a8c929f9c19fd158738a971915f33e78236/solana/bridge/program/tests/integration.rs#L370)):*
> ```XML
>         let message: Vec<u8> = (0..rand::random::<u8>())
>             .map(|_| rand::random::<u8>())
>             .collect();
> ```

nit: I think using `fill` could be more efficient



_Suggestion:_
```XML
        let mut message = vec![0u8; rand::random::<u8>().into());
        rand::thread_rng().fill(&mut message);
```

hendrikhofstadt:
@jynnantonix reverted changes on account resizing and rebased

jynnantonix (approved):
> Reviewed 19 of 19 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: all files reviewed, 1 unresolved discussion (waiting on @evan-gray, @hendrikhofstadt, @Reisen, and @sekoje)

kcsongor (dismissed):

jynnantonix (commented):
> Reviewed 10 of 11 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: all files reviewed, 5 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @Reisen, and @sekoje)

___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 37 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5YfjBv3v-tQePclapZ:-N5YfjBv3v-tQePclap_:bbvhmxz) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L37)):*
> ```XML
> 
> pub type PostedMessageUnreliable<'a, const State: AccountState> =
>     Data<'a, PostedMessageData, { State }>;
> ```

Is this supposed to use `PostedMessageUnreliableData`?  If not, please add a comment explaining why it's not using PostedMessageUnreliableData.
___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 89 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5Yf8CT81avihMjM_Xu:-N5Yf8CT81avihMjM_Xv:b2ux1n4) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L89)):*
> ```XML
> impl BorshDeserialize for PostedMessageData {
>     fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
>         let expected = "msg".as_bytes();
> ```

style nit: byte string literals can be declared like:



_Suggestion:_
```XML
   let expected = b"msg";
```
___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 90 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5YfdtW90zl5XCpFu06:-N5YfdtW90zl5XCpFu07:b-52dtu0) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L90)):*
> ```XML
>     fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
>         let expected = "msg".as_bytes();
>         let magic: &[u8] = &buf[0..3];
> ```

Do we need to check the length here first?  Otherwise this will panic.
___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 177 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5YhR7cBcP-Io57ilHC:-N5YhR7cBcP-Io57ilHD:bb9smxx) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L177)):*
> ```XML
> 
>     fn deref(&self) -> &Self::Target {
>         unsafe { std::mem::transmute(&self.data) }
> ```

Why do we need a transmute here?  Can't we just return `&self.data`? 
___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 183 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5YhX7IEy5RcMS2_7o9:-N5YhX7IEy5RcMS2_7oA:bulp9ms) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L183)):*
> ```XML
> impl DerefMut for PostedMessageUnreliableData {
>     fn deref_mut(&mut self) -> &mut Self::Target {
>         unsafe { std::mem::transmute(&mut self.data) }
> ```

Same comment here.  We can just return `&mut self.data`

kcsongor (commented):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: all files reviewed, 5 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @jynnantonix, @Reisen, and @sekoje)

___
*[`solana/bridge/program/src/accounts/posted_message.rs` line 37 at r6](https://reviewable.io/reviews/certusone/wormhole/1077#-N5YfjBv3v-tQePclapZ:-N5Yu7jm1KGNXPAednnh:b-x4ceq7) ([raw file](https://github.com/certusone/wormhole/blob/59236e26d8dd19fd1493c3ed79eb9526f8e25e45/solana/bridge/program/src/accounts/posted_message.rs#L37)):*

Is this supposed to use `PostedMessageUnreliableData`?  If not, please add a comment explaining why it's not using PostedMessageUnreliableData.
</blockquote></details>

Yes, will fix.

jynnantonix (dismissed):
> Reviewed 2 of 2 files at r7, 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: 24 of 25 files reviewed, 3 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @jynnantonix, @Reisen, and @sekoje)

___
*[`solana/bridge/program/tests/common.rs` line 248 at r8](https://reviewable.io/reviews/certusone/wormhole/1077#-N5Z75Bm5CfxaJtakm2c:-N5Z75Bm5CfxaJtakm2d:borp542) ([raw file](https://github.com/certusone/wormhole/blob/b518f3411e23b92b76db4b744a57a9f3aed92542/solana/bridge/program/tests/common.rs#L248)):*
> ```XML
>             Some(keypair) => keypair,
>             None => new_message_pair
>         };
> ```

nit: this could be a one-liner



_Suggestion:_
```XML
        let message = message.unwrap_or(new_message_pair);
```
___
*[`solana/bridge/program/tests/common.rs` line 249 at r8](https://reviewable.io/reviews/certusone/wormhole/1077#-N5Z6vww52DLueDrRjia:-N5Z6vww52DLueDrRjib:b646cok) ([raw file](https://github.com/certusone/wormhole/blob/b518f3411e23b92b76db4b744a57a9f3aed92542/solana/bridge/program/tests/common.rs#L249)):*
> ```XML
>             None => new_message_pair
>         };
>             // Keypair::new();
> ```

nit: leftover comment

kcsongor (approved):

node/pkg/solana/client.go
kcsongor: Added this check

kcsongor (approved):

jynnantonix (approved):
> Reviewed 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1077)* status: all files reviewed, 3 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @Reisen, and @sekoje)

PR #1091, opened 2022-04-19
algo/contracts: Initial core and token_bridge algorand contracts for wormhole

jumpsiegel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1091)


algorand/portal_core.py
```diff
@@ -0,0 +1,550 @@
...
+                App.globalPut(Bytes("booted"), Bytes("true")),
+
+                # This opportunity comes once in a lifetime
+                checkForDuplicate(),
+
+                # You can do anything you set your mind to...
+                hdlGovernance(Int(1))
+            ])
+
+        def verifySigs():
```
kcsongor, line 310: what's the purpose of this function?

algorand/portal_core.py
```diff
@@ -0,0 +1,550 @@
...
+                    num_sigs.load() > ((total_guardians.load() * Int(2)) / Int(3)),
+                    )),
+
+
+                # Point it at the start of the signatures in the VAA
+                off.store(Int(6)),
+
+                For(
+                        i.store(Int(0)),
+                        i.load() <= Txn.group_index(),
```
kcsongor, line 401: Why `<=` here? Down below we have to now check
```
                                [a.load() == Bytes("verifyVAA"), Seq([])],
```
which is the `==` case. Might as well do `<` and remove that case? I don't think it's possible to have
```
verifySigs
verifySigs
verifyVAA
verifySigs
...
verifyVAA
```

algorand/portal_core.py
kcsongor: I think I understand... this is a dummy function that doesn't do anything... except that it has to be signed by the `vaa_verify` logic sig. So later, you check that the sender for this contract call is indeed the `vaa_verify` sig. Is that right?

jumpsiegel: Which function?  The verifySigs function?  all the processing is done in the logicSig so the code executed in the memPool doesn't have to do anything

algorand/portal_core.py
jumpsiegel: I want the same asserts executed on verifyVAA on lines 404-408 as I want for all the verifySigs.  If I make it just a <, then I'd have to copy those asserts to another spot to execute them for Txn.  Doing in this way saves code

jumpsiegel: Now as another interesting point, if we were to modify this to not start at 0 but instead to walk backwards from the current index...   then you would be "allowed" to put two wormhole ops into a single txn group so that you could do two operations transactionally...

kcsongor: That's sounds good, do you think it's worth switching over to that pattern? It would also mean better composition not just between wormhole txs but other contracts too

jumpsiegel: Yeah, except... 1) We'd have to tell the auditors about the change and they would  not like that we are changing the contract they have been auditing,  2) We would work backwards until we saw a function we didn't understand OR we have consumed all the signatures in the vaa?  lets chat after the meeting

algorand/portal_core.py
jumpsiegel: Yes, that is all correct

kcsongor (approved):
Still haven't finished reading through all the code, but we can merge it in, put into testnet, and keep iterating if necessary

PR #1102, opened 2022-04-22
sdk: improve type safety + add documentation

kcsongor:
The main goal of this PR is to improve the type safety of some of the sdk functionality.
- `isEVMChain` now performs type refinement. This way, we can achieve exhaustivity checking in downstream code that relies on this check.
- This in turn allows the address conversion functions to have safer types, along with exhaustivity checking. The end result is that when a new chain is added, these functions will raise a type error, indicating that the new chain ids are not handled. No `null` value is returned from using the functions in an intended way (that is, passing in well-typed arguments). This obviates the need for callers to coalesce null cases to values like `""` which results in a debugging nightmare.
- Added some return types, in particular some of these types were inferred to be things like `Promise<any>`, which is not that helpful.
- Added some documentation

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1102)

### DIFF
--- a/sdk/js/src/utils/array.ts
+++ b/sdk/js/src/utils/array.ts
@@ -102,7 +102,7 @@ export const tryHexToNativeAssetString = (h: string, c: ChainId): string =>
  * Convert an address in a wormhole's 32-byte hex representation into a chain's native
  * string representation.
  *
- * @deprecated since 0.2.8, use [[tryHexToNativeString]] instead.
+ * @deprecated since 0.3.0, use [[tryHexToNativeString]] instead.
  */
 export const hexToNativeAssetString = (
   h: string | undefined,
@@ -135,7 +135,7 @@ export const tryHexToNativeString = (
  * Convert an address in a wormhole's 32-byte hex representation into a chain's native
  * string representation.
  *
- * @deprecated since 0.2.8, use [[tryHexToNativeString]] instead.
+ * @deprecated since 0.3.0, use [[tryHexToNativeString]] instead.
  */
 export const hexToNativeString = (
   h: string | undefined,
@@ -197,7 +197,7 @@ export const tryNativeToHexString = (
  * Convert an address in a chain's native representation into a 32-byte hex string
  * understood by wormhole.
  *
- * @deprecated since 0.2.8, use [[tryNativeToHexString]] instead.
+ * @deprecated since 0.3.0, use [[tryNativeToHexString]] instead.
  * @throws if address is a malformed string for the given chain id
  */
 export const nativeToHexString = (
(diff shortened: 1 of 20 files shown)


sdk/js/src/utils/array.ts
```diff
@@ -1,72 +1,97 @@
...
+ * are encoded in the Portal payloads by the setting the first byte to 1.  This
+ * encoding is safe, because the first 12 bytes of the 32-byte wormhole address
+ * space are not used on Terra otherwise, as cosmos addresses are 20 bytes wide.
+ */
+export const isHexNativeTerra = (h: string): boolean => h.startsWith("01");
 
-export const isHexNativeTerra = (h: string) => h.startsWith("01");
-export const nativeTerraHexToDenom = (h: string) =>
+export const nativeTerraHexToDenom = (h: string): string =>
   Buffer.from(stripZeros(hexToUint8Array(h.substr(2)))).toString("ascii");
```
evan-gray, line 35: Should this now call nativeTerraUint8ArrayToDenom internally?

sdk/js/src/utils/array.ts
```diff
@@ -1,72 +1,97 @@
...
+ *
+ * Convert an address in a wormhole's 32-byte array representation into a chain's
+ * native string representation.
+ */
+export const uint8ArrayToNative = (
+  a: Uint8Array,
+  chainId: ChainId
+): string => {
+  if (isEVMChain(chainId)) {
+    return hexZeroPad(a, 20)
```
evan-gray, line 56: Is it safe to have removed the hex value here?

evan-gray (changes requested):
Nice uses of never! Three requests:
1. Can we add some verbose tests for the array utils?
2. Please update the CHANGELOG when making SDK changes. 
3. This will require a minor version bump. Can you please update the package to reflect this?

sdk/js/src/utils/array.ts
kcsongor: No, it's not... looks like `hexZeroPad` expects the string to start with `0x`...

kcsongor: oh, except that we now pass in a `Uint8Array` which it handles differently - so this is fine. But I'll add tests

evan-gray (approved):

sdk/js/src/utils/array.ts
```diff
@@ -105,13 +179,60 @@ export const nativeToHexString = (
...
+  const chainId = coalesceChainId(chain);
+  return hexToUint8Array(tryNativeToHexString(address, chainId));
+}
+
+/**
+ *
+ * Convert an address in a chain's native representation into a 32-byte hex string
+ * understood by wormhole.
+ *
+ * @deprecated since 0.2.8, use [[tryUint8ArrayToNative]] instead.
```
kcsongor, line 233: 0.3.0 here too

evan-gray (dismissed):

panoel (approved):

PR #1104, opened 2022-04-24
fix gas for NFT minting with Karura/Acala

shunjizhan:
## Change
When minting NFT we also need these custom gas params, since the default gas param's storageLimit isn't enough for an NFT, so added gas overrides for it.

## Test
Tested locally: 
- before the change: redeeming NFT will throw error: `outOfStorrage`
- after the change: NFT transferring works bidirectionally Karura <=> Goerli

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1104)


evan-gray (approved):
Looks good, thanks!

justinschuldt (approved):

PR #1110, opened 2022-04-25
cloud_functions: Split notional-tvl into read and write (to cache) entry points

kev1n-peters:
ComputeTVL is intended to be called periodically which writes the computed result to notional-tvl.json. TVL reads from notional-tvl.json and returns it as the response.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1110)


ckeun (approved):

justinschuldt (approved):
LGTM

PR #1114, opened 2022-04-27
node/cmd/guardiand: improve send-observation-request arg parser

leoluk:
Accept chain names (i.e. "solana") and base58 tx hashes.

---

**Stack**:
- #1115
- #1114 ⮜

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1114)

### DIFF
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -4,17 +4,20 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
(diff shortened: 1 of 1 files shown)


SEJeff (approved):

kcsongor (approved):
👍 
One nit, which is outside of the scope of this PR, is that the usage strings should say `<arg>` instead of `[arg]` since these arguments are required, not optional

PR #1115, opened 2022-04-27
node/pkg/processor: fix ignoring of existing VAAs

leoluk:
`k.Timestamp.Sub(existing.Timestamp)` is always zero
and this feature never worked. Fix that and make it
conditional on the guardian set being identical.

---

**Stack**:
- #1115 ⮜
- #1114

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1115)

### DIFF
--- a/node/pkg/processor/message.go
+++ b/node/pkg/processor/message.go
@@ -3,9 +3,11 @@ package processor
 import (
 	"context"
 	"encoding/hex"
+	"time"
+
+	"github.com/mr-tron/base58"
 
 	"github.com/certusone/wormhole/node/pkg/db"
-	"github.com/mr-tron/base58"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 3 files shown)


node/pkg/processor/observation.go
```diff
@@ -318,12 +319,24 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(ctx context.Context, m *gos
...
+		if err != nil {
+			// Undefined state: This can't happen unless the DB is corrupted.
+			panic(fmt.Sprintf("failed to unmarshal VAA: %v", err))
+		}
+
+		// Accept new VAA if the new one has a different guardian set index.
+		//
+		// This allows messages re-signed by a different guardian set to be persisted by nodes
+		// even on nodes that failed to re-observe the message.
+		if existing.GuardianSetIndex == v.GuardianSetIndex {
```
SEJeff, line 334: This is sort of a rhetorical question, but should the logic to re-sign only happen if the different guardian set index is higher than the existing one? I'm trying to think if it is possible for there to be hijinks with someone trying to use a lower index and think the answer is no, but just wanted to think about it.

node/pkg/processor/message.go
```diff
@@ -105,7 +107,11 @@ func (p *Processor) handleMessage(ctx context.Context, k *common.MessagePublicat
 			panic("failed to unmarshal VAA from db")
 		}
 
-		if k.Timestamp.Sub(existing.Timestamp) > settlementTime {
+		if time.Now().Sub(existing.Timestamp) > settlementTime &&
```
SEJeff, line 110: Could a guardian mess with things if they were to change their system time? Also, unlikely, just trying to consider all of the ways this could be used.

SEJeff (approved):
Code seems sensible. Just a few questions.

node/pkg/processor/observation.go
kcsongor: This is a potential DoS vector, say you have a signed VAA by guardian set N, then there's an upgrade to N+1 and a reobservation request for the VAA. Now, as long as a guardian holds on to the VAA signed by N, they can keep broadcasting it, kicking out the N+1 VAA from storage. Pretty unlikely event, but theoretically possible, so we should only override with newer guardian sets.

leoluk:

kcsongor (changes requested):
Can we make it so that the state is only overridden by newer guardian sets

leoluk:

PR #1155, opened 2022-05-05
bridge_ui: Added WalletConnect support for Ethereum

kev1n-peters:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1155)


bridge_ui/src/contexts/EthereumProviderContext.tsx
```diff
@@ -42,91 +58,201 @@ export const EthereumProviderProvider = ({
...
+          });
         }
-      })
-      .catch(() => {
-        setProviderError("Please install MetaMask");
-      });
+      } catch (error) {
+        console.error(error);
+      }
+      if (process.env.REACT_APP_INFURA_API_KEY) {
```
kev1n-peters, line 81: Maybe we should set this for Netlify builds?

panoel (dismissed):

bridge_ui/src/hooks/useIsWalletReady.ts
```diff
@@ -56,7 +64,18 @@ function useIsWalletReady(
...
+        EVM_RPC_MAP[correctEvmNetwork] === undefined
+      ) {
+        // WalletConnect requires a rpc url for this chain
+        // Force user to switch connect type
+        disconnect();
+        return;
+      }
+
       try {
+        console.log("switching");
```
panoel, line 78: Not sure where we are on log messages.  Doesn't bother me.  But a one word log may not provide much context.

panoel (approved):

evan-gray (commented):
Tested with Solana, Avalanche, and Goerli on testnet and seems to be working well! One thing I noticed was that events in MetaMask still fired even after it was disconnected (and swapped for WalletConnect) so we should probably remove all listeners on disconnect.

bridge_ui/src/hooks/useIsWalletReady.ts
evan-gray: agree, we should probably remove the log

evan-gray (approved):

panoel (approved):

PR #1161, opened 2022-05-10
Wormhole Receiver

evan-gray:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1161)


examples/receiver/evm/contracts/Setup.sol
```diff
@@ -0,0 +1,35 @@
...
+    ) public {
+        require(initialGuardians.length > 0, "no guardians specified");
+
+        setOwner(msg.sender);
+
+        Structs.GuardianSet memory initialGuardianSet = Structs.GuardianSet({
+            keys : initialGuardians,
+            expirationTime : 0
+        });
+
```
kcsongor, line 26: This doesn't set the chain id, which means that it will be initialised to 0. Even if we do want it to be 0, I think we should do it explicitly. I think reserving a high number for it would be better though (it's fine if that number is shared between all receiver contracts on all chains), chain id 0 still has some special meaning which could lead to surprises down the line.

kcsongor:
What's the difference between this contract and the original core contract? It seems like the only difference is the upgrade mechanism, whereby instead of upgrade VAAs, the contract can be upgraded by an owner instead of guardian VAAs. (We should make this explicit in the README.)

This makes sense, given that the idea is that guardians won't care to generate upgrade VAAs for this chain. This raises a question though: what if eventually the guardians decide to support the network properly? It sounds like it would mean that a completely separate contract will need to be deployed, which is slightly unfortunate. Maybe that can be worked around by upgrading to the Setup contract then reinitialising this one to ensure the state is zeroed out properly etc. But maybe that's too dangerous. 

An implementation detail: as it stands, this new contract copies all of the code from the existing core contract. Could we achieve this with inheritance instead? That would greatly reduce the maintenance burden of this contract. Something like

```
contract Receiver is Implementation {
 ... do ownership stuff here
}
```
where `Implementation` could just be imported from the existing path.

kcsongor (changes requested):

examples/receiver/evm/contracts/Setup.sol
evan-gray: What “chain id”?

evan-gray:
> What's the difference between this contract and the original core contract? It seems like the only difference is the upgrade mechanism, whereby instead of upgrade VAAs, the contract can be upgraded by an owner instead of guardian VAAs. (We should make this explicit in the README.)
> 
> This makes sense, given that the idea is that guardians won't care to generate upgrade VAAs for this chain. This raises a question though: what if eventually the guardians decide to support the network properly? It sounds like it would mean that a completely separate contract will need to be deployed, which is slightly unfortunate. Maybe that can be worked around by upgrading to the Setup contract then reinitialising this one to ensure the state is zeroed out properly etc. But maybe that's too dangerous.
> 
> An implementation detail: as it stands, this new contract copies all of the code from the existing core contract. Could we achieve this with inheritance instead? That would greatly reduce the maintenance burden of this contract. Something like
> 
> ```
> contract Receiver is Implementation {
>  ... do ownership stuff here
> }
> ```
> 
> where `Implementation` could just be imported from the existing path.

Will update the readme. This is intended to be an unofficial light client simply to parse and verify wormhole messages. It has no chain id and no directed governance. The only governance it accepts is guardian set upgrades. I don’t think it’s worthwhile to refactor any other contract at this time.

examples/receiver/evm/contracts/Setup.sol
kcsongor: hmm, I guess I got confused when I saw this: https://github.com/certusone/wormhole/pull/1161/files#diff-361bc6909361dc7c00584b3f67d164b2d33e2e8f5808f499bf63f35363c19d4dR35 but indeed it doesn't seem to be mentioned anywhere else. Looking at it closer, the whole interface file is dead code it seems?

evan-gray:
Closed in favor of adding to the examples repo.
https://github.com/certusone/wormhole-examples/tree/main/receiver/evm

PR #1165, opened 2022-05-10
bridge_ui: changed formatting for token picker

panoel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1165)

### DIFF
--- a/bridge_ui/src/utils/balancePretty.ts
+++ b/bridge_ui/src/utils/balancePretty.ts
@@ -1,0 +1,17 @@
+export const balancePretty = (uiString: string) => {
+  const numberString = uiString.split(".")[0];
+  const nsLen = numberString.length;
+  if (nsLen > 9) {
+    // Billion case
+    const num = numberString.substring(0, nsLen - 9);
+    const fract = numberString.substring(nsLen - 9, nsLen - 9 + 2);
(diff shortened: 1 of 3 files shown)


bridge_ui/src/utils/balancePretty.ts
```diff
@@ -0,0 +1,17 @@
...
+    // Billion case
+    const num = numberString.substring(0, nsLen - 9);
+    const fract = numberString.substring(nsLen - 9, nsLen - 9 + 2);
+    return num + "." + fract + " B";
+  } else if (nsLen > 6) {
+    // Million case
+    const num = numberString.substring(0, nsLen - 6);
+    const fract = numberString.substring(nsLen - 6, nsLen - 6 + 2);
+    return num + "." + fract + " M";
+  } else {
```
evan-gray, line 14: I think this loses the previous limitation that the string not be > 8 characters when less than 1 M (which saves us from nasty 18 decimal values. Could you add that conditional back?

evan-gray (changes requested):
Great change, like the tests, one small piece of functionality to restore. 🙂

evan-gray (approved):

kev1n-peters (approved):

PR #1169, opened 2022-05-11
algo/guardian v2

jumpsiegel:
… depending on Rewrite algorand guardian watcher to directly hit the node instead of depending on the indexer

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1169)


jumpsiegel:
We are switching the guardian over to directly hitting the node to retrieve its info instead of using the indexer.  It only uses the indexer on a obsRequest to track down the block of the txid supplied at which point it then goes to the local node to retrieve the data.

panoel (approved):

bruce-riley (approved):

PR #1170, opened 2022-05-11
EVM: replace `transfer()` with `call{value:}("")`

0x0ece:
Following [this recommendation](https://ethereum.stackexchange.com/a/19343), replacing `transfer()` with `call()` should slightly reduce the cost of the tx.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1170)

### DIFF
--- a/ethereum/contracts/bridge/Bridge.sol
+++ b/ethereum/contracts/bridge/Bridge.sol
@@ -86,7 +86,8 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
         // refund dust
         uint dust = amount - deNormalizeAmount(normalizedAmount, 18);
         if (dust > 0) {
-            payable(msg.sender).transfer(dust);
+            (bool dustSuccess, ) = payable(msg.sender).call{value:dust}("");
+            require(dustSuccess, "Transfer dust failed.");
         }
 
         // deposit into WETH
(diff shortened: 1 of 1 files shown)


calebrate:
.call() forwards all gas to the target contract and can lead to reentrancy attacks.
For transfers that are just supposed to transfer ether, not allow another contract to be executed its recommended to use .transfer()

calebrate:
.call() forwards all gas to the target contract and can lead to reentrancy attacks.
For transfers that are just supposed to transfer ether, not allow another contract to be executed its recommended to use .transfer()

From the [Solidity Documentation](https://docs.soliditylang.org/en/v0.8.17/security-considerations.html?highlight=.transfer#sending-and-receiving-ether) :

> There is a way to forward more gas to the receiving contract using addr.call{value: x}(""). This is essentially the same as addr.transfer(x), only that it forwards all remaining gas and opens up the ability for the recipient to perform more expensive actions (and it returns a failure code instead of automatically propagating the error). This might include calling back into the sending contract or other state changes you might not have thought of. So it allows for great flexibility for honest users but also for malicious actors.

calebrate:
.call() forwards all gas to the target contract and can lead to reentrancy attacks.
For transfers that are just supposed to transfer ether, not allow another contract to be executed its recommended to use .transfer()

From the [Solidity Documentation](https://docs.soliditylang.org/en/v0.8.17/security-considerations.html?highlight=.transfer#sending-and-receiving-ether) :

> There is a way to forward more gas to the receiving contract using addr.call{value: x}(""). This is essentially the same as addr.transfer(x), only that it forwards all remaining gas and opens up the ability for the recipient to perform more expensive actions (and it returns a failure code instead of automatically propagating the error). This might include calling back into the sending contract or other state changes you might not have thought of. So it allows for great flexibility for honest users but also for malicious actors.

PR #1183, opened 2022-05-13
node/repair: allow multiple sequences in a tx

kcsongor:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1183)

### DIFF
--- a/node/hack/repair_terra/repair.go
+++ b/node/hack/repair_terra/repair.go
@@ -23,13 +23,15 @@ import (
 	"github.com/certusone/wormhole/node/pkg/vaa"
 
 	"github.com/tidwall/gjson"
+	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 )
 
 var (
-	adminRPC   = flag.String("adminRPC", "/run/guardiand/admin.socket", "Admin RPC address")
-	terraAddr  = flag.String("terraProgram", "terra1dq03ugtd40zu9hcgdzrsq6z2z4hwhc9tqk2uy5", "Terra program address")
-	dryRun     = flag.Bool("dryRun", true, "Dry run")
+	adminRPC     = flag.String("adminRPC", "/run/guardiand/admin.socket", "Admin RPC address")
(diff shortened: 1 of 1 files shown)


leoluk (approved):

SEJeff (approved):

node/hack/repair_terra/repair.go
```diff
@@ -47,41 +47,42 @@ func getAdminClient(ctx context.Context, addr string) (*grpc.ClientConn, error,
...
-		return 0, fmt.Errorf("terra tx has no events")
+		return []uint64{}, fmt.Errorf("terra tx has no events")
 	}
 	msgs := EventsToMessagePublications(*terraAddr, txHash, events.Array())
 	// Should only ever be 1 message. Stole the above function from watcher.go
-	if len(msgs) != 1 {
-		return 0, fmt.Errorf("EventsToMessagePublications returned %d msgs", len(msgs))
+	var sequences = []uint64{}
+	for _, msg := range msgs {
+		sequences = append(sequences, msg.Sequence)
```
evan-gray, line 83: We should filter out emitters that don't match so we don't exit earlier on lower sequence numbers (or incorrectly match up the wrong one). @SEJeff to take a look 🙂

evan-gray (changes requested):

node/hack/repair_terra/repair.go
kcsongor: another thing we should do is make sure we only do a single reobservation request per transaction (even if it contains multiple sequence numbers)

evan-gray (approved):
Looks improved!

node/hack/repair_terra/repair.go
SEJeff: I implemented your suggestion @evan-gray in c3ae93dab87015d810baea41bcdad89c3c5066bf.

We can do your suggestion in a followup @kcsongor, but what we have works perfectly.

SEJeff (approved):
Tested and verified working

node/hack/repair_terra/repair.go
evan-gray: I agree we don’t need to change further right now 🙂

PR #1185, opened 2022-05-13
Klaytn mainnet deploy

bruce-riley:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1185)


evan-gray (approved):

SEJeff (approved):
Perhaps as a follow up we need to make a repair_klaytn that is native to their scope app.

PR #1187, opened 2022-05-13
Algo/abi:  adding changes to support abi method selector

jumpsiegel:
algorand has a newer standard : https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md

The change here is to make the called payload3 contract follow the newer standard so that the pyteal/stub stuff just "does the right thing"

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1187)

### DIFF
--- a/algorand/admin.py
+++ b/algorand/admin.py
@@ -21,7 +21,7 @@ from gentest import GenTest
 
 from algosdk.v2client.algod import AlgodClient
 from algosdk.kmd import KMDClient
-from algosdk import account, mnemonic
+from algosdk import account, mnemonic, abi
 from algosdk.encoding import decode_address, encode_address
 from algosdk.future import transaction
 from pyteal import compileTeal, Mode, Expr
@@ -1006,11 +1006,12 @@ class PortalCore:
                 txns[-1].fee = txns[-1].fee * 2
 
             if p["Meta"] == "TokenBridge Transfer With Payload":
+                m = abi.Method("portal_transfer", [abi.Argument("byte[]")], abi.Returns("byte[]"))
                 txns.append(transaction.ApplicationCallTxn(
                     sender=sender.getAddress(),
                     index=int.from_bytes(bytes.fromhex(p["ToAddress"]), "big"),
                     on_complete=transaction.OnComplete.NoOpOC,
-                    app_args=[b"completeTransfer", vaa],
+                    app_args=[m.get_selector(), m.args[0].type.encode(vaa)],
                     foreign_assets = foreign_assets,
                     sp=sp
                 ))
--- a/algorand/test/test.py
+++ b/algorand/test/test.py
@@ -134,10 +134,13 @@ class AlgoTest(PortalCore):
         while True:
             nexttoken = ""
             while True:
-                response = self.myindexer.search_transactions( min_round=self.INDEXER_ROUND, note_prefix=self.NOTE_PREFIX, next_page=nexttoken)
+                response = self.myindexer.search_transactions( min_round=self.INDEXER_ROUND, next_page=nexttoken)
 #                pprint.pprint(response)
                 for x in response["transactions"]:
 #                    pprint.pprint(x)
+                    if 'inner-txns' not in x:
+                        continue
+
                     for y in x["inner-txns"]:
                         if "application-transaction" not in y:
                             continue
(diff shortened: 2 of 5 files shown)


kcsongor:
could you extract this to a function like `encodeSelector` or something like that instead of hardcoding the hashes, it would be a little easier to read in the future

barnjamin:
I think we'll have to change one more thing, the ABI defines a dynamic array as having the length of the array encoded as a uint16 in the first 2 bytes so the second arg for the app call should be encoded as `uint16(len(vaa))|vaa`

algorand/test_contract.py
```diff
@@ -134,7 +134,7 @@ def mint():
         [METHOD == Bytes("test1"), test1()],
         [METHOD == Bytes("setup"), setup()],
         [METHOD == Bytes("mint"), mint()],
-        [METHOD == Bytes("completeTransfer"), completeTransfer()],
+        [METHOD == Bytes("base16", "0x903f4535"), completeTransfer()],
```
panoel, line 137: Is this a magic number for "completeTransfer"?

algorand/token_bridge.py
```diff
@@ -522,8 +522,8 @@ def completeTransfer():
                     tidx.store(Txn.group_index() + Int(1)),
                     MagicAssert(And(
                         Gtxn[tidx.load()].type_enum() == TxnType.ApplicationCall,
-                        Gtxn[tidx.load()].application_args[0] == Txn.application_args[0],
-                        Gtxn[tidx.load()].application_args[1] == Txn.application_args[1],
+                        Gtxn[tidx.load()].application_args[0] == Bytes("base16", "0x903f4535"), # sha256("portal_transfer(byte[])byte[]")[:4]
```
panoel, line 525: This is the same magic number as above.  Would life be easier for you if it was #defined?

panoel (approved):

sdk/js/src/algorand/Algorand.ts
```diff
@@ -871,9 +874,11 @@ export async function _submitVAAAlgorand(
     if (meta === "TokenBridge Transfer With Payload") {
       txs[txs.length - 1].tx.appForeignApps = [aid];
 
+      let m = abi.ABIMethod.fromSignature("portal_transfer(byte[])byte[]").getSelector();
+      
       txs.push({
         tx: makeApplicationCallTxnFromObject({
-          appArgs: [textToUint8Array("completeTransfer"), vaa],
+          appArgs: [m, concatUint8Arrays(bigIntToBytes(vaa.length, 2), vaa)],
```
barnjamin, line 881: Should be possible to do something like
```ts
m.args[0].encode(vaa)
```

algorand/test_contract.py
```diff
@@ -33,6 +33,9 @@
 
 import sys
 
+# sha256("portal_transfer(byte[])byte[]")[:4]
+portal_transfer_selector = Bytes("base16", "0x903f4535")
```
barnjamin, line 37: Can have something similar to the js-sdk one where you construct from Method from method signature and get selector

algorand/test/test.py
```diff
@@ -134,10 +134,13 @@ def getVAA(self, client, sender, sid, app):
         while True:
             nexttoken = ""
             while True:
-                response = self.myindexer.search_transactions( min_round=self.INDEXER_ROUND, note_prefix=self.NOTE_PREFIX, next_page=nexttoken)
+                response = self.myindexer.search_transactions( min_round=self.INDEXER_ROUND, next_page=nexttoken)
```
barnjamin, line 137: tweaked this line for my own env, might not be what you want?

jumpsiegel: It does not hurt since it is test net only.. just makes things slower

panoel (approved):

kcsongor (approved):

algorand/admin.py
```diff
@@ -21,7 +21,7 @@
 
 from algosdk.v2client.algod import AlgodClient
 from algosdk.kmd import KMDClient
-from algosdk import account, mnemonic
```
LToxiCs, line 24: Mnemonic

algorand/test_contract.py
```diff
@@ -134,7 +136,7 @@ def mint():
         [METHOD == Bytes("test1"), test1()],
         [METHOD == Bytes("setup"), setup()],
         [METHOD == Bytes("mint"), mint()],
-        [METHOD == Bytes("completeTransfer"), completeTransfer()],
+        [METHOD == portal_transfer_selector, completeTransfer()],
```
LToxiCs, line 139: [METHOD == account_transfer_complete,()],

PR #1205, opened 2022-05-19
Add unit-test coverage for marshaling

claudijd:
This PR does two things:

- Fixes a bug in the existing test VAA helper to initialize Signatures as non-nil (so marshalling and unmarshalling are exact matches).
- Adds unit-tests for both Marshal and Unmarshal operations

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1205)


SEJeff (approved):
> Fixes a bug in the VAA helper to initialize Signatures as non-nil (so marshalling and unmarshalling are exact matches)

Fixes a bug in the existing test VAA helper to initialize Signatures as non-nil (so marshalling and unmarshalling are exact matches).

For a moment I thought there was a bug this fixed in the actual vaa marshalling code.

node/pkg/vaa/structs_test.go
```diff
@@ -207,6 +207,42 @@ func TestHexDigest(t *testing.T) {
...
+	vaa.Payload = payload
+
+	// Let's marshal the VAA to bytes to unmarshaled
+	marshalBytes, err := vaa.Marshal()
+	assert.Nil(t, err)
+
+	// Let's now unmarshal the oversized VAA and cause it to panic
+	vaa2, err2 := Unmarshal(marshalBytes)
+	assert.Nil(t, vaa2)
+	assert.NotNil(t, err2)
```
claudijd, line 243: This should fail, seeking feedback on the desired behavior before I adjust the implementation.

node/pkg/vaa/structs.go
```diff
@@ -271,11 +271,15 @@ func Unmarshal(data []byte) (*VAA, error) {
 		return nil, fmt.Errorf("failed to read commitment: %w", err)
 	}
 
-	payload := make([]byte, 1000)
+	payload := make([]byte, 1001)
```
claudijd, line 274: @hendrikhofstadt can you provide some context on why this was set to 1000?

claudijd: Feedback I heard on this was that Solana has a limit of 1200 or so, so this seems like the rationale.

evan-gray (changes requested):
Having tested creating a payload larger than 1000 gets emitted, I'm pretty sure this unmarshall code is only used by the go code to read things that aren't the payload. We should likely put a warning comment _not_ throw an error.

claudijd:
@evan-gray fixed per request, makes sense.

evan-gray (approved):

node/pkg/vaa/structs.go
```diff
@@ -203,6 +203,10 @@ const (
 )
 
 // Unmarshal deserializes the binary representation of a VAA
+//
+// WARNING: Unmarshall will truncate payloads at 1000 bytes, this is done mainly to avoid denial of service
+//   - If you need to access the full payload, consider parsing VAA from Bytes instead of Unmarshal
+//
```
SEJeff, line 209: If we are truncating stuff, does it make sense to add a log? I'm not sure.

node/pkg/vaa/structs_test.go
```diff
@@ -207,6 +207,48 @@ func TestHexDigest(t *testing.T) {
...
+	assert.Equal(t, &vaa1, vaa2)
+}
+
+func TestUnmarshalTooBig(t *testing.T) {
+	vaa := getVaa()
+
+	// Overwrite an oversized payload for the VAA that we cannot unmarshal
+	var payload []byte
+	for i := 0; i < 2000; i++ {
+		payload = append(payload, 'a')
```
SEJeff, line 232: infinite screaming: aaaaaaaaaaaaaaaaaaaaaaaaaaa

SEJeff (approved):

node/pkg/vaa/structs.go
claudijd: going to pass on this now in the primative, I'd like to make a sweep once we have better monitoring setup, and then make this more robust.

PR #1207, opened 2022-05-19
Ensure that we return if we fail to store the VAA in handleObservation

claudijd:
If we fail to store the signedVAA in the DB, we shouldn't continue further, right?

This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1207)

### DIFF
--- a/node/pkg/processor/observation.go
+++ b/node/pkg/processor/observation.go
@@ -233,6 +233,7 @@ func (p *Processor) handleObservation(ctx context.Context, m *gossipv1.SignedObs
 
 			if err := p.db.StoreSignedVAA(signed); err != nil {
 				p.logger.Error("failed to store signed VAA", zap.Error(err))
+				return
 			}
 
 			p.broadcastSignedVAA(signed)


kcsongor:
I don't see a reason why we shouldn't continue - if for whatever reason, the VAA can't be stored, it's still better to broadcast it so other guardians can store it. The condition for quorum doesn't require that all guardians store the VAA, only that they sign it. cc @leoluk

SEJeff (commented):
I'm not actually sure if this is the right behavior or not.

After logging it was signed with quorum, it fails to get saved to the db, and you choose NOT to broadcast the fact that it was signed. It was observed, but failed to write to the db. The find-missing-messages admin command has a `--backfill` flag that I believe will update the db from public rpc:

https://github.com/certusone/wormhole/blob/594674db5ef3b6712db7c839f8aec60e557becad/node/cmd/guardiand/adminclient.go#L44-L45

So I think the existing code might be a bit weird, but probably is the right thing. What do you think?

tbjump:
I'm thinking that it might not make a big difference either way? Certainly no security implications. If everything else works properly, other guardians will also aggregate enough signatures for a quorum for that VAA, broadcast the SignedVAAWithQuorum, and then we get another shot at persisting it to the DB in observation.go:handleInboundSignedVAAWithQuorum

kcsongor:
Sounds like we should close this PR. Thoughts @claudijd?

PR #1213, opened 2022-05-20
evm: Make finality a state variable

kcsongor:
The main thing to review is the finalities assigned to chain ids in `BridgeImplementation.sol` and `NFTBridgeImplementation.sol`.

This is what they are:
```
    | chain name | chain id | finality |
    |------------+----------+----------|
    | ethereum   |        2 |       15 |
    | binance    |        4 |       15 |
    | polygon    |        5 |       15 |
    | avalanche  |        6 |        1 |
    | oasis      |        7 |        1 |
    | aurora     |        9 |        1 |
    | fantom     |       10 |        1 |
    | karura     |       11 |        1 |
    | acala      |       12 |        1 |
    | klaytn     |       13 |        1 |
    | celo       |       14 |        1 |
    | moonbeam   |       16 |        1 |
    | neon       |       17 |       32 |
```

This modification introduces a new state variable, which means it needs a custom `initialize()` function. This also means that the upgrade has to be done on every single supported contract (including testnet contracts) so that the field gets initialised correctly everywhere. For new deploys, we set this in the `Setup` logic.

Starting with this patch, new contract upgrades will no longer need to be special cased based on whether the chain is 15 conf or 1 conf (or other). Once this patch has landed on every chain, the custom `initialize()` should be removed (though it is idempotent).

NOTE: the `finality` variable is added to the `Provider` struct, next to similar variables. Luckily, there was enough space in the first storage slot to fit this in (as verified by the ethereum-upgrade test).

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1213)


ethereum/contracts/nft/NFTBridgeImplementation.sol
```diff
@@ -17,6 +17,82 @@ contract NFTBridgeImplementation is NFTBridge {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
+        uint8 finality;
+        uint256 chain = block.chainid;
+
+        // Chain ids explicitly enumerated (from truffle-config.js)
+
+        if (chain == 1) {
```
bruce-riley, line 25: This might be cleaner if it used wormhole chain ID.

bruce-riley (dismissed):

clients/js/evm.ts
```diff
@@ -30,6 +30,7 @@ export async function query_contract_evm(
 
   switch (module) {
     case "Core":
+      // TODO: add finality (need new sdk release)
```
evan-gray, line 33: why is this TODO under "Core"?

ethereum/contracts/bridge/BridgeState.sol
```diff
@@ -9,6 +9,8 @@ contract BridgeStorage {
     struct Provider {
         uint16 chainId;
         uint16 governanceChainId;
+        // Required number of block confirmations to assume finality
+        uint8 finality;
```
evan-gray, line 13: does this not affect the existing state by adding this in the middle?

evan-gray (changes requested):
Please test the getter in the `should be initialized with the correct signers and values` test for bridge and nft

ethereum/contracts/bridge/BridgeState.sol
kcsongor: see the note in the PR description (and the upgrade test)

ethereum/contracts/bridge/BridgeImplementation.sol
```diff
@@ -17,6 +17,28 @@ contract BridgeImplementation is Bridge {
...
+        } else if (chain == 6)  { finality = 1;  // avalanche
+        } else if (chain == 7)  { finality = 1;  // oasis
+        } else if (chain == 9)  { finality = 1;  // aurora
+        } else if (chain == 10) { finality = 1;  // fantom
+        } else if (chain == 11) { finality = 1;  // karura
+        } else if (chain == 12) { finality = 1;  // acala
+        } else if (chain == 13) { finality = 1;  // klaytn
+        } else if (chain == 14) { finality = 1;  // celo
+        } else if (chain == 16) { finality = 1;  // moonbeam
+        } else if (chain == 17) { finality = 1;  // neon
```
bruce-riley, line 36: I think Neon should be 32.

ethereum/contracts/nft/NFTBridgeImplementation.sol
```diff
@@ -17,6 +17,28 @@ contract NFTBridgeImplementation is NFTBridge {
...
+        } else if (chain == 6)  { finality = 1;  // avalanche
+        } else if (chain == 7)  { finality = 1;  // oasis
+        } else if (chain == 9)  { finality = 1;  // aurora
+        } else if (chain == 10) { finality = 1;  // fantom
+        } else if (chain == 11) { finality = 1;  // karura
+        } else if (chain == 12) { finality = 1;  // acala
+        } else if (chain == 13) { finality = 1;  // klaytn
+        } else if (chain == 14) { finality = 1;  // celo
+        } else if (chain == 16) { finality = 1;  // moonbeam
+        } else if (chain == 17) { finality = 1;  // neon
```
bruce-riley, line 36: 32 again.

bruce-riley (approved):

bruce-riley (approved):

PR #1226, opened 2022-05-24
Solana fix warnings

kcsongor:
`rustc` raises a lot of warnings when building the solana contracts. The first commit fixes them, most of them being redundant imports.
The second commit fixes all of the `clippy` hints. Most of them are about using `write_all` instead of `write` and spurious lifetime parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1226)

### DIFF
--- a/solana/modules/token_bridge/client/src/main.rs
+++ b/solana/modules/token_bridge/client/src/main.rs
@@ -1,6 +1,5 @@
-
+#![allow(incomplete_features)]
 #![feature(adt_const_params)]
-#![allow(warnings)]
 
 use std::{
     fmt::Display,
@@ -8,7 +7,6 @@ use std::{
     process::exit,
 };
 
-use borsh::BorshDeserialize;
 use clap::{
     crate_description,
     crate_name,
@@ -19,7 +17,6 @@ use clap::{
     Arg,
     SubCommand,
 };
-use hex;
 use solana_clap_utils::{
     input_parsers::{
         keypair_of,
@@ -36,26 +33,22 @@ use solana_client::{
     rpc_client::RpcClient,
     rpc_config::RpcSendTransactionConfig,
 };
-use solana_program::account_info::AccountInfo;
 use solana_sdk::{
     commitment_config::{
         CommitmentConfig,
         CommitmentLevel,
     },
     native_token::*,
-    program_error::ProgramError::AccountAlreadyInitialized,
     pubkey::Pubkey,
     signature::{
         read_keypair_file,
         Keypair,
         Signer,
     },
-    system_instruction::transfer,
     transaction::Transaction,
 };
 use solitaire::{
     processors::seeded::Seeded,
-    AccountState,
     Info,
 };
 use solitaire_client::Derive;
@@ -70,6 +63,10 @@ struct Config {
 type Error = Box<dyn std::error::Error>;
 type CommmandResult = Result<Option<Transaction>, Error>;
 
+// [`get_recent_blockhash`] is deprecated, but devnet deployment hangs using the
+// recommended method, so allowing deprecated here. This is only the client, so
+// no risk.
+#[allow(deprecated)]
 fn command_init_bridge(config: &Config, bridge: &Pubkey, core_bridge: &Pubkey) -> CommmandResult {
     println!("Initializing Token bridge {}", bridge);
 
@@ -79,13 +76,13 @@ fn command_init_bridge(config: &Config, bridge: &Pubkey, core_bridge: &Pubkey) -
 
     let ix = token_bridge::instructions::initialize(*bridge, config.owner.pubkey(), *core_bridge)
         .unwrap();
-    println!("config account: {}, ", ix.accounts[1].pubkey.to_string());
+    println!("config account: {}, ", ix.accounts[1].pubkey);
     let mut transaction = Transaction::new_with_payer(&[ix], Some(&config.fee_payer.pubkey()));
 
     let (recent_blockhash, fee_calculator) = config.rpc_client.get_recent_blockhash()?;
     check_fee_payer_balance(
         config,
-        minimum_balance_for_rent_exemption + fee_calculator.calculate_fee(&transaction.message()),
+        minimum_balance_for_rent_exemption + fee_calculator.calculate_fee(transaction.message()),
     )?;
     transaction.sign(&[&config.fee_payer, &config.owner], recent_blockhash);
     Ok(Some(transaction))
@@ -127,7 +124,7 @@ fn command_create_meta(
     );
     let mut transaction = Transaction::new_with_payer(&[ix], Some(&config.fee_payer.pubkey()));
 
-    let (recent_blockhash, _) = config.rpc_client.get_recent_blockhash()?;
+    let recent_blockhash = config.rpc_client.get_latest_blockhash()?;
     transaction.sign(&[&config.fee_payer, &config.owner], recent_blockhash);
     Ok(Some(transaction))
 }
@@ -146,7 +143,7 @@ fn main() {
                 .global(true)
                 .help("Configuration file to use");
             if let Some(ref config_file) = *solana_cli_config::CONFIG_FILE {
-                arg.default_value(&config_file)
+                arg.default_value(config_file)
             } else {
                 arg
             }
@@ -338,7 +335,8 @@ fn main() {
             )
             .0;
             let meta_info = config.rpc_client.get_account(&meta_acc).unwrap();
-            let meta_info = spl_token_metadata::state::Metadata::from_bytes(&meta_info.data).unwrap();
+            let meta_info =
+                spl_token_metadata::state::Metadata::from_bytes(&meta_info.data).unwrap();
             println!("Key: {:?}", meta_info.key);
             println!("Mint: {}", meta_info.mint);
             println!("Metadata Key: {}", meta_acc);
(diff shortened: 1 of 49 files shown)


kcsongor:
Interesting that there's a failing test 🤔 will investigate

solana/modules/token_bridge/program/src/messages.rs
```diff
@@ -386,6 +372,7 @@ impl DeserializeGovernancePayload for GovernancePayloadUpgrade {
 }
 
 #[cfg(feature = "no-entrypoint")]
+#[allow(unused_imports)]
```
jynnantonix, line 376: Why not remove the unused imports here as well?

solana/modules/token_bridge/program/src/instructions.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::too_many_arguments)]
```
jynnantonix, line 1: If there aren't a lot of functions where this applies we can put this right next to those functions rather than allowing it for the entire crate.

solana/modules/token_bridge/program/src/messages.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::manual_memcpy)]
```
jynnantonix, line 1: Same here.  Probably better to put this next to the call site so that it can be removed if we ever change it.

jynnantonix (approved):
Thanks a lot for doing this.  It looks great!

We still have `#![allow(warnings)]` in a few places (solitaire, clients, token-metadata).  Do we want to fix those here as well?

solana/modules/token_bridge/program/src/messages.rs
kcsongor: This was a weird case, the compiler was raising a lot of spurious unused import warnings in the test code.

kcsongor:
> We still have `#![allow(warnings)]` in a few places (solitaire, clients, token-metadata).  Do we want to fix those here as well?

Absolutely! I missed those, but I'll get them too, thanks.

jynnantonix (commented):
> Reviewed 21 of 29 files at r1, 26 of 26 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1226)* status: all files reviewed, 7 unresolved discussions (waiting on @evan-gray, @hendrikhofstadt, @kcsongor, and @Reisen)

___
*[`solana/bridge/client/src/main.rs` line 75 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N338GHs33kRkFA5kPJO:-N338GHs33kRkFA5kPJP:bbuz6el) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/bridge/client/src/main.rs#L75)):*
> ```XML
> type CommmandResult = Result<Option<Transaction>, Error>;
> 
> // [`get_recent_blockhash`] is deprecated, but devnet deployment hangs using the
> ```

https://goomics.net/50/ 



_Code quote:_
```XML
// [`get_recent_blockhash`] is deprecated, but devnet deployment hangs using the
// recommended method, so allowing deprecated here. This is only the client, so
```
___
*[`solana/modules/nft_bridge/program/src/messages.rs` line 127 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33J07j0J_GRCIqAlMj:-N33J07j0J_GRCIqAlMk:b-nbfhjv) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/nft_bridge/program/src/messages.rs#L127)):*
> ```XML
>         for i in 0..self.symbol.len() {
>             symbol[i] = self.symbol.as_bytes()[i];
>         }
> ```

This seems like a legitimate issue because the current code has a risk of panicking with an out-of-bounds access if `self.symbol.len()` or `self.name.len()` is longer than 32 bytes.  Rather than a loop we should use `copy_from_slice`:



_Suggestion:_
```XML
        let mut symbol: [u8; 32] = [0; 32];
        let count = cmp::min(symbol.len(), self.symbol.len());
        symbol[..count].copy_from_slice(self.symbol[..count].as_bytes());
```
___
*[`solana/modules/nft_bridge/program/tests/common.rs` line 671 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33DtD931_dKzVZmUm2:-N33DtD931_dKzVZmUm3:bp48xfu) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/nft_bridge/program/tests/common.rs#L671)):*
> ```XML
>     }
> 
>     #[allow(clippy::too_many_arguments)]
> ```

I'm personally not a fan of this lint so I'd also be ok with turning it off globally via a clippy.toml file in the workspace root.
___
*[`solana/modules/nft_bridge/program/tests/integration.rs` line 218 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33EvBR-uDCPRwwk8qq:-N33EeQV4NrVTsoy8hOt:btvhcpq) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/nft_bridge/program/tests/integration.rs#L218)):*
> ```XML
>         mint_authority: _,
>         mint: _,
>         token_authority: _,
> ```

nit: I think we can also just delete these lines and this should still work (because of the `..` on the line below).
___
*[`solana/modules/token_bridge/program/src/messages.rs` line 266 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33FFqX9T5Bqn1f9o18:-N33FFqX9T5Bqn1f9o19:b-s3c1d3) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/token_bridge/program/src/messages.rs#L266)):*
> ```XML
>         for i in 0..self.symbol.len() {
>             symbol[i] = self.symbol.as_bytes()[i];
>         }
> ```

Same comment here as in the other file:  this will panic if `self.symbol.len() > 32` so we should replace it with a `copy_from_slice` instead.
___
*[`solana/modules/token_bridge/program/tests/integration.rs` line 1 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33FVd13WIw1FjIlKHN:-N33FVd13WIw1FjIlKHO:b-fzsxnr) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/token_bridge/program/tests/integration.rs#L1)):*
> ```XML
> #![allow(warnings)]
> ```

Can this be removed now?
___
*[`solana/modules/token_bridge/token-metadata/src/lib.rs` line 2 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33FinG6QuDc99xqz_e:-N33FinG6QuDc99xqz_f:b-bbbg7z) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/modules/token_bridge/token-metadata/src/lib.rs#L2)):*
> ```XML
> // The solana_program::declare_id! macro generates spurious import statements.
> #![allow(unused_imports)]
> ```

This will disable this warning for the entire crate.  I think we can just do `#[allow(unused_imports)]` (without the `!`) right above the `declare_id!` line to suppress the warnings for the macro but keep them enabled everywhere else.
___
*[`solana/solitaire/program/src/error.rs` line 75 at r2](https://reviewable.io/reviews/certusone/wormhole/1226#-N33GZCx6iS5RSBWMEYC:-N33GZCx6iS5RSBWMEYD:boszbn3) ([raw file](https://github.com/certusone/wormhole/blob/1018e135649818144a4864a9f2f1eb46bb2bf099/solana/solitaire/program/src/error.rs#L75)):*
> ```XML
>         }
>     }
> }
> ```

I think this one might be legitimate as well.  Implementing `From` automatically gets you an `Into` impl via the [blanket impl in the standard library](https://doc.rust-lang.org/std/convert/trait.Into.html#impl-Into%3CU%3E) but not the reverse.  The usual convention is to implement `From` for your type but use `Into` in trait bounds.

I think the From impl for this should be pretty straightforward:



_Suggestion:_
```XML
impl From<SolitaireError> for ProgramError {
    fn from(err: SolitaireError) -> ProgramError {
        match err {
            SolitaireError::ProgramError(e) => e,
            _ => ProgramError::Custom(0),
        }
    }
}
```

jynnantonix (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1226)* status: all files reviewed, 1 unresolved discussion (waiting on @evan-gray, @hendrikhofstadt, @kcsongor, and @Reisen)

claudijd (approved):

solana/modules/token_bridge/client/src/main.rs
```diff
@@ -70,6 +63,10 @@ struct Config {
 type Error = Box<dyn std::error::Error>;
 type CommmandResult = Result<Option<Transaction>, Error>;
 
+// [`get_recent_blockhash`] is deprecated, but devnet deployment hangs using the
```
Reisen, line 66: Might be worth rewording the comment to suggest it only _may_ hang in devnet as the coode actually does have the updated method below.

Reisen (approved):

PR #1228, opened 2022-05-25
Add more unit-tests for attestation_events

claudijd:
Supercedes: https://github.com/certusone/wormhole/pull/1040

This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1228)


SEJeff (commented):
It's not immediately obvious to me without digging through the non-test code what those [magic numbers](https://www.codebyamir.com/blog/software-anti-patterns-magic-numbers) are for.

Also if those ids are the same every time, are they really unique? 😄

PR #1232, opened 2022-05-27
Guardian parse transfer VAA payload

bruce-riley:
Change-Id: I23136807e5101ae823e9d198ac52d3446a94b0f2

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1232)

### DIFF
--- a/node/pkg/vaa/structs.go
+++ b/node/pkg/vaa/structs.go
@@ -500,11 +500,17 @@ func MustWrite(w io.Writer, order binary.ByteOrder, data interface{}) {
 
 // StringToAddress converts a hex-encoded adress into a vaa.Address
 func StringToAddress(value string) (Address, error) {
+	if len(value) >= 2 && value[0] == '0' && value[1] == 'x' {
+		value = value[2:]
+	}
 	var address Address
 	res, err := hex.DecodeString(value)
 	if err != nil {
 		return address, err
 	}
-	copy(address[:], res)
+	if len(res) > 32 {
+		return address, fmt.Errorf("value must be no more than 32 bytes")
+	}
+	copy(address[32-len(res):], res)
 	return address, nil
 }
--- a/node/pkg/vaa/structs.go
+++ b/node/pkg/vaa/structs.go
@@ -502,17 +502,29 @@ func MustWrite(w io.Writer, order binary.ByteOrder, data interface{}) {
 
 // StringToAddress converts a hex-encoded address into a vaa.Address
 func StringToAddress(value string) (Address, error) {
-	if len(value) >= 2 && value[0] == '0' && value[1] == 'x' {
+	var address Address
+
+	// Make sure we have enough to decode
+	if len(value) < 2 {
+		return address, fmt.Errorf("value must be at least 1 byte")
+	}
+
+	// Trim any preceding "0x" to the address
+	if value[0:2] == "0x" {
 		value = value[2:]
 	}
-	var address Address
+
+	// Decode the string from hex to binary
 	res, err := hex.DecodeString(value)
 	if err != nil {
 		return address, err
 	}
+
+	// Make sure we don't have too many bytes
 	if len(res) > 32 {
 		return address, fmt.Errorf("value must be no more than 32 bytes")
 	}
 	copy(address[32-len(res):], res)
+
 	return address, nil
 }
--- a/node/pkg/vaa/structs.go
+++ b/node/pkg/vaa/structs.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"math/big"
 	"strings"
 	"time"
 
@@ -58,6 +59,16 @@ type (
 	}
 
 	SignatureData [65]byte
+
+	TransferPayloadHdr struct {
+		Type         uint8
+		TokenChainID ChainID
+		TokenAddress Address
+		ToChainID    ChainID
+		ToAddress    Address
+		Amount       *big.Int
+		Fee          *big.Int
+	}
 )
 
 func (a Address) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


bruce-riley:
This will be used for the chain governor feature.

node/pkg/vaa/structs.go
```diff
@@ -58,6 +59,16 @@ type (
...
 	SignatureData [65]byte
+
+	TransferPayloadHdr struct {
+		Type         uint8
+		TokenChainID ChainID
+		TokenAddress Address
+		ToChainID    ChainID
+		ToAddress    Address
+		Amount       *big.Int
+		Fee          *big.Int
```
evan-gray, line 70: Probably don't need / won't want the fee field after the payload 3 update, and I don't think we planned on using it anyway.

node/pkg/vaa/structs.go
```diff
@@ -415,6 +426,65 @@ func (v *VAA) AddSignature(key *ecdsa.PrivateKey, index uint8) {
 	})
 }
 
+func (v *VAA) IsTransfer() bool {
+	return len(v.Payload) > 0 && (v.Payload[0] == 1) || (v.Payload[0] == 3)
```
evan-gray, line 430: Not sure of the context but you probably need to check that the emitter chain + contract is a token bridge. 🤔 not sure if we _explicitly_ have that available, but we could make it so.

evan-gray (changes requested):

node/pkg/vaa/structs.go
bruce-riley: I removed the fee.

node/pkg/vaa/structs.go
bruce-riley: In my case, I will not get this far if it's not from the token bridge. I don't want to add the complexity of making that check here (since it would mean listing all of the token bridge addresses just for that). Instead, I just added a comment saying that the caller should ensure this is only called for a VAA from the token bridge.

justinschuldt:
`amount` in an Ethereum payload is [uint256](https://github.com/certusone/wormhole/blob/dev.v2/ethereum/contracts/bridge/BridgeStructs.sol#L10-L11), I'm not sure what would happen with this code if it tried to process a number larger than `big.Int`, I think it would either throw an exception, or drop a bunch of digits. 

I used the `github.com/holiman/uint256` package when [dealing with amounts in cloud functions](https://github.com/certusone/wormhole/blob/dev.v2/event_database/cloud_functions/process-vaa.go#L320-L324), however that implementation just checks if the amount overflows `unit64`.

node/pkg/vaa/structs.go
```diff
@@ -424,11 +490,17 @@ func MustWrite(w io.Writer, order binary.ByteOrder, data interface{}) {
 
 // StringToAddress converts a hex-encoded adress into a vaa.Address
```
claudijd, line 491: typo on comments, not yours, but we should fix while we're here.

node/pkg/vaa/structs.go
```diff
@@ -415,20 +425,100 @@ func (v *VAA) AddSignature(key *ecdsa.PrivateKey, index uint8) {
...
+
+	// Make sure we have enough to decode
+	if len(value) < 2 {
+		return address, fmt.Errorf("value must be at least 1 byte")
+	}
+
+	// Trim any preceding "0x" to the address
+	if value[0:2] == "0x" {
+		value = value[2:]
+	}
```
claudijd, line 509: The previous logic was overloading a condition where you could supply a sub-1 byte hex string which would trip on `hex.DecodeString() ` step.

https://github.com/certusone/wormhole/blob/fa49e2a375a721a2715cd6f1e1c0de03291ab8be/node/pkg/vaa/structs.go#L493-L495

node/pkg/vaa/structs.go
```diff
@@ -415,20 +425,100 @@ func (v *VAA) AddSignature(key *ecdsa.PrivateKey, index uint8) {
...
+	if len(payload) > 0 {
+		if payload[0] == 1 || payload[0] == 3 {
+			return true
+		}
+	}
+
+	return false
+}
+
+func DecodeTransferPayloadHdr(payload []byte) (*TransferPayloadHdr, error) {
```
claudijd, line 439: I am curious where we plan to start calling DecodeTransferPayloadHdr, as this will have potentially more implications into how robust this parsing needs to be and what the failure conditions will be, but we can of course settle that once we bake these primitives out.

node/pkg/vaa/structs_test.go
```diff
@@ -488,8 +489,157 @@ func TestVerifySignaturesFuzz(t *testing.T) {
...
+		result  bool
+	}
+
+	tests := []Test{
+		{label: "empty", payload: []byte{}, result: false},
+		{label: "non-valid payload", payload: []byte{0x0}, result: false},
+		{label: "payload 1", payload: []byte{0x1}, result: true},
+		{label: "payload 2", payload: []byte{0x2}, result: false},
+		{label: "payload 3", payload: []byte{0x3}, result: true},
+		{label: "payload 4", payload: []byte{0x4}, result: false},
```
claudijd, line 637: Would there be cases where we'd expect that a Wormhole message might be using a payload type and providing malicious inputs here to muck with the stats or do we only expect payloads at this point to be fully trusted?

I'm thinking if this contributes to a limit, can I forge arbitrary payloads to run the numbers up?  If so, how should/would we guard against that?

node/pkg/vaa/structs.go
bruce-riley: This will be used in the chain governor to determine if the VAA should be limited. You can definitely throw rocks at it when we get there.

claudijd (dismissed):

node/pkg/vaa/structs.go
```diff
@@ -58,6 +59,15 @@ type (
...
 
 	SignatureData [65]byte
+
+	TransferPayloadHdr struct {
+		Type         uint8
+		TokenChainID ChainID
+		TokenAddress Address
+		ToChainID    ChainID
+		ToAddress    Address
+		Amount       *big.Int
```
kcsongor, line 69: Shouldn't this be a https://github.com/holiman/uint256 instead? (we've used this type elsewhere)

kcsongor: it would be nicer if these fields appeared in the same order as they do in the binary payload

node/pkg/vaa/structs_test.go
```diff
@@ -491,8 +492,161 @@ func TestVerifySignaturesFuzz(t *testing.T) {
...
+					assert.Equal(t, 133, len(vaa.Payload))
+
+					payload, err := DecodeTransferPayloadHdr(vaa.Payload)
+					assert.Nil(t, err)
+					assert.Equal(t, tc.payloadType, payload.Type)
+					assert.Equal(t, tc.tokenChainId, payload.TokenChainID)
+					assert.Equal(t, expectedTokenAddr, payload.TokenAddress)
+					assert.Equal(t, tc.toChainId, payload.ToChainID)
+					assert.Equal(t, expectedToAddr, payload.ToAddress)
+					assert.Equal(t, expectedAmount.Cmp(payload.Amount), 0)
```
SEJeff, line 619: I almost wonder if this wouldn't make more sense to just have something more like:

```
assert.Equal(t.payload, tc, payload)
```

You'd need to change the Test struct to be dramatically simplified then.

SEJeff (dismissed):
Only nits, but it should make the code a tad simpler. Feel free to merge as-is or fix them and I'll ack them.

SEJeff (approved):

claudijd (approved):

node/pkg/vaa/structs.go
```diff
@@ -421,20 +431,94 @@ func (v *VAA) AddSignature(key *ecdsa.PrivateKey, index uint8) {
...
 // MustWrite calls binary.Write and panics on errors
 func MustWrite(w io.Writer, order binary.ByteOrder, data interface{}) {
 	if err := binary.Write(w, order, data); err != nil {
 		panic(fmt.Errorf("failed to write binary data: %v", data).Error())
 	}
 }
 
-// StringToAddress converts a hex-encoded adress into a vaa.Address
+// StringToAddress converts a hex-encoded address into a vaa.Address
 func StringToAddress(value string) (Address, error) {
```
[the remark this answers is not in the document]
claudijd: might be overkill, there's also a bit of weirdness here, in that one is measuring hex chars and one is measuring bytes.

2 hex chars (1) == 1 byte

vs. 

32 byte (32) == 1 byte

PR #1234, opened 2022-05-28
Add scheduled job to alert the team when new Solidity bugs come out

claudijd:


claudijd:
Moving away from this in repo.

PR #1239, opened 2022-06-01
Zero-pad from the left in vaa.StringToAddress

jynnantonix:
The doc comment for `vaa.Address` says

> If the address data type of a chain is < 32bytes the value is zero-padded on the left.

but `vaa.StringToAddress` was actually adding zero-padding on the right.
Fix it and add some tests and documentation for various edge cases.

It looks like this function is used by the repair scripts but it's unclear if they depend on the current zero-pad on the right behavior.

### DIFF
--- a/node/pkg/vaa/structs_test.go
+++ b/node/pkg/vaa/structs_test.go
@@ -5,13 +5,14 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"encoding/hex"
+	"reflect"
+	"testing"
+	"time"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"reflect"
-	"testing"
-	"time"
 )
 
 func TestChainIDFromString(t *testing.T) {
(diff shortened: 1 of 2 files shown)


node/pkg/vaa/structs_test.go
```diff
@@ -488,8 +489,51 @@ func TestVerifySignaturesFuzz(t *testing.T) {
...
+			value:    "",
+			expected: Address{},
+		},
+		{
+			name:  "longer than 32",
+			value: "62d54ccad4d27ae172d0f6312dba686808e08da08367dae932fc58d020238da4f467358e9b9670a6",
+			expected: Address{
+				0x62, 0xd5, 0x4c, 0xca, 0xd4, 0xd2, 0x7a, 0xe1, 0x72, 0xd0, 0xf6, 0x31,
+				0x2d, 0xba, 0x68, 0x68, 0x08, 0xe0, 0x8d, 0xa0, 0x83, 0x67, 0xda, 0xe9,
+				0x32, 0xfc, 0x58, 0xd0, 0x20, 0x23, 0x8d, 0xa4,
```
claudijd, line 517: Should the expectation be that we truncate on longer than or should it error?

node/pkg/vaa/structs_test.go
```diff
@@ -488,8 +489,51 @@ func TestVerifySignaturesFuzz(t *testing.T) {
...
+		},
+		{
+			name:  "zero pad left",
+			value: "3c6fa05be708cb0e2662f7811d0f6809099fe11c",
+			expected: Address{
+				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+				0x3c, 0x6f, 0xa0, 0x5b, 0xe7, 0x08, 0xcb, 0x0e, 0x26, 0x62, 0xf7, 0x81,
+				0x1d, 0x0f, 0x68, 0x09, 0x09, 0x9f, 0xe1, 0x1c,
+			},
+		},
```
claudijd, line 528: Can we add a few negative tests and validate the error condition?

claudijd (changes requested):
Would like to see negative test cases and I'm not sure if we should be silently truncating

kcsongor (approved):

PR #1241, opened 2022-06-01
apply and enforce gofumpt

tbjump:
What do y'all think about using and enforcing gofumpt? Would reduce the amount of trivial changes in PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1241)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,6 +85,9 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: "1.17.5"
+      # ensure that code is formatted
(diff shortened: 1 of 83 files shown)


kcsongor:
I like this idea a lot, but I'd like to reiterate the same point as in https://github.com/certusone/wormhole/pull/1243#issuecomment-1149822961. For this kind of stuff to not get in the way, we need a uniform workflow that can easily be done locally before making a commit without having to remember different tools and flags. Sounds trivial, but it makes a huge difference :)

tbjump:
we did `goimports` instead

PR #1243, opened 2022-06-02
fix goimports and enforce it

tbjump:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1243)


claudijd (approved):
Looks good.

kcsongor:
Looks great!

It would be great to be able to run the linters & formatters in a unified way both locally and on CI. If we move forward with adding more formatters and linters for each project in the monorepo (and we should!), having to remember the different invocations of each tool will become daunting.
I think the ideal setup would be having a script that just runs the lint/fmt command with the right arguments. Then a docker file that simply installs the tools and calls the script (currently `lint.sh` just calls docker, I think this should be the other way around, so the tools can be run without docker). Then CI would also just install the dependency and run the script.

tbjump:
@kcsongor I'll make those improvements in a new PR. For now, let's get this merged so I don't have to keep rebasing and resolve conflicts?

tbjump:
This PR is superseded by https://github.com/certusone/wormhole/pull/1263

PR #1266, opened 2022-06-10
Modular relayer support

SEJeff:
This is my very first typescript project, so please review thoroughly. I'm always up for some learnings and have thick skin.

The listener bits are hooked in, but I'd like to ensure that the actual approach is sound before I test more heavily. It seems like I should use the type guard function here to properly cast the return from `validate()`: https://github.com/certusone/wormhole/blob/3940c95bfb97f7866df92837d6b2f05b9591a24f/relayer/spy_relayer/src/backends/definitions.ts#L19-L20

The relayer bits are a W.I.P. that aren't ready to push up yet, but soon. Most of my initial pain was just learning how to switch from vim --> vscode to be productive while also learning a very new language.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1266)


relayer/spy_relayer/src/backends/default/listener.ts
```diff
@@ -0,0 +1,272 @@
...
+      this.logger.debug("Failed to parse the transfer payload.");
+      throw new Error("Could not parse the transfer payload.");
+    }
+    return parsedPayload;
+  }
+
+  /** Verify this is a VAA we want to relay. */
+  public async validate(
+    rawVaa: Uint8Array
+  ): Promise<ParsedVaa<ParsedTransferPayload> | string> {
```
SEJeff, line 161: I'm not actually sure which is the best approach to write idiomatic typescript here:

* Just throwing `Error()` on any problem
* The return being the correct type OR a string

relayer/spy_relayer/src/backends/default/listener.ts
```diff
@@ -0,0 +1,272 @@
...
+      filters.push(typedFilter);
+      this.logger.info("Getting spyServiceFiltere " + i);
+    }
+    return filters;
+  }
+
+  /** Process and validate incoming VAAs from the spy. */
+  public async process(rawVaa: Uint8Array): Promise<void> {
+    const validationResults: ParsedVaa<ParsedTransferPayload> | string =
+      await this.validate(rawVaa);
```
SEJeff, line 232: This is one of the lines that I might need to build a type guard for both `ParsedVaa<ParsedTransferPayload>` as I've only got one currently for `ParsedTransferPayload`: https://github.com/certusone/wormhole/blob/3940c95bfb97f7866df92837d6b2f05b9591a24f/relayer/spy_relayer/src/listener/validation.ts#L38-L49

relayer/spy_relayer/src/backends/default/listener.ts
```diff
@@ -0,0 +1,272 @@
...
+      return;
+    }
+
+    const key = getKey(parsedVaa.payload.originChain, originAddressNative);
+
+    const isQueued = await checkQueue(key);
+    if (isQueued) {
+      this.logger.error(`Not storing in redis: ${isQueued}`);
+      return;
+    }
```
SEJeff, line 259: This should likely be turned into a generic function outside of this class, but the `checkQueue()` logic needs to work better first.

relayer/spy_relayer/src/backends/default/relayer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    isComplete(): boolean {
+      this.logger.info("Check if relay is complete for a given vaa")
+      return true
+    }
+
+    targetChain(): ChainId {
+      return 1
+    }
+  }
```
SEJeff, line 26: W.I.P.

relayer/spy_relayer/src/helpers/redisHelper.ts
```diff
@@ -235,54 +240,6 @@ export function resetPayload(storePayload: StorePayload): StorePayload {
...
-      storePayloadToJson(storePayload) +
-      "]"
-  );
-
-  await storeInRedis(
-    storeKeyToJson(storeKey),
-    storePayloadToJson(storePayload)
-  );
-}
-
```
SEJeff, line 285: This is the logic in the backend `Listener.process()` and `Listener.store()`

relayer/spy_relayer/src/helpers/redisHelper.ts
```diff
@@ -428,3 +385,51 @@ export async function monitorRedis(metrics: PromHelper) {
...
+    if (!rClient) {
+      logger.error("Failed to connect to redis");
+      return null;
+    }
+    /**
+     * TODO: Pretty sure this code never ever worked for checking if a key is in redis.
+     *
+     * The `key` variable is `chainId + ":" + address`, but the actual redis keys come
+     * from serializing the StoreKey type into a stringified json representation.
+     */
```
SEJeff, line 412: I'd love to fix this and the easiest way (though my least fav) is making the memory queue use the same json representation as the json set as the redis key, I just don't love that.

More preferable to me would be setting this format however, it would mess up anything outstanding in redis when the upgrade happens:

    {chainId}/{emitterAddress}/{sequenceNumber}

relayer/spy_relayer/src/listener/spy_listen.ts
```diff
@@ -142,39 +82,4 @@ export async function run(ph: PromHelper) {
...
-    return await getEmitterAddressSolana(emitterAddressStr);
-  }
-
-  if (myChainId === CHAIN_ID_TERRA) {
-    return await getEmitterAddressTerra(emitterAddressStr);
-  }
-
-  return getEmitterAddressEth(emitterAddressStr);
-}
+}
```
SEJeff, line 85: Taken over by https://github.com/certusone/wormhole/blob/3940c95bfb97f7866df92837d6b2f05b9591a24f/relayer/spy_relayer/src/backends/definitions.ts#L22-L23

relayer/spy_relayer/src/listener/validation.ts
```diff
@@ -1,207 +1,5 @@
-import {
```
SEJeff, line 1: This file will be removed and all of the types / functions will go into a utils.ts or something similar. Eventually they should make their way back into the sdk.

relayer/spy_relayer/src/listener/validation.ts
```diff
@@ -230,9 +28,22 @@ export type ParsedVaa<T> = {
 
 export type ParsedTransferPayload = {
   amount: BigInt;
-  originAddress: Uint8Array; //hex
+  originAddress: string; // hex
   originChain: ChainId;
-  targetAddress: Uint8Array; //hex
+  targetAddress: string; // hex
```
SEJeff, line 33: I changed these to make this type match this parse function perfectly:

https://github.com/certusone/wormhole/blob/3940c95bfb97f7866df92837d6b2f05b9591a24f/sdk/js/src/utils/parseVaa.ts#L55-L62

The `ParseTransferPayload` type is only within the relayer codebase and is not used anywhere else.

relayer/spy_relayer/src/listener/validation.ts
```diff
@@ -230,9 +28,22 @@ export type ParsedVaa<T> = {
...
+  payload: any
+): payload is ParsedTransferPayload {
+  return (
+    typeof (payload as ParsedTransferPayload).amount == "bigint" &&
+    typeof (payload as ParsedTransferPayload).originAddress == "string" &&
+    typeof (payload as ParsedTransferPayload).originChain == "number" &&
+    typeof (payload as ParsedTransferPayload).targetAddress == "string" &&
+    typeof (payload as ParsedTransferPayload).targetChain == "number"
+  );
+}
```
SEJeff, line 49: If this stays, I'll need to add a matching one for `IsParsedVaaWithParsedTransferPayload()` to do type guards on:

    ParsedVaa<ParsedVaaTransferPayload>

This might be necessary for the `Listener.validate()` return:
 https://github.com/certusone/wormhole/blob/3940c95bfb97f7866df92837d6b2f05b9591a24f/relayer/spy_relayer/src/backends/definitions.ts#L19-L20

relayer/spy_relayer/src/helpers/redisHelper.ts
```diff
@@ -428,3 +385,51 @@ export async function monitorRedis(metrics: PromHelper) {
     await sleep(TEN_SECONDS);
   }
 }
+
+/** Check to see if a queue is in the listener memory queue before redis */
+export async function checkQueue(key: string): Promise<string | null> {
+  try {
+    const backupQueue = getBackupQueue();
+    const queuedRecord = backupQueue.find((record) => {
```
swimricky, line 393: i think you need to either remove the `{}` here or explicitly put `return` or else in this current implementation this will always return undefined

https://replit.com/@swimricky/findarrowfn?v=1

SEJeff: Fixed in ae9e2fb7b5bf5ab5aeda27928dc9a22785ec9da5, keep them coming :)

bruce-riley (approved):

relayer/spy_relayer/src/backends/default/relayer.ts
```diff
@@ -0,0 +1,394 @@
...
+            "Checking timestamps:  now: " +
+              now.toISOString() +
+              ", old: " +
+              old.toISOString() +
+              ", delta: " +
+              timeDelta
+          );
+          if (timeDelta > TEN_MINUTES) {
+            // Deal with this item
+            if (storePayload.status === Status.FatalError) {
```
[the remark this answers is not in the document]
SEJeff: The auditor's purpose is to ensure there wasn't a chain reorg / rollback that killed this relay. Do you think this should be lower?

relayer/spy_relayer/src/backends/default/listener.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /** Get spy filters for all emitters we care about */
+  public async getEmitterFilters(): Promise<TypedFilter[]> {
+    let env = getListenerEnvironment();
+    let filters: {
+      emitterFilter: { chainId: ChainId; emitterAddress: string };
+    }[] = [];
+    for (let i = 0; i < env.spyServiceFilters.length; i++) {
+      this.logger.info("Getting spyServiceFiltera " + i);
```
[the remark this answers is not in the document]
SEJeff: This was code that I moved around, but never fully understood. I've removed all but one of those log lines as they don't make a lot of sense to me.

relayer/spy_relayer/src/backends/default/listener.ts
```diff
@@ -0,0 +1,300 @@
...
+  /**
+   * @throws - when the listener environment setup fails
+   */
+  constructor() {
+    this.logger = getScopedLogger(["TokenBridgeListener"]);
+  }
+
+  /** Verify this payload is version 1. */
+  verifyIsPayloadV1(parsedVaa: ParsedVaa<Uint8Array>): boolean {
+    const isCorrectPayloadType = parsedVaa.payload[0] === 1;
```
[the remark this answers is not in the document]
SEJeff: This is how the previous code did it I think, but I'll collapse it.

relayer/spy_relayer/src/main.ts
```diff
@@ -75,49 +75,60 @@ if (!foundOne) {
...
-    promClient = new PromHelper("spy_relay", promPort, PromMode.Relay);
-  } else if (runWalletMonitor) {
-    promClient = new PromHelper("spy_relay", promPort, PromMode.WalletMonitor);
-  } else {
-    logger.error("Invalid run mode for Prometheus");
-    promClient = new PromHelper("spy_relay", promPort, PromMode.All);
-  }
+const runAll: boolean = runListen && runRelayWorker && runWalletMonitor;
+if (runListen && !spyListener.init()) {
+  process.exit(1);
```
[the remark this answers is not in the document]
SEJeff: I've verified this seems to work as expected. Let me know if you think this isn't good enough.

With this diff:

```diff
diff --git a/relayer/spy_relayer/src/listener/spy_listen.ts b/relayer/spy_relayer/src/listener/spy_listen.ts
index 7dd798b0..d5e43cd8 100644
--- a/relayer/spy_relayer/src/listener/spy_listen.ts
+++ b/relayer/spy_relayer/src/listener/spy_listen.ts
@@ -25,7 +25,7 @@ export function init(): boolean {
     return false;
   }

-  return true;
+  return false;
 }

 export async function run(ph: PromHelper) {
diff --git a/relayer/spy_relayer/src/main.ts b/relayer/spy_relayer/src/main.ts
index 24db4516..8943a8a5 100644
--- a/relayer/spy_relayer/src/main.ts
+++ b/relayer/spy_relayer/src/main.ts
@@ -77,6 +77,7 @@ if (!foundOne) {

 const runAll: boolean = runListen && runRelayWorker && runWalletMonitor;
 if (runListen && !spyListener.init()) {
+  logger.error("listener init failed");
   process.exit(1);
 }
 
 ```

I got it to fail and log for me that it was going to fail:

```bash
$ npm run build

> spy_relay@1.0.0 build
> tsc

$ node lib/main.js --listen_only
loading config file [.env.sample]
spy_relay is logging to the console at level [debug]
2022-07-13 14:29:11.951|info|main: spy_relay is running in listen only mode
2022-07-13 14:29:11.954|info|main: Getting SPY_SERVICE_FILTERS...
2022-07-13 14:29:11.954|info|main: nativeToHexString: 95f83a27e90c622a98c037353f271fd8f5f57b4dc18ebf5ff75a934724bd0491
2022-07-13 14:29:11.954|info|main: nativeToHexString: 0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16
2022-07-13 14:29:11.954|info|main: nativeToHexString: 000000000000000000000000784999135aaa8a3ca5914468852fdddbddd8789d
2022-07-13 14:29:11.954|info|main: nativeToHexString: 0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16
2022-07-13 14:29:11.954|info|main: Getting REST_PORT...
2022-07-13 14:29:11.955|info|main: Getting SPY_NUM_WORKERS...
2022-07-13 14:29:11.955|info|main: Getting SUPPORTED_TOKENS...
2022-07-13 14:29:11.955|info|main: Setting the listener backend...
2022-07-13 14:29:11.955|error|main: listener init failed
```

bruce-riley (dismissed):

relayer/spy_relayer/src/helpers/redisHelper.ts
kev1n-peters: Could we use `const storeKeyJson = storeKeyToJson(storeKeyFromParsedVAA(parsedVaa));` as the `key` - making the memory queue use the same key format as the redis store. I don't see why it'd be so bad

SEJeff: I can totally do that, just think the other way is easier to read. But you've made a totally legit suggestion, so I'm going to go ahead and do it. Thanks!

bruce-riley (approved):

PR #1282, opened 2022-06-16
node: terra2 support

evan-gray:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1282)

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -110,6 +110,10 @@ var (
 	terraLCD      *string
 	terraContract *string
 
+	terra2WS       *string
+	terra2LCD      *string
+	terra2Contract *string
+
 	algorandIndexerRPC   *string
 	algorandIndexerToken *string
 	algorandAlgodRPC     *string
@@ -209,6 +213,10 @@ func init() {
 	terraWS = NodeCmd.Flags().String("terraWS", "", "Path to terrad root for websocket connection")
 	terraLCD = NodeCmd.Flags().String("terraLCD", "", "Path to LCD service root for http calls")
 	terraContract = NodeCmd.Flags().String("terraContract", "", "Wormhole contract address on Terra blockchain")
+
+	terra2WS = NodeCmd.Flags().String("terra2WS", "", "Path to terrad root for websocket connection")
+	terra2LCD = NodeCmd.Flags().String("terra2LCD", "", "Path to LCD service root for http calls")
+	terra2Contract = NodeCmd.Flags().String("terra2Contract", "", "Wormhole contract address on Terra 2 blockchain")
 
 	algorandIndexerRPC = NodeCmd.Flags().String("algorandIndexerRPC", "", "Algorand Indexer RPC URL")
 	algorandIndexerToken = NodeCmd.Flags().String("algorandIndexerToken", "", "Algorand Indexer access token")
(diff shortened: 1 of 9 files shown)


node/cmd/guardiand/node.go
```diff
@@ -534,6 +543,15 @@ func runNode(cmd *cobra.Command, args []string) {
...
 	if *testnetMode || *unsafeDevMode {
+		if *terra2WS == "" {
+			logger.Fatal("Please specify --terra2WS")
+		}
+		if *terra2LCD == "" {
+			logger.Fatal("Please specify --terra2LCD")
+		}
+		if *terra2Contract == "" {
+			logger.Fatal("Please specify --terra2Contract")
+		}
```
bruce-riley, line 554: This says we must run terra2 in testnet, so as soon as this gets merged the next testnet deploy must have terra2. Is that what we want?

node/pkg/terra/watcher.go
```diff
@@ -115,21 +136,21 @@ func (e *Watcher) Run(ctx context.Context) error {
...
 
 	// Wait for the success response
 	_, _, err = c.ReadMessage()
 	if err != nil {
-		p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDTerra, 1)
+		p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
 		terraConnectionErrors.WithLabelValues("event_subscription_error").Inc()
 		return fmt.Errorf("event subscription failed: %w", err)
 	}
 	logger.Info("subscribed to new transaction events")
```
bruce-riley, line 151: In the Eth watcher, we have a networkName which is included in all the log messages so we can tell which chain the error is coming from. Do we want to start doing that for Terra now?

bruce-riley (approved):

node/cmd/guardiand/node.go
SEJeff: Yes I think so. The terra 2 node was setup today and should be done being configured tomorrow in theory.

node/pkg/terra/watcher.go
```diff
@@ -83,12 +92,24 @@ func NewWatcher(
...
+	chainID vaa.ChainID) *Watcher {
+
+	// CosmWasm 1.0.0
+	contractAddressFilterKey := "execute._contract_address"
+	contractAddressLogKey := "_contract_address"
+	if chainID == vaa.ChainIDTerra {
+		// CosmWasm <1.0.0
+		contractAddressFilterKey = "execute_contract.contract_address"
+		contractAddressLogKey = "contract_address"
+	}
```
SEJeff, line 106: These bits are utter magic to me.

no idea dog

node/pkg/terra/watcher.go
SEJeff: Yes, yes we absolutely do!

SEJeff (approved):
I'd like to see you take a stab and the node/hack/repair_terra port to terra2.

GIven how much fun we have, I'm strongly hesitant to add chains that we can't easily repair.

node/pkg/terra/watcher.go
evan-gray: lol that's what was in the watcher before (for terra classic) but CosmWasm 1.0.0 has different fields for the contract address on logs and filters https://github.com/terra-money/core/blob/main/WASM_MIGRATE.md#execute

node/pkg/terra/watcher.go
evan-gray: thanks! added `terra_network` to the metric (possibly to be renamed in the future)

evan-gray:
Note: the following prometheus metrics now include `terra_network`
- `wormhole_terra_connection_errors_total`
- `wormhole_terra_messages_confirmed_total`
- `wormhole_terra_current_height`
- `wormhole_terra_query_latency`
This also changes the middle two to `Vec`s.

SEJeff (approved):

SEJeff (approved):

bruce-riley (approved):

PR #1301, opened 2022-06-23
Injective cosmwam test

panoel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1301)


jynnantonix (commented):
> Reviewed 3 of 8 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1301)* status: 5 of 9 files reviewed, all discussions resolved (waiting on @evan-gray and @kcsongor)

___
*[`cosmwasm/contracts/wormhole/src/contract.rs` line 85 at r2](https://reviewable.io/reviews/certusone/wormhole/1301#-N5K0djk6lxQBh0KEjBT:-N5K0djk6lxQBh0KEjBU:b-dpg67e) ([raw file](https://github.com/certusone/wormhole/blob/ee5a0deb306d45c6c6b7569163bd8e0cc63ae3f0/cosmwasm/contracts/wormhole/src/contract.rs#L85)):*
> ```XML
> const FEE_AMOUNT: u128 = 0;
> 
> pub fn get_chain() -> u16 {
> ```

const_panic has been stable since rust 1.57 so if we make this function (and `get_denom`) const then the panic will be a compile-time error rather than a runtime error.

kcsongor:
> const_panic has been stable since rust 1.57 so if we make this function (and `get_denom`) const then the panic will be a compile-time error rather than a runtime error.

that's great, I wasn't aware that functions could be const, I was originally planning to rewrite these as macros to achieve the same thing, but const functions are much more lightweight

jynnantonix (dismissed):
> Reviewed 1 of 1 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1301)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on @bruce-riley, @evan-gray, @kcsongor, and @panoel)

*[a discussion](https://reviewable.io/reviews/certusone/wormhole/1301#-N6l0FY49vWGcyLhdEZW:-N6l0FY49vWGcyLhdEZX:b-ttahra) (no related file):*
Can't comment on the js/ts parts but the rust part looks good to me with the exception of making the 2 functions const.

___
___
*[`cosmwasm/contracts/wormhole/src/contract.rs` line 85 at r2](https://reviewable.io/reviews/certusone/wormhole/1301#-N5K0djk6lxQBh0KEjBT:-N6l-zVY9Kl2Mi3lFwdR:b9u518a) ([raw file](https://github.com/certusone/wormhole/blob/ee5a0deb306d45c6c6b7569163bd8e0cc63ae3f0/cosmwasm/contracts/wormhole/src/contract.rs#L85)):*

const_panic has been stable since rust 1.57 so if we make this function (and `get_denom`) const then the panic will be a compile-time error rather than a runtime error.
</blockquote></details>

This function (and `get_denom`) needs to be changed to const.

evan-gray (changes requested):
with the merge of #1369 , we can remove the dockerfile, makefile, and contract changes. but we should keep the test changes and update them accordingly 🙂

evan-gray (approved):
Likely need to update the instantiate messages at next go around

evan-gray (changes requested):
had approved, but actually should do a quick clean-up pass after the sdk is updated in case anything changes. might move these changes to that PR when it opens. adding this change request just to hold.

cosmwasm/test/injective.ts
```diff
@@ -0,0 +1,717 @@
...
+// wasm contracts
+const wormholeWasm: string = "wormhole.wasm";
+const tokenBridgeWasm: string = "token_bridge_terra_2.wasm";
+const cw20WrappedWasm: string = "cw20_wrapped_2.wasm";
+const mockBridgeWasm: string = "mock_bridge_integration_2.wasm";
+const cw20BaseWasm: string = "cw20_base.wasm";
+
+function createWasmFullPath(w: string) {
+  const pathToWasms: string =
+    "/home/pnoel/git/injective/wormhole/cosmwasm/artifacts/";
```
evan-gray, line 35: can we avoid the full path here. not everyone will be pnoel 🙃

panoel: Done

panoel:
This PR is out of date and no longer needed.

PR #1309, opened 2022-06-27
solana: Allow specifying wallet instead of ATA in VAA

kcsongor:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1309)

### DIFF
--- a/solana/modules/token_bridge/program/src/api/complete_transfer.rs
+++ b/solana/modules/token_bridge/program/src/api/complete_transfer.rs
@@ -105,9 +105,9 @@ pub fn complete_native(
     if accs.vaa.to_chain != CHAIN_ID_SOLANA {
         return Err(InvalidChain.into());
     }
-    if accs.vaa.to != accs.to.info().key.to_bytes() {
-        return Err(InvalidRecipient.into());
-    }
+
+    verify_recipient(&accs.vaa.to, &accs.to)?;
+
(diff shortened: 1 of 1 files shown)


evan-gray:
This is a good change that would benefit integrators and end-users by making the mechanics clearer. Let’s get this rebased and add tests. Maybe something @a5-pickle can pick up.

solana/modules/token_bridge/program/src/api/complete_transfer.rs
```diff
@@ -262,3 +262,21 @@ pub fn complete_wrapped(
 
     Ok(())
 }
+
+/// Verify that the SPL token account `token_account` can receive tokens sent to
+/// `vaa_recipient`.
+/// For this to be the case, one of the following conditions have to hold:
+/// a) `vaa_recipient` matches `token_account`
+/// b) `vaa_recipient` owns `token_account`
```
kcsongor, line 270: maybe we should strengthen this condition to instead of just an ownership check, check if it's the associated token account. allowing putting it into any owned account might open up some dos attack surface

evan-gray:
Upon further review, it was found that maintaining the existing functionality of specifying a token account while also enabling the field to be interpreted as a wallet (i.e. the owner of an associated token account) opens up a DoS / loss-of-funds vector where a malicious actor could attempt to redeem a transfer to the associated token account of a token account. :crying_cat_face:

PR #1313, opened 2022-06-27
Add more testing and resilience to structs and public rpc server

claudijd:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1313)

### DIFF
--- a/node/pkg/publicrpc/publicrpcserver.go
+++ b/node/pkg/publicrpc/publicrpcserver.go
@@ -82,11 +82,6 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 		return nil, status.Error(codes.InvalidArgument, "invalid chain specified")
 	}
 
-	// Sequence input validation
-	if req.MessageId.Sequence == 0 {
-		return nil, status.Error(codes.InvalidArgument, "no sequence specified")
-	}
-
 	b, err := s.db.GetSignedVAABytes(db.VAAID{
 		EmitterChain:   vaa.ChainID(req.MessageId.EmitterChain.Number()),
 		EmitterAddress: addr,
--- a/node/pkg/publicrpc/publicrpcserver.go
+++ b/node/pkg/publicrpc/publicrpcserver.go
@@ -77,13 +77,13 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 	copy(addr[:], address)
 
 	// Chain input validation
-	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain))
-	if err != nil || chainId == vaa.ChainIDUnset {
+	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain.Number()))
+	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, "invalid chain specified")
 	}
 
 	b, err := s.db.GetSignedVAABytes(db.VAAID{
-		EmitterChain:   vaa.ChainID(req.MessageId.EmitterChain.Number()),
+		EmitterChain:   chainId,
 		EmitterAddress: addr,
 		Sequence:       req.MessageId.Sequence,
 	})
--- a/node/pkg/publicrpc/publicrpcserver.go
+++ b/node/pkg/publicrpc/publicrpcserver.go
@@ -64,6 +64,7 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 		return nil, status.Error(codes.InvalidArgument, "no message ID specified")
 	}
 
+	// Address input validation
 	address, err := hex.DecodeString(req.MessageId.EmitterAddress)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("failed to decode address: %v", err))
@@ -74,6 +75,17 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 
 	addr := vaa.Address{}
 	copy(addr[:], address)
+
+	// Chain input validation
+	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain))
+	if err != nil || chainId == vaa.ChainIDUnset {
+		return nil, status.Error(codes.InvalidArgument, "invalid chain specified")
+	}
+
+	// Sequence input validation
+	if req.MessageId.Sequence == 0 {
+		return nil, status.Error(codes.InvalidArgument, "no sequence specified")
+	}
 
 	b, err := s.db.GetSignedVAABytes(db.VAAID{
 		EmitterChain:   vaa.ChainID(req.MessageId.EmitterChain.Number()),
--- a/node/pkg/publicrpc/publicrpcserver.go
+++ b/node/pkg/publicrpc/publicrpcserver.go
@@ -77,13 +77,12 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 	copy(addr[:], address)
 
 	// Chain input validation
-	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain.Number()))
-	if err != nil {
+	if !vaa.IsValidChainId(vaa.ChainID(req.MessageId.EmitterChain)) {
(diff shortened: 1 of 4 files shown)


node/pkg/publicrpc/publicrpcserver.go
```diff
@@ -75,6 +76,17 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 	addr := vaa.Address{}
 	copy(addr[:], address)
 
+	// Chain input validation
+	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain))
+	if err != nil || chainId == vaa.ChainIDUnset {
+		return nil, status.Error(codes.InvalidArgument, "invalid chain specified")
+	}
```
SEJeff, line 83: The UX of `vaa.ChainIDFromInt()` really tickles my code smell spidey sense here. If the return'd `chainId` here is 0, both of these conditions will be true:

```
if err != nil || chainId == vaa.ChainIDUnset {
```

Due to: https://github.com/certusone/wormhole/blob/ec0a32b0a82a0c0d4906bdea45e2ac634fe77db2/node/pkg/vaa/structs.go#L179

So you have a duplicate return AND a duplicate check, both of which are unnecessary. Simplify, simplify, simplify!

If I understand this PR correctly, you're verifying the passed in chainId is valid. If that's correct, this is just a boolean check and doesn't need an error.

So the above code could become simply

```suggestion
	if !vaa.IsValidChainId(uint16(req.MessageId.EmitterChain) {
		return nil, status.Error(codes.InvalidArgument, "invalid chain specified")
	}
```

node/pkg/publicrpc/publicrpcserver.go
```diff
@@ -75,6 +76,17 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
...
+
+	// Sequence input validation
+	if req.MessageId.Sequence == 0 {
+		return nil, status.Error(codes.InvalidArgument, "no sequence specified")
+	}
```
SEJeff, line 88:
```suggestion
```

This logic is invalid as sequence 0 is (much to my chagrin) actually valid.

A quick example is from Aurora's token bridge:
https://github.com/certusone/wormhole/blob/0018bf7250954a56bb8f9953c70f0a44d05be3bf/node/pkg/common/mainnet_consts.go#L74

See this link for the example vaa with sequence 0 on aurora's token bridge: [/v1/signed_vaa/9/00000000000000000000000051b5123a7b0F9b2bA265f9c4C8de7D78D52f510F/0](https://wormhole-v2-mainnet-api.certus.one/v1/signed_vaa/9/00000000000000000000000051b5123a7b0F9b2bA265f9c4C8de7D78D52f510F/0)

And [here is](https://europe-west3-wormhole-315720.cloudfunctions.net/mainnet-readrow?emitterChain=9&emitterAddress=00000000000000000000000051b5123a7b0f9b2ba265f9c4c8de7d78d52f510f&sequence=0000000000000000) the decoded payload which for this vaa.

It's much easier to read with curl and jq:

```
$ curl -sq 'https://europe-west3-wormhole-315720.cloudfunctions.net/mainnet-readrow?emitterChain=9&emitterAddress=00000000000000000000000051b5123a7b0f9b2ba265f9c4c8de7d78d52f510f&sequence=0000000000000000' | jq '.EmitterChain, .Sequence, .TransferDetails'
"aurora"
"0"
{/v1/signed_vaa/9/00000000000000000000000051b5123a7b0F9b2bA265f9c4C8de7D78D52f510F/0  "Amount": ".749002",
  "Decimals": "6",
  "NotionalUSDStr": "0.750372",
  "TokenPriceUSDStr": "1.001829",
  "TransferTimestamp": "2022-04-13 21:18:22 +0000 UTC",
  "OriginSymbol": "UST",
  "OriginName": "UST",
  "OriginTokenAddress": "ust"
}
```

node/pkg/vaa/structs.go
```diff
@@ -180,6 +180,49 @@ func ChainIDFromString(s string) (ChainID, error) {
...
+		return ChainIDNeon, nil
+	case 18:
+		return ChainIDTerra2, nil
+	case 10001:
+		return ChainIDEthereumRopsten, nil
+	default:
+		return ChainIDUnset, fmt.Errorf("unknown chain ID: %v", num)
+	}
+}
+
```
SEJeff, line 225:
```suggestion
func IsValidChainId(id ChainID) bool {
	validChains := []ChainID{
		ChainIDSolana,
		ChainIDEthereum,
		ChainIDTerra,
		ChainIDBSC,
		ChainIDPolygon,
		ChainIDAvalanche,
		ChainIDOasis,
		ChainIDAlgorand,
		ChainIDAurora,
		ChainIDFantom,
		ChainIDKarura,
		ChainIDAcala,
		ChainIDKlaytn,
		ChainIDCelo,
		ChainIDMoonbeam,
		ChainIDNeon,
		ChainIDTerra2,
	}
	for chainId := range validChains {
		if id == ChainID(chainId) {
			return true
		}
	}
	return false
}
```

Much simpler, just as obvious, easier to maintain. The caveat to both of these approaches is that a developer might miss adding a new chain to that function. The solution is to simply add a call to this function somewhere important early in the watcher initialization or somewhere so it crashes early and loudly. It *guarantees* this function will be kept up to date.

SEJeff (changes requested):
A Sequence number of 0 is valid I'm afraid and let me know what you think of my idea to simplify this even further.

node/pkg/publicrpc/publicrpcserver.go
claudijd: Yeah, will probably need to remove this validation, as a nil/omission will default to 0.

node/pkg/vaa/structs.go
claudijd: The nauce here is that I'm trying to also solve for an invalid int before we've parse it into a ChainID.  Examples here: https://github.com/certusone/wormhole/pull/1313/files#diff-a4248d958e0773722a811482448d2bad30ea20c8e8467f6a06d639e5619306d0R189

node/pkg/publicrpc/publicrpcserver.go
```diff
@@ -75,6 +76,12 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 	addr := vaa.Address{}
 	copy(addr[:], address)
 
+	// Chain input validation
+	chainId, err := vaa.ChainIDFromInt(uint16(req.MessageId.EmitterChain))
```
tbjump, line 80: Here we're validating if `uint16(req.MessageId.EmitterChain)` is valid, but then in L86 we're reading `vaa.ChainID(req.MessageId.EmitterChain.Number())`. I'm not sure if there is a difference between the two, but there could be and we should make sure that we're validating what we're actually using at the end.

tbjump (changes requested):

node/pkg/publicrpc/publicrpcserver.go
claudijd: I'll change them to be consistent.

node/pkg/vaa/structs.go
claudijd: I ended up refactoring all the chain handling to maps, so we can stop duplicating the canonical source.

node/pkg/publicrpc/publicrpcserver.go
claudijd: made them consistent.

SEJeff (commented):
I like that you copied almost verbatim how it is done in the javascript sdk, but actually @jayantk has upstaged you.

Would you be ok with redoing this PR ontop of #1317? It is about as clean and readable as can be.

node/pkg/vaa/structs.go
```diff
@@ -91,93 +135,35 @@ func (a SignatureData) String() string {
...
-	case ChainIDMoonbeam:
-		return "moonbeam"
-	case ChainIDNeon:
-		return "neon"
-	case ChainIDTerra2:
-		return "terra2"
-	default:
+	chainString := chainIntStringMap[int(c)]
+
+	if chainString == "" {
```
bruce-riley, line 140: Is it better to do "chainString, exists := chainIntStringMap[int(c)]" an check exists, rather than checking for the returned value being empty?

Also, you don't need the else below.

bruce-riley (approved):

node/pkg/vaa/structs.go
SEJeff: Yes, your approach is the most idiomatic go.

PR #1277, opened 2022-06-15
Chain governor

bruce-riley:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1277)

PR #1316, opened 2022-06-28
node: processor: Make observation state generic

jynnantonix:
We need to reuse almost all of the gossip infrastructure for accounting
transactions, with the only difference being that accounting will use a
`Transfer` message rather than a `VAA`.

Make the observation stored in the processor state generic so that it
can be either a VAA or a Transfer.  The rest of the code is shared.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1316)

### DIFF
--- a/node/pkg/processor/broadcast.go
+++ b/node/pkg/processor/broadcast.go
@@ -23,15 +23,18 @@ var (
 		})
 )
 
-func (p *Processor) broadcastSignature(v *vaa.VAA, signature []byte, txhash []byte) {
-	digest := v.SigningMsg()
-
+func (p *Processor) broadcastSignature(
+	o Observation,
+	signature []byte,
+	txhash []byte,
+) {
+	digest := o.SigningMsg()
 	obsv := gossipv1.SignedObservation{
 		Addr:      crypto.PubkeyToAddress(p.gk.PublicKey).Bytes(),
 		Hash:      digest.Bytes(),
 		Signature: signature,
 		TxHash:    txhash,
-		MessageId: v.MessageID(),
+		MessageId: o.MessageID(),
 	}
 
 	w := gossipv1.GossipMessage{Message: &gossipv1.GossipMessage_SignedObservation{SignedObservation: &obsv}}
@@ -46,18 +49,18 @@ func (p *Processor) broadcastSignature(v *vaa.VAA, signature []byte, txhash []by
 	// Store our VAA in case we're going to submit it to Solana
 	hash := hex.EncodeToString(digest.Bytes())
 
-	if p.state.vaaSignatures[hash] == nil {
-		p.state.vaaSignatures[hash] = &vaaState{
+	if p.state.signatures[hash] == nil {
+		p.state.signatures[hash] = &state{
 			firstObserved: time.Now(),
 			signatures:    map[ethcommon.Address][]byte{},
 			source:        "loopback",
 		}
 	}
 
-	p.state.vaaSignatures[hash].ourVAA = v
-	p.state.vaaSignatures[hash].ourMsg = msg
-	p.state.vaaSignatures[hash].source = v.EmitterChain.String()
-	p.state.vaaSignatures[hash].gs = p.gs // guaranteed to match ourVAA - there's no concurrent access to p.gs
+	p.state.signatures[hash].ourObservation = o
+	p.state.signatures[hash].ourMsg = msg
+	p.state.signatures[hash].source = o.GetEmitterChain().String()
+	p.state.signatures[hash].gs = p.gs // guaranteed to match ourObservation - there's no concurrent access to p.gs
 
 	// Fast path for our own signature
 	go func() { p.obsvC <- &obsv }()
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/certusone/wormhole/node/pkg/common"
 	"github.com/certusone/wormhole/node/pkg/db"
+	"github.com/certusone/wormhole/node/pkg/notify/discord"
 	"github.com/certusone/wormhole/node/pkg/vaa"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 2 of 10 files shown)


node/pkg/processor/broadcast.go
```diff
@@ -46,18 +50,18 @@ func (p *Processor) broadcastSignature(v *vaa.VAA, signature []byte, txhash []by
...
 		}
 	}
 
-	p.state.vaaSignatures[hash].ourVAA = v
-	p.state.vaaSignatures[hash].ourMsg = msg
-	p.state.vaaSignatures[hash].source = v.EmitterChain.String()
-	p.state.vaaSignatures[hash].gs = p.gs // guaranteed to match ourVAA - there's no concurrent access to p.gs
+	p.state.signatures[hash].ourObservation = o
+	p.state.signatures[hash].ourMsg = msg
+	p.state.signatures[hash].source = source
```
hendrikhofstadt, line 63: isn't source still available from the observation?

hendrikhofstadt (approved):

node/pkg/processor/observation.go
```diff
@@ -218,38 +204,60 @@ func (p *Processor) handleObservation(ctx context.Context, m *gossipv1.SignedObs
...
+		Nonce:            v.Nonce,
+		Sequence:         v.Sequence,
+		EmitterChain:     v.EmitterChain,
+		EmitterAddress:   v.EmitterAddress,
+		Payload:          v.Payload,
+		ConsistencyLevel: v.ConsistencyLevel,
+	}
+	vaaBytes, err := signed.Marshal()
+	if err != nil {
+		panic(err)
```
SEJeff, line 242:
```suggestion
		panic(fmt.Errorf("problem unmarshalling signed VAA: %w", err))
```

Not necessarily better, but maybe?

SEJeff (approved):
The code looks sensible to me and is well written. For fairly lower level "guts" changes like this, it does strike me as sad that there aren't any tests, but we can come back and write those later. We'll just want to bake this code in testnet for a bit as I don't want to have any surprises later with runtime panics (of which we've found a few recently)

jynnantonix (commented):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1316)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @hendrikhofstadt, @kcsongor, and @leoluk)

___
*[`node/pkg/processor/broadcast.go` line 63 at r1](https://reviewable.io/reviews/certusone/wormhole/1316#-N5cv9Y2o4XOJxfHclLp-r1-63:-N5nW3kS8Ku1Zv1z8fUT:bfmomts) ([raw file](https://github.com/certusone/wormhole/blob/949a2cb0839dd4265c148599ea7ee7a162e6940b/node/pkg/processor/broadcast.go#L63)):*
<details><summary><i>Previously, hendrikhofstadt (Hendrik Hofstadt) wrote…</i></summary><blockquote>

isn't source still available from the observation?
</blockquote></details>

Yes, good catch.  Removed `source` as a parameter to this function.

node/pkg/processor/broadcast.go
jynnantonix: It seems reviewable doesn't update comments directly.  I removed `source` as a parameter for this function in favor of getting it from the observation.

Please re-open if that's not what you meant.

node/pkg/processor/broadcast.go
```diff
@@ -23,15 +23,18 @@ var (
 		})
 )
 
-func (p *Processor) broadcastSignature(v *vaa.VAA, signature []byte, txhash []byte) {
-	digest := v.SigningMsg()
-
+func (p *Processor) broadcastSignature(
+	o Observation,
+	digest ethcommon.Hash,
```
hendrikhofstadt, line 28: Is there a reason you're not using the digest available on the Observation interface (it's already available as hex, why not as binary digest)?

hendrikhofstadt (changes requested):

node/pkg/processor/broadcast.go
jynnantonix: Good point.  I didn't realize that HexDigest was just the hex of SigningMsg.  I've updated the Observation interface to use SigningMsg instead and removed this parameter from broadcastSignature.

kcsongor (dismissed):
> Reviewed 3 of 9 files at r1, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1316)* status: all files reviewed, 3 unresolved discussions (waiting on @hendrikhofstadt, @jynnantonix, and @leoluk)

___
*[`node/pkg/processor/observation.go` line 208 at r3](https://reviewable.io/reviews/certusone/wormhole/1316#-N6JN0V_4KF4FH-6rwOf:-N6JN0V_4KF4FH-6rwOg:b6hs1si) ([raw file](https://github.com/certusone/wormhole/blob/cb394d33828f20bdf0492eed9f51e43353271d56/node/pkg/processor/observation.go#L208)):*
> ```Go
> 
> 		if len(sigs) >= quorum && !p.state.signatures[hash].submitted {
> 			switch o := p.state.signatures[hash].ourObservation.(type) {
> ```

would it make sense to make `handleQuorum` an interface function of `Observation` instead of type casing here? That would statically enforce implementing this function, otherwise it feels like a refactoring hazard.

jynnantonix (commented):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1316)* status: 3 of 10 files reviewed, 3 unresolved discussions (waiting on @hendrikhofstadt, @kcsongor, and @leoluk)

___
*[`node/pkg/processor/observation.go` line 208 at r3](https://reviewable.io/reviews/certusone/wormhole/1316#-N6JN0V_4KF4FH-6rwOf:-N6zGYtT8IDYZmPwNrKl:b-8v5neh) ([raw file](https://github.com/certusone/wormhole/blob/cb394d33828f20bdf0492eed9f51e43353271d56/node/pkg/processor/observation.go#L208)):*
<details><summary><i>Previously, kcsongor (Csongor Kiss) wrote…</i></summary><blockquote>

would it make sense to make `handleQuorum` an interface function of `Observation` instead of type casing here? That would statically enforce implementing this function, otherwise it feels like a refactoring hazard.
</blockquote></details>

This is a good suggestion.  I've added `HandleQuorum` as an interface function for `Observation`.  I wasn't able to figure out a good way to implement it in the vaa package without leaking a bunch of details of the processor package so I instead added a `VAA` type to the processor package that embeds a `*vaa.VAA` and implemented `HandleQuorum` there.  This should get us the type safety without much overhead so please take another look.

We could go one step further and make `Observation` a private interface for the processor package but I don't really see much benefit in doing that whereas keeping it as a public interface will let us use dummy types in tests (whenever we add those).

hendrikhofstadt (approved):

bruce-riley (approved):

kcsongor (approved):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1316)* status: 3 of 10 files reviewed, 3 unresolved discussions (waiting on @hendrikhofstadt, @jynnantonix, @kcsongor, @leoluk, and @SEJeff)

node/pkg/processor/vaa.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"encoding/hex"
+
+	"github.com/certusone/wormhole/node/pkg/vaa"
+	"go.uber.org/zap"
+)
+
+type VAA struct {
+	vaa.VAA
+}
```
SEJeff, line 12: This is a reasonable tradeoff to make future testing easier.

SEJeff (approved):
Depending on which of this or #1277 make it in first, one is going to need some refactoring I suspect (mentioned by @evan-gray).

node/pkg/processor/cleanup.go
```diff
@@ -56,45 +57,46 @@ const (
...
+
 		case !s.settled && delta > settlementTime:
-			// After 30 seconds, the VAA is considered settled - it's unlikely that more observations will
+			// After 30 seconds, the observation is considered settled - it's unlikely that more observations will
 			// arrive, barring special circumstances. This is a better time to count misses than submission,
 			// because we submit right when we quorum rather than waiting for all observations to arrive.
 			s.settled = true
 
-			// Use either the most recent (in case of a VAA we haven't seen) or stored gs, if available.
+			// Use either the most recent (in case of a observation we haven't seen) or stored gs, if available.
```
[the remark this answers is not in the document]
jynnantonix: github is going to auto-dismiss all the approvals if I make this change here so I'm going to do it in a separate PR

node/pkg/processor/vaa.go
jynnantonix: Ack

jynnantonix:
> Depending on which of this or https://github.com/certusone/wormhole/pull/1277 make it in first, one is going to need some refactoring I suspect (mentioned by @evan-gray).

Yes, but it should be pretty straightforward. We should be able to remove `GetEmitterAddressForChain` from that PR and replace it with a map lookup.

PR #1318, opened 2022-06-28
Remove Recursive `Peel` to reduce Claim complexity.

Reisen:
This PR contains two commits, both atomic changes. The change requires understanding `FromAccounts` and `Peel` so this PR is written to be as helpful as possible in explaining the change.

Commit 1 - [f44088a](https://github.com/certusone/wormhole/pull/1318/commits/f44088a7a9c39522661bf42e5573f65e8339ecd2)
================================================================================

This commit aims to remove the recursive `FromAccount` behavior. The commit is medium sized as it also includes changes to all instructions affected by the change so all programs still compile.

**Justification**

Removing this code helps with several things:

- It allows removing the `'c` lifetime sprinkled through the entire codebase which is difficult to work with.
- It would allow removing `rocksalt` in future entirely as the macros no longer need to be as complex.
- It simplifies reading the code in the future for audits and any general reader of the code.
- It makes `ClaimableVAA`'s behavior explicit and easier to reason about.

**Commit Explanation**

Solitaire parses accounts using two traits, `FromAccounts` and `Peel`. The `FromAccounts` derive macro generates a function, `FromAccounts::from()`, which repeatedly calls `Peel::peel` to construct each field in the struct:

```
let example = Example {
    foo: Peel::peel(next_account_iter(accs)?),
    bar: Peel::peel(next_account_iter(accs)?),
    baz: Peel::peel(next_account_iter(accs)?),
    ...,
}
```

The `FromAccounts` derivation also attempts to implement Peel for the structure itself however, which means `Example` itself can be embedded as a field in another accounts struct. This is currently only done with `ClaimableVAA` which is a large source of confusion.

This commit removes the recursion by:

1) Removing the `impl Peel` which the `FromAccounts` macro emits.
2) Implementing `Peel` manually on `ClaimableVAA` so it continues to work the same way.
3) In all instructions that take a `ClaimableVAA`, replace with the two underlying accounts.
4) Change verification call sites to explicitly use claim/messages.

With recursion gone, the `Context` used by Peel can be simplified as it now only ever parses a single `AccountInfo` for each field being parsed. This change is performed by:

1) Removing the `AccountInfo` iterator from Context as it is no longer needed.
2) Adds the current parsed account to `Context` instead, safe thanks to (2).

Commit 2 - [0ffc00b](https://github.com/certusone/wormhole/pull/1318/commits/0ffc00b87351651ac0c59ccbc538a0e85dc9bc54)
================================================================================

This commit simplifies `Claim` further. In Commit 1 we keep all the code identical and add a `Peel` implementation to keep the code identical to before removing the recursion. In this commit, the code for claiming is cleaned up slightly through the following steps:

1) No longer implement `ClaimVAA` as a struct, removing a layer of Peel complexity.
2) Removing `verify()` as it is never used alone without an associated `claim()`.
3) Move the account and verification logic to `accounts/claim.rs` so it's co-located with the account data.
4) In all locations use the `claim::consume` (the new verify/claim name) function instead.

This change also makes it more explicitly clear that the claim account being initialized is how we know a claim is consumed, instead of the flag/value.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1318)

### DIFF
--- a/solana/bridge/program/src/accounts/claim.rs
+++ b/solana/bridge/program/src/accounts/claim.rs
@@ -1,5 +1,28 @@
-//! ClaimData accounts are one off markers that can be combined with other accounts to represent
-//! data that can only be used once.
+//! Claim accounts add replay protection to Messages.
+//!
+//! Claim accounts work by having the constraint that they must be uninitialized. This means once a
+//! Claim account has been created, they can no longer be passed to an instruction. This gives us
+//! the behaviour of replay protection.
+//!
+//! Example usage:
+//!
+//! ```rust,noplayground,no_run
+//! struct ExampleAccounts {
+//!     message: PayloadMessage<'info, Example>,
+//!     claim:   Claim<'info>,
+//!     payer:   Mut<Signer<'info>>,
+//! }
+//!
+//! // Note that as a Claim must be uninitialized, only the first time this instruction is called
+//! // will succeed, subsequent calls will fail the `Uninitialized` check.
+//! fn read_message(
+//!    ctx:  &ExecutionContext,
+//!    accs: &mut ExampleAccounts,
+//!    data: (),
+//! ) {
+//!    claim::consume(ctx, &accs.payer.key, &mut accs.claim, &accs.vaa)?;
+//! }
+//! ```
 
 use borsh::{
     BorshDeserialize,
@@ -9,15 +32,63 @@ use serde::{
     Deserialize,
     Serialize,
 };
+use solana_program::pubkey::Pubkey;
 use solitaire::{
     processors::seeded::Seeded,
     AccountOwner,
-    AccountState,
+    AccountState::*,
+    CreationLamports::Exempt,
     Data,
     Owned,
+    Result,
+    *,
 };
 
-pub type Claim<'a, const State: AccountState> = Data<'a, ClaimData, { State }>;
+use crate::{
+    DeserializePayload,
+    PayloadMessage,
+};
+
+pub type Claim<'a> = Data<'a, ClaimData, { Uninitialized }>;
+
+/// Consume a claim by initializing the account. Initialized claims act as an indicator proving
+/// that a message has been consumed.
+#[must_use]
+pub fn consume<T>(
+    ctx: &ExecutionContext,
+    payer: &Pubkey,
+    claim: &mut Claim,
+    message: &PayloadMessage<T>,
+) -> Result<()>
+where
+    T: DeserializePayload,
+{
+    // Verify that the claim account is derived correctly before claiming.
+    claim.verify_derivation(
+        ctx.program_id,
+        &ClaimDerivationData {
+            emitter_address: message.meta().emitter_address,
+            emitter_chain: message.meta().emitter_chain,
+            sequence: message.meta().sequence,
+        },
+    )?;
+
+    // Claim the account by initializing it with a value.
+    claim.create(
+        &ClaimDerivationData {
+            emitter_address: message.meta().emitter_address,
+            emitter_chain: message.meta().emitter_chain,
+            sequence: message.meta().sequence,
+        },
+        ctx,
+        payer,
+        Exempt,
+    )?;
+
+    claim.claimed = true;
+
+    Ok(())
+}
 
 #[derive(Default, Clone, Copy, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
 pub struct ClaimData {
@@ -36,7 +107,7 @@ pub struct ClaimDerivationData {
     pub sequence: u64,
 }
 
-impl<'b, const State: AccountState> Seeded<&ClaimDerivationData> for Claim<'b, { State }> {
+impl<'b> Seeded<&ClaimDerivationData> for Claim<'b> {
     fn seeds(data: &ClaimDerivationData) -> Vec<Vec<u8>> {
         return vec![
             data.emitter_address.to_vec(),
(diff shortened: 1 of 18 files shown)


kcsongor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1318)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @claudijd, @hendrikhofstadt, @jynnantonix, and @Reisen)

___
*[`solana/bridge/program/src/accounts/claim.rs` line 32 at r2](https://reviewable.io/reviews/certusone/wormhole/1318#-N5fKtnT8-qYMKW7jU1L:-N5fKtnT8-qYMKW7jU1M:b-794xms) ([raw file](https://github.com/certusone/wormhole/blob/91fbd44935ef3382bab0f98ba103fb22d6f765f8/solana/bridge/program/src/accounts/claim.rs#L32)):*
> ```XML
> };
> 
> pub type Claim<'a, const S: AccountState> = Data<'a, ClaimData, { S }>;
> ```

`S` is always instantiated to `Uninitialized`, is there a reason why we still want to keep it parameterised?
___
*[`solana/bridge/program/src/accounts/claim.rs` line 32 at r2](https://reviewable.io/reviews/certusone/wormhole/1318#-N5fPpyG5jv9xlgP0uj0:-N5fPpyG5jv9xlgP0uj1:bkjo505) ([raw file](https://github.com/certusone/wormhole/blob/91fbd44935ef3382bab0f98ba103fb22d6f765f8/solana/bridge/program/src/accounts/claim.rs#L32)):*
> ```XML
> };
> 
> pub type Claim<'a, const S: AccountState> = Data<'a, ClaimData, { S }>;
> ```

This is only tangentially related to this PR, but there's something throwing me off about this module. The doc at the top says that `ClaimData` accounts are supposed to be generic and combinable with other accounts. Yet the `Seeded` instance very clearly specialises this to claiming VAAs messages. Since the `Claim` account and the `PayloadMessage` account no longer sit together in a dedicated struct, the connection between them is harder to spot when reading the code.

How about doing it this way:

```rust
pub type Claim<'a, T, const State: AccountState> = Data<'a, ClaimData<T>, { State }>;

#[derive(Default, Clone, Copy, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
pub struct ClaimData<T> {
    pub claimed: bool,
    phantom: PhantomData<T>
}
```

then in the account list we can write

```rust
    pub vaa: PayloadMessage<'b, GovernancePayloadUpgrade>,

    /// An Uninitialized Claim account to consume the VAA.
    pub claim: Mut<Claim<'b, PayloadMessage<'b, GovernancePayloadUpgrade>>>,
```

Reisen:
Ah yes, I also noticed this. In fact I started making this exact change before I made this PR, but what stopped me is when it became this general it seemed like it should probably be in the solitaire module instead.

I left everything in `accounts/claim.rs` instead because then It's an API feature of the bridge. The problem then is just the comment and maybe the name isn't specific enough, so your point is 100% right.

There's two options basically, what do you think:

1) Update the comment and/or account name to make it clearer it's for the bridge only.
2) Go full generalised and move it up to solitaire.

jynnantonix (dismissed):
> Reviewed 6 of 13 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1318)* status: all files reviewed, 2 unresolved discussions (waiting on @claudijd, @hendrikhofstadt, and @Reisen)

*[a discussion](https://reviewable.io/reviews/certusone/wormhole/1318#-N5naozr3cv97tDgxh0g:-N5naozr3cv97tDgxh0h:bm6h50v) (no related file):*
lgtm % Csongor's comments

___

Reisen:
> > 1. Update the comment and/or account name to make it clearer it's for the bridge only.
> > 2. Go full generalised and move it up to solitaire.
> 
> my heart says 2. but my brain say 1. Less overhead if we explicitly specialise it to the current (singular) use case, and we can always generalise later if need be.

Yeah that was my thought too, heart says 2 but brain said 1. I've updated the file with better documentation, and also removed the state parameter that was a good call as it stops them being misused accidentally.

kcsongor (changes requested):
> Reviewed 14 of 14 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1318)* status: all files reviewed, 1 unresolved discussion (waiting on @claudijd, @hendrikhofstadt, and @Reisen)

___
*[`solana/modules/token_bridge/program/src/api/governance.rs` line 41 at r3](https://reviewable.io/reviews/certusone/wormhole/1318#-N6JIO568p1t0KcQnA13:-N6JIO568p1t0KcQnA14:btl1dgk) ([raw file](https://github.com/certusone/wormhole/blob/0ffc00b87351651ac0c59ccbc538a0e85dc9bc54/solana/modules/token_bridge/program/src/api/governance.rs#L41)):*
> ```XML
> };
> 
> // Confirm that a ClaimableVAA came from the correct chain, signed by the right emitter.
> ```

Comment outdated now

jynnantonix (approved):
> Reviewed 14 of 14 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1318)* status: all files reviewed, 1 unresolved discussion (waiting on @claudijd, @hendrikhofstadt, and @Reisen)

kcsongor (approved):

PR #1319, opened 2022-06-28
cloud_functions: Override terra-classic LUNA CoinGeckoId

kev1n-peters:


kev1n-peters:
LUNA on CoinGecko resolves to the new chain's token

evan-gray (approved):
wonder if we should change the name and symbol to LUNC but 🤷

ckeun (approved):

PR #1324, opened 2022-06-29
CLI: add missing VAA parsers

kcsongor:
The goal is to support all of the VAA formats.

- [x] Core bridge
- [x] Token bridge
- [x] NFT bridge

Since this tool might be useful for external integrators too, I've added some [golden tests](https://en.wikipedia.org/wiki/Characterization_test) to ensure we're aware of any format changes. Other than the extended parsing functionality, the other change is that the CLI now expects address arguments to be in the native format of the given chain.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1324)


clients/js/README.md
```diff
@@ -1,97 +1,95 @@
...
+worm [command]
 
-    Commands:
-      worm generate      generate VAAs (devnet and testnet only)
-      worm parse <vaa>   Parse a VAA
-      worm submit <vaa>  Execute a VAA
+Commands:
+  worm generate                             generate VAAs (devnet and testnet
+                                            only)
+  worm parse <vaa>                          Parse a VAA
```
spacemandev-git, line 21: mention the format the vaa's can be in (hex and b64, right?) and if anything is needed to specify the format as a flag when passing in

clients/js/README.md
```diff
@@ -1,97 +1,95 @@
...
+  worm evm                                  EVM utilites
+  worm submit <vaa>                         Execute a VAA
 
-    Options:
-      --help     Show help                                  [boolean]
-      --version  Show version number                        [boolean]
+Options:
+  --help     Show help                                                 [boolean]
+  --version  Show version number                                       [boolean]
+```
```
spacemandev-git, line 31: --network should be added here 
maybe allow for passing in *any* gaurdian endpoint so people can use this against their own networks?

clients/js/README.md
```diff
@@ -103,19 +101,73 @@ For VAAs that don't have a specific target chain (like registrations or guardian
...
+    "ropsten": "0x0000000000000000000000000000000000000000000000000000000000000000"
+  }
+}
+
+```
+### Misc
+
+To get the contract address for a module:
+
+    $ worm contract mainnet bsc NFTBridge
```
spacemandev-git, line 168: probably should use the --network flag to keep things uniform

scripts/devnet-consts.json
```diff
@@ -12,11 +12,11 @@
             "rpcUrlLocal": "http://localhost:8899",
             "rpcPort": "8899",
             "contracts": {
-                "coreEmitterAddress": "074c6d4fcaf24e4b33b3624d7c340c09e46af6fe0a86759efeb25552e46a0e2b",
```
spacemandev-git, line 15: when submitting Emitter address into a contract, we need to pass in the Wormhole one rather than the Emitter hex right? in which case isn't it more useful to have the hexified version of the address rather than native?

spacemandev-git (commented):
parser and other functions should unify --network flag rather than having it be a flag in one place and an arg in another

clients/js/README.md
kcsongor: Agreed that `--network` should be a global flag (right now most commands take it, but not all - so didn't want to make it a global required flag. That said, we can just declare the flag globally as optional, then do the requirement check only in the commands that actually need it). Happy to come back to this in another PR (see my other comment below). FWIW the CLI currently offers no commands that communicate with guardian endpoints, the network flag instead is used for selecting contract addresses, and chain rpc endpoints. So when you say `worm submit <vaa>`, the chain id in the VAA determines which chain to submit to, and the `--network` flag is then used to look up what the contract address and chain RPC endpoint are. Overriding the contract address and the RPC is possible in certain subcommands, though we could probably take a more methodical approach to these flags.

Something that would be interesting is overriding these values from a json config file rather than through flags.

clients/js/README.md
kcsongor: yeah, both hex and base64 are accepted, and no additional flags are needed to specify -- updated the help string

bruce-riley (dismissed):

clients/js/parse_tests/base64-2.expected
```diff
@@ -0,0 +1,22 @@
...
+  timestamp: 1656354705,
+  nonce: 0,
+  emitterChain: 2,
+  emitterAddress: '0x000000000000000000000000f19a2a01b70519f67adb309a994ec8c69a967e8b',
+  sequence: 0n,
+  consistencyLevel: 1,
+  payload: {
+    type: 'Other',
+    payload: '46726f6d3a2065766d305c6e4d73673a2048656c6c6f20576f726c6421',
+    dump: 'From: evm0\\nMsg: Hello World!'
```
kcsongor, line 19: @spacemandev-git this now works too

bruce-riley (dismissed):

evan-gray (approved):

PR #1326, opened 2022-06-30
evm: verify bytecode

kcsongor:
Similarly to `terra/verify` and `solana/verify`, this PR adds an `ethereum/verify` script that can be used to ensure that the deployed bytecode on an EVM chain matches the local build artifact. This is useful to streamline the verification process when doing contract upgrades. In the future, we can build additional checks on top of it to find out of date contracts etc.

### DIFF
--- a/clients/js/networks.ts
+++ b/clients/js/networks.ts
@@ -148,11 +148,11 @@ const TESTNET = {
     key: get_env_var("ETH_KEY_TESTNET"),
   },
   karura: {
-    rpc: "http://103.253.145.222:8545",
+    rpc: "https://karura-dev.aca-dev.network/eth/http",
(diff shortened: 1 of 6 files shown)


jumpsiegel (approved):

bruce-riley (approved):

clients/js/networks.ts
```diff
@@ -148,11 +148,11 @@ const TESTNET = {
     key: get_env_var("ETH_KEY_TESTNET"),
   },
   karura: {
-    rpc: "http://103.253.145.222:8545",
+    rpc: "https://karura-dev.aca-dev.network/eth/http",
```
claudijd, line 151: thank you for fixing these.

PR #1327, opened 2022-06-30
Guardian support for Injective

bruce-riley:
I plan to rename the node/pkg/terra folder to be cosmwasm, but didn't want to generate a ton of unnecessary diffs with this PR. Once this PR is merged, I'll do another PR that simply does the rename and nothing else.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -41,7 +41,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 
-	"github.com/certusone/wormhole/node/pkg/terra"
+	cosmwasm "github.com/certusone/wormhole/node/pkg/terra"
 
 	"github.com/certusone/wormhole/node/pkg/algorand"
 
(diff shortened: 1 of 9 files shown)


node/cmd/guardiand/node.go
```diff
@@ -904,16 +932,24 @@ func runNode(cmd *cobra.Command, args []string) {
 
 		logger.Info("Starting Terra watcher")
 		if err := supervisor.Run(ctx, "terrawatch",
-			terra.NewWatcher(*terraWS, *terraLCD, *terraContract, lockC, setC, chainObsvReqC[vaa.ChainIDTerra], common.ReadinessTerraSyncing, vaa.ChainIDTerra).Run); err != nil {
+			cosmwasm.NewWatcher(*terraWS, *terraLCD, *terraContract, lockC, setC, chainObsvReqC[vaa.ChainIDTerra], common.ReadinessTerraSyncing, vaa.ChainIDTerra).Run); err != nil {
```
panoel, line 935: I'm not sure if this is a thing, but is cosmwasm version 1.0.0 and only applies to Terra2 and Injective?  Andy old Terra should stay on the old version of cosmwasm?  Does that apply here?

panoel (approved):

SEJeff (approved):
We need to remember to update this before injective goes to mainnet:

https://github.com/certusone/wormhole/blob/d54c174974de07e59fb953abb4fbc4dde09158e4/node/hack/repair_terra/repair.go#L30-L43

It would be super easy to forget.

PR #1331, opened 2022-07-02
Optimise guardian set init script

kcsongor:
Use prebuilt CLI executable instead of invoking with 'ts-node'. This way the script runs 10 times
faster, 40s->4s.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1331)


justinschuldt (approved):
LG

hendrikhofstadt (approved):

SEJeff (approved):

PR #1332, opened 2022-07-02
node: Prepare development binary release

kcsongor:
In preparation for a development release of the node image, this PR introduces several changes:

* remove the cruft from the docker image by copying only the necessary files into a scratch stage
* add a `GO_BUILD_ARGS` build arg to the docker container which by default sets the `-race` flag, but allows turning it off
* enforce that `--unsafeDevMode` is enabled when running a development build of the binary.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1332)

### DIFF
--- a/.github/workflows/guardiand-docker.yml
+++ b/.github/workflows/guardiand-docker.yml
@@ -1,0 +1,45 @@
+name: Publish guardiand development docker image
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: certusone/guardiand
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - run: DOCKER_BUILDKIT=1 docker build --target go-export -f Dockerfile.proto -o type=local,dest=node .
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: ./node
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          target: export
+          build-args: GO_BUILD_ARGS=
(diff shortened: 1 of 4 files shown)


node/Dockerfile
```diff
@@ -21,4 +20,18 @@ RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
   go build -race -gcflags="all=-N -l" --ldflags '-extldflags "-Wl,--allow-multiple-definition"' -mod=readonly -o /guardiand github.com/certusone/wormhole/node && \
```
leoluk, line 20: We don't want to run with `-race` in production

node/Dockerfile
```diff
@@ -21,4 +20,18 @@ RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
...
   cp /go/pkg/mod/github.com/!cosm!wasm/wasmvm@v0.16.2/api/libwasmvm.so /usr/lib/
 
+# Only export the final binary (+ shared objects). This reduces the image size
+# from ~1GB to ~150MB.
+FROM scratch as export
+
+# guardiand can't (easily) be statically linked due to the C dependencies, so we
+# have to copy all the dynamic libraries
+COPY --from=build /bin/* /bin/
```
leoluk, line 29: We'll want `/dlv` too if this same image should also run in tilt. Or specify that we want the `build` stage when building in Tilt

leoluk (changes requested):
Perhaps a separate Dockerfile (or at least separate stages) would be easier?

node/Dockerfile
kcsongor: I'll specify that tilt should run `build`

node/Dockerfile
kcsongor: Do you have a preferred solution for doing this? I'm thinking a build arg like `NO_SANITIZER` that's set to false by default

kcsongor:
> Perhaps a separate Dockerfile (or at least separate stages) would be easier?

I've kept the single docker file but made it a bit more configurable to support building without `-race`. Also added a way for the binary to check if it was built in development mode, and require the `--unsafeDevMode` flag in that case. Let me know what you think of doing it this way

SEJeff (approved):
Strong +1, nice!

SEJeff (approved):

.github/workflows/guardiand-docker.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
claudijd, line 28: Should we use the default GITHUB_TOKEN permissions or generate a special key for this action and scope down it's permissions to a minimal set of scope?

kcsongor: No need for that, I've specified in the workflow file that the token should only have write access to the packages (where it publishes): https://github.com/certusone/wormhole/pull/1332/files#diff-0646c51b93a786d348c46b78d911a1c9d066b8263e4f0dfc817839bdf86fd0c1R15-R17

reference: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
> You can use the `permissions` key in your workflow file to modify permissions for the GITHUB_TOKEN for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. When the `permissions` key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

leoluk (approved):

PR #1334, opened 2022-07-04
Optimise solana dockerfiles

kcsongor:
This PR optimises the solana docker images by caching and removing unnecessary dependencies. Also published prebuilt base images, which make building faster, and more robust against upstream dependency changes.

```
| Dockerfile | before | after |
|------------+--------+-------|
| --no-cache | 320s   | 130s  |
| cache      | 112s   | 25s   |

Code changes will no longer trigger a fully dependency tree rebuild.


<!-- Reviewable:start -->
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/certusone/wormhole/1334)
<!-- Reviewable:end -->

### CODE REVIEW

File: solana/Dockerfile.wasm [hunk 1]
```diff
@@ -1,18 +1,21 @@
...
+COPY cert.pem* /certs/
 # Debian
 RUN if [ -e /certs/cert.pem ]; then cp /certs/cert.pem /etc/ssl/certs/ca-certificates.crt; fi
 
-RUN cargo install wasm-pack --vers 0.9.1
+COPY rust-toolchain .
+
+# Install wasm-pack binary, saves about 2 minutes relative to building it from
+# source
+ARG WASM_PACK_VER=v0.9.1
```
hendrikhofstadt, line 15: I'm not sure we want to trust another binary artifact here for security reasons

hendrikhofstadt (dismissed):

solana/Dockerfile.wasm
kcsongor: verified that they produce identical wasms and added a checksum

kcsongor: @claudijd how do you feel about this?

SEJeff: Seems ok to me so long as we do a manual build from a git checkout to verify the checksum before uprevving this.

SEJeff (dismissed):
Quicker builds means happier developers

solana/Dockerfile.wasm
kcsongor: I have done that manual bit, if that counts

solana/Dockerfile.wasm
```diff
@@ -1,18 +1,28 @@
...
+#
+# Alternatively it could be built from source this way:
+# RUN cargo install wasm-pack --vers ${WASM_PACK_VER}
+# If you upgrade the version, it's a good idea to compare the result against
+# the source build to make sure the distributed binary indeed corresponds to the correct version
+ENV WASM_PACK_CHECKSUM=d478bd20811067566bfc88141afcc857e7713b5385c684f6d50e7c2d549847f7
+RUN curl -L https://github.com/rustwasm/wasm-pack/releases/download/v${WASM_PACK_VER}/wasm-pack-v${WASM_PACK_VER}-x86_64-unknown-linux-musl.tar.gz --output wasm-pack.tar.gz && \
+    echo "$WASM_PACK_CHECKSUM wasm-pack.tar.gz" | sha256sum --check && \
+    tar xf wasm-pack.tar.gz --strip-components 1 -C /usr/local/cargo/bin && \
+    rm wasm-pack.tar.gz
```
kcsongor, line 25: @claudijd @leoluk  tagging you here - this should be safe, I verified that the binary distribution produces identical wasms to the source build, but wanted to get your eyes on it too.

SEJeff (dismissed):

leoluk:
The alternative way would be to use buildkit caching layers built on dev.v2, so we wouldn't have to hardcode images and rebuild cleanly when desired (such as for releases).

But this works too.

PR #1337, opened 2022-07-05
Docker FROM image pinning

claudijd:
This PR introduces the following:

- Pins all our unpinned Docker image dependancies
- Adds a script to check the repo for unpinned dependancies
- Adds a job in the CI build workflow that runs the script, which will fail the build if we have an unpinned Docker image dep

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1337)

### DIFF
--- a/scripts/check-docker-pin.sh
+++ b/scripts/check-docker-pin.sh
@@ -4,7 +4,7 @@
 #   - https://nickjanetakis.com/blog/docker-tip-18-please-pin-your-docker-image-versions
 #   - https://snyk.io/blog/10-docker-image-security-best-practices/ (Specifically: USE FIXED TAGS FOR IMMUTABILITY)
 #
-find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | grep -v scratch | grep -v sha256
+find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | grep -v scratch | grep -v sha256 | grep -v "solana AS builder" | grep -v "solana AS ci_tests" | grep -v "node_modules"
 if [ $? -eq 0 ]; then
    echo "[!] Unpinned docker files"
    exit 1
--- a/scripts/check-docker-pin.sh
+++ b/scripts/check-docker-pin.sh
@@ -4,7 +4,7 @@
 #   - https://nickjanetakis.com/blog/docker-tip-18-please-pin-your-docker-image-versions
 #   - https://snyk.io/blog/10-docker-image-security-best-practices/ (Specifically: USE FIXED TAGS FOR IMMUTABILITY)
 #
-find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | grep -v scratch | egrep -v 'scratch|sha256|solana AS builder|solana AS ci_tests|node_module'
+find . -name 'Dockerfile*' -print0  -type f | xargs -0 grep -s "FROM" {} | grep -v scratch | egrep -v 'scratch|sha256|solana AS builder|solana AS ci_tests|node_module'
 if [ $? -eq 0 ]; then
    echo "[!] Unpinned docker files" >&2
    exit 1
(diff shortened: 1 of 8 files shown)


scripts/check-docker-pin.sh
```diff
@@ -0,0 +1,14 @@
+# This script is checks to that all our Docker images are pinned to a specific SHA256 hash
+#
+# References as to why...
+#   - https://nickjanetakis.com/blog/docker-tip-18-please-pin-your-docker-image-versions
+#   - https://snyk.io/blog/10-docker-image-security-best-practices/ (Specifically: USE FIXED TAGS FOR IMMUTABILITY)
+#
+find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | grep -v scratch | grep -v sha256 | grep -v "solana AS builder" | grep -v "solana AS ci_tests" | grep -v "node_modules"
```
SEJeff, line 7: You might change the `solana AS foo` lines with just `AS` since AS is different.

```suggestion
find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | egrep -v 'scratch|sha256|solana AS builder|solana AS ci_tests|node_module'
```

I'd probably do it more like this, which might be a better approach:

```
[jschroeder@gusc1a-ossdev-jsl3 wormhole]$ find . -name 'Dockerfile*' -print0  -type f | xargs -0 grep FROM | egrep -v 'sha256|as|AS|node_modules'
./third_party/redis/Dockerfile:FROM alpine:3.14
./algorand/sandbox-algorand/images/indexer/Dockerfile:FROM golang:$GO_VERSION-alpine
./algorand/sandbox-algorand/images/Dockerfile:FROM ubuntu:20.04
./algorand/Dockerfile:FROM docker.io/python:3.10
```

scripts/check-docker-pin.sh
```diff
@@ -0,0 +1,14 @@
...
+if [ $? -eq 0 ]; then
+   echo "[!] Unpinned docker files"
+   exit 1
+else
+   echo "[+] No unpinned docker files"
+   exit 0
+fi
```
SEJeff, line 14: Please update your editor to add trailing `\n`

scripts/check-docker-pin.sh
```diff
@@ -0,0 +1,14 @@
+# This script is checks to that all our Docker images are pinned to a specific SHA256 hash
+#
+# References as to why...
+#   - https://nickjanetakis.com/blog/docker-tip-18-please-pin-your-docker-image-versions
+#   - https://snyk.io/blog/10-docker-image-security-best-practices/ (Specifically: USE FIXED TAGS FOR IMMUTABILITY)
+#
+find ~+ -type f -name "Docker*" -print0 | xargs -0 grep -s "FROM" {} | grep -v scratch | grep -v sha256 | grep -v "solana AS builder" | grep -v "solana AS ci_tests" | grep -v "node_modules"
+if [ $? -eq 0 ]; then
+   echo "[!] Unpinned docker files"
```
SEJeff, line 9: errors go to stderr:

```suggestion
   echo "[!] Unpinned docker files" >&2
```

SEJeff (dismissed):

SEJeff (dismissed):

SEJeff (dismissed):

kcsongor (changes requested):
Great idea!

One nit: the `grep -v`s currently serve multiple purposes:
1. Ignoring files in `node_modules` folders.
2. Ignoring files that do have sha256s
3. Ignoring specific `FROM` statements that build on previous images

`node_modules` in the first point is actually a red herring, we simply want to ignore any file that's not in the git tree. For that `git ls-files` is better than `find`. The third point is in turn going to be a little painful because the regexp will have to be extended for each new dockerfile, and feels a little ad hoc. Maybe we can only look at the first `FROM` in each file, which is what tends to be the base image? Something like

```bash
git ls-files | grep "Dockerfile*" | xargs -i sh -c 'printf "{}: "; grep -s -e "^FROM" {} | head -n1' | grep -v sha256
```

kcsongor (approved):

SEJeff (approved):

PR #1338, opened 2022-07-05
Move from ADD to COPY on Dockerfiles

claudijd:
Reference: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

COPY is generally preferred unless we need to make use of the more powerful features within ADD, which it appears we are not using.  The most common example case is that ADD can pull in remote files and auto-magically explode tarballs, neither of which we are doing here, so it's nice to just sidestep those two use cases, which could be abused.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1338)

### DIFF
--- a/event_database/functions_server/Dockerfile
+++ b/event_database/functions_server/Dockerfile
@@ -3,7 +3,7 @@ FROM docker.io/golang:1.17.0@sha256:06e92e576fc7a7067a268d47727f3083c0a564331bfc
 
 WORKDIR /app
 
-ADD . .
+COPY . .
 
 WORKDIR /app/functions_server
 RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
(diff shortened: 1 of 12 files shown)


event_database/functions_server/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM docker.io/golang:1.17.0@sha256:06e92e576fc7a7067a268d47727f3083c0a564331bfc
 
 WORKDIR /app
 
-ADD . .
+COPY . .
```
claudijd, line 6: Not a huge fan of ". ." usage when copying, it has the potential to leak secrets when publishing images.  However, since we are not doing this, it's probably ok for the time being.

claudijd: We should aim to be explicit as possible in the relative context we are working in.

kcsongor (approved):
These changes make sense, though it will result in a more painful than usual pull because it invalidates each docker cache.

event_database/functions_server/Dockerfile
SEJeff: We can use `.dockerignore` files that could contain things like `.env*`

SEJeff (approved):

PR #1339, opened 2022-07-06
cloud_functions: tokenAllowlist implementation

kev1n-peters:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1339)


evan-gray (approved):

evan-gray (approved):

evan-gray (dismissed):

justinschuldt:
LG! can you check in a `tokenAllowlist.json` so that people can run the functions locally without it `log.Fatal("tokenAllowlist is empty")`

justinschuldt:
in prod, will `tokenAllowlist.json` be stored in GCS like the cache documents, or does it get uploaded with the Go files at deploy time?  I ask because if it gets deployed with Go files, we'll have to remove the [gcloudignore for *.json](https://github.com/certusone/wormhole/blob/dev.v2/event_database/cloud_functions/.gcloudignore#L14)

kev1n-peters:
> in prod, will `tokenAllowlist.json` be stored in GCS like the cache documents, or does it get uploaded with the Go files at deploy time? I ask because if it gets deployed with Go files, we'll have to remove the [gcloudignore for *.json](https://github.com/certusone/wormhole/blob/dev.v2/event_database/cloud_functions/.gcloudignore#L14)

In prod it'll be uploaded to the cache bucket

evan-gray (approved):

justinschuldt (approved):

PR #1345, opened 2022-07-11
node: optional chains in tilt

evan-gray:


Tiltfile
```diff
@@ -39,7 +39,10 @@ config.define_string("webHost", False, "Public hostname for port forwards")
 
 # Components
 config.define_bool("algorand", False, "Enable Algorand component")
+config.define_bool("eth2", False, "Enable second Eth component")
```
SEJeff, line 42: Normally names, don't matter, but eth2 literally is the name of the ethereum beacon chain that is going to be the "consensus layer" post-merge.

I'd suggest finding a better name because this is confusing a bit.

[An overview of eth2](https://www.gfinityesports.com/cryptocurrency/ethereum-2-release-date-eth2-roadmap-phases-is-ethereum-2-new-coin-serenity/)

It is a bit more unweildy, but perhaps: `second_eth`

devnet/node.yaml
```diff
@@ -93,34 +93,34 @@ spec:
...
             - --solanaContract
             - Bridge1p5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o
-            - --solanaWS
-            - ws://solana-devnet:8900
-            - --solanaRPC
-            - http://solana-devnet:8899
+            # - --solanaWS
+            # - ws://solana-devnet:8900
+            # - --solanaRPC
+            # - http://solana-devnet:8899
```
SEJeff, line 123: Are you commenting these so other people can see what they'll look like? I'm generally :-1: on updating code and committing it with a comment, but I'm guessing this is for users to be able to see how to run with everything?

node/cmd/guardiand/node.go
```diff
@@ -561,35 +567,35 @@ func runNode(cmd *cobra.Command, args []string) {
...
 	}
-	if *terra2Contract == "" {
+	if *terra2Contract == "" && !*unsafeDevMode {
 		logger.Fatal("Please specify --terra2Contract")
 	}
-	if *testnetMode || *unsafeDevMode {
+	if *testnetMode {
 		if *algorandIndexerRPC == "" {
 			logger.Fatal("Please specify --algorandIndexerRPC")
 		}
```
SEJeff, line 601: I'd be tempted do simply wrap all of this logic in a single:

```
if *unsaeDevMode {
   ... logic here ...
}
```

As it would be a bit cleaner and less to update. When someone adds the next optional chain it's harder to break it if they put this logic inside a single conditional.

SEJeff (dismissed):

bruce-riley (dismissed):

Tiltfile
evan-gray: agreed, this made me uncomfortable for the same reason the whole time I was coding it (like that's not what eth2 means). I'll change it to `second_eth` even though it sounds dumb 😛

devnet/node.yaml
evan-gray: Was on the fence about this, because it does kinda act as a one stop shop for parameters and is really confusing to any sane person who wouldn't think to check the yaml _and_ the tiltfile

SEJeff (dismissed):

SEJeff (approved):

kcsongor (approved):

PR #1349, opened 2022-07-11
cloud_functions: terra2 support

kev1n-peters:


event_database/cloud_functions/process-transfer.go
```diff
@@ -62,26 +72,40 @@ func transformHexAddressToNative(chain vaa.ChainID, address string) string {
...
 	case vaa.ChainIDAlgorand:
 		// TODO
 		return ""
+	case vaa.ChainIDTerra2:
+		if addressType == AddressTypeContract {
+			// handle terra2 native assets manually
+			if val, ok := tokenAddressExceptions[address]; ok {
+				return val
+			}
+			// TODO: other terra2 asset types
```
kev1n-peters, line 85: For now we are just supporting uluna on terra2 (that's the only asset in the allowlist anyway). But in the future we should make a contract query to get the native address.

event_database/cloud_functions/addresses-transferred-to.go
```diff
@@ -77,7 +77,7 @@ func fetchAddressRowsInInterval(tbl *bigtable.Table, ctx context.Context, prefix
 						t.DestinationAddress = string(item.Value)
 					}
 				}
-				t.DestinationAddress = transformHexAddressToNative(chainIdStringToType(t.DestinationChain), t.DestinationAddress)
+				t.DestinationAddress = transformHexAddressToNative(chainIdStringToType(t.DestinationChain), t.DestinationAddress, AddressTypeWallet)
```
evan-gray, line 80: are destinations guaranteed to be wallets?

evan-gray (commented):
I'm not sure if the wallet vs contract logic is going to work right - you might be better off checking if the first 24 hex are `0` like the sdk.

evan-gray (approved):

event_database/cloud_functions/external-data.go
```diff
@@ -107,6 +107,8 @@ func chainIdToCoinGeckoPlatform(chain vaa.ChainID) string {
 		return "klay-token"
 	case vaa.ChainIDCelo:
 		return "celo"
+	case vaa.ChainIDTerra2:
+		return "" // TODO
```
SEJeff, line 111: This seems bad? Should this not return`terra2`?

kev1n-peters: There isn't a CoinGecko asset_platform for terra2. When we add support for contract queries to get the contract's symbol, this shouldn't even need to be called for terra2

SEJeff (approved):

PR #1350, opened 2022-07-11
Update injective url/s

SEJeff:
This change ensures that the injective watcher uses the correct url for getting the blockheight.

It also removes the guardian set info bits from the terra^Wcosmwasm watcher since those only make sense on Ethereum.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1350)

### DIFF
--- a/node/pkg/terra/watcher.go
+++ b/node/pkg/terra/watcher.go
@@ -35,7 +35,6 @@ type (
 		contract string
 
 		msgChan chan *common.MessagePublication
-		setChan chan *common.GuardianSet
 
 		// Incoming re-observation requests from the network. Pre-filtered to only
 		// include requests for our chainID.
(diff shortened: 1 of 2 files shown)


node/pkg/terra/watcher.go
```diff
@@ -156,6 +156,11 @@ func (e *Watcher) Run(ctx context.Context) error {
 	readiness.SetReady(e.readiness)
 
 	go func() {
+		latestBlockURL := fmt.Sprintf("%s/blocks/latest", e.urlLCD)
```
evan-gray, line 159: should we add this logic to the watcher struct like `contractAddressFilterKey` and `contractAddressLogKey` so that all the chain differing logic stays in one place?

bruce-riley (approved):

bruce-riley (approved):

evan-gray (approved):

PR #1356, opened 2022-07-14
solana: Fix verify script

kcsongor:
Had a stale reference to the account dump file

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1356)


solana/verify
```diff
@@ -74,5 +74,7 @@ if [ "$hash1" == "$hash2" ]; then
   exit 0;
 else
   printf "\033[0;31mFailed to verify\033[0m\n" >&2;
+  echo "JSON: $account_json" >&2;
```
evan-gray, line 77: do we really want to echo the whole json and dump?

evan-gray (approved):

SEJeff (approved):

solana/verify
kcsongor: it just prints the file name (so you can see it, and open if you want)

PR #1376, opened 2022-07-20
Foundry

kcsongor:
Foundry (https://getfoundry.sh/) is a new toolchain for working with Solidity (EVM) smart contracts, allowing a more efficient testing workflow and some unique features such as writing shell scripts in Solidity.

For now, the goal is not to replace the existing Truffle toolchain, but instead augment it with additional capabilities.

I've found foundry's automatic compiler installation process a little unreliable, so I've included a script to handle that explicitly.

Some stuff already uses foundry (the upgrade simulation test), so this PR is more about making it a more integral part of the pipeline.

Added a script to derive token constructor arguments (for verification) and a test for signature verification.

This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1376)


bruce-riley (dismissed):

ethereum/contracts/Messages.sol
```diff
@@ -80,6 +80,14 @@ contract Messages is Getters {
             require(i == 0 || sig.guardianIndex > lastIndex, "signature indices must be ascending");
             lastIndex = sig.guardianIndex;
 
+            /// @dev Ensure that the provided signature index is within the
+            /// bounds of the guardianSet. This is implicitly checked by the array
```
dsterioti, line 84: Nit-picky, but this comment style is not consistent with the rest of the contract.

ethereum/contracts/Messages.sol
```diff
@@ -80,6 +80,14 @@ contract Messages is Getters {
...
+            /// index operation below, so this check is technically redundant.
+            /// However, reverting explicitly here ensures that a bug is not
+            /// introduced accidentally later due to the nontrivial storage
+            /// semantics of solidity.
+            require(sig.guardianIndex < guardianSet.keys.length, "guardian index out of bounds");
```
dsterioti, line 89: Should we store guardianSet.keys.length as a local variable instead of accessing .length in each loop?

kcsongor: yes 👍

ethereum/contracts/Messages.sol
kcsongor: In what way? It looks consistent to me, insofar as it's a natspec style comment in the wrong place (the function body).

dsterioti: I didn't realize natspec would interpret multiline comments using "///" instead of "/**"

kcsongor: we should probably remove the natspecs from inside the function bodies, as I don't think they actually get parsed, but that's outside of the scope of this PR

bruce-riley (approved):

PR #1379, opened 2022-07-22
solana: SDK upgrade to v1.10.31

drozdziak1:
The SDK upgrade caused a stack offset violation error in the tests, due to additional stack allocation resulting from the SDK upgrade. The solution is to allocate account data on the heap instead.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379)


drozdziak1:
I marked this as WIP because of the change to Rust container instead of the Certus One container. Please let me know how we want to handle the dependency on it

Tiltfile
```diff
@@ -39,6 +39,7 @@ config.define_string("webHost", False, "Public hostname for port forwards")
 
 # Components
 config.define_bool("algorand", False, "Enable Algorand component")
+config.define_bool("pyth", False, "Enable Pyth") # TODO(sdrozd): remove this before contributing
```
kcsongor, line 42: 👀

solana/Dockerfile
```diff
@@ -1,5 +1,15 @@
 #syntax=docker/dockerfile:1.2@sha256:e2a8561e419ab1ba6b2fe6cbdf49fd92b95912df1cf7d313c3e2230a333fdbcc
-FROM ghcr.io/certusone/solana:1.9.4@sha256:5389eccba0ba59ae119bc1438b61be8904707e26df8f0732116a8dce0f64eb52 AS solana
+# FROM ghcr.io/certusone/solana:1.9.4@sha256:5389eccba0ba59ae119bc1438b61be8904707e26df8f0732116a8dce0f64eb52 AS solana
+FROM rust:1.60 as solana
+
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - && apt-get install -y nodejs
```
kcsongor, line 5: will push a new image so we don't have to do this install

kcsongor (dismissed):

solana/Dockerfile.wasm
```diff
@@ -3,7 +3,7 @@ FROM docker.io/library/rust:1.49@sha256:a50165ea96983c21832578afb1c8c028674c965b
 
 RUN apt-get update && apt-get install -y libssl-dev libudev-dev pkg-config zlib1g-dev llvm clang
 RUN rustup component add rustfmt
-RUN rustup default nightly-2022-01-02
+RUN rustup default nightly-2022-07-12
```
kcsongor, line 6: instead of setting this here, we should copy `rust-toolchain` from the host too

jynnantonix (changes requested):
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379)* status: all files reviewed, 9 unresolved discussions (waiting on @drozdziak1)

*[a discussion](https://reviewable.io/reviews/certusone/wormhole/1379#-N7mubc87vemR4ZaKy_2:-N7mubc87vemR4ZaKy_3:b-73tsmi) (no related file):*
Also, looks like the integration tests need to be updated:

```
 solana-tests │ #14 113.7 error[E0308]: mismatched types
 solana-tests │ #14 113.7   --> bridge/program/tests/common.rs:66:5
 solana-tests │ #14 113.7    |
 solana-tests │ #14 113.7 66 | /     client
 solana-tests │ #14 113.7 67 | |         .process_transaction_with_commitment(transaction, commitment_level)
 solana-tests │ #14 113.7 68 | |         .await
 solana-tests │ #14 113.7    | |______________^ expected enum `solana_sdk::transport::TransportError`, found enum `BanksClientError`
 solana-tests │ #14 113.7    |
 solana-tests │ #14 113.7    = note: expected enum `Result<_, solana_sdk::transport::TransportError>`
 solana-tests │ #14 113.7               found enum `Result<_, BanksClientError>`
 solana-tests │ #14 113.7 
```

___

jynnantonix (dismissed):
lgtm w/ nits

> Reviewed 5 of 5 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379#-:-N8WCOEODYTZ1JP5PaBF:bjn82aj)* status: all files reviewed, 4 unresolved discussions (waiting on @drozdziak1)

___
*[`solana/bridge/program/Cargo.toml` line 33 at r2](https://reviewable.io/reviews/certusone/wormhole/1379#-N7mt5N-CdkdR8MqJkZI:-N8WCic3AUvf1gi4TCJa:b-teb5e0) ([raw file](https://github.com/certusone/wormhole/blob/901eae12efc713982860bbcdf5eb267f7e558daf/solana/bridge/program/Cargo.toml#L33)):*

Unused dependency?  Is there a corresponding change to an rs file that uses this crate?
</blockquote></details>

Need to drop this unused dep.

jynnantonix (dismissed):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379)* status: all files reviewed, 4 unresolved discussions (waiting on @drozdziak1)

solana/Dockerfile.wasm
drozdziak1: Yup!

jynnantonix (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379)* status: all files reviewed, 4 unresolved discussions (waiting on @drozdziak1 and @kcsongor)

kcsongor:
Squashed the commits and added instructions on future SDK ugprades.

kcsongor:
@Reisen can you please take a look too, in particular at the heap allocation code.

jynnantonix (approved):
Left some suggestions but I'm ok with merging this as is.

> Reviewed 25 of 25 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379#-:-N8uu4o-8if9aFtnYNou:bl9f07y)* status: all files reviewed, 4 unresolved discussions (waiting on @drozdziak1, @kcsongor, and @Reisen)

___
*[`solana/UPGRADE.md` line 29 at r10](https://reviewable.io/reviews/certusone/wormhole/1379#-N8ul8i-008q-XOPM_xq:-N8ul8i-008q-XOPM_xr:bshd7vd) ([raw file](https://github.com/certusone/wormhole/blob/bd53ae6fea50d391b7d40d39fcd552ff6d162c12/solana/UPGRADE.md#L29)):*
> ```Markdown
> 
> ```sh
> find . -name "Cargo.toml" -exec sed -i 's/=1.9.4/=1.10.31/g' {} \;
> ```

FYI, the `cargo upgrade` command from the [cargo-edit crate](https://crates.io/crates/cargo-edit) can automate most of this.  So in this case we would run `cargo upgrade --workspace --pinned solana-sdk@=1.10.31` and then fix the dependency resolution conflicts as they come up.  When I ran this upgrade locally my final commandline was

```
cargo upgrade --workspace --pinned solana-sdk@=1.10.31 solana-client@=1.10.31 solana-program@=1.10.31 \
           solana-clap-utils@=1.10.31 solana-program-test@=1.10.31 borsh@=0.9.3 spl-token@=3.3.0 \
           solana-cli-config@=1.10.31
```
___
*[`solana/solitaire/program/src/lib.rs` line 86 at r10](https://reviewable.io/reviews/certusone/wormhole/1379#-N8unVxsEA_Sl7gqhzli:-N8unVxsEA_Sl7gqhzlj:bufsosy) ([raw file](https://github.com/certusone/wormhole/blob/bd53ae6fea50d391b7d40d39fcd552ff6d162c12/solana/solitaire/program/src/lib.rs#L86)):*
> ```XML
> /// references. A list of dependent accounts is produced as a side effect of the parsing stage.
> pub trait FromAccounts<'a, 'b: 'a> {
>     fn from<T>(_: &'a Pubkey, _: &mut Iter<'a, AccountInfo<'b>>, _: &'a T) -> Result<Box<Self>>
> ```

I think if we wanted to avoid forcing the Box on all implementors of this trait we could use `mem::size_of` in the macro to only box the types that are too big.  `size_of` is a const function so the unused branch should be compiled out of the final binary.  Rust doesn't have `decltype` though so I think we'd have to change the macro to take in the necessary type as a parameter.  Something like:

```rust
pub fn execute<'a, 'b: 'a, 'c>(p: &Pubkey, a: &'c [AccountInfo<'b>], d: &[u8]) -> Result<()> {
    let ix_data = BorshDeserialize::try_from_slice(d).map_err(|e| SolitaireError::InstructionDeserializeFailed(e))?;
    if ::std::mem::size_of::<$A>() > STACK_THRESHOLD {
        let mut accounts = FromAccounts::from(p, &mut a.iter(), &()).map(Box::new)?;
        $fn(&ExecutionContext{program_id: p, accounts: a}, &mut accounts, ix_data)?;
        Persist::persist(accounts.as_ref(), p)
    } else {
        let mut accounts = FromAccounts::from(p, &mut a.iter(), &())?;
        $fn(&ExecutionContext{program_id: p, accounts: a}, &mut accounts, ix_data)?;
        Persist::persist(accounts.as_ref(), p)
    }
}
```


On a separate note, since we're taking a reference to `T` here the generic bound should be `T: ?Sized`.  Currently `T` cannot be an unsized type like a trait object (`dyn Foo`) or a slice `[u8]`, which doesn't really matter for now but could be an issue in the future.

drozdziak1:
@jynnantonix While this is an interesting idea, unless $A somehow represents the whole stack frame, it will be impossible to use it as a reliable trigger for using a `Box`. i.e. if any code other than the accounts struct grows, we run a risk of this code allocating on stack when it shouldn't. Besides, this stack limit is very low-level,, as in LLVM level. There's no chance we'll have that kind of introspection at Rust level. Independently, we can't guarantee that the bytecode will never choose to bump the stack when branching off below STACK_THRESHOLD. The compiler may choose to make the program pre-allocate the struct by bumping the stack pointer before the condition is even evaluated.

jynnantonix:
> unless $A somehow represents the whole stack frame

`$A` in this case would be the type that implements `FromAccounts`, passed in as a parameter to the macro.

> if any code other than the accounts struct grows, we run a risk of this code allocating on stack when it shouldn't

Yes, I agree that we don't know what else is going on in the stack so `STACK_THRESHOLD` would be have to be some heuristic value picked by us (512 bytes, 1K, etc) and not based on anything else.

> The compiler may choose to make the program pre-allocate the oversize struct anyway by bumping the stack pointer before the condition is even evaluated.

`mem::size_of::<T>()` is a const function, which means the branch is evaluated _at compile time_.  There is no branch in the emitted bytecode.  It will either get the boxed version or the stack version but never both (except for maybe in debug builds with no optimizations).

drozdziak1 (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1379#-:-N91WZv8-rA2GK3HlEfs:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @jynnantonix, @kcsongor, and @Reisen)

___
*[`Tiltfile` line 42 at r1](https://reviewable.io/reviews/certusone/wormhole/1379#-N7a0tOh7jGIP17dyEub-r1-42:-N91WhZmEpjm9Q3Aey15:b-896fix) ([raw file](https://github.com/certusone/wormhole/blob/b7f6afb36ee157659cf2a7253415ac6566c29a05/Tiltfile#L42)):*
<details><summary><i>Previously, kcsongor (Csongor Kiss) wrote…</i></summary><blockquote>

👀
</blockquote></details>

Done.
___
*[`solana/Dockerfile` line 5 at r1](https://reviewable.io/reviews/certusone/wormhole/1379#-N7a0tOh7jGIP17dyEue-r1-5:-N91WmT56UBhDXunRkxV:b-896fix) ([raw file](https://github.com/certusone/wormhole/blob/b7f6afb36ee157659cf2a7253415ac6566c29a05/solana/Dockerfile#L5)):*
<details><summary><i>Previously, kcsongor (Csongor Kiss) wrote…</i></summary><blockquote>

will push a new image so we don't have to do this install
</blockquote></details>

Done.
___
*[`solana/bridge/program/Cargo.toml` line 33 at r2](https://reviewable.io/reviews/certusone/wormhole/1379#-N7mt5N-CdkdR8MqJkZI:-N91WsHyBQLibYAJNJS5:b-896fix) ([raw file](https://github.com/certusone/wormhole/blob/901eae12efc713982860bbcdf5eb267f7e558daf/solana/bridge/program/Cargo.toml#L33)):*

Need to drop this unused dep.
</blockquote></details>

Done.

kcsongor (approved):

PR #1381, opened 2022-07-26
Wormhole chain

jynnantonix:
This is a rebase of the wormhole-chain branch on dev.v2.  The original commits have been left unmodified as much as possible except for places where I had to fix merge conflicts or broken tests.

I debated doing a merge commit instead so it would be more obvious where I made changes but considering that almost all the original code didn't really go through code review it didn't seem like there was much benefit in separating out my changes from the original code.

This doesn't include the accounting changes.  If this lands first then I will rebase the accounting PR on dev.v2.  Otherwise if the accounting PR lands first I will update this PR with the changes from that branch.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1381)


jynnantonix:
There are a lot of files in this change, many of which come from our copy of cosmos-sdk or were autogenerated by starport.  The things that are the highest priority for a closer look are the x/wormhole and x/tokenbridge modules in the wormhole_chain directory as well as the changes to the guardiand node.

node/pkg/wormchain/watcher.go
```diff
@@ -0,0 +1,517 @@
...
+			//TODO do we also need to retrieve the addresses via a second webcall, or can this be omitted?
+			//addresses := gjson.Get(json, "result.addresses.#.bytes")
+
+			logger.Debug("current guardian set on Wormchain",
+				zap.Any("guardianSetIndex", guardianSetIndex))
+			
+
+			resp.Body.Close()
+
+			// We do not send guardian changes to the processor - ETH guardians are the source of truth.
```
bruce-riley, line 310: We recently ripped the guardian set stuff out of the Terra / Injective watcher because it was only used for logging anyway. You can probably do the same thing here.

jynnantonix: Done

jynnantonix:
I've rebased the branch on the latest dev.v2 and pulled in the js sdk that contains the wormholechain constants.

I also have a range diff showing the difference between this branch and the wormhole-chain branch in the forked repo but github refuses to let me upload it.  Please let me know if you'd like me to share the patch with you directly.

jynnantonix:
I've squashed most of the changes to the wormhole_chain directory into a single commit except for audit related changes, which I've left separate in order to make it easier to link to them.  The squashed commit includes the following changes:

 * Replace the vendored cosmos-sdk directory with a replace directive in
   the go.mod file pointing to the wormhole-foundation/cosmos-sdk repo.
 * Update the cosmos-sdk version to v0.45.7.
 * Rename starport -> ignite
 * Update to ignite version 0.23.0
 * Update go version in dockerfile to 1.19.0
 * Upgrade github.com/cosmos/ibc-go v1.2.0 => v1.2.2
 * Upgrade github.com/tendermint/spm v0.1.5 => v0.1.9
 * Add missing module names to the
   SetOrder{BeginBlockers,EndBlockers,InitGenesis} functions.
 * Fix the gentx memo in the genesis.json file.

I've also squashed all the integration changes into a single commit.

hendrikhofstadt:
Is it intended that the `build` folder is checked in?

jynnantonix:
> Is it intended that the `build` folder is checked in?

I think so, though "build" might be a misnomer.  That directory is meant to act as the home directory for dev builds and has the appropriate genesis.json, devnet validator key, etc

hendrikhofstadt:
I was surprised because the `development.md` reads:

```
This command creates a `build` directory and in particular, the
`build/wormhole-chaind` binary, which can be used to run and interact with the
blockchain.
```

hendrikhofstadt (dismissed):

hendrikhofstadt (approved):

bruce-riley (approved):
I only reviewed the guardian and sdk/vaa.

wormhole_chain/x/tokenbridge/types/asset.go
```diff
@@ -0,0 +1,158 @@
...
+		return big.NewInt(1), nil
+	}
+}
+
+var uwormChain uint16 = 1
+var uwormAddress = [32]byte{0x16, 0x58, 0x09, 0x73, 0x92, 0x40, 0xa0, 0xac, 0x03, 0xb9, 0x84, 0x40, 0xfe, 0x89, 0x85, 0x54, 0x8e, 0x3a, 0xa6, 0x83, 0xcd, 0x0d, 0x4d, 0x9d, 0xf5, 0xb5, 0x65, 0x96, 0x69, 0xfa, 0xa3, 0x01}
+
+func IsWORMToken(tokenChain uint16, tokenAddress [32]byte) bool {
+	// TODO(csongor): figure out WORM token address on Solana
+	// TODO(csongor): this should be configurable in the genesis
```
kcsongor, line 70: I think we should add the wormtoken origin info into the genesis config instead

kcsongor (approved):

wormhole_chain/x/tokenbridge/types/asset.go
jynnantonix: Sounds good to me but I'd like to do this in a follow-up change.

jynnantonix: Actually, since we're going to be dropping this entire module maybe this isn't necessary.

node/cmd/guardiand/node.go
```diff
@@ -1023,6 +1034,15 @@ func runNode(cmd *cobra.Command, args []string) {
 			}
 		}
 
+		// Start Wormchain watcher only if configured
+		if *wormchainWS != "" && *wormchainLCD != "" {
```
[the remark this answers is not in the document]
jynnantonix: I will do this in a follow-up change.

wormhole_chain/notes-for-chase.txt
```diff
@@ -0,0 +1,140 @@
+--------------------------------------------------------------------------------
```
conorpp, line 1: let's delete this file?

conorpp (commented):
Reviewed all keeper modules + genesis.
- There are a number of steps in VAA checking that are duplicated across message handlers.
- The max number of validators is set to 1 in genesis.json
- Should be some unit tests for message handlers evaluating VAA's.

I've included fixes for these in https://github.com/wormhole-foundation/wormhole/pull/1674

jynnantonix:
For VAA verification, we want to delete everything here and just use the code from the wormhole go sdk directly.

PR #1386, opened 2022-07-28
Add wormhole chain to the sdk

jynnantonix:
Add support for wormhole chain to the sdk.  Since the `humanAddress` and `canonicalAddress` functions in the terra package are also needed for wormhole chain, move them into a new cosmos package.  Also, stop hard coding "terra" in `humanAddress` and take the human-readable part as a parameter instead.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1386)

### DIFF
--- a/sdk/js/src/cosmos/address.ts
+++ b/sdk/js/src/cosmos/address.ts
@@ -1,8 +1,8 @@
 import { bech32 } from "bech32";
 
 export function canonicalAddress(humanAddress: string) {
-    return new Uint8Array(bech32.fromWords(bech32.decode(humanAddress).words));
+  return new Uint8Array(bech32.fromWords(bech32.decode(humanAddress).words));
 }
 export function humanAddress(hrp: string, canonicalAddress: Uint8Array) {
-    return bech32.encode(hrp, bech32.toWords(canonicalAddress));
+  return bech32.encode(hrp, bech32.toWords(canonicalAddress));
 }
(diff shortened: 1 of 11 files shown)


kev1n-peters:
PR looks good. Only comment is that contributors use the [VS Code prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) with default settings for formatting typescript code. A CI job that enforces consistent typescript formatting will likely be added in the future.

kev1n-peters (approved):

sdk/js/src/cosmos/address.ts
```diff
@@ -0,0 +1,8 @@
+import { bech32 } from "bech32";
+
+export function canonicalAddress(humanAddress: string) {
+  return new Uint8Array(bech32.fromWords(bech32.decode(humanAddress).words));
+}
+export function humanAddress(hrp: string, canonicalAddress: Uint8Array) {
```
bruce-riley, line 6: Just thinking out loud here, and you have probably already thought about it, so you can feel free to just close this issue. 

This is a breaking change (adding a parameter). Would it be better to leave this function as is and create a separate one that takes the extra parameter?

bruce-riley (approved):

sdk/js/src/cosmos/address.ts
kev1n-peters: It's been decided to break backwards compatibility and add the new parameter to the function since it's no longer terra specific

evan-gray (approved):

PR #1387, opened 2022-07-28
cloud_functions: Use CoinGecko API IDs from allowlist in TVL calculations

kev1n-peters:
Specify the CoinGecko API ID for tokens in the allowlist.
It's useful to manually specify the API ID to in certain cases.

---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1387)


ckeun (approved):

event_database/cloud_functions/process-transfer.go
```diff
@@ -217,11 +217,6 @@ func ProcessTransfer(ctx context.Context, m PubSubMessage) error {
 			nativeTokenAddress = string(item.Value)
 		}
 	}
-	if coinId == "" {
```
justinschuldt, line 220: just wondering, do we want to write transfer details to the row if we do not have coinId or price?  I'm wondering if that will have downstream effects, code that expects a coinId or price if the transfer details family exists.

justinschuldt (approved):

event_database/cloud_functions/process-transfer.go
kev1n-peters: The downstream/calculate functions now look up the CoinGecko API ID in the allowlist and don't depend on the one set in the transfer details column family or the price. I think that it should be okay for the transfer details column family to exist even if the coinId or price is empty/0

PR #1390, opened 2022-07-29
node: read checkpoints on Ethereum for polygon blocks

bruce-riley:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1390)

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -72,8 +72,10 @@ var (
 	bscRPC      *string
 	bscContract *string
 
-	polygonRPC      *string
-	polygonContract *string
+	polygonRPC                      *string
+	polygonContract                 *string
+	polygonRootChainRpc             *string
+	polygonRootChainContractAddress *string
 
 	ethRopstenRPC      *string
 	ethRopstenContract *string
@@ -191,6 +193,8 @@ func init() {
 
 	polygonRPC = NodeCmd.Flags().String("polygonRPC", "", "Polygon RPC URL")
 	polygonContract = NodeCmd.Flags().String("polygonContract", "", "Polygon contract address")
+	polygonRootChainRpc = NodeCmd.Flags().String("polygonRootChainRpc", "", "Polygon root chain RPC")
+	polygonRootChainContractAddress = NodeCmd.Flags().String("polygonRootChainContractAddress", "", "Polygon root chain contract address")
 
 	ethRopstenRPC = NodeCmd.Flags().String("ethRopstenRPC", "", "Ethereum Ropsten RPC URL")
 	ethRopstenContract = NodeCmd.Flags().String("ethRopstenContract", "", "Ethereum Ropsten contract address")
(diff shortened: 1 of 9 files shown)


bruce-riley:
This code reads the Root Chain contract on Ethereum to get the latest confirm blocks on polygon.

node/pkg/ethereum/polygonimpl.go
```diff
@@ -0,0 +1,240 @@
...
+	err       chan error
+	quit      chan error
+	unsubDone chan struct{}
+}
+
+var ErrUnsubscribedForPolygonPolling = errors.New("unsubscribed")
+
+func (sub *PolygonPollSubscription) Err() <-chan error {
+	return sub.err
+}
```
SEJeff, line 132: Why not name this `Error()`? That's more in line with most Golang stuff.

node/pkg/ethereum/polygonimpl.go
```diff
@@ -0,0 +1,240 @@
...
+		Hash ethCommon.Hash `json:"hash"`
+	}
+
+	var m Marshaller
+	err := e.rawClient.CallContext(ctx, &m, "eth_getBlockByNumber", numStr, false)
+	if err != nil {
+		e.logger.Error("RCP: failed to get block", zap.String("eth_network", e.BaseEth.NetworkName),
+			zap.String("requested_block", numStr), zap.Error(err))
+		return nil, err
+	}
```
SEJeff, line 237:
```suggestion
	if err := e.rawClient.CallContext(ctx, &m, "eth_getBlockByNumber", numStr, false); err != nil {
		e.logger.Error("RCP: failed to get block", zap.String("eth_network", e.BaseEth.NetworkName),
			zap.String("requested_block", numStr), zap.Error(err))
		return nil, err
	}
```

If you want, you can collapse this since you only use `err` inside that conditional.

SEJeff (dismissed):

evan-gray:
Discussed with @bruce-riley and @kcsongor  that the polygon watcher should allow an option for both waiting a number of polygon blocks and waiting for a checkpoint. Moving this to draft status while that is worked out.

tbjump (changes requested):
This implementation relies on two separate connections. One to the Polygon RPC node and one to the Ethereum RPC node. The way I understand it is that this code subscribes to Wormhole events on Polygon as usual, notes their block height, and then separately polls getLastChildBlock() from the Polygon contract on Ethereum to watch for new blocks. Then once we are at a block height sufficiently higher than the initial observation, we give the green light. 

I think there is a problem here: There is nothing linking the Polygon node state to the snapshot hashes submitted to Ethereum. E.g. say we observe a wormhole event on Polygon and then our polygon node stalls. The polygon mainnet doesn't end up including that transaction in the snapshot, but keeps producing new blocks. We will end up green-lighting that transaction after a number of new blocks even though it is not part of the checkpoint. 

On other chains we don't have that problem since there is only one RPC node we connect to. See the comment here: 
https://github.com/certusone/wormhole/blob/5111b40ae0cbcb3446cb69f60783aca1b2f13048/node/pkg/ethereum/watcher.go#L466

node/cmd/guardiand/node.go
```diff
@@ -932,73 +938,83 @@ func runNode(cmd *cobra.Command, args []string) {
...
+			var extraParams []string
+			var minConfirmations uint64 = 512
+			if *polygonRootChainRpc != "" {
+				if *polygonRootChainContractAddress == "" {
+					log.Fatal("if \"polygonRootChainRpc\" is specified then \"polygonRootChainContractAddress\" is required", zap.Error(err))
+				}
+				extraParams = []string{*polygonRootChainRpc, *polygonRootChainContractAddress}
+				minConfirmations = 1
+			} else if *testnetMode {
+				minConfirmations = 64
```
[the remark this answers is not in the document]
bruce-riley: It will stay 512 if we are running in mainnet without the root chain feature enabled. Leaving as is.

node/pkg/ethereum/polygonimpl.go
bruce-riley: Changed.

node/pkg/ethereum/polygonimpl.go
bruce-riley: Err() is the standard name used for subscription errors in geth, so I am staying consistent.

PR #1397, opened 2022-08-04
node: add near support

jumpsiegel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1397)


bruce-riley (approved):

node/pkg/near/watcher.go
```diff
@@ -0,0 +1,381 @@
...
+			case <-ctx.Done():
+				return
+			case r := <-e.obsvReqC:
+				if vaa.ChainID(r.ChainId) != vaa.ChainIDNear {
+					panic("invalid chain ID")
+				}
+
+				txHash := base58.Encode(r.TxHash)
+
+				logger.Info("Received obsv request", zap.String("tx_hash", txHash))
```
SEJeff, line 306: Can you add the chain name to these logs please? It makes it much harder for some of these more generic messages to troubleshoot or just see what's happening.

node/pkg/near/watcher.go
```diff
@@ -0,0 +1,381 @@
...
+					errC <- err
+					return
+				}
+				parsedFinalBody := gjson.ParseBytes(finalBody)
+				lastBlock := parsedFinalBody.Get("result.chunks.0.height_created").Uint()
+
+				logger.Info("lastBlock", zap.Uint64("lastBlock", lastBlock), zap.Uint64("next_round", e.next_round))
+
+				if lastBlock < e.next_round {
+					logger.Error("Went backwards... ")
```
SEJeff, line 328: That's not a more fatal error? What could cause this? Failing over to a different rpc node? or does the near chain go backwards?

SEJeff (approved):
This looks great. I wish that the watcher logs would contain the chain name however as that would make it a lot easier to parse the very spammy guardian logs with watcher logs for every single chain.

node/pkg/near/watcher.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	outcomes := gjson.ParseBytes(t).Get("result.receipts_outcome")
+
+	if !outcomes.Exists() {
+		return nil
+	}
```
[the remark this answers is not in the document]
jumpsiegel: No, just a empty outcome block... I see it all the time in devnet when nobody sends any txn's

node/pkg/near/watcher.go
jumpsiegel: devnet behavior when we reset the chain

node/pkg/near/watcher.go
```diff
@@ -0,0 +1,381 @@
...
+		nearRPC:          nearRPC,
+		wormholeContract: wormholeContract,
+		msgChan:          lockEvents,
+		obsvReqC:         obsvReqC,
+		next_round:       0,
+	}
+}
+
+func (e *Watcher) getBlock(block uint64) ([]byte, error) {
+	s := fmt.Sprintf(`{"id": "dontcare", "jsonrpc": "2.0", "method": "block", "params": {"block_id": %d}}`, block)
```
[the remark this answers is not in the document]
jumpsiegel: The id is required but since we only have one call at a time, we dontcare.  They use it in their own examples as well

node/pkg/near/watcher.go
jumpsiegel: isn't the filename in the logs?  Is there a standard on how you want to see the chain?  I don't see the chain in the "message observed" message for solana either

node/pkg/near/watcher.go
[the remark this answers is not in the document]
SEJeff: Oh I love it. Wasn't saying get rid of it.

node/pkg/near/watcher.go
SEJeff: Yeah, this is one of those things we really badly need to just burn through and refactor. We can do it in a followup.

node/pkg/near/watcher.go
SEJeff: Seems like that would be a bfd if it happened on mainnet.

PR #1405, opened 2022-08-08
Proposed Governor mainnet config

bruce-riley:
Change-Id: Ia84a9bd79bb15a4915e54cf019be23f8eee0fdcf

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1405)


ckeun (approved):

tbjump (approved):
I like the initial proposal of a flat $500k/24h limit for the lower-activity chains. This is 3 orders of magnitude less than the overall TVL of ~$500M and still much higher than historic daily bridging activity. This should be at least good enough to get the first iteration of this feature out of the door. Over time, we can iterate on criteria for setting the limit and fine-tune it.

PR #1410, opened 2022-08-09
sdk/js: Added getGovernorIsVAAEnqueued functions

kev1n-peters:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1410)

### DIFF
--- a/sdk/js/src/rpc/getGovernorIsVAAEnqueuedWithRetry.ts
+++ b/sdk/js/src/rpc/getGovernorIsVAAEnqueuedWithRetry.ts
@@ -1,0 +1,35 @@
+import { ChainId, ChainName, getGovernorIsVAAEnqueued } from "..";
+import { coalesceChainId } from "../utils";
+
+export async function getGovernorIsVAAEnqueuedWithRetry(
+  hosts: string[],
+  emitterChain: ChainId | ChainName,
+  emitterAddress: string,
+  sequence: string,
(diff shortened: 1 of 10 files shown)


sdk/js/src/rpc/getGovernorIsVAAEnqueuedWithRetry.ts
```diff
@@ -0,0 +1,35 @@
...
+  emitterChain: ChainId | ChainName,
+  emitterAddress: string,
+  sequence: string,
+  extraGrpcOpts = {},
+  retryTimeout = 1000,
+  retryAttempts?: number
+) {
+  let currentWormholeRpcHost = -1;
+  const getNextRpcHost = () => ++currentWormholeRpcHost % hosts.length;
+  let result;
```
panoel, line 15: Is this guaranteed to be initialized to a false equivalent or should it be explicitly initialized so that the while loop is guaranteed to run the first time?

panoel (dismissed):

sdk/js/src/rpc/getGovernorIsVAAEnqueuedWithRetry.ts
kev1n-peters: `result` === `undefined` here so the while loop will exec the first time

evan-gray (approved):

panoel (approved):

PR #1418, opened 2022-08-10
Governor should handle duplicate coin gecko ids

bruce-riley:
The governor was not properly handling the case where there are multiple tokens using the same CoinGecko symbol.

The governor previously had a map from CoinGecko symbol to a reference to a token object. The change is to make that a map from the CoinGecko symbol to an array of references to tokens, and to update all of those tokens when the CoinGecko price is udpated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1418)

### DIFF
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -105,7 +105,7 @@ type ChainGovernor struct {
 	logger              *zap.Logger
 	mutex               sync.Mutex
 	tokens              map[tokenKey]*tokenEntry
-	tokensByCoinGeckoId map[string]*tokenEntry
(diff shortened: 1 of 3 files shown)


tbjump:
Good catch! To help review, can you point me to the Coingecko API spec, or better add a link as a comment in the code?

evan-gray (approved):

SEJeff (approved):

evan-gray:
> Good catch! To help review, can you point me to the Coingecko API spec, or better add a link as a comment in the code?

https://www.coingecko.com/en/api/documentation 🙂

PR #1425, opened 2022-08-12
add EVM contract for batched VAA development

justinschuldt:
Added an EVM contract that calls wormhole core's `publishMessage` several times, in order to produce VAAs that should be batched.

Also added a truffle script that invokes the new contract's method, and some tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1425)

### DIFF
--- a/devnet/eth-devnet.yaml
+++ b/devnet/eth-devnet.yaml
@@ -53,7 +53,7 @@ spec:
           command:
             - /bin/sh
             - -c
-            - "npm run migrate && npx truffle exec scripts/deploy_test_token.js && npx truffle exec scripts/register_solana_chain.js && npx truffle exec scripts/register_terra_chain.js && npx truffle exec scripts/register_terra2_chain.js && npx truffle exec scripts/register_bsc_chain.js && npx truffle exec scripts/register_algo_chain.js && nc -lkp 2000 0.0.0.0"
+            - "npm run migrate && npm run deploy-batched-vaa-sender && npx truffle exec scripts/deploy_test_token.js && npx truffle exec scripts/register_solana_chain.js && npx truffle exec scripts/register_terra_chain.js && npx truffle exec scripts/register_terra2_chain.js && npx truffle exec scripts/register_bsc_chain.js && npx truffle exec scripts/register_algo_chain.js && nc -lkp 2000 0.0.0.0"
           readinessProbe:
             periodSeconds: 1
             failureThreshold: 300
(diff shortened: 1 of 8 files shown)


devnet/eth-devnet.yaml
```diff
@@ -53,7 +53,7 @@ spec:
           command:
             - /bin/sh
             - -c
-            - "npm run migrate && npx truffle exec scripts/deploy_test_token.js && npx truffle exec scripts/register_solana_chain.js && npx truffle exec scripts/register_terra_chain.js && npx truffle exec scripts/register_terra2_chain.js && npx truffle exec scripts/register_bsc_chain.js && npx truffle exec scripts/register_algo_chain.js && nc -lkp 2000 0.0.0.0"
+            - "npm run migrate && npm run deploy-batched-vaa-sender && npx truffle exec scripts/deploy_test_token.js && npx truffle exec scripts/register_solana_chain.js && npx truffle exec scripts/register_terra_chain.js && npx truffle exec scripts/register_terra2_chain.js && npx truffle exec scripts/register_bsc_chain.js && npx truffle exec scripts/register_algo_chain.js && nc -lkp 2000 0.0.0.0"
```
kcsongor, line 56: I would put this after `deploy_test_token.js`. That has some logic to pad the deployer nonce to allow `npm run migrate` to grow: https://github.com/certusone/wormhole/blob/dev.v2/ethereum/scripts/deploy_test_token.js#L18-L28.
And I'd add a line like https://github.com/certusone/wormhole/blob/8818d4b8f0471095dd48fa6f5da9c315cbfc9b52/ethereum/scripts/deploy_test_token.js#L99-L102 into the deploy script to ensure that it's deployed to the correct address (it's a pain to hunt down when these addresses change)

kcsongor (commented):
Looks good, might be interesting to consider writing the sender logic as a forge script: https://book.getfoundry.sh/tutorials/solidity-scripting
the script can initiate transactions without it itself having to be deployed as a contract on chain (see `vm.startBroadcast()`). If you don't want to give that a go then the js script you wrote looks good as is.

devnet/eth-devnet.yaml
justinschuldt: Sure, not a bad idea, I added it.

We don't really need this contract address to be deterministic, the client that invokes the contract finds the address in the ABI from the last deploy, but it doesn't hurt.

kcsongor (approved):

dsterioti (approved):

PR #1427, opened 2022-08-12
algo/prod: prod changes for going live

jumpsiegel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1427)

### DIFF
--- a/algorand/admin.py
+++ b/algorand/admin.py
@@ -630,7 +630,7 @@ class PortalCore:
             )
         ]
 
-        self.sendTxn(client, sender, txns, True)
+        return self.sendTxn(client, sender, txns, True)
 
     def decodeLocalState(self, client, sender, appid, addr):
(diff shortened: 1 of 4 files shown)


algorand/admin.py
```diff
@@ -1333,6 +1333,7 @@ def testnet(self):
         self.tokenid = self.args.tokenid
 
     def mainnet(self):
+        #self.ALGOD_ADDRESS = "http://localhost:4001"
```
panoel, line 1336: Probably can remove

panoel:
Is the teal code generated?  Because there is no way I understand it.

bruce-riley (approved):

algorand/admin.py
jumpsiegel: Going to make this a settable argument for testing in another branch

PR #1429, opened 2022-08-13
node: don't allow obs reqs if the queue is full

bruce-riley:
Change-Id: Ifb0d038fa3adeddc6226e2289fe9dbfc8e39b4e7

Jeff noticed a while ago that too many reobservation requests can hang the guardian. This is because an attempt to write to the channel hangs if the channel has reached capacity. This PR addresses by making sure the channel is not full before writing to it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1429)

### DIFF
--- a/node/pkg/common/obsvReqSendC.go
+++ b/node/pkg/common/obsvReqSendC.go
@@ -1,0 +1,19 @@
+package common
+
+import (
+	"fmt"
+
+	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+)
+
+const ObsvReqChannelSize = 50
+const ObsvReqChannelFullError = "channel is full"
+
+func PostObservationRequest(obsvReqSendC chan *gossipv1.ObservationRequest, req *gossipv1.ObservationRequest) error {
+	if len(obsvReqSendC) >= cap(obsvReqSendC) {
+		return fmt.Errorf(ObsvReqChannelFullError)
+	}
+
+	obsvReqSendC <- req
+	return nil
+}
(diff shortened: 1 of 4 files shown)


leoluk (approved):

SEJeff (approved):
Love it! Do we want to consider bumping this channel size up a bit? Say maybe 100? During the worst events I've ever seen, we once had about 80 or so when the L1 was going bananas and nodes were dropping left and right.

Trying to rapid fire reobservations via some unspeakable shell one liners is how I found the "feature". This looks great!

bruce-riley:
We discussed increasing the queue size and decided to leave it as is for now.

node/pkg/common/obsvReqSendC.go
```diff
@@ -0,0 +1,19 @@
...
+const ObsvReqChannelSize = 50
+const ObsvReqChannelFullError = "channel is full"
+
+func PostObservationRequest(obsvReqSendC chan *gossipv1.ObservationRequest, req *gossipv1.ObservationRequest) error {
+	if len(obsvReqSendC) >= cap(obsvReqSendC) {
+		return fmt.Errorf(ObsvReqChannelFullError)
+	}
+
+	obsvReqSendC <- req
+	return nil
```
jynnantonix, line 18: Since this was already merged I don't think we need to change it but I believe the idiomatic way to do this in go is to use a select statement:

```suggestion
select {
	obsvReqSendC <- req:
		return nil
	default:
		return ObsvReqChannelFullError
}
```

Since any goroutine can push to a channel the channel could become full immediately after the length check, which would cause the send to block.  `select` avoids this issue because if the actual send blocks then it will fall through to the default statement.

I think the part of guardiand that calls this method is single-threaded so it doesn't matter in practice but we need to make sure we update this if we ever start pushing into this channel from multiple goroutines.

PR #1435, opened 2022-08-15
node: reobserver

bruce-riley:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1435)

### DIFF
--- a/node/pkg/processor/processor.go
+++ b/node/pkg/processor/processor.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/certusone/wormhole/node/pkg/db"
 	"github.com/certusone/wormhole/node/pkg/governor"
+	"github.com/certusone/wormhole/node/pkg/reobserver"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -112,8 +113,9 @@ type Processor struct {
 	// cleanup triggers periodic state cleanup
 	cleanup *time.Ticker
 
-	notifier *discord.DiscordNotifier
-	governor *governor.ChainGovernor
+	notifier  *discord.DiscordNotifier
+	governor  *governor.ChainGovernor
+	reobserve *reobserver.Reobserver
 }
 
 func NewProcessor(
@@ -133,6 +135,7 @@ func NewProcessor(
 	attestationEvents *reporter.AttestationEventReporter,
 	notifier *discord.DiscordNotifier,
 	g *governor.ChainGovernor,
+	ro *reobserver.Reobserver,
 ) *Processor {
 
 	return &Processor{
@@ -153,10 +156,11 @@ func NewProcessor(
 
 		notifier: notifier,
 
-		logger:   supervisor.Logger(ctx),
-		state:    &aggregationState{observationMap{}},
-		ourAddr:  crypto.PubkeyToAddress(gk.PublicKey),
-		governor: g,
+		logger:    supervisor.Logger(ctx),
+		state:     &aggregationState{observationMap{}},
+		ourAddr:   crypto.PubkeyToAddress(gk.PublicKey),
+		governor:  g,
+		reobserve: ro,
 	}
 }
 
@@ -164,7 +168,7 @@ func (p *Processor) Run(ctx context.Context) error {
 	p.cleanup = time.NewTicker(30 * time.Second)
 
 	// Always initialize the timer so don't have a nil pointer in the case below. It won't get rearmed after that.
-	govTimer := time.NewTimer(time.Minute)
+	t := time.NewTimer(time.Minute)
 
 	for {
 		select {
@@ -190,7 +194,7 @@ func (p *Processor) Run(ctx context.Context) error {
 			p.handleInboundSignedVAAWithQuorum(ctx, m)
 		case <-p.cleanup.C:
 			p.handleCleanup(ctx)
-		case <-govTimer.C:
+		case <-t.C:
 			if p.governor != nil {
 				toBePublished, err := p.governor.CheckPending()
 				if err != nil {
@@ -201,7 +205,14 @@ func (p *Processor) Run(ctx context.Context) error {
 						p.handleMessage(ctx, k)
 					}
 				}
-				govTimer = time.NewTimer(time.Minute)
+			}
+
+			if p.reobserve != nil {
+				p.reobserve.CheckForReobservations()
+			}
+
+			if p.governor != nil || p.reobserve != nil {
+				t = time.NewTimer(time.Minute)
 			}
 		}
 	}
--- a/node/pkg/reobserver/reobserver_monitoring.go
+++ b/node/pkg/reobserver/reobserver_monitoring.go
@@ -1,0 +1,104 @@
+// This file contains the code to monitor the reobservation monitor. This includes the following functions:
+// - Admin commands
+// - Prometheus metrics
+
+// The reobservation monitor supports the following admin client commands:
+//   - reobserver-status - displays the status of the reobservation monitor.
+//   - reobserver-drop-vaa [VAA_ID] - removes the specified VAA from the reobservation list.
+//
+// The VAA_ID is of the form "2/0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16/3", which is "emitter chain / emitter address / sequence number".
+
+// The reobservation monitor also supports the following Prometheus metrics:
+//
+// guardian_reobserver_pending_vaas 1
+// - This metric lists the number of VAAs the reobservation monitor is attempting to reobserver.
+//
+// guardian_reobserver_successful_reobservations 1
+// - This metric lists the number of VAAs the reobservation monitor successfully reobserved.
+//
+// guardian_reobserver_failed_reobservation_attempts 1
+// - This metric lists the number of VAAs the reobservation monitor failed to reobserver.
+
+package reobserver
+
+import (
+	"fmt"
+
+	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// Admin command to display status.
+func (reob *Reobserver) Status() string {
+	reob.mutex.Lock()
+	defer reob.mutex.Unlock()
+
+	foundSomething := false
+	var resp string
+	for _, oe := range reob.observations {
+		if !oe.quorumReached {
+			foundSomething = true
+			s1 := fmt.Sprintf("quorum not reached on message: %v, numRetries: %v, timeStamp: %v", oe.msgId, oe.numRetries, oe.timeStamp.String())
+			s2 := fmt.Sprintf("reobserver: %v", s1)
+			resp += s1 + "\n"
+			reob.logger.Info(s2)
+		}
+	}
+
+	if !foundSomething {
+		return "There are no reobservations in progress."
+	}
+
+	return resp
+}
+
+// Admin command to remove a VAA from the list.
+func (reob *Reobserver) DropVAA(msgId string) (string, error) {
+	reob.mutex.Lock()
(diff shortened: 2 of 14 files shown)


node/pkg/reobserver/reobserver_monitoring.go
```diff
@@ -0,0 +1,104 @@
...
+// The reobservation monitor supports the following admin client commands:
+//   - reobserver-status - displays the status of the reobservation monitor.
+//   - reobserver-drop-vaa [VAA_ID] - removes the specified VAA from the reobservation list.
+//
+// The VAA_ID is of the form "2/0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16/3", which is "emitter chain / emitter address / sequence number".
+
+// The reobservation monitor also supports the following Prometheus metrics:
+//
+// guardian_reobserver_pending_vaas 1
+// - This metric lists the number of VAAs the reobservation monitor is attempting to reobserver.
```
SEJeff, line 14:
```suggestion
// - This metric lists the number of VAAs the reobservation monitor is attempting to reobserve.
```

node/pkg/reobserver/reobserver_monitoring.go
```diff
@@ -0,0 +1,104 @@
...
+			Name: "guardian_reobserver_successful_reobservations",
+			Help: "Total number of successful reobservation attempts",
+		})
+
+	metricFailedReobservationAttempts = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Name: "guardian_reobserver_failed_reobservation_attempts",
+			Help: "Total number of failed reobservation attempts",
+		})
+)
```
SEJeff, line 90: Would you mind adding a tag name for the `chain_name`? That way we can have breakdowns by chain of missed vaas and successfully reobserved vaas.

This is generally indicative of a problem with the chain and is nice to be able to visualize in grafana.

proto/node/v1/node.proto
```diff
@@ -189,3 +195,17 @@ message ChainGovernorReleasePendingVAARequest {
...
+  string response = 1;
+}
+
+message ReobserverDropVAARequest {
+  string vaa_id = 1;
+}
+
+message ReobserverDropVAAResponse {
+  string response = 1;
+}
```
SEJeff, line 211: Trailing `\n` pretty please.

node/pkg/processor/processor.go
```diff
@@ -201,7 +205,14 @@ func (p *Processor) Run(ctx context.Context) error {
 						p.handleMessage(ctx, k)
 					}
 				}
-				govTimer = time.NewTimer(time.Minute)
+			}
+
+			if p.reobserve != nil {
+				p.reobserve.CheckForReobservations()
```
SEJeff, line 211:
```suggestion
				err := p.reobserve.CheckForReobservations()
				if err != nil {
					return err
				}
```

node/pkg/reobserver/reobserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+	reob.QuorumReached(msgId)
+
+	// Make sure it was received and has been marked completed.
+	assert.Equal(t, 1, len(reob.observations))
+	oe, exists := reob.observations[msgId]
+	assert.Equal(t, true, exists)
+	assert.Equal(t, true, oe.localMsgReceived())
+	assert.Equal(t, true, oe.quorumReached)
+	assert.Equal(t, true, oe.completed)
+	assert.Equal(t, 0, oe.numRetries)
```
SEJeff, line 36: For this sort of thing, would you mind just constructing an expected observationEntry struct and doing a single assertion against it?

Maybe something vaguely like this?

```
expected := observationEntry{
	msgId:		msgId
	chainId:	vaa.ChainIDSolana
	txHash:		common.Hash{}
	numRetries:	0
	quorumReached:	true
	completed:	true
}
assert.Equal(t, expected, oe)
```

I strongly preference this approach so if for some reason extra fields are added in the future, the test doesn't miss them. Worst case, just zero out the timestamp.

SEJeff (commented):
Please add a bit of jitter (random time period) for when the observation requests are sent. Also add some exponential backoff to the retries.

If multiple guardians decide to enable this feature, it makes flooding the network at the same time less likely.

jynnantonix (changes requested):
I had some small nits for the code changes but the bigger issue is that it's not clear to me what new behavior this package is adding.  AFAICT, it's doing the same thing that the cleanup code in the processor package does.

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/certusone/wormhole/1435#-:-N9doMlp6XWkW0qwU2Ra:b-5o2r4c)* status: all files reviewed, 10 unresolved discussions (waiting on @bruce-riley and @kev1n-peters)

___
*[`node/cmd/guardiand/adminclient.go` line 144 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9cxhA52l21qQof652O:-N9cxhA52l21qQof652P:b-gu6rrf) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/cmd/guardiand/adminclient.go#L144)):*
> ```Go
> 	Use:   "reobserver-drop-vaa [VAA_ID]",
> 	Short: "Removes the specified VAA (chain/emitter/seq) from the reobserver list",
> 	Run:   runReobserverDropVAA,
> ```

`cobra.Command` also has a `RunE` field if we want to return an error from the run function rather than `log.Fatalf`.
___
*[`node/cmd/guardiand/node.go` line 1114 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9czCKvBOU6HWO-zTQl:-N9czCKvBOU6HWO-zTQm:bkqwn11) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/cmd/guardiand/node.go#L1114)):*
> ```Go
> 			if err != nil {
> 				log.Fatal("failed to create reobservation monitor", zap.Error(err))
> 			}
> ```

style nit: local variables can be initialized in an if statement (https://go.dev/doc/effective_go#if)



_Suggestion:_
```Go
			if err := reobserve.Run(ctx); err != nil {
				log.Fatal("failed to create reobservation monitor", zap.Error(err))
			}
```
___
*[`node/cmd/spy/spy.go` line 302 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9d-wS56c1Ll1fsh4AL:-N9d-wS56c1Ll1fsh4AM:ba03taa) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/cmd/spy/spy.go#L302)):*
> ```Go
> 	// Run supervisor.
> 	supervisor.New(rootCtx, logger, func(ctx context.Context) error {
> 		if err := supervisor.Run(ctx, "p2p", p2p.Run(obsvC, nil, nil, sendC, signedInC, priv, nil, gst, *p2pPort, *p2pNetworkID, *p2pBootstrap, "", false, rootCtxCancel, nil, nil)); err != nil {
> ```

This isn't something we need to do in this PR but this parameter list is long enough that we might want to consider switching to the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).  It's basically impossible to figure out just from this call site what the various nil, false, and empty string parameters are supposed to represent.
___
*[`node/pkg/common/obsvReqSendC.go` line 10 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9czcpG8VPXPkbUkiUF:-N9czcpG8VPXPkbUkiUG:b-jxioyl) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/pkg/common/obsvReqSendC.go#L10)):*
> ```Go
> 
> const ObsvReqChannelSize = 50
> const ObsvReqChannelFullError = "channel is full"
> ```

nit: We can use `errors.New` here so we don't need to call `fmt.Errorf` below:

```go
var ObsvReqChannelFullError = errors.New("reobservation request channel is full")
```
___
*[`node/pkg/processor/processor.go` line 171 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9di0l55k17mDTxMhp9:-N9di0l55k17mDTxMhpA:b-rxxxt7) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/pkg/processor/processor.go#L171)):*
> ```Go
> 
> 	// Always initialize the timer so don't have a nil pointer in the case below. It won't get rearmed after that.
> 	t := time.NewTimer(time.Minute)
> ```

Seems like we should be using `time.NewTicker(time.Minute)` here so that we don't have to keep re-creating the timer every time it expires.
___
*[`node/pkg/reobserver/reobserver.go` line 5 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9dekP5BFvdXMi5C9g9:-N9dekP5BFvdXMi5C9gA:b6fp5fc) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/pkg/reobserver/reobserver.go#L5)):*
> ```Go
> // that reach quorum. If an observation does not reach quorum in a timely manner (as determined by retryIntervalInMinutes), a reobservation
> // request is submitted for it. Once expirationIntervalInMinutes minutes have elapsed since an observation reaches quorum, or maxRetries
> // reobservation attemps have failed, it is deleted from the cache.
> ```

This seems to have a lot of overlap with what the processor does in cleanup.go: https://github.com/certusone/wormhole/blob/7a1b1344a19db22935d8532135ac249da128b4e7/node/pkg/processor/cleanup.go#L172-L207

I can't actually find a difference in the behavior.  Both places try to re-broadcast an observation over the p2p network until it reaches quorum or some deadline expires.  Both places ignore remote observations that don't have a local message.  The only difference appears to be in the parameter configuration (5 minutes / 120 hours for the processor and 5 minutes / 1 hour for the reobserver).  Am I missing something?

Personally, I like having this behavior in a separate package because it's more easily tested but if we keep this package we should also delete the relevant code from the processor package so that we don't end up with 2 code paths that do the same thing.
___
*[`node/pkg/reobserver/reobserver.go` line 203 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9dmAGg50eTbjlTOQQI:-N9dmAGg50eTbjlTOQQJ:b-df0apx) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/pkg/reobserver/reobserver.go#L203)):*
It's safe to delete entries from a map while iterating over the map so we don't need to keep a separate list.  From the [language spec](https://go.dev/ref/spec#For_statements):

> If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced.
___
*[`node/pkg/reobserver/reobserver_test.go` line 40 at r1](https://reviewable.io/reviews/certusone/wormhole/1435#-N9dnWec5UvkUuXMrJI5:-N9dnWec5UvkUuXMrJI6:b-1ezgkn) ([raw file](https://github.com/certusone/wormhole/blob/986e074d0013c688c6ff8e3f001540de86a0af19/node/pkg/reobserver/reobserver_test.go#L40)):*
> ```Go
> 	// Make sure it gets expired.
> 	err := reob.checkForReobservationsForTime(now.Add(time.Minute * time.Duration(expirationIntervalInMinutes+1)))
> 	assert.Nil(t, err)
> ```

nit: `assert.NoError(t, err)`.  This gives a more specialized message for errors.

jynnantonix:
So, looking at this more closely I think the fact that messages are being dropped is a bug in the processor code.  It tries to reach quorum on an observation by repeatedly re-broadcasting the local message but that doesn't help when not enough guardians observed the tx the first time it happened.

I think I've fixed this in #1438.  Since I think this is just a bug fix, it's not an optional feature and doesn't have separate metrics.  Please take a look.

bruce-riley:
It looks like this feature is not really needed, and can more easily be handle by the bug fix in PR 1438.

I am putting this PR on hold and converting it to a draft until we decide if it is needed at all.

Please do not spend any more time looking at this PR.

PR #1438, opened 2022-08-17
Re-observation

jynnantonix:
This PR fixes a bug in the processor package where the guardian network would never reach quorum on an observation if not enough guardians observed the tx when it first happened.  Instead the guardians that did observe the tx would keep re-broadcasting their signatures over the network every 5 minutes for 120 hours or until someone manually submitted a re-observation request.

The fix is to send the re-observation request whenever we re-broadcast a local observation.  This allows guardians that missed the tx the first time to re-observe it in their local RPC node.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1438)

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1102,6 +1102,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			setC,
 			sendC,
 			obsvC,
+			obsvReqSendC,
 			injectC,
 			signedInC,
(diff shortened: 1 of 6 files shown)


jynnantonix:
I haven't actually testing this beyond running unit tests but I wanted to send this out early to get feedback.

bruce-riley (approved):

SEJeff (approved):

PR #1441, opened 2022-08-18
Add a wormhole go sdk

jynnantonix:
The wormhole sdk is a new go module in the sdk/ directory.  This
initially contains the *_consts.go files from the common package in the
top-level sdk package and the entire vaa package as a sub-package.

For go reasons this needs to be in the sdk directory itself (rather than
a sdk/go subdir).  To prevent the go tooling from looking into the other
non-go subdirs, add an empty go.mod file in each one.  See
https://github.com/golang/go/issues/42965 for more details on why we
can't have nice things.

- - -
This change is [Reviewable](https://reviewable.io/reviews/wormhole-foundation/wormhole/1441)


jynnantonix:
This PR just moves stuff around.  Once it lands, I will add some more changes to make signature verification easier to use but I can't do it now because we're not allowed to use tools that make it easy to do dependent changes.  When those are in we can tag a release for the sdk module.

kcsongor (approved):
The directory structure is unfortunate, but it's just go being go. Could you add a short README.md to the sdk folder explaining the directory structure? Otherwise looks good.

jynnantonix:
Added a readme.  Also renamed "certusone" -> "wormhole-foundation".  PTAL.

bruce-riley (approved):

jynnantonix:
Had  to rebase to fix merge conflicts.  Please take another look.

hendrikhofstadt (approved):

kcsongor (dismissed):

kcsongor (approved):

hendrikhofstadt (approved):

kcsongor (approved):

hendrikhofstadt (approved):
Assuming rebase

PR #1448, opened 2022-08-20
Aptos/integration

optke3:
**Aptos Wormhole Core Bridge Testing Framework + Implementation + Updates**

- we use Aptos CLI to spin up local network with single validator, compile + test + deploy modules 
-  we can deploy Core Bridge module using CLI command `aptos move publish `
-  we can initialize Wormhole Core Bridge from client-side through `init_wormhole.ts` script
-  we call publishWormholeMessage through `publish_wormhole_message.ts` and read corresponding event in `read.ts` 
- we can call various on-chain functions using a TypeScript client and check the transaction status / changes made

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1448)


SEJeff (changes requested):
I didn’t look at this change past the fact you seem to have added the testnet data directory, logs, and all to this pr? That is not the kind of thing that should be in this repo at all.

Also if possible please make smaller incremental changes vs these big PRs. It is harder to review accurately and more difficult to digest.

optke3:
removed the testnet data directory, as well as NFTBridge, TokenBridge (will add these back in later after they are complete and in incremental PRs as per your suggestion).

It's only the Core Bridge now, which should be self-contained.

Tiltfile
```diff
@@ -693,7 +693,7 @@ if aptos:
         dockerfile = "aptos/Dockerfile",
         only = ["Dockerfile", "node_builder.sh", "start_node.sh", "README.md", "cert.pem"],
     )
-
+
```
SEJeff, line 696: You're adding trailing whitespace. Please fix your editor. I like to set it to strip trailing whitespace on buffer save.

Tiltfile
```diff
@@ -58,12 +58,12 @@ namespace = cfg.get("namespace", "wormhole")
...
+algorand = cfg.get("algorand", False)
 aptos = cfg.get("aptos", True)
-evm2 = cfg.get("evm2", True)
-solana = cfg.get("solana", True)
-terra_classic = cfg.get("terra_classic", True)
-terra2 = cfg.get("terra2", True)
+evm2 = cfg.get("evm2", False)
+solana = cfg.get("solana", False)
+terra_classic = cfg.get("terra_classic", False)
+terra2 = cfg.get("terra2", False)
```
SEJeff, line 66: Fair to say you didn't mean to do any of this in this PR?

aptos/Dockerfile.base
```diff
@@ -9,3 +9,6 @@ WORKDIR /tmp
 RUN ./node_builder.sh
 
 COPY start_node.sh  /tmp
+
+
+
```
SEJeff, line 14: Don't add bogus whitespace like this.

aptos/scripts/README.txt
```diff
@@ -0,0 +1,5 @@
+1. bring up local net using ```aptos node run-local-testnet --with-faucet --force-restart```
+2. initialize account using ```aptos account fund-with-faucet --account 277fa055b6a73c42c0662d5236c65c864ccbf2d4abd21f174a30c8b786eab84b```
+3. publish modules using ```aptos move publish --private-key 0x537c1f91e56891445b491068f519b705f8c0f1a1e66111816dd5d4aa85b8113d --profile default```
+4. run ```init_wormhole.ts```
+5. run ```publish_wormhole_message.ts```
```
SEJeff, line 5: Please name this file `README.md` to match the rest of the repo.

aptos/scripts/deploy.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// export async function main(){
+//     const alice = new AptosAccount();
+//     publishModule(alice, )
+// }
+
+// if (require.main === module) {
+//     main().then((resp) => console.log(resp));
+//   }
```
SEJeff, line 43: Please don't add commented code to version control. Add it when it is ready.

SEJeff (changes requested):
So a couple of major questions here.

1. Do we actually need to create a submodule of aptos in the wormhole repo? We don't do this for any other blockchain and I'm not sure that we want to. Can we not setup an aptos node from a release with a Dockerfile or something for tilt?

2. You add a lot of commented code. Unless absolutely necessary, don't do this. Learn to use git patch mode (`git add -p`) to only add / commit the stuff you're ready to commit.

3. Can you put the config for aptos in the top level aptos directory and just specify the config path instead of adding a `.aptos` directory?

I'm not the right guy to review the move contracts.

Tiltfile
optke3: I did this to make the tilt build time faster, but didn't end up using tilt at all to write/test the Aptos contracts. Will revert these.

PR #1455, opened 2022-08-23
node: configure p2p routing table with bootstrapping nodes

kev1n-peters:
Configure the bootstrapping nodes that we will connect to to seed and refresh
the Routing Table if it becomes empty.

Fixes an issue where the spy stops receiving gossiped messages after the node
is restarted, because the Routing Table becomes empty and is never refreshed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1455)

### DIFF
--- a/node/pkg/p2p/p2p.go
+++ b/node/pkg/p2p/p2p.go
@@ -92,17 +92,40 @@ func Run(obsvC chan *gossipv1.SignedObservation, obsvReqC chan *gossipv1.Observa
 			// Let's prevent our peer from having too many
 			// connections by attaching a connection manager.
 			libp2p.ConnectionManager(connmgr.NewConnManager(
-				100,         // Lowwater
+				100,         // LowWater
 				400,         // HighWater,
 				time.Minute, // GracePeriod
 			)),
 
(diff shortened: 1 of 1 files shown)


node/pkg/p2p/p2p.go
```diff
@@ -138,49 +157,6 @@ func Run(obsvC chan *gossipv1.SignedObservation, obsvReqC chan *gossipv1.Observa
...
-				logger.Info("We're a bootstrap node")
-				bootstrapNode = true
-				continue
-			}
-
-			if err = h.Connect(ctx, *pi); err != nil {
-				logger.Error("Failed to connect to bootstrap peer", zap.String("peer", addr), zap.Error(err))
-			} else {
-				successes += 1
-			}
```
SEJeff, line 174: Why was this block removed? Is it not necessary? It doesn't appear to be in the code block above.

Does this mean the bootstrap nodes will try to connect to themselves? What happens if there is a failure to connect to a bootstrap peer?

SEJeff (approved):

bruce-riley (approved):

PR #1456, opened 2022-08-23
bridge_ui: Recovery should await fetchSignedVAA fix

kev1n-peters:
Otherwise if there's an exception, the `handleError` case will never be executed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1456)


kcsongor (approved):
implicit effects is why we can't have nice things...

panoel (approved):

PR #1462, opened 2022-08-24
node: Deduplicate re-observation requests

jynnantonix:
Commit "1753bb3: Send re-observation request when re-broadcasting local
observations" changed the processor code to automatically send
re-observation requests.  If multiple guardians do this around the same
time that would lead to multiple re-observation requests for the same
transaction.

Deduplicate re-observation requests by keeping a cache of the requests
that were received in the last 11 minutes.  Any request that's already
in the cache will not be forwarded to the chain-specific observation
code.  The 11 minute value was chosen because the guardians will send
these requests approximately every 5 minutes.

Also move the re-observation code into a standalone function so that it
can be more easily tested.

- - -
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1462)

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -3,7 +3,6 @@ package guardiand
 import (
 	"context"
 	"encoding/base64"
-	"encoding/hex"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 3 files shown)


bruce-riley (approved):

SEJeff (approved):
Nice!

PR #1465, opened 2022-08-24
near/audit: audit feedback

jumpsiegel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1465)


bruce-riley (dismissed):

bruce-riley (approved):

near/contracts/ft/src/lib.rs
```diff
@@ -14,6 +14,7 @@ use {
         },
         collections::LazyOption,
         env,
+        utils::assert_one_yocto,
```
panoel, line 17: I'm not familiar with what a yocto is?

near/contracts/ft/src/lib.rs
```diff
@@ -190,6 +194,10 @@ impl FTContract {
 
         let mut deposit: Balance = env::attached_deposit();
 
+        if deposit == 0 {
```
panoel, line 197: Can deposit be negative and should you check for it?

panoel (approved):

near/contracts/ft/src/lib.rs
jumpsiegel: one yocto is the smallest possible amount you can transfer

PR #1471, opened 2022-08-26
near/audit2: more audit feedback plus additional upgrade tests

jumpsiegel:
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1471)


near/contracts/token-bridge/src/lib.rs
```diff
@@ -1652,32 +1654,30 @@ impl TokenBridge {
...
-            emitter_registration: old_state.emitter_registration,
-            last_asset:           old_state.last_asset,
-            upgrade_hash:         old_state.upgrade_hash,
-            tokens:               old_state.tokens,
-            key_map:              old_state.key_map,
-            hash_map:             old_state.hash_map,
-            bank:                 old_state.bank,
-        }
-    }
+//    #[init(ignore_state)]
```
kcsongor, line 1657: do we not need this stuff anymore?

panoel, line 1657: Should this just go away?

near/contracts/wormhole/src/lib.rs
```diff
@@ -601,30 +601,28 @@ impl Wormhole {
...
-            guardian_set_expirity: old_state.guardian_set_expirity,
-            guardian_set_index:    old_state.guardian_set_index,
-            owner_pk:              old_state.owner_pk,
-            upgrade_hash:          old_state.upgrade_hash,
-            message_fee:           old_state.message_fee,
-            bank:                  old_state.bank,
-        }
-    }
+//    #[init(ignore_state)]
+//    #[payable]
```
panoel, line 605: Should this just go away?

near/contracts/token-bridge/src/lib.rs
jumpsiegel: yes,   in the future we might want to do a migration so I wanted to keep it around BUT I didn't want it in the contract in case there was a unknown security issue with this...  pure paranoia which is why I commented it out

jumpsiegel: yes, in the future we might want to do a migration so I wanted to keep it around BUT I didn't want it in the contract in case there was a unknown security issue with this... pure paranoia which is why I commented it out

near/contracts/wormhole/src/lib.rs
jumpsiegel: yes, in the future we might want to do a migration so I wanted to keep it around BUT I didn't want it in the contract in case there was a unknown security issue with this... pure paranoia which is why I commented it out

near/contracts/token-bridge/src/lib.rs
jumpsiegel: if people feel strongly, we can just assume I would pull it back out of git history...

bruce-riley (approved):

PR #1478, opened 2022-08-26
aptos: fix npm dependency

kcsongor:
We shouldn't be using local dependencies


kcsongor:
Superseded by https://github.com/wormhole-foundation/wormhole/pull/1480

PR #1483, opened 2022-08-29
aptos: upgradeability

kcsongor:
Requires https://github.com/aptos-labs/aptos-core/pull/3568 as we need the package metadata for on-chain publishing. Then compilation will need to be run with

```shell
aptos move compile --save-metadata
```

### DIFF
--- a/aptos/Makefile
+++ b/aptos/Makefile
@@ -1,0 +1,3 @@
+build:
+	make -C deployer build
+	make -C contracts build
--- a/aptos/contracts/Makefile
+++ b/aptos/contracts/Makefile
@@ -1,0 +1,13 @@
+-include ../../Makefile.help
+
+.PHONY: artifacts
+artifacts: build
+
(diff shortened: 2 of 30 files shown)


kcsongor:
Update: there is now https://github.com/aptos-labs/aptos-core/pull/3622/ which adds better upstream support for publishing modules under a resource account. I think for now we can just keep what's here, because our code only makes use of stuff that's already in devnet and considered more stable, and it's also valuable to keep the documentation here for future reference.

There's also a small implementation difference, namely that in our version the `SignerCapability` is transferred directly into the resource account (though the account still has to explicitly claim it in a tx signed by the deployer) while the upstream version puts them into a map object under the deployer account's memory space. It's not a huge difference, especially in practice, but in theory it feels unnecessary to take write locks on a single object when publishing unrelated packages.

jumpsiegel (approved):

PR #1485, opened 2022-08-30
node: governor: fix broken tests

jynnantonix:
The token list has 126 entries, not 123.  Also include the chain id when
checking for duplicates in the token list.


kcsongor (approved):

jynnantonix:
Looks like the checks were stale because a couple of these tests are broken on dev.v2.  See #1485 for the fix.

kcsongor (approved):

claudijd:
@bruce-riley already has a fix for this in his auto-merge branch, should land shortly, I think this can be closed and if we want to add the other test improvement, let's base it on dev.v2 once it lands.

PR #1489, opened 2022-08-30
ethereum: EVM fork protection

kev1n-peters:
Added a `uint256 evmChainId` state variable to the Core, Token and NFT Bridge
contracts. `evmChainId` is compared to `block.chainid` in the `chainId`
function. When they don't match, a "bad fork" Chain ID is returned (max uint16),
otherwise the Wormhole Chain ID is returned.


ethereum/contracts/nft/NFTBridge.sol
```diff
@@ -21,6 +21,8 @@ contract NFTBridge is NFTBridgeGovernance {
 
     // Initiate a Transfer
     function transferNFT(address token, uint256 tokenID, uint16 recipientChain, bytes32 recipient, uint32 nonce) public payable returns (uint64 sequence) {
+        require(evmChainId() == block.chainid, "invalid evmChainId");
```
calebrate, line 24: Should we create a function modifier for this check?

ethereum/contracts/Implementation.sol
```diff
@@ -32,6 +32,31 @@ contract Implementation is Governance {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
+        uint256 evmChainId;
+        uint16 chain = _state.provider.chainId;
+
+        // Wormhole chain ids explicitly enumerated
```
calebrate, line 38: Should we actually put this into the core bridge or just application layer?

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -217,6 +221,8 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
      *  @notice Initiate a transfer
      */
     function _transferTokens(address token, uint256 amount, uint256 arbiterFee) internal returns (BridgeStructs.TransferResult memory transferResult) {
+        require(evmChainId() == block.chainid, "invalid evmChainId");
```
calebrate, line 224: Is it necessary to have the check (or modifier) on outgoing transfers as well or could incoming be enough? (as forks will not be watched by guardians)

ethereum/contracts/Implementation.sol
calebrate: In case we want to put it into the core bridge, we could also think about a getter function isForked() on the core bridge that the application layer could refer to
cc @hendrikhofstadt

ethereum/contracts/nft/NFTBridge.sol
kev1n-peters: I think this is a good idea to reduce duplicated code

ethereum/contracts/Implementation.sol
kev1n-peters: The core bridge could refuse attestations on a forked chain, but I think it should allow them for governance VAAs in case we want to upgrade to support the forked chain.

ethereum/contracts/bridge/Bridge.sol
kev1n-peters: I don't think it's necessary for outgoing transfers, because, as you mentioned, the fork won't even be watched.

hendrikhofstadt: I think this is actualy a nice "protection" against users doing transfers that will go to nowhere and then complaining. Only concern is the SLOAD gas cost that is added to all outbound transfers (2100 gas).

ethereum/contracts/Getters.sol
```diff
@@ -27,9 +27,16 @@ contract Getters is State {
     }
 
     function chainId() public view returns (uint16) {
+        if (evmChainId() != block.chainid) {
+            return type(uint16).max;
```
hendrikhofstadt, line 31: This makes all upgrades compatible between all forks. Do we want to introduce some shifting here e.g. `uint16.max -32 + block.chainid % 32` to reduce the odds of collisions?

ethereum/contracts/State.sol
```diff
@@ -41,6 +41,9 @@ contract Storage {
         mapping(address => bool) initializedImplementations;
 
         uint256 messageFee;
+
+        // EIP-155 Chain ID
+        uint256 evmChainId;
```
calebrate, line 46: This could go into the provider struct

ethereum/contracts/Getters.sol
calebrate: Latest approach we discussed with @evan-gray was we should differenciate between the chains using `block.chainid` when applying an ugrade.
I have no strong opinion here. This could make an upgrade on a fork easier but i think we should still have the `block.chainid` check.

ethereum/contracts/State.sol
kev1n-peters: Wouldn't that shift the storage variables declared after provider down a slot and cause issues or am I missing something?

ethereum/contracts/bridge/BridgeGetters.sol
```diff
@@ -27,9 +27,17 @@ contract BridgeGetters is BridgeState {
     }
 
     function chainId() public view returns (uint16){
+        if (evmChainId() != block.chainid) {
+            // reduce the likelihood of forked chain ID collisions
+            return type(uint16).max - 32 + uint16(block.chainid % 32);
```
kcsongor, line 32: same as above

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -20,10 +20,15 @@ import "./token/TokenImplementation.sol";
 contract Bridge is BridgeGovernance, ReentrancyGuard {
     using BytesLib for bytes;
 
+    modifier onlyEvmChainId() {
```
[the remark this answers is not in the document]
kev1n-peters: Perhaps `noFork`, `noForkedChain`, `onlyUnforkedChain`, `onlyRootChain`, `disableForkedChain` would be better? 🤷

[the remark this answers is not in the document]
kcsongor: I couldn't come up with a single alternative and here you are giving me 5 🥇 `noFork` and `onlyRootChain` are my favourites

ethereum/contracts/Getters.sol
evan-gray: Chatted with @calebrate and @kcsongor and we fall in favor of the static return type(uint16).max; (e.g. all chains ending with 01 end up the same 🙁)

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -96,11 +118,11 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
 
         chain.module = encoded.toBytes32(index);
         index += 32;
-        require(chain.module == module, "invalid RegisterChain: wrong module");
+        require(chain.module == module, "wrong module");
```
kev1n-peters, line 121: Had to cut down on some of these strings so circumvent the 24KB contract size limit. Would rather do this than lower the optimizer runs (default: 200).

ethereum/contracts/Governance.sol
```diff
@@ -37,7 +37,8 @@ abstract contract Governance is GovernanceStructs, Messages, Setters, ERC1967Upg
         require(upgrade.module == module, "Invalid Module");
 
         // Verify the VAA is for this chain
-        require(upgrade.chain == chainId(), "Invalid Chain");
+        uint16 chainId = chainId();
+        require(upgrade.chain == chainId && chainId != type(uint16).max, "Invalid Chain");
```
kcsongor, line 41: if we already revert here... why don't we just revert in `chainId()` directly? and instead of checking for `chainId() == type(uint16).max` in `submitRecoverChainId` we require `evmChainId() != block.chainid` there.

ethereum/contracts/GovernanceStructs.sol
```diff
@@ -52,6 +52,14 @@ contract GovernanceStructs {
         bytes32 recipient;
     }
 
+    struct RecoverChainId {
+        bytes32 module;
+        uint8 action;
```
kcsongor, line 57: maybe this ship has sailed, but I don't think `module` and `action` should be in these structs at all, it's something the parser could check and then throw away

ethereum/contracts/Governance.sol
kev1n-peters: I would prefer to not revert in the `chainId()` getter and instead let the caller revert. For example, we don't want to revert here https://github.com/wormhole-foundation/wormhole/blob/8f094b4d8b9f97c9b9b8490d8e9b222b1b32256b/ethereum/contracts/Governance.sol#L91 (in case we have to update the guardian set on a forked chain) and I'd rather not read the chainId variable directly there (instead of through the getter). What do you think?

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -484,7 +489,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     }
 
     // Execute a Transfer message
-    function _completeTransfer(bytes memory encodedVm, bool unwrapWETH) internal returns (bytes memory) {
+    function _completeTransfer(bytes memory encodedVm, bool unwrapWETH) internal onlyEvmChainId returns (bytes memory) {
```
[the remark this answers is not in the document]
kev1n-peters: The `noFork` modifier is now on the `verifyBridgeVM` function, affecting `completeTransfer` and `createWrapped`.

ethereum/contracts/Governance.sol
kev1n-peters: Thought more about it, and `chainId()` doesn't even need to return `type(uint16).max` for "bad forks". We can simplify by checking if `evmChainId() != block.chainid` in the appropriate places. This simplifies the code and makes it easier to understand IMO.

ethereum/contracts/Governance.sol
```diff
@@ -25,6 +25,8 @@ abstract contract Governance is GovernanceStructs, Messages, Setters, ERC1967Upg
      * @dev Upgrades a contract via Governance VAA/VM
      */
     function submitContractUpgrade(bytes memory _vm) public {
+        require(!isFork(), "bad fork");
```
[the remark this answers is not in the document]
kev1n-peters: Changed

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -137,6 +160,27 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+    function parseRecoverChainId(bytes memory encodedRecoverChainId) public pure returns (BridgeStructs.RecoverChainId memory rci) {
+        uint index = 0;
+
+        rci.module = encodedRecoverChainId.toBytes32(index);
+        index += 32;
+        require(rci.module == module, "wrong module");
+
+        rci.action = encodedRecoverChainId.toUint8(index);
+        index += 1;
+        require(rci.action == 5, "wrong action");
```
calebrate, line 176: Should this be 3 for the bridge?

ethereum/contracts/bridge/BridgeImplementation.sol
```diff
@@ -17,27 +17,31 @@ contract BridgeImplementation is Bridge {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
-        address tokenContract;
-        uint16 chain = chainId();
+        uint256 evmChainId;
+        uint16 chain = _state.provider.chainId;
```
calebrate, line 21: Why not use the chainId() getter?

ethereum/contracts/Implementation.sol
```diff
@@ -32,6 +32,31 @@ contract Implementation is Governance {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
+        uint256 evmChainId;
+        uint16 chain = _state.provider.chainId;
```
calebrate, line 36: Why not use the chainId() getter?

ethereum/contracts/bridge/BridgeGovernance.sol
kev1n-peters: Yes

ethereum/contracts/bridge/BridgeImplementation.sol
kev1n-peters: Will change

ethereum/contracts/Implementation.sol
kev1n-peters: Will change

ethereum/contracts/Implementation.sol
```diff
@@ -32,6 +32,31 @@ contract Implementation is Governance {
...
+        } else if (chain == 16) { evmChainId = 1284;       // moonbeam
+        } else if (chain == 17) { evmChainId = 245022934;  // neon
+        } else if (chain == 23) { evmChainId = 42161;      // arbitrum
+        } else if (chain == 24) { evmChainId = 10;         // optimism
+        } else if (chain == 25) { evmChainId = 100;        // gnosis
+        } else {
+            revert("Unknown chain id.");
+        }
+
+        setEvmChainId(evmChainId);
```
calebrate, line 59: We should compare it to block.chainid before applying

kev1n-peters: We do this check in the actual setter

calebrate (approved):

kcsongor (approved):

PR #1490, opened 2022-08-30
event_database: Added XCN to allow list

ckeun:


ckeun:
already added in another PR.

PR #1496, opened 2022-08-31
ethereum: remove wormhole from token name

evan-gray:


ethereum/migrations/7_deploy_token_implementation_only.js
```diff
@@ -0,0 +1,13 @@
+// run with:
+// npm run deploy-token-implementation-only
+// e.g. Ethereum Mainnet
+// INFURA_KEY="" MNEMONIC="" npm run deploy-token-implementation-only -- --network mainnet
+// e.g. BSC
+// MNEMONIC="" npm run deploy-token-implementation-only -- --network binance
+// e.g. Polygon
+// MNEMONIC="" npm run deploy-token-implementation-only -- --network polygon
+const TokenImplementation = artifacts.require("TokenImplementation");
```
kcsongor, line 9: maybe consider adding this file as a regular truffle script, like: https://github.com/wormhole-foundation/wormhole/blob/3325e8ba02598b666a95947337a1714d6b3512f6/ethereum/scripts/deploy_token_bridge.js since the migration flow doesn't make that much sense for these one-off deploys

evan-gray:
The diff in `ethereum-upgrade` is expected 🙂 and the point of the migration ✅ 
https://github.com/wormhole-foundation/wormhole/runs/8122634122?check_suite_focus=true#step:5:274

ethereum/migrations/7_deploy_token_implementation_only.js
evan-gray: Will consider adding another regular truffle script in a future PR. I think (token) bridge has both which is 🤷 . Personally I like the simulation, but agree it's not super necessary (and is weightier).

kcsongor (approved):

bruce-riley (approved):

PR #1499, opened 2022-09-01
scripts: Ignore cosmos-sdk for docker checks

jynnantonix:


jynnantonix:
We're not vendoring the cosmos-sdk repo anymore so this isn't necessary.

PR #1502, opened 2022-09-01
near/upgradePhase1: fix some upgrade issues to make it easier to migrate in the future

jumpsiegel:


jumpsiegel:
Without this change,   if you change your state there is no way to get refunded the balance of your upgrade deposit because the upgrade_done fails because it cannot deserialize the state

panoel (approved):

kcsongor (approved):

PR #1505, opened 2022-09-02
Solitaire: Use sysvar for rent instead of default

ali-bahjati:
Using default has some risks such as:
- Network having a different values (such as Pythnet)
- Network changes the value

It avoids using default as fallback to have more explicit Errors. There are some risks that under rare conditions (e.g., a bug) the sysvar return Err and expose the above risks.


hendrikhofstadt (approved):

hendrikhofstadt:
Let's make sure this doesn't break any of the Wormhole tests.

solana/solitaire/program/src/lib.rs
```diff
@@ -72,10 +73,15 @@ pub enum CreationLamports {
...
-    pub fn amount(self, size: usize) -> u64 {
+    pub fn amount(self, size: usize) -> Result<u64>  {
         match self {
-            CreationLamports::Exempt => Rent::default().minimum_balance(size),
-            CreationLamports::Amount(v) => v,
+            // Rent::get() is used over Rent::default() here in case on-chain Rent has diverged from
+            // the Solana defaults. Unlike `default()`, the `get()` call only works on-chain and in
+            // Solana's test framework, it will fail in unit tests. In those environments one can use
+            // the following:
+            // `CreationLamports::Amount(Rent::default().minimum_balance(size))`
```
jynnantonix, line 82: Can we just add `#[cfg(test)]` and `#[cfg(not(test))]` branches here?  Otherwise it's going to be very difficult for someone to figure out why their tests are failing unless they know to look here.

Reisen: I think letting this bubble up to the user is better than having conditionally compiled behaviour. We have it in some places in the program already and It’s just as easy to trip someone up. While most developers are only targetting Solana itself, allowing users to test against `Rent::default` allows them to continue to write programs that could potentially break later.

If we want to make sure there’s discoverability perhaps we can add some sort of `WormholeError::RentUnavailable` error to WormholeError and bubble that up so that it isn’t surprising behaviour?

jynnantonix (approved):

Reisen (approved):

ripatel-jump:
FYI -- The sysvar can also be retrieved using the [sol_get_rent_sysvar](https://bpf.wtf/sol-0x04-syscalls/#solgetrentsysvar) syscall. This is more robust, as it doesn't rely on untrusted user input (accounts).

PR #1506, opened 2022-09-02
aptos: use keccak256

kcsongor:


aptos/contracts/sources/vaa.move
```diff
@@ -202,7 +202,7 @@ module wormhole::vaa {
         serialize::serialize_u64(&mut bytes, vaa.sequence);
         serialize::serialize_u8(&mut bytes, vaa.consistency_level);
         serialize::serialize_vector(&mut bytes, vaa.payload);
-        hash::sha3_256(bytes)
+        aptos_hash::keccak256(bytes)
```
jumpsiegel, line 205: Ah, this is where I saw it called once

jumpsiegel (approved):

aptos/contracts/sources/vaa.move
kcsongor: right... this function should probably be removed

PR #1508, opened 2022-09-02
Script to generate change registration governance

bruce-riley:
Change-Id: If99e78b031dcd4049a9ee40b09914eeb93e7518a

Also changing the guardian verify VAA command to print the digest in the same format as the execute VAA command.

Much of the magic in register-chain-governance.sh was taken directly from contract-upgrade-governance.sh, which has previously been reviewed.

### DIFF
--- a/clients/js/main.ts
+++ b/clients/js/main.ts
@@ -10,6 +10,9 @@ import {
   CONTRACTS,
   setDefaultWasm,
   hexToUint8Array,
+  getEmitterAddressSolana,
+  getEmitterAddressTerra,
+  getEmitterAddressEth,
 } from "@certusone/wormhole-sdk";
 import { execute_solana } from "./solana";
 import {
@@ -254,11 +257,18 @@ yargs(hideBin(process.argv))
           describe: "Module to query",
           type: "string",
           choices: ["Core", "NFTBridge", "TokenBridge"],
+        })
+        .option("emitter", {
+          alias: "e",
+          describe: "Print in emitter address format",
+          type: "boolean",
+          default: false,
+          required: false,
         });
     },
     async (argv) => {
       assertChain(argv["chain"]);
-      assertEVMChain(argv["chain"]);
+      // assertEVMChain(argv["chain"]);
       const network = argv.network.toUpperCase();
       if (
         network !== "MAINNET" &&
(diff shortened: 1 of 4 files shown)


clients/js/main.ts
```diff
@@ -254,11 +257,18 @@ yargs(hideBin(process.argv))
...
+          describe: "Print in emitter address format",
+          type: "boolean",
+          default: false,
+          required: false,
         });
     },
     async (argv) => {
       assertChain(argv["chain"]);
-      assertEVMChain(argv["chain"]);
+      // assertEVMChain(argv["chain"]);
```
panoel, line 271: Do you want to just remove this line or comment why we would use it in the future?

bruce-riley: I will remove it.

clients/js/main.ts
```diff
@@ -267,20 +277,44 @@ yargs(hideBin(process.argv))
...
+      if (argv["emitter"]) {
+        if (chain === "solana" || chain === "pythnet") { // TODO: Create an isSolanaChain()
+          addr = await getEmitterAddressSolana(addr)
+        } else if (isTerraChain(chain)) {
+          addr = await getEmitterAddressTerra(addr)
+        } else if (chain === "algorand") {
+          if (network !== "MAINNET") {
+            throw Error(`unable to look up algorand emitter address for ${network}`);
+          }
+          addr = "25e716e0618d9f38b603a97cc42db659069c0f5185230e5e61679fa876191ec4"
```
panoel, line 305: Just out of curiosity, why isn't prettier putting semicolons at the end of some lines?

node/cmd/guardiand/adminverify.go
```diff
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/protobuf/encoding/prototext"
 
 	nodev1 "github.com/certusone/wormhole/node/pkg/proto/node/v1"
+	"github.com/status-im/keycard-go/hexutils"
```
panoel, line 17: Is the certusone reference in line 16 still valid?

panoel (dismissed):

node/cmd/guardiand/adminverify.go
bruce-riley: I see it all over the imports in the guardian, so I assume so?

clients/js/main.ts
```diff
@@ -267,20 +277,44 @@ yargs(hideBin(process.argv))
...
+        } else if (chain === "near") {
+          if (network !== "MAINNET") {
+            throw Error(`unable to look up near emitter address for ${network}`);
+          }
+          addr = "148410499d3fcda4dcfd68a1ebfcdddda16ab28326448d4aae4d2f0465cdfcb7"          
+        } else {
+          addr = getEmitterAddressEth(addr)
+        }
+
+        addr += " " + toChainId(chain)
```
kcsongor, line 315: what's the use case for putting the chain id into the output? My worry is that it will make composability a little more cumbersome when used in shell scripts. For example a common use case for me is doing things like
```shell
cast call $(worm contract mainnet ethereum Core) "chainId()" --rpc-url $(worm rpc mainnet ethereum)
```
but with this change I would have to do
```shell
cast call $(worm contract mainnet ethereum Core | cut -d' ' -f1) "chainId()" --rpc-url $(worm rpc mainnet ethereum)
```

kcsongor (commented):
This is awesome!

clients/js/main.ts
bruce-riley: Dunno.

clients/js/main.ts
bruce-riley: My script needs both the emitter address and the chain ID and I got lazy...

clients/js/main.ts
```diff
@@ -267,20 +276,58 @@ yargs(hideBin(process.argv))
...
+          addr = "148410499d3fcda4dcfd68a1ebfcdddda16ab28326448d4aae4d2f0465cdfcb7";
+        } else {
+          addr = getEmitterAddressEth(addr);
+        }
+      }
+      console.log(addr);
+    }
+  )
+  .command(
+    "chain_id <chain>",
```
kcsongor, line 318: naming nit: the other commands/args use hyphens instead of underscores, so chain-id would be better here

kcsongor (approved):

clients/js/main.ts
bruce-riley: Fixed.

kcsongor (approved):

clients/js/main.ts
```diff
@@ -267,20 +276,58 @@ yargs(hideBin(process.argv))
...
+        } else if (chain === "algorand") {
+          if (network !== "MAINNET") {
+            throw Error(`unable to look up algorand emitter address for ${network}`);
+          }
+          addr = "25e716e0618d9f38b603a97cc42db659069c0f5185230e5e61679fa876191ec4";
+        } else if (chain === "near") {
+          if (network !== "MAINNET") {
+            throw Error(`unable to look up near emitter address for ${network}`);
+          }
+          addr = "148410499d3fcda4dcfd68a1ebfcdddda16ab28326448d4aae4d2f0465cdfcb7";
```
panoel, line 309: Josh just added this functionality in his PR.  Don't know if you want to take it.

panoel (approved):

clients/js/main.ts
bruce-riley: Will make that change the next time around, after Josh's change makes it into an SDK release.

PR #1509, opened 2022-09-02
sdk/js-proto*: 0.0.4 version bump

kev1n-peters:


sdk/js-proto-node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@certusone/wormhole-sdk-proto-node",
-  "version": "0.0.3",
+  "version": "0.0.4",
```
panoel, line 3: Is there a .MD file that needs to be updated for release notes?

panoel (approved):

sdk/js-proto-node/package.json
kev1n-peters: Not currently for the proto packages, but we could consider adding them in the future

evan-gray (approved):

PR #1513, opened 2022-09-03
aptos: finish core bridge

kcsongor:
See the individual commits

### DIFF
--- a/aptos/contracts/sources/contract_upgrade.move
+++ b/aptos/contracts/sources/contract_upgrade.move
@@ -11,6 +11,7 @@ module wormhole::contract_upgrade {
     const E_UNEXPECTED_HASH: u64 = 1;
     const E_INVALID_MODULE: u64 = 2;
     const E_INVALID_ACTION: u64 = 3;
+    const E_INVALID_TARGET: u64 = 4;
 
     // TODO(csongor): document how this works
     struct UpgradeAuthorized has key {
@@ -33,9 +34,8 @@ module wormhole::contract_upgrade {
         let action = deserialize::deserialize_u8(&mut cur);
         assert!(action == 0x01, E_INVALID_ACTION);
 
-        let _chain = deserialize::deserialize_u16(&mut cur);
-        // TODO(csongor): check it's the current chain
-        // assert!(chain == u16::from_u64(0x00), E_INVALID_TARGET);
+        let chain = deserialize::deserialize_u16(&mut cur);
+        assert!(chain == state::get_chain_id(), E_INVALID_TARGET);
 
         let hash = deserialize::deserialize_vector(&mut cur, 32);
 
@@ -49,6 +49,7 @@ module wormhole::contract_upgrade {
     ) acquires UpgradeAuthorized {
         let vaa = vaa::parse_and_verify(vaa);
         vaa::assert_governance(&vaa);
+        vaa::replay_protect(&vaa);
 
         authorize_upgrade(parse_payload(vaa::destroy(vaa)));
     }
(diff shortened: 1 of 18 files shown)


aptos/contracts/sources/wormhole.move
```diff
@@ -1,55 +1,108 @@
...
+    use wormhole::u32::{Self, U32};
+    use wormhole::emitter;
 
-    friend wormhole::contract_upgrade;
+// -----------------------------------------------------------------------------
+// Sending messages
 
-    // TODO(csongor): maybe merge the different capabilities into the same key
-    // to reduce storage access
+    public entry fun publish_message(
```
optke3, line 13: maybe get rid of "entry" here, because this function can't be called from the client? (because of the non-primitive-type arg emitter)

optke3: I think I read somewhere that the only args you can pass in to entry functions are primitives (bools, addresses, uints...)

aptos/contracts/sources/wormhole.move
```diff
@@ -1,55 +1,108 @@
...
+            governance_contract,
+            initial_guardian,
+            u32::from_u64(86400),
+        )
+    }
 
-        set_chain_id(u16::from_u64(chainId));
-        set_governance_chain_id(u16::from_u64(governance_chain_id));
-        set_governance_contract(governance_contract);
+    entry fun init_internal(
```
optke3, line 61: same here? remove `entry`?

aptos/contracts/sources/state.move
```diff
@@ -132,75 +130,61 @@ module wormhole::state {
...
+        let current_index = get_current_guardian_set_index();
+        let now = timestamp::now_seconds();
+
+        index == current_index ||
+            u32::to_u64(structs::get_guardian_set_expiry(guardian_set)) > now
+    }
+
     public(friend) fun set_governance_action_consumed(hash: vector<u8>) acquires WormholeState {
         let state = borrow_global_mut<WormholeState>(@wormhole);
         table::add(&mut state.consumed_governance_actions, hash, true);
```
optke3, line 167: are you relying on `table::add` to error out if the hash has already been consumed? (since it would already be in the table)

optke3:
Thought you said before you wanted to rename `/contracts` to `/wormhole`. Don't want to do that anymore?

aptos/contracts/sources/contract_upgrade.move
```diff
@@ -5,20 +5,25 @@ module wormhole::contract_upgrade {
...
     const E_INVALID_MODULE: u64 = 2;
     const E_INVALID_ACTION: u64 = 3;
+    const E_INVALID_TARGET: u64 = 4;
 
     // TODO(csongor): document how this works
     struct UpgradeAuthorized has key {
         hash: vector<u8>
     }
 
+    struct Hash {
```
optke3, line 21: What is the rationale behind passing around a Hash {vector<u8>} instead of simply a vector<u8>?

aptos/examples/core_messages/sources/sender.move
```diff
@@ -0,0 +1,36 @@
+/// A simple contracts that demonstrates how to send messages with wormhole.
+module core_messages::sender {
+    use wormhole::wormhole;
+
+    struct State has key {
+        emitter_cap: wormhole::emitter::EmitterCapability,
+    }
+
+    entry fun init_module(core_messages: &signer) {
```
optke3, line 9: nice

optke3 (dismissed):
looks good, really like the use of vaa:replay_protect and vaa:assert_governance

possibly delete `entry` in some places?

aptos/contracts/sources/state.move
kcsongor: yes, that's right

kcsongor: I've added unit tests for this to make sure it works as expected

aptos/contracts/sources/contract_upgrade.move
kcsongor: this is personal preference, I generally like to create wrapper types for things that are semantically well-defined, like a "hash". IMO it makes the code more readable because it's more clear what parse payload returns now - it's a little odd to me if the parser takes unstructured data (vector<u8>) and returns the same type of unstructured data - as if it didn't do anything!
It's also easier to see that authorize_upgrade takes the output of parse_payload just by looking at the types

jumpsiegel (approved):

PR #1518, opened 2022-09-06
node: Update libp2p to v0.22.0

jynnantonix:
Needed to build with newer go toolchains (up to 1.19).  Many of the
libp2p subpackages have been merged as sub-directories of the main
libp2p package so update the imports to use the newer versions.


kcsongor (approved):
Great! I think this was one of the dependencies (maybe the only one?) that was blocking the go upgrade to 1.18. Does this mean we can do that upgrade soon?

bruce-riley (approved):

kev1n-peters (approved):

jynnantonix:
> Great! I think this was one of the dependencies (maybe the only one?) that was blocking the go upgrade to 1.18. Does this mean we can do that upgrade soon?

I think so.  I tested this PR with 1.19 so we could even skip past 1.18.

PR #1521, opened 2022-09-06
aptos: token bridge initial scaffold

kcsongor:
Opening PR for discussion.


aptos/scripts/init_wormhole.ts
```diff
@@ -0,0 +1,53 @@
+import { AptosAccount, TxnBuilderTypes, BCS, HexString, MaybeHexString, AptosClient, FaucetClient, AptosAccountObject } from "aptos";
```
kcsongor, line 1: @optke3 we should remove this file. The chain can now be initialised by calling
```
aptos/scripts $ ./deploy
```
which sets up everything correctly, including the core bridge init stuff.

aptos/token_bridge/sources/bridge_implementation.move
```diff
@@ -0,0 +1,138 @@
...
+    public entry fun create_wrapped<CoinType>(vaa: vector<u8>): address {
+        let vaa = parse_and_verify(vaa);
+        let _token_bridge_signer = token_bridge_signer();
+        let seed = b"blah"; //TODO - how to set seed?
+        let (new_signer, new_cap) = create_resource_account(&_token_bridge_signer, seed);
+        let token_address = address_of(&new_signer);
+        deploy_coin(&new_signer);
+        bridge_state::set_wrapped_asset_signer_capability(to_bytes(&token_address), new_cap);
+
+        let _asset_meta: AssetMeta = bridge_structs::parse_asset_meta(vaa::get_payload(&vaa));
```
kcsongor, line 82: why `vaa::get_payload` instead of `vaa:destroy` here? We don't need the VAA later anyway.

aptos/token_bridge/sources/bridge_implementation.move
```diff
@@ -0,0 +1,138 @@
...
+        //let lhs = u256::add(outstanding, normalized_amount);
+        //assert!(u256::compare(lhs, &(2<<128-1))==1, 0); //LHS is less than RHS
+        //setOutstandingBridged(token, u256::add(outstanding, normalized_amount));
+    }
+
+    fun bridged_in(token: vector<u8>, normalized_amount: U256) {
+        set_outstanding_bridged(token, u256::sub(outstanding_bridged(token), normalized_amount));
+    }
+
+    fun verify_bridge_vm(vm: &VAA): bool{
```
kcsongor, line 131: I think this should not return a bool, but do an assertion instead, and we should call it
`assert_known_emitter`

optke3 (approved):
looks good

PR #1526, opened 2022-09-06
Aptos/token bridge continued

optke3:
add mint_wrapped function

### DIFF
--- a/aptos/token_bridge/sources/bridge_state.move
+++ b/aptos/token_bridge/sources/bridge_state.move
@@ -1,16 +1,14 @@
 module token_bridge::bridge_state {
-
     use std::table::{Self, Table};
     use aptos_framework::type_info::{TypeInfo, type_of, account_address};
     use aptos_framework::account::{SignerCapability, create_signer_with_capability};
-    use aptos_framework::coin::{Coin, MintCapability, BurnCapability, FreezeCapability, mint, initialize, transfer};
+    use aptos_framework::coin::{Self, Coin, MintCapability, BurnCapability, FreezeCapability, mint, initialize};
     use aptos_framework::aptos_coin::{AptosCoin};
     use aptos_framework::string::{utf8};
     use aptos_framework::bcs::{to_bytes};
-    use aptos_framework::vector::{Self};
 
     use wormhole::u256::{Self, U256};
-    use wormhole::u16::{Self, U16};
+    use wormhole::u16::{U16};
     use wormhole::emitter::{EmitterCapability};
     use wormhole::state::{get_chain_id, get_governance_contract};
     use wormhole::wormhole;
@@ -26,6 +24,8 @@ module token_bridge::bridge_state {
 
     const E_IS_NOT_WRAPPED_ASSET: u64 = 0;
     const E_COIN_CAP_DOES_NOT_EXIST: u64 = 1;
+    const E_COIN_NOT_REGISTERED: u64 = 2;
+    const E_FEE_EXCEEDS_AMOUNT: u64 = 3;
 
     struct Asset has key, store {
         chain_id: U16,
@@ -39,7 +39,7 @@ module token_bridge::bridge_state {
     }
 
     // the native chain and address of a wrapped token
-    struct NativeInfo has store, copy, drop {
+    struct OriginInfo has store, copy, drop {
         token_address: vector<u8>,
         token_chain: U16,
     }
@@ -55,15 +55,15 @@ module token_bridge::bridge_state {
         // (chainId, nativeAddress) => wrappedAddress?
         // that would be more efficient since it's a single hash and a single lookup
         //
-        // Mapping of wrapped assets (chain_id => native_address => wrapped_address)
+        // Mapping of wrapped assets (chain_id => origin_address => wrapped_address)
         //
         // A Wormhole wrapped coin on Aptos is identified by a single address, because
         // we assume it was initialized from the CoinType "deployer::coin::T", where the module and struct
         // names are fixed.
         //
-        native_info_to_wrapped_assets: Table<NativeInfo, vector<u8>>,
+        origin_info_to_wrapped_assets: Table<OriginInfo, vector<u8>>,
 
-        wrapped_assets_to_native_info: Table<vector<u8>, NativeInfo>,
+        wrapped_assets_to_origin_info: Table<vector<u8>, OriginInfo>,
 
         // https://github.com/aptos-labs/aptos-core/blob/devnet/aptos-move/framework/aptos-stdlib/sources/type_info.move
         // Mapping of native asset TypeInfo sha3_256 hash (32 bytes) => TypeInfo
@@ -110,14 +110,14 @@ module token_bridge::bridge_state {
         return state.governance_contract
     }
 
-    public entry fun wrapped_asset(native_info: NativeInfo): vector<u8> acquires State {
-        let native_info_to_wrapped_assets = &borrow_global<State>(@token_bridge).native_info_to_wrapped_assets;
-        *table::borrow(native_info_to_wrapped_assets, native_info)
+    public entry fun wrapped_asset(native_info: OriginInfo): vector<u8> acquires State {
+        let origin_info_to_wrapped_assets = &borrow_global<State>(@token_bridge).origin_info_to_wrapped_assets;
+        *table::borrow(origin_info_to_wrapped_assets, native_info)
     }
 
-    public entry fun native_info(token_address: vector<u8>): NativeInfo acquires State {
-        let wrapped_assets_to_native_info = &borrow_global<State>(@token_bridge).wrapped_assets_to_native_info;
-        *table::borrow(wrapped_assets_to_native_info, token_address)
+    public entry fun native_info(token_address: vector<u8>): OriginInfo acquires State {
+        let wrapped_assets_to_origin_info = &borrow_global<State>(@token_bridge).wrapped_assets_to_origin_info;
+        *table::borrow(wrapped_assets_to_origin_info, token_address)
     }
 
     public entry fun native_asset(token_address: vector<u8>): TypeInfo acquires State {
@@ -181,7 +181,7 @@ module token_bridge::bridge_state {
         // 2. wrapped address => (native chain, native address)
         let token_address = bridge_structs::get_token_address(& _asset_meta);
         let token_chain = bridge_structs::get_token_chain(& _asset_meta);
-        let native_info = NativeInfo {token_address: token_address, token_chain: token_chain};
+        let native_info = OriginInfo {token_address: token_address, token_chain: token_chain};
         set_native_info(token_address, &native_info);
         set_wrapped_asset(&native_info, token_address);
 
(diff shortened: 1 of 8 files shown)


aptos/token_bridge/sources/bridge_state.move
```diff
@@ -125,6 +131,15 @@ module token_bridge::bridge_state {
          *table::borrow(&state.is_wrapped_asset, token)
     }
 
+    fun mint_wrapped<CoinType>(amount:u64, token: vector<u8>): Coin<CoinType> acquires CoinCapabilities, State{
+        assert!(is_wrapped_asset(token), 0);
```
kcsongor, line 135: please define custom incremental error codes for these instead of using `0`.

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,6 +147,92 @@ module token_bridge::bridge_state {
...
+        // store coin capabilities
+        let _token_bridge_signer = token_bridge_signer();
+        let coin_caps = CoinCapabilities<CoinType> { mint_cap: mint_cap, freeze_cap: freeze_cap, burn_cap: burn_cap};
+        move_to(&_token_bridge_signer, coin_caps);
+
+        vaa::destroy(vaa);
+    }
+
+    // transfer a native or wraped token from sender to token_bridge
+    public fun transfer_tokens<CoinType>(_sender: &signer, _amount: u128, _arbiter_fee: u128): TransferResult acquires State {
```
kcsongor, line 197: these arguments are used now, so we should drop the underscores

kcsongor: this function should not take a signer but the coins instead directly, and could have a wrapper version that takes the signer, and the caller can decide which one they want to call

kcsongor: shouldn't it be an entry call?

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,6 +147,92 @@ module token_bridge::bridge_state {
...
+        // TODO: assertions and checks,
+        // check if CoinType is registered with @token_bridge
+        let token_address = hash_type_info<CoinType>();
+
+        // transfer tokens from sender to token_bridge
+        transfer<CoinType>(_sender, @token_bridge, (_amount as u64));
```
kcsongor, line 203: this might fail if the token bridge is not registered for the coin yet

kcsongor: instead of casting `amount as u64` the function should just take a u64 amount as an argument -- precise type information should always be pushed outwards as much as possible

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,6 +147,92 @@ module token_bridge::bridge_state {
...
+            native_chain = get_chain_id();
+            native_address = token_address;
+        } else {
+            // TODO - if unregistered asset, then register the asset in the native_asset map?
+        };
+
+        // TODO - normalize amount by using helpers from utils.move
+        let normalized_amount = u256::from_u128(_amount);
+        // TODO - normalize arbiter fee
+        let normalized_arbiter_fee = u256::from_u128(_arbiter_fee);
```
kcsongor, line 223: let's rename "arbiter" to "relayer" everywhere -- the former is a legacy name in the EVM contracts

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,6 +147,92 @@ module token_bridge::bridge_state {
...
+        let wormhole_fee = u256::from_u64(0);
+
+        let transfer_result: TransferResult = create_transfer_result(
+            native_chain,
+            native_address,
+            normalized_amount,
+            normalized_arbiter_fee,
+            wormhole_fee
+        );
+        transfer_result
```
kcsongor, line 233: this function should publish a message at the end

optke3: this function is meant to be transfer_tokens_internal, going to make the "wrapper" function now

aptos/token_bridge/sources/bridge_state.move
optke3: got it

aptos/token_bridge/sources/bridge_state.move
optke3: left the check for coin registration as a todo, but will do the assertion today!

aptos/token_bridge/sources/bridge_state.move
optke3: What's the reasoning behind having two functions with and without the signer again (which both do the same thing)? how about I name them
transfer_tokens_with_signer, transfer_tokens_with_coins?

aptos/token_bridge/sources/token_bridge.move
```diff
@@ -8,7 +9,76 @@ module token_bridge::token_bridge {
     /// bridge resource account)
     entry fun init_module(token_bridge: &signer) {
         let signer_cap = claim_signer_capability(token_bridge, @token_bridge);
+        init_internal(signer_cap);
+    }
+
+    public entry fun init_internal(signer_cap: SignerCapability){
```
kcsongor, line 15: this seems dangerous to be public

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,13 +148,127 @@ module token_bridge::bridge_state {
...
+        token_address: vector<u8>,
+        amount: U256,
+        recipient_chain: U16,
+        recipient: vector<u8>,
+        fee: U256,
+        message_fee: Coin<AptosCoin>,
+        nonce: u64
+    ): u64 acquires State{
+        let fee_value = coin::value<AptosCoin>(&message_fee);
+        assert!(u256::compare(&u256::from_u64(fee_value), &amount)==1, E_FEE_EXCEEDS_AMOUNT);
```
kcsongor, line 255: use `u256::LESS_THAN` instead of `1`

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,13 +148,127 @@ module token_bridge::bridge_state {
...
+        // TODO - payloadID is 1 for token transfer?
+        let transfer = bridge_structs::create_transfer(1, amount, token_address, token_chain, recipient, recipient_chain, fee);
```
kcsongor, line 257: create_transfer should not take the payload id as an argument, and the payload id should not be part of the struct. but we can fix that later (I think this is a "bug" in the EVM contracts, at least from a coding style perspective)

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -118,13 +148,127 @@ module token_bridge::bridge_state {
...
+        );
+        transfer_result
+    }
+
+    public entry fun transfer_tokens_with_signer<CoinType>(sender: &signer, amount: u64, relayer_fee: u128): TransferResult acquires State {
+        let coins = coin::withdraw<CoinType>(sender, amount);
+        transfer_tokens<CoinType>(coins, relayer_fee)
+    }
+
+    public fun log_transfer(
```
kcsongor, line 244: this being public seems like a major security risk

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -159,10 +303,18 @@ module token_bridge::bridge_state {
         table::upsert(&mut state.bridge_implementations, chain_id, bridge_contract);
     }
 
-    public entry fun set_wrapped_asset(token_chain_id: U16, token_address: vector<u8>, wrapper: vector<u8>) acquires State {
+    public entry fun set_wrapped_asset(native_info: &OriginInfo, wrapper: vector<u8>) acquires State {
```
kcsongor, line 306: did we want this `public` (and `entry`)? It looks like it allows anyone to override the state

aptos/token_bridge/sources/bridge_state.move
```diff
@@ -159,10 +303,18 @@ module token_bridge::bridge_state {
...
+        let state = borrow_global_mut<State>(@token_bridge);
+        let origin_info_to_wrapped_assets = &mut state.origin_info_to_wrapped_assets;
+        table::upsert(origin_info_to_wrapped_assets, *native_info, wrapper);
+        let is_wrapped_asset = &mut state.is_wrapped_asset;
+        table::upsert(is_wrapped_asset, wrapper, true);
+    }
+
+    public entry fun set_native_info(wrapper: vector<u8>, native_info: &OriginInfo) acquires State {
```
kcsongor, line 314: likewise, I don't think this should be `public` or `entry`

aptos/token_bridge/sources/bridge_state.move
optke3: made this public(friend)

aptos/token_bridge/sources/bridge_structs.move
```diff
@@ -26,6 +26,26 @@ module token_bridge::bridge_structs {
         // Amount of tokens (big-endian uint256) that the user is willing to pay as relayer fee. Must be <= Amount.
         fee: U256, //should be u256
     }
+    public fun create_transfer(
```
kcsongor, line 29: should be `public(friend)` - we should protect/hide our struct constructors as a general principle

aptos/token_bridge/sources/bridge_structs.move
```diff
@@ -54,11 +74,27 @@ module token_bridge::bridge_structs {
         // Amount being transferred (big-endian uint256)
         normalized_amount: U256,
         // Amount of tokens (big-endian uint256) that the user is willing to pay as relayer fee. Must be <= Amount.
-        normalized_arbiter_fee: U256, // should be u256
+        normalized_relayer_fee: U256, // should be u256
         // Portion of msg.value to be paid as the core bridge fee
         wormhole_fee: U256,
     }
 
+    public fun create_transfer_result(
```
kcsongor, line 82: private

aptos/token_bridge/sources/bridge_structs.move
```diff
@@ -159,6 +195,18 @@ module token_bridge::bridge_structs {
         }
     }
 
+    // Construct a seed using AssetMeta fields for creating a new resource account 
+    // N.B. seed is product of coin native chain and native address
+    public fun create_seed(asset_meta: &AssetMeta): vector<u8>{
```
kcsongor, line 200: private

aptos/token_bridge/sources/bridge_state.move
optke3: made these public(friend) and non-entry

optke3: some state-modifying functions are split between `bridge_state` and `bridge_implementation`, which is why it's not totally private

aptos/token_bridge/sources/bridge_structs.move
optke3: done

aptos/token_bridge/sources/bridge_structs.move
optke3: done

aptos/token_bridge/sources/bridge_structs.move
optke3: **done**

aptos/token_bridge/sources/bridge_state.move
optke3: ok

aptos/token_bridge/sources/token_bridge.move
optke3: made it private

aptos/token_bridge/sources/bridge_state.move
optke3: u256.move(69, 11): Constants are internal to their module, and cannot can be accessed outside of their module

aptos/token_bridge/sources/bridge_state.move
optke3: made it private

PR #1527, opened 2022-09-06
aptos: token bridge VAA utils

kcsongor:
Added a slightly cleaner way to do replay protection by replacing the `Table<vector<u8>, bool>` with a custom `Set` type, which should just make the code a little easier to reason about (in particular now there's no such thing as a key being present but the value being false).

Also added a `token_bridge::vaa` module with some common utilities. In particular, `parse_verify_and_replay_protect` should be called by all token bridge functions that take VAA bytes, because it does the parsing, the verification, emitter checking, and replay protection all in one go. These checks need to be done by all token bridge functions so might as well share this bit out.


optke3 (approved):
lgtm!

PR #1536, opened 2022-09-06
cosmwasm: simplify quorum calculation

tbjump:
No functional changes. Just making a complicated formula more easy to read and make it match the formula in other places of the codebase.

### DIFF
--- a/cosmwasm/contracts/wormhole/src/state.rs
+++ b/cosmwasm/contracts/wormhole/src/state.rs
@@ -216,7 +216,7 @@ impl GuardianSetInfo {
         if self.addresses.is_empty() {
             return 0;
         }
-        ((self.addresses.len() * 10 / 3) * 2) / 10 + 1
+        ((self.addresses.len() * 2) / 3) + 1
     }
 }
 
--- a/near/contracts/wormhole/src/lib.rs
+++ b/near/contracts/wormhole/src/lib.rs
@@ -47,7 +47,7 @@ pub struct GuardianSetInfo {
 
 impl GuardianSetInfo {
     pub fn quorum(&self) -> usize {
-        ((self.addresses.len() * 10 / 3) * 2) / 10 + 1
+        ((self.addresses.len() * 2) / 3) + 1
     }
 }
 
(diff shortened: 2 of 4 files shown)


kcsongor (approved):

jumpsiegel (approved):

cosmwasm/contracts/wormhole/src/state.rs
```diff
@@ -238,7 +238,7 @@ impl GuardianSetInfo {
         if self.addresses.is_empty() {
             return 0;
         }
-        ((self.addresses.len() * 10 / 3) * 2) / 10 + 1
+        ((self.addresses.len() * 2) / 3) + 1
```
claudijd, line 241: Do each of the chains we're modifying quorum calculation here have independant quorum unit-tests to ensure that this doesn't cause a regression?  I'm thinking of what exists for Solidity / Guardians:

- https://github.com/wormhole-foundation/wormhole/blob/dev.v2/node/pkg/processor/quorum_test.go
- https://github.com/wormhole-foundation/wormhole/blob/dev.v2/ethereum/forge-test/Messages.t.sol#L13-L29

Where possible, I'd love to see the core trust assumptions, like quorum, well unit-tested on all chains.  This is especially true when making medications to those implementations to help prevent regression scenarios.

claudijd: That said, the math is pretty straight forward, so as a function of the review, I agree with the implementation changes, would just like to see associated test coverage.

kcsongor (approved):

evan-gray (changes requested):
I would approve, but given the criticality of the code can you provide the unit tests that cover each of these?

tbjump:
Unit tests are here: https://github.com/wormhole-foundation/wormhole/blob/dedcea34af2c6760882c019957d51359aef186f7/cosmwasm/contracts/wormhole/src/testing/tests.rs#L6

kcsongor (dismissed):

jynnantonix (approved):

jynnantonix (approved):

jynnantonix (dismissed):

hendrikhofstadt:
Are we 100% sure this behaves the same in the Solana bpf runtime?

evan-gray:
Discussed with @jumpsiegel and @hendrikhofstadt and we would prefer not to change these critical calculations at this time.

PR #1543, opened 2022-09-07
Solana CI improvements

jynnantonix:
This PR improves our solana pre-submit checks by moving them out of Tilt and into a dedicated github action.  It also adds formatting and linting (clippy) checks so that we can catch more issues in pre-submit.

Most of the changes should be no-ops but the replacement for the deprecated function in the nft-bridge could use some extra scrutiny.  Since the deprecation function suggests the replacement, it should be a drop-in replacement but the actual implementation is slightly different (the new function constructs a slightly different instruction).  An alternative is to just keep using the deprecated function and silence the warning.


.github/workflows/build.yml
```diff
@@ -81,6 +81,112 @@ jobs:
...
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('solana/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Run `cargo fmt`
+        run: cargo fmt --check --all --manifest-path solana/Cargo.toml
```
kcsongor, line 138: how about adding these commands as make targets so they can more conveniently be invoked locally too?

jynnantonix: Done.

kcsongor (dismissed):

kcsongor (approved):

hendrikhofstadt (approved):

PR #1549, opened 2022-09-07
near/upgradeCheck: upgrade test

jumpsiegel:


near/Dockerfile.build
```diff
@@ -3,8 +3,10 @@ FROM debian@sha256:2ce44bbc00a79113c296d9d25524e15d423b23303fdbbe20190d2f96e0aeb
 RUN apt-get update && apt-get install apt-utils && apt-get install -y python3 npm curl --no-install-recommends
 RUN apt-get install -y build-essential git
 
-ADD . .
+ADD setup-rust.sh .
 RUN ./setup-rust.sh
+ADD . .
+RUN make clean
```
panoel, line 9: Just out of curiosity, do you need to "make clean" on a brand spanking new docker?  Or does this cover rebuild/cache cases?

near/contracts/wormhole/src/lib.rs
```diff
@@ -602,6 +602,7 @@ impl Wormhole {
         true
     }
 
+    #[private]
```
panoel, line 605: If you make the migrate function private, can you still migrate?

panoel (dismissed):

near/Dockerfile.build
jumpsiegel: The add . .   copied in all my prebuild files and I wanted them to be built entirely in the dockerfile

near/contracts/wormhole/src/lib.rs
jumpsiegel: The migrate function is called automatically during a contract upgrade.  The Real question that keeps me up at night is... if it is not private, can people do any harm since it basically overwrites the contract state with itself..

near/Dockerfile.build
kcsongor: you can explicitly ignore those in the `.dockerignore` file next to the docker file

panoel (approved):

bruce-riley (approved):

PR #1550, opened 2022-09-07
make evm watcher more robust to correctly observing messages with low consistencyLevel

tbjump:


node/pkg/ethereum/watcher.go
```diff
@@ -522,9 +545,9 @@ func (e *Watcher) Run(ctx context.Context) error {
 								zap.Stringer("current_block", ev.Number),
 								zap.Stringer("current_blockhash", currentHash),
 								zap.String("eth_network", e.networkName))
-							delete(e.pending, key)
```
bruce-riley, line 525: Why don't we want to delete this? If we don't, we will generate the log message and peg the count multiple times until it gets dropped for being too old. Also, if this is a "SECURITY CRITICAL" error, should this be more than a logger.Info?

node/pkg/ethereum/watcher.go
```diff
@@ -99,6 +99,9 @@ type (
 		// Minimum number of confirmations to accept, regardless of what the contract specifies.
 		minConfirmations uint64
 
+		// Number of blocks we will wait until we abandon a previously observed message that is not on the current chain anymore
+		messageBlocksTillAbandon uint64
```
bruce-riley, line 103: There is quite a variety of values being passed in for messageBlocksTillAbandon (as opposed to before when it was four times minConfirmations). Is there logic to how you came up with these values? If so, could you put a comment explaining how to come up with the value?

bruce-riley (approved):

tbjump:
Closing because not as relevant anymore after The Merge and ethereum switching to PoS.

PR #1552, opened 2022-09-07
Aptos/wrapped coin type

optke3:


kcsongor (approved):

kcsongor (approved):

aptos/coin/Move.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "WrappedCoin"
+version = "0.0.1"
+upgrade_policy = "compatible"
```
kcsongor, line 4: maybe immutable would be better here?

kcsongor: we can come back to this later

PR #1564, opened 2022-09-08
aptos/token_bridge: implement transfer_tokens

optke3:


aptos/token_bridge/sources/bridge_state.move
```diff
@@ -193,46 +274,68 @@ module token_bridge::bridge_state {
...
+        nonce: u64,
+        payload: vector<u8>
+    ): u64 acquires State{
+        let transfer = transfer_with_payload::create(
+            amount,
+            token_address,
+            token_chain,
+            recipient,
+            recipient_chain,
+            to_bytes<address>(&@token_bridge), //TODO - is token bridge the only one who will ever call log_transfer_with_payload?
```
kcsongor, line 331: no, we'll have to roll something similar to `EmitterCapability`, but at the token bridge level. for now I reused `EmitterCapability` in `complete_transfer` but we should have a token bridge version of that stuff really.

aptos/token_bridge/sources/utils.move
```diff
@@ -19,28 +19,35 @@ module token_bridge::utils {
...
-    // public entry fun normalize_amount(amount: U256, decimals: u8): U256 {
-    //     if (decimals > 8) {
-    //         amount = u256::div(amount, 10 ** (decimals - 8));
-    //     };
-    //     amount
-    // }
+    public entry fun normalize_amount(amount: U256, decimals: u8): U256 {
+         if (decimals > 8) {
+            let n = decimals - 8;
+            while (n > 0){
```
kcsongor, line 25: not sure if we need a loop here, just
```
amount = u256::div(amount, from_u64(10^n));
```

aptos/token_bridge/sources/bridge_state.move
optke3: we'll probably have to discuss this later, I'm not sure what that implementation would look like

aptos/token_bridge/sources/utils.move
optke3: tried this and it doesn't work, I don't think ^ is exponentiation, couldn't figure out the syntax for it or find it...

kcsongor: ah looks like `^` is exclusive or

kcsongor (approved):
👍

PR #1565, opened 2022-09-09
Fix CI race conditions

justinschuldt:
I noticed intermittent failures of `sdk-ci-tests` in CI, found that the tests were sometimes running before Algorand was ready, and that the guardian was not waiting for Algorand to come up.

This change fixes the race condition by:
- making `algorand` a `resource_dep` of the guardian
- making `guardian` a `resource_dep` of `sdk-ci-tests` & `spydk-ci-tests`


evan-gray (approved):
Great catch, thanks!

ckeun (approved):

PR #1572, opened 2022-09-11
aptos/wormhole: external address

optke3:


aptos/token_bridge/sources/structs/transfer.move
```diff
@@ -13,7 +13,7 @@ module token_bridge::transfer {
         deserialize_vector
     };
     use wormhole::cursor::{Self};
-
+    use wormhole::external_address::{ExternalAddress, from_vector, get_bytes};
```
kcsongor, line 16: I think it's a little easier to read code if the import statement only grabs types unqualified, but not functions, so
```suggestion
    use wormhole::external_address::{Self, ExternalAddress};
```

because then looking at the code later, `external_address:get_bytes(...)` is clearer than just `get_bytes(...)`

aptos/token_bridge/sources/structs/transfer.move
```diff
@@ -65,19 +65,19 @@ module token_bridge::transfer {
...
-            token_chain,
-            to,
-            to_chain,
-            fee,
+            amount: amount,
+            token_address: token_address,
+            token_chain: token_chain,
+            to: to,
+            to_chain: to_chain,
+            fee: fee,
```
kcsongor, line 80: any reason for spelling out these field names explicitly?
```suggestion
            amount,
            token_address,
            token_chain,
            to,
            to_chain,
            fee,
```
in rust, the compiler would warn about this (https://rust-lang.github.io/rfcs/1682-field-init-shorthand.html), so it's not unreasonable to expect a move compiler warning about this some time down the line.

aptos/token_bridge/sources/structs/transfer_result.move
```diff
@@ -1,21 +1,22 @@
...
-        token_address: vector<u8>,
+        token_address: ExternalAddress,
         // Amount being transferred (big-endian uint256)
         normalized_amount: U256,
         // Amount of tokens (big-endian uint256) that the user is willing to pay as relayer fee. Must be <= Amount.
         normalized_relayer_fee: U256,
     }
 
-    public fun destroy(a: TransferResult): (U16, vector<u8>, U256, U256) {
+    public fun destroy(a: TransferResult): (U16, ExternalAddress, U256, U256) {
```
kcsongor, line 19: this type is now so much nicer to look at 🙂

aptos/wormhole/sources/state.move
```diff
@@ -44,7 +45,7 @@ module wormhole::state {
         governance_chain_id: U16,
 
         /// Address of governance contract on governance chain
-        governance_contract: vector<u8>, //(TODO: create custom type for wormhole addresses)
+        governance_contract: ExternalAddress, //(TODO: create custom type for wormhole addresses)
```
kcsongor, line 48:
```suggestion
        governance_contract: ExternalAddress,
```

aptos/wormhole/sources/external_address.move
```diff
@@ -0,0 +1,139 @@
...
+            if ((length-i)<=32){
+                vector::push_back(&mut res, *vector::borrow(&bytes, i));
+            };
+            i = i +1;
+        };
+        res
+    }
+
+    public fun from_vector(bytes: vector<u8>): ExternalAddress {
+        left_pad(&take_32(bytes))
```
kcsongor, line 48: I'm tempted to suggest that we should not allow longer than 32 bytes coming in here, and only do left-padding.

In contrast, for `String32` it's reasonable to expect truncation semantics because `String` represents free-form text, which we want to squeeze into our payloads, but an `ExternalAddress` should never come in longer than 32 bytes, because then truncation loses some information which becomes unrecoverable. At best, this is a denial of service, at worst it's a security risk.

Instead, if the 'address' is longer, then we need a preprocessing mechanism for fitting it into 32 bytes first, like the hashing we do for native token addresses. But that should not be the responsibility of the `ExternalAddress` module I think (as the preprocessing step is domain-specific). So it should just reject anything longer than 32 bytes.

aptos/token_bridge/sources/structs/transfer.move
optke3: it was because I had `from_vector(token_address)` originally in place of token_address, so had to do `token_address:from_vector(token_address)`, and just named all the other fields for consistency. Can change this back.

aptos/token_bridge/sources/structs/transfer.move
optke3: done

aptos/token_bridge/sources/structs/transfer_result.move
optke3: agree!

aptos/wormhole/sources/state.move
optke3: done!

aptos/wormhole/sources/external_address.move
optke3: got rid of take_32

kcsongor (approved):
lgtm 👍

PR #1592, opened 2022-09-13
cosmwasm/contracts: Adding shutdown contracts

panoel:
Added shutdown contracts and remove migrate functionality.
Basically, copied the wormhole and token bridge contracts and stripped out functionality.
You might be better off diff'ing the shutdown contracts with the base contracts to see what was removed.  
A directory diff would work well.


kcsongor:
see here for an alternative approach: https://github.com/wormhole-foundation/wormhole/pull/1598

PR #1598, opened 2022-09-14
cosmwasm: Add shutdown feature flag to token bridge

kcsongor:


cosmwasm/contracts/token-bridge/Cargo.toml
```diff
@@ -10,6 +10,9 @@ crate-type = ["cdylib", "rlib"]
 
 [features]
 backtraces = ["cosmwasm-std/backtraces"]
+# This feature flag does a 'shutdown' mode compilation, which disables most
+# functionality except for governance
+shutdown = []
```
jynnantonix, line 15: Cargo features are supposed to be [additive](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification):

> When a dependency is used by multiple packages, Cargo will use the union of all features enabled on that dependency when building it.
> ...
> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#semver-compatibility).

Yes, this means all the crypto crates using features for mutually exclusive things are _wrong_.
 
I think the usual convention for something like this would be to define a "full" feature that's enabled by default:

```
[features]
full = []
default = ["full"]
```

and then the shutdown contract would depend on this crate with default features disabled:

```
[dependencies]
token-bridge-terra-2 = { path = "../token-bridge", default-features = false }
```

kcsongor: thanks, I wasn't aware of this, but makes perfect sense

panoel:
This feature needs to have shutdown contracts for both core bridge and token bridge.  The shutdown token bridge should be tied to the shutdown core bridge.  If the shutdown token bridge is implemented as shown in this PR, then how do we tell it to use the shutdown-wormhole lib instead of the normal wormhole lib?  Is there a way to do that in the Cargo.toml file?

jynnantonix:
> This feature needs to have shutdown contracts for both core bridge and token bridge. The shutdown token bridge should be tied to the shutdown core bridge. If the shutdown token bridge is implemented as shown in this PR, then how do we tell it to use the shutdown-wormhole lib instead of the normal wormhole lib? Is there a way to do that in the Cargo.toml file? The token-bridge/Cargo.toml file has the following dependency: `wormhole-bridge-terra-2 = { path = "../wormhole", features = ["library"] }` Is there a way to make that variable?

I think the way to do that is the same as the tokenbridge.  The wormhole crate defines a "full" (or some other name) feature in the default set:

```toml
[features]
backtraces = ["cosmwasm-std/backtraces"]
library = []
full = []
default = ["full"]
```

The "full" feature would enable all the non-shutdown functionality of the contract and the tokenbridge contract's "full" feature would depend on it:

```toml
[features]
backtraces = ["cosmwasm-std/backtraces"]
library = []
full = ["wormhole-bridge-terra-2/full"]
default = ["full"]
```

Then in shutdown mode we turn default-features off, which will automatically turn off non-shutdown features for the wormhole contract.


On a separate note, it seems like the design of the library feature is also backwards: enabling the feature removes functionality (the entrypoints), which is not how cargo features are supposed to be used.  Instead we should have an "entrypoint" feature that enables entrypoints when it is enabled.

cosmwasm/contracts/shutdown-token-bridge/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Wormhole Project Contributors"]
+edition = "2018"
+description = "Shutdown Wormhole token bridge on cosmwasm"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+token-bridge-terra-2 = { path = "../token-bridge", features = [] }
```
jynnantonix, line 12: We need to set `default-features = false` here.  `features = []` doesn't do anything.

```suggestion
token-bridge-terra-2 = { path = "../token-bridge", default-features = false }
```

cosmwasm/contracts/shutdown-wormhole/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Wormhole Project Contributors"]
+edition = "2018"
+description = "Shutdown Wormhole core bridge on cosmwasm"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+wormhole-bridge-terra-2 = { path = "../wormhole", features = [] }
```
jynnantonix, line 12: Same here

```suggestion
wormhole-bridge-terra-2 = { path = "../wormhole", default-features = false }
```

jynnantonix (commented):
I think it might be worth adding integration tests to ensure that the non-shutdown functionality is actually disabled in shutdown mode.

cosmwasm/contracts/shutdown-token-bridge/Cargo.toml
panoel: Fixed

cosmwasm/contracts/shutdown-wormhole/Cargo.toml
panoel: Fixed

panoel:
Closing this PR and opening a new one so that @panoel can author it and @kcsongor can be a reviewer.  This PR will be referenced by the new PR.

jynnantonix (approved):

panoel (approved):

jynnantonix (approved):

PR #1605, opened 2022-09-16
node: processor: Fix incorrect fallthrough in `handleCleanup`

jynnantonix:
When processing pending observations, the `handleCleanup` function
checks if we already have a stored quorum VAA and deletes the in-memory
observation if one is found.  If a stored quorum VAA is not found, then
we're supposed to continue evaluating the other conditions and take
the appropriate action.  This was implemented using a `fallthrough`
statement.

Unfortunately, this is not how `fallthrough` works.  `fallthrough`
simply tells the compiler to execute the body of the next branch in
the switch block, *without evaluating the condition*.  It also doesn't
evaluate the conditions for any of the other branches in the switch.

In practice what this meant is that for local observations that didn't
have quorum we would always take the first branch, fall through to the
second branch, and then exit the switch.  Only once we had a quorum
(`s.submitted == true`) would we actually consider any of the other
branches in the switch.  It also meant that there was no case where we
would take the branch for re-observing messages that hadn't reached
quorum.

Fix this by moving the stored quorum VAA check into an if statement and
then falling through to the switch statement if one is not found.

This code badly needs better testing, which I will probably work on next
week.  We can either wait to merge this until I have tests ready or we can
try to merge as is if we want to get the fix out sooner.

### DIFF
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -66,8 +66,7 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 	for hash, s := range p.state.signatures {
 		delta := time.Since(s.firstObserved)
 
-		switch {
-		case !s.submitted && s.ourObservation != nil && delta > settlementTime:
+		if !s.submitted && s.ourObservation != nil && delta > settlementTime {
 			// Expire pending VAAs post settlement time if we have a stored quorum VAA.
 			//
 			// This occurs when we observed a message after the cluster has already reached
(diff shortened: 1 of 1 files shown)


node/pkg/processor/cleanup.go
```diff
@@ -81,16 +80,17 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 					p.logger.Info("Expiring late VAA", zap.String("digest", hash), zap.Duration("delta", delta))
 					aggregationStateLate.Inc()
 					delete(p.state.signatures, hash)
-					break
+					continue
```
kcsongor, line 83: looking at this code, `continue` _seems_ to be the right thing to do. Do you know why it did a `break` before?

kcsongor (approved):

node/pkg/processor/cleanup.go
jynnantonix: It was an early break out of the switch case.

kcsongor: makes sense, the first resource I found on `break` only discussed loops, but but now I see that it applies to switch and select too

hendrikhofstadt (approved):
Great catch!

In favour of including this in the next release even without full testing. No security risk and this is clearly more correct behaviour even if the whole system still isn't 100% correct pre-tests.

evan-gray:
Great find. Concur with getting this fix out. 🙂

PR #1606, opened 2022-09-16
aptos: cap symbol length

kcsongor:
aptos recently added a max limit of 10 on symbol length (and 32 on symbol name).


optke3 (approved):
lgtm

PR #1609, opened 2022-09-16
Complete Rust SDK

Reisen:
This PR picks up some of the old SDK code that is currently unused and puts in some of the final touches to get it cleaned up. While the PR looks large, It's mostly repetitive serialization implementations. The best way to read this PR, instead of commit by commit, is to start browsing from the [core crate](https://github.com/wormhole-foundation/wormhole/tree/reisen/sdk-cleanup/sdk/rust/core/src).

The goal of this `core` crate is for on-chain and off-chain Rust programs the ability to parse and interact with core Wormhole structures, without any chain-specific code.

Changes to the original SDK:

- Drops the `sdk/` module with chain-specific code, provides only a thin core.
- Adds a lot more documentation.
- `Chain` enum extended to cover all chains, and uses `From` instead of `repr(u16)`.
- The `Error` enum extended to cover as many known Wormhole errors as possible.
- Uses `nom` currently to do parsing due to good error feedback and efficient parsing [^1]
- Exposes more common Wormhole datatypes, like `GuardianSet` and chain headers/consts.
- Exposes all known Wormhole actions.

The last bullet point is the largest change, it exposes an `Action` trait which can be implemented for a programs action enum to go from a `VAA` to a concrete action. The parsers here can be used by the current wormhole module implementations as this crate has zero dependencies on Solitaire, Anchor, or anything else chain specific.

This would allow for example, a token bridge implementation on any Rust based chain (Cosmwasm for example) to write their VAA/Action code as follows:

```rust
use wormhole::Action;
use wormhole::Error;
use wormhole::token_bridge::Action as TokenAction;

match TokenAction::from_vaa(&vaa, Chain::Terra)? {
    RegisterChain(a)       => handle_register_chain(a)?,
    ContractUpgrade(a)     => handle_contract_upgrade(a)?,
    AssetMeta(a)           => handle_asset_meta(a)?,
    Transfer(a)            => handle_transfer(a)?,
    TransferWithPayload(a) => handle_transfer_with_payload(a)?,
}
```

Note that the `from_vaa` function can do error handling, and in the current PR will do all the checks necessary to parse GovHeaders and verify chain destination, module header and so on.

[^1]: Though should/could use #1001 instead. @jynnantonix


jynnantonix (approved):
Generally looks good to me though I obviously I'm biased towards using the serde-based solution (#1001).  serde is one of the few crates that's widely adopted across the entire rust ecosystem and it gives users of the sdk more flexibility with using other data formats (like json). 

I suspect that people will eventually ask us to provide the standard serde Serialize and Deserialize impls at which point the benefit of using a bespoke parsing solution is unclear to me.

hendrikhofstadt:
@jynnantonix @Reisen can we close this? As far as I understand @jynnantonix has picked up some of the changes into the other SDK PRs?

evan-gray:
Handled in #1919

PR #1618, opened 2022-09-19
Refactor ethwatcher

hendrikhofstadt:
This PR modularizes the eth watcher and reduces code duplication


node/pkg/evm/watcher.go
```diff
@@ -155,59 +143,109 @@ func NewEthWatcher(
...
-	}
 
-	err = e.ethIntf.NewAbiCaller(e.contract)
-	if err != nil {
-		panic(err)
+	var err error
+	if w.chainID == vaa.ChainIDCelo && !w.unsafeDevMode {
+		// When we are running in mainnet or testnet, we need to use the Celo ethereum library rather than go-ethereum.
+		// However, in devnet, we currently run the standard ETH node for Celo, so we need to use the standard go-ethereum.
+		w.ethConn, err = connectors.NewEthereumConnector(timeout, w.networkName, w.url, w.contract, logger)
```
bruce-riley, line 172: Should this be NewCeloConnector?

bruce-riley:
Thanks for doing this! I think it will make things much easier going forward.

node/pkg/evm/connectors/logpoller.go
```diff
@@ -0,0 +1,158 @@
...
+
+		return fmt.Errorf("GetLogsQuery: failed to query for log messages: %w", err)
+	}
+
+	if len(logs) == 0 {
+		return nil
+	}
+
+	for _, log := range logs {
+		if log.Topics[0] == logsLogMessageTopic {
```
jynnantonix, line 140: This is more of a personal preference but we can invert this condition to reduce nesting, which I personally find to be more readable:

```
for _, log := range logs {
	if log.Topics[0] != logsLogMessageTopic {
		continue
	}

	ev, err := l.ParseLogMessagePublished(log)
	if err != nil {
		// logging
		continue
	}

	l.messageFeed.Send(ev)
}
```

This makes the normal code path have the lowest indentation and only the exceptional paths have extra indentation. Also see https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+	supervisor.Signal(ctx, supervisor.SignalHealthy)
+
+	var errorCount int
+	for {
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return ctx.Err()
+		case <-timer.C:
+			var errorOccurred bool
```
jynnantonix, line 69: readability nit: maybe refactor this case block into a separate named function

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+			if errorOccurred {
+				errorCount++
+				if errorCount > 1 {
+					b.errFeed.Send(fmt.Errorf("polling encountered multiple errors"))
+				}
+			} else {
+				errorCount = 0
+			}
+
+			timer = time.NewTimer(time.Duration(b.DelayInMs) * time.Millisecond)
```
jynnantonix, line 134: I think this is fine the way it is but we know the timer has expired and we've drained the channel (because we're in the `case <-timer.C` block) so we could also do `timer.Reset(time.Duration(b.DelayInMs) * time.Millisecond)`

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+
+type PollFinalizer interface {
+	IsBlockFinalized(ctx context.Context, block *NewBlock) (bool, error)
+}
+
+// BlockPollConnector polls for new blocks instead of subscribing when using SubscribeForBlocks. It allows to specify a
+// finalizer which will be used to only return finalized blocks on subscriptions.
+type BlockPollConnector struct {
+	Connector
+	DelayInMs           int
```
jynnantonix, line 26: Seems like we multiply this by `time.Millisecond` in the only place it's used so can this just be a `time.Duration` instead?

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+				errorCount = 0
+			}
+
+			timer = time.NewTimer(time.Duration(b.DelayInMs) * time.Millisecond)
+		}
+	}
+}
+
+func (b *BlockPollConnector) SetEthSwitched() {
+	b.hasEthSwitchedToPoS = true
```
jynnantonix, line 140: Now that eth has switched do we still need this?

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+	innerErrSub := b.errFeed.Subscribe(innerErrSink)
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				innerSub.Unsubscribe()
+				innerErrSub.Unsubscribe()
+				return
+			case <-sub.quit:
```
jynnantonix, line 163: Same comment here about `sub.quit` being uninitialized.

node/pkg/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+	b.hasEthSwitchedToPoS = true
+}
+
+func (b *BlockPollConnector) SubscribeForBlocks(ctx context.Context, sink chan<- *NewBlock) (ethereum.Subscription, error) {
+	sub := &PollSubscription{
+		err: make(chan error, 1),
+	}
+
+	innerSink := make(chan *NewBlock, 10)
+	innerSub := b.blockFeed.Subscribe(innerSink)
```
jynnantonix, line 149: Is there a reason we're not passing in `sink` directly here?  Is it so that we can guarantee that it's buffered?  Otherwise, the only thing `innerSink` is used for is forwarding to `sink`.

node/pkg/evm/watcher.go
```diff
@@ -155,59 +143,109 @@ func NewEthWatcher(
...
+			p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
+			return fmt.Errorf("creating poll connector failed: %w", err)
+		}
+	} else {
+		w.ethConn, err = connectors.NewEthereumConnector(timeout, w.networkName, w.url, w.contract, logger)
+		if err != nil {
+			ethConnectionErrors.WithLabelValues(w.networkName, "dial_error").Inc()
+			p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
+			return fmt.Errorf("dialing eth client failed: %w", err)
+		}
```
jynnantonix, line 230: style nit: maybe refactor into a separate function so that we don't have to keep repeating the 

```go
ethConnectionErrors.WithLabelValues(w.networkName, "dial_error").Inc()
p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
```

lines for every branch.  Something like:

```go
if err := w.initializeConnector(); err != nil {
	ethConnectionErrors.WithLabelValues(w.networkName, "dial_error").Inc()
	p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
	return err
}
```

jynnantonix (commented):
Nice!  I really like the modular design here.  Most of my comments are minor nits except for one issue with uninitialized channels.

Also, since we're refactoring this code anyway it might be worth adding some tests? ;-p

node/pkg/evm/connectors/ethereum.go
```diff
@@ -0,0 +1,146 @@
...
+func (e *EthereumConnector) NetworkName() string {
+	return e.networkName
+}
+
+func (e *EthereumConnector) ContractAddress() ethCommon.Address {
+	return e.address
+}
+
+func (e *EthereumConnector) GetCurrentGuardianSetIndex(ctx context.Context) (uint32, error) {
+	if e.caller == nil {
```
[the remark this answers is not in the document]
hendrikhofstadt: Yes!

node/pkg/evm/connectors/poller.go
hendrikhofstadt: Great point!

node/pkg/evm/connectors/poller.go
hendrikhofstadt: We don't, but I'd recommend we change that in a separate PR.

node/pkg/evm/connectors/poller.go
hendrikhofstadt: Did that and cleaned up the logic there

node/pkg/evm/watcher.go
hendrikhofstadt: Yeah I was thinking about that but given the different error cases with different adapters I didn't find a nice common denominator to extract into a function.

bruce-riley (approved):

jynnantonix (approved):

PR #1627, opened 2022-09-20
node: prevent reobservation of unreliable messages

hendrikhofstadt:
Given solana unreliable messages overwrite the account, they cannot be reobserved. 
This change prevents sending reobservation requests for unreliable messages.

### DIFF
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -177,12 +177,19 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 			aggregationStateTimeout.Inc()
 		case !s.submitted && delta.Minutes() >= 5:
 			// Poor observation has been unsubmitted for five minutes - clearly, something went wrong.
-			// If we have previously submitted an observation, we can make another attempt to get it over
-			// the finish line by sending a re-observation request to the network and rebroadcasting our
+			// If we have previously submitted an observation, and it was reliable, we can make another attempt to get
+			// it over the finish line by sending a re-observation request to the network and rebroadcasting our
 			// sig. If we do not have an observation, it means we either never observed it, or it got
 			// revived by a malfunctioning guardian node, in which case, we can't do anything about it
 			// and just delete it to keep our state nice and lean.
 			if s.ourMsg != nil {
+				// Unreliable observations cannot be resubmitted and can be considered failed after 5 minutes
+				if !s.ourObservation.IsReliable() {
+					p.logger.Info("expiring unsubmitted unreliable observation", zap.String("digest", hash), zap.Duration("delta", delta))
+					delete(p.state.signatures, hash)
+					aggregationStateTimeout.Inc()
+					break
+				}
 				p.logger.Info("resubmitting observation",
 					zap.String("digest", hash),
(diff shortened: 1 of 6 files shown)


node/pkg/processor/cleanup.go
```diff
@@ -177,12 +177,19 @@ func (p *Processor) handleCleanup(ctx context.Context) {
...
-			// the finish line by sending a re-observation request to the network and rebroadcasting our
+			// If we have previously submitted an observation, and it was reliable, we can make another attempt to get
+			// it over the finish line by sending a re-observation request to the network and rebroadcasting our
 			// sig. If we do not have an observation, it means we either never observed it, or it got
 			// revived by a malfunctioning guardian node, in which case, we can't do anything about it
 			// and just delete it to keep our state nice and lean.
 			if s.ourMsg != nil {
+				// Unreliable observations cannot be resubmitted and can be considered failed after 5 minutes
+				if s.ourObservation.IsUnreliable() {
+					p.logger.Info("expiring unsubmitted unreliable observation after exhausting retries", zap.String("digest", hash), zap.Duration("delta", delta))
```
jynnantonix, line 188: nit: the log message is a bit inaccurate since we didn't really retry anything

node/pkg/processor/processor.go
```diff
@@ -31,6 +31,8 @@ type (
 		// SigningMsg returns the hash of the signing body of the observation. This is used
 		// for signature generation and verification.
 		SigningMsg() ethcommon.Hash
+		// IsUnreliable returns whether this message is considered unreliable meaning it cannot be reobserved.
+		IsUnreliable() bool
```
jynnantonix, line 35: naming nit: how about `IsReliable`?  Otherwise parsing a double negative (`!v.IsUnreliable()`) adds some unnecessary cognitive complexity.

jynnantonix (approved):
lgtm.  I had a couple of small nits but also fine with merging as is.

node/pkg/solana/client.go
```diff
@@ -534,6 +536,16 @@ func (s *SolanaWatcher) processMessageAccount(logger *zap.Logger, data []byte, a
...
 	copy(txHash[:], acc[:])
 
+	var reliable bool
+	switch string(data[:3]) {
+	case accountPrefixReliable:
+		reliable = true
+	case accountPrefixUnreliable:
+		reliable = false
+	default:
+		panic("invalid prefix")
```
bruce-riley, line 546: Do we really want to panic here, rather than just log an error and assume reliable? Are there upgrade scenarios where some guardians might see a new message type that they don't support yet?

bruce-riley:
I was just thinking about adding this flag to help with purging pyth messages from the db. Before we can do that though, we will need to persist this flag to the DB. (Obviously not part of this PR, but something to think about.)

Looks good!

bruce-riley (approved):

node/pkg/solana/client.go
hendrikhofstadt: we can safely panic here as we only reach this code after having checked this right now

bruce-riley (approved):

jynnantonix (approved):

PR #1630, opened 2022-09-21
WH Typescript SDK for Aptos

heyitaki:
This PR contains a preliminary implementation of a Typescript SDK for Wormhole actions on Aptos. Usage of the SDK is demonstrated in `sdk/js/src/aptos/example.ts`.

Next steps:
- Need to write tests for each SDK method
- Add additional endpoints to get read-only state from Aptos
- Documentation & examples
- Dynamically compute gas
- Eventually move on to a framework that will auto-generate SDK from Move contracts (like what is happening with EVM currently)


sdk/js/src/aptos/example.ts
```diff
@@ -0,0 +1,23 @@
+import { AptosAccount } from "aptos";
+import { AptosApi } from "./api";
+import { AptosClientWrapper } from "./client";
+import { AptosNetwork } from "./constants";
+import { hex } from "./utils";
+
```
optke3, line 6: Maybe we could add a `testnet` example as well, so end users don't have to learn how to spin up a local validator before testing things out. 

Publishing a message through core bridge or doing a token transfer might be useful examples.

optke3 (approved):
nice, lgtm!

sdk/js/src/aptos/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+  },
+  [AptosNetwork.Testnet]: {
+    token_bridge: "0xb4ec6ea1bff962721cc376be9a9aea840c485f9ceb10f31f523828fd6f4ca95a",
+    core: "0x25de93a587d5dc2d7e673663554b7e1d5b00de5d1d38341a896a2141bba5c5c9",
+  },
+  [AptosNetwork.Devnet]: {
+    token_bridge: "0x4450040bc7ea55def9182559ceffc0652d88541538b30a43477364f475f4a4ed",
+    core: "0x251011524cd0f76881f16e7c2d822f0c1c9510bfd2430ba24e1b3d52796df204",
+  },
+};
```
kcsongor, line 20: it would be great to move these to https://github.com/wormhole-foundation/wormhole/blob/6611ffff879c1d90114f6eb2ff594db743c98766/sdk/js/src/utils/consts.ts so everything is in one place within the sdk. The string enum you defined looks cool, we should add it to the consts file too I think.

sdk/js/src/aptos/client.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor(nodeUrl: string) {
+    this.client = new AptosClient(nodeUrl);
+  }
+
+  executeEntryFunction = async (
+    sender: AptosAccount,
+    payload: Types.EntryFunctionPayload,
+    opts?: Partial<Types.SubmitTransactionRequest>
+  ) => {
```
kcsongor, line 14: can you add return types explicitly? it makes the code a little easier to read outside of an IDE

sdk/js/src/aptos/api/tokenBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+      function: `${this.address}::transfer_tokens::submit_vaa`,
+      type_arguments: [`${assetContract}::coin::T`],
+      arguments: [amount, recipientChain, recipient, relayerFee, wormholeFee, nonce],
+    };
+    return this.client.executeEntryFunction(sender, payload);
+  };
+
+  // Created wrapped coin
+
+  createWrappedCoinType = (sender: AptosAccount, vaa: Uint8Array) => {
```
kcsongor, line 97: maybe `createWrappedCoinType` and `createWrappedCoin` could be done in a single function? You'd never want to do them separately anyway, the fact that they have to be done in two different transactions is an implementation detail of the contract

sdk/js/src/aptos/api/tokenBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+    tokenAddress: string,
+    amount: number | bigint,
+    recipientChain: number | bigint,
+    recipient: Uint8Array,
+    relayerFee: number | bigint,
+    wormholeFee: number | bigint,
+    nonce: number | bigint,
+  ) => {
+    assertChain(tokenChain);
+    const assetContract = deriveWrappedAssetAddress(this.address, tokenChain, tokenAddress);
```
kcsongor, line 86: currently this only handles wrapped assets, but it will need to handle native assets too. For native assets (i.e. when `tokenChain == aptos`), the `tokenAddress` field should be a fully qualified `address::module::type` string

sdk/js/src/aptos/api/tokenBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+      arguments: [vaa],
+    };
+    return this.client.executeEntryFunction(sender, payload);
+  };
+
+  // Transfer tokens
+
+  transferTokensWithSigner = (
+    sender: AptosAccount,
+    tokenChain: number,
```
kcsongor, line 76: this should be a `ChainId | ChainName` (from consts.ts)

sdk/js/src/aptos/api/tokenBridge.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  completeTransfer = (
+    sender: AptosAccount,
+    tokenChain: number,
+    tokenAddress: string,
+    vaa: Uint8Array,
+    feeRecipient: string,
+  ) => {
+    assertChain(tokenChain);
+    const assetContract = deriveWrappedAssetAddress(this.address, tokenChain, tokenAddress);
```
kcsongor, line 24: same as below, this should support both native and wrapped assets

kcsongor (changes requested):

sdk/js/src/aptos/client.ts
heyitaki: ahh yes I'll start doing this to accommodate my non-vscode brothers

kev1n-peters (changes requested):
Can we expose the aptos sdk functions like how we do for the other chains? Integrators are already familiar with this pattern. For example, it would be nice to have a `export async function transferTokenFromAptos` function exported in sdk/js/token_bridge/transfer.ts along with the other sdk functions. Open to discussion

sdk/js/src/aptos/api/tokenBridge.ts
heyitaki: good call

sdk/js/src/utils/consts.ts
```diff
@@ -316,8 +317,9 @@ const TESTNET = {
     nft_bridge: undefined,
   },
   aptos: {
-    core: undefined,
-    token_bridge: undefined,
+    core: "0x25de93a587d5dc2d7e673663554b7e1d5b00de5d1d38341a896a2141bba5c5c9",
```
kcsongor, line 320: let's not put the mainnet stuff in until it's actually deployed

sdk/js/src/utils/aptos.ts
```diff
@@ -0,0 +1,20 @@
+import { sha3_256 } from "js-sha3";
+import { ChainId, CHAIN_ID_APTOS, hex } from "../utils";
+
+export const deriveWrappedAssetAddress = (
```
kcsongor, line 4: looking at this function, should it now be called `deriveAssetAddress`? since it now supports non-wrapped assets too

sdk/js/src/aptos/api/tokenBridge.ts
```diff
@@ -0,0 +1,139 @@
...
+    nonce: number | bigint,
+  ): Promise<string> => {
+    const assetContract = deriveWrappedAssetAddress(
+      this.address,
+      coalesceChainId(tokenChain),
+      tokenAddress,
+    );
+    const payload = {
+      function: `${this.address}::transfer_tokens::submit_vaa`,
+      type_arguments: [`${assetContract}::coin::T`],
```
kcsongor, line 103: the `${address}::coin::T` type name is only guaranteed for wrapped coins. non-wrapped (i.e. native) coins may have an arbitrary module name (instead of `coin`) and an arbitrary type-name (instead of `T`), so they may look like `0xABCD::my_module::MyCoin`. This applies to `complete_transfer` too

sdk/js/src/utils/consts.ts
heyitaki: @kcsongor this was your change, taking it out

kev1n-peters (approved):

kcsongor (approved):

PR #1636, opened 2022-09-22
node: add command to purge pythnet VAAs

bruce-riley:
This adds a guardiand admin command to delete pythnet VAAs that are more than X days old. Although we are also working on avoiding writing pythnet messages to the db (a much better solution), this command can still be used to remove pythnet messages that have already been written.


node/cmd/guardiand/adminclient.go
```diff
@@ -137,6 +139,13 @@ var ClientChainGovernorResetReleaseTimerCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(1),
 }
 
+var PurgePythNetVaasCmd = &cobra.Command{
+	Use:   "purge-pythnet-vaas [DAYS_OLD] <logonly>",
+	Short: "Deletes PythNet VAAs from the database that are more than [DAYS_OLD] days only (if logonly is specified, doesn't delete anything)",
+	Run:   runPurgePythNetVaas,
+	Args:  cobra.RangeArgs(1, 2),
```
panoel, line 146: The answer to this is probably "no", but is there a way to specify that if there is only 1 arg, it must be the DAYS_OLD arg?

node/cmd/guardiand/adminclient.go
```diff
@@ -408,3 +417,43 @@ func runChainGovernorResetReleaseTimer(cmd *cobra.Command, args []string) {
 
 	fmt.Println(resp.Response)
 }
+
+func runPurgePythNetVaas(cmd *cobra.Command, args []string) {
+	daysOld, err := strconv.Atoi(args[0])
+	if err != nil {
```
panoel, line 423: This handles the case where someone puts in "logonly" and not the DAYS_OLD argument?

node/pkg/db/purge_vaas.go
```diff
@@ -0,0 +1,78 @@
...
+		it := txn.NewIterator(badger.DefaultIteratorOptions)
+		defer it.Close()
+		prefix := prefix.EmitterPrefixBytes()
+		for it.Seek(prefix); it.ValidForPrefix(prefix); it.Next() {
+			item := it.Item()
+			key := item.Key()
+			err := item.Value(func(val []byte) error {
+				v, err := vaa.Unmarshal(val)
+				if err != nil {
+					return fmt.Errorf("failed to unmarshal VAA for %s: %v", string(key), err)
```
panoel, line 33: Would you ever want to just delete what you can, continue instead of returning on the first error?  Either way there would be some deleted and some not?

panoel (approved):

node/pkg/db/purge_vaas.go
bruce-riley: I think it is okay to just throw an error and give up.

claudijd (dismissed):

SEJeff (approved):

SEJeff (approved):

panoel (approved):

SEJeff (approved):

PR #1637, opened 2022-09-22
algo/typo: Fix a typo and a round-off

jumpsiegel:
algo/typo: type fix, algorand tests, near tests

### DIFF
--- a/algorand/package.json
+++ b/algorand/package.json
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "@certusone/wormhole-sdk": "file:../sdk/js",
-    "algosdk": "^1.14.0",
+    "algosdk": "^1.21.0",
     "ts-node": "^10.7.0",
     "typescript": "^4.6.2",
     "varint": "^6.0.0",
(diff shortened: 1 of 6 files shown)


algorand/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "@certusone/wormhole-sdk": "file:../sdk/js",
-    "algosdk": "^1.14.0",
+    "algosdk": "^1.21.0",
```
panoel, line 4: Does this have any effect on sdk/js?  Like if tests pass here because 1.21.0 is being used, can same things fail in sdk/js because it is using older version?

algorand/test/tvl.ts
```diff
@@ -0,0 +1,77 @@
+const algosdk = require("algosdk");
+
+async function firstTransaction() {
+  // Connect your client
+  const algodToken = "";
+//  const algodServer = "http://localhost";
+  const algodServer = "https://mainnet-api.algonode.cloud";
```
panoel, line 7: Testing against mainnet and not testnet?

algorand/package.json
jumpsiegel: No, this should have no effect... but we might want to upgrade the algosdk used in our sdk...

bruce-riley (approved):

PR #1643, opened 2022-09-22
node: don't store pythnet VAAs in the database

bruce-riley:
This PR stores PythNet VAAs in memory rather than writing them to the guardiand database. It does this for ChainID == Pythnet rather than using the new IsUnreliable flag because that flag is not present in the SignedVAAWithQuorum gossip message (and changing that message might break various things).

Change-Id: Ief4357ab4c909d25dc9182490c322ef253ac23d3

### DIFF
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -73,20 +73,26 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 			// This occurs when we observed a message after the cluster has already reached
 			// consensus on it, causing us to never achieve quorum.
 			if ourVaa, ok := s.ourObservation.(*VAA); ok {
-				if _, err := p.db.GetSignedVAABytes(*db.VaaIDFromVAA(&ourVaa.VAA)); err == nil {
-					// If we have a stored quorum VAA, we can safely expire the state.
-					//
-					// This is a rare case, and we can safely expire the state, since we
-					// have a quorum VAA.
-					p.logger.Info("Expiring late VAA", zap.String("digest", hash), zap.Duration("delta", delta))
-					aggregationStateLate.Inc()
-					delete(p.state.signatures, hash)
-					continue
-				} else if err != db.ErrVAANotFound {
-					p.logger.Error("failed to look up VAA in database",
-						zap.String("digest", hash),
-						zap.Error(err),
-					)
+				if ourVaa.VAA.EmitterChain == vaa.ChainIDPythNet {
+					if p.deletePythNetVAA(&ourVaa.VAA) {
+						p.logger.Info("Expiring late pythnet VAA", zap.String("digest", hash), zap.Duration("delta", delta))
+					}
+				} else {
+					if _, err := p.db.GetSignedVAABytes(*db.VaaIDFromVAA(&ourVaa.VAA)); err == nil {
+						// If we have a stored quorum VAA, we can safely expire the state.
+						//
+						// This is a rare case, and we can safely expire the state, since we
+						// have a quorum VAA.
+						p.logger.Info("Expiring late VAA", zap.String("digest", hash), zap.Duration("delta", delta))
+						aggregationStateLate.Inc()
+						delete(p.state.signatures, hash)
+						continue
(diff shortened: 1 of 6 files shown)


node/pkg/processor/processor.go
```diff
@@ -216,3 +224,21 @@ func (p *Processor) Run(ctx context.Context) error {
...
+	key := fmt.Sprintf("%v/%v", v.EmitterAddress, v.Sequence)
+	ret, exists := p.pythnetVaas[key]
+	return ret.v, exists
+}
+
+func (p *Processor) deletePythNetVAA(v *vaa.VAA) bool {
+	key := fmt.Sprintf("%v/%v", v.EmitterAddress, v.Sequence)
+	_, exists := p.pythnetVaas[key]
+	delete(p.pythnetVaas, key)
+	return exists
```
jynnantonix, line 243: nit: we can skip the delete if the entry doesn't exist in the map:

```suggestion
	if _, exists := p.pythnetVaas[key]; exists {
		delete(p.pythnetVaas, key)
		return true
	}
	
	return false
```

node/pkg/publicrpc/publicrpcserver.go
```diff
@@ -68,6 +68,13 @@ func (s *PublicrpcServer) GetSignedVAA(ctx context.Context, req *publicrpcv1.Get
 		return nil, status.Error(codes.InvalidArgument, "no message ID specified")
 	}
 
+	chainID := vaa.ChainID(req.MessageId.EmitterChain.Number())
+
+	// This interface is not supported for PythNet messages because those VAAs are not stored in the database.
```
jynnantonix, line 73: Do we really need to treat this specially?  Getting the signed VAA for pythnet will return a "VAA not found" error, which seems fine?

node/pkg/processor/observation.go
```diff
@@ -286,30 +286,41 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(ctx context.Context, m *gos
...
-	} else if err != db.ErrVAANotFound {
-		p.logger.Error("failed to look up VAA in database",
-			zap.String("digest", hash),
-			zap.Error(err),
-		)
-		return
-	}
+	if v.EmitterChain == vaa.ChainIDPythNet {
+		_, exists := p.getPythNetVAA(v)
+		if exists {
```
jynnantonix, line 291: style nit: the usual convention in go is to combine these checks into a single line

```suggestion
		if _, exists := p.getPythNetVAA(v); exists {
```

See [Effective Go](https://go.dev/doc/effective_go#if) for more details.

node/pkg/processor/cleanup.go
```diff
@@ -73,20 +73,26 @@ func (p *Processor) handleCleanup(ctx context.Context) {
...
-					p.logger.Info("Expiring late VAA", zap.String("digest", hash), zap.Duration("delta", delta))
-					aggregationStateLate.Inc()
-					delete(p.state.signatures, hash)
-					continue
-				} else if err != db.ErrVAANotFound {
-					p.logger.Error("failed to look up VAA in database",
-						zap.String("digest", hash),
-						zap.Error(err),
-					)
+				if ourVaa.VAA.EmitterChain == vaa.ChainIDPythNet {
```
jynnantonix, line 76: I feel like adding an emitter chain check every place we handle a vaa is going to become a maintenance hassle.  We'll have to remember to add this check for every new place that deals with concrete vaas.

I think extending the `Observation` interface is a less brittle way to do this.  Basically, we would define a `PythnetVAA` type and then implement the `Observation` interface for it.  This will automatically get us type safety because any place that downcasts to the `*VAA` type directly (like here on line 75) will not catch `PythnetVAA`s.  We would implement `HandleQuorum` by storing the vaa in the `pythnetVaas` map, which would guarantee at the type level that pythnet vaas don't end up in the db.

In general, any place where we're doing an emitter chain check for pythnet should be a method on the observation interface instead (other than the initial place where the struct is created).

jynnantonix (approved):

hendrikhofstadt (approved):

PR #1650, opened 2022-09-26
algo/clawback: allow algorand clawback

jumpsiegel:


panoel (approved):

panoel:
Is there an NFT bridge?  If so, does the same change apply?

claudijd (approved):

PR #1651, opened 2022-09-26
terra: Add fromAddress to payload 3

kcsongor:
backport of #1292 for terra classic

### DIFF
--- a/terra/contracts/token-bridge/src/state.rs
+++ b/terra/contracts/token-bridge/src/state.rs
@@ -185,9 +185,9 @@ impl TokenBridgeMessage {
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct TransferInfo {
     pub amount: (u128, u128),
-    pub token_address: Vec<u8>,
+    pub token_address: [u8; 32],
     pub token_chain: u16,
-    pub recipient: Vec<u8>,
+    pub recipient: [u8; 32],
     pub recipient_chain: u16,
     pub fee: (u128, u128),
 }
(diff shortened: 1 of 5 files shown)


kev1n-peters (approved):

kev1n-peters (dismissed):

terra/contracts/token-bridge/src/state.rs
```diff
@@ -231,30 +231,72 @@ impl TransferInfo {
...
-    pub fn get_payload(data: &Vec<u8>) -> Vec<u8> {
-        return data[132..].to_vec();
+
+    pub fn get_payload(data: &[u8]) -> Vec<u8> {
+        data[132..].to_vec()
+    }
+
+    /// Convert [`TransferWithPayloadInfo`] into [`TransferInfo`] for the
+    /// purpose of handling them uniformly. Transfers with payload have 0 fees.
+    pub fn as_transfer_info(&self) -> TransferInfo {
```
jynnantonix, line 291: The usual convention for this is to implement `From` instead unless you need to borrow the original:

```rust
impl From<TransferWithPayloadInfo> for TransferInfo {
    fn from(tx: TransferWithPayloadInfo) -> Self {
        TransferInfo {
            amount: tx.amount,
            token_address: tx.token_address.clone(),
            token_chain: tx.token_chain,
            recipient: tx.recipient,
            recipient_chain: tx.recipient_chain,
            fee: (0, 0),
        }
    }
}
```

From what I can tell none of the places this is currently used need to borrow the original.  We would need to re-order the reply handler a bit:

```rust
let mut info = TransferWithPayloadInfo::deserialize(&token_bridge_message.payload)?;
let payload = (mem::take(&mut info.payload), info.sender_address);
Ok((info.into(), TransferType::WithPayload { payload }))
```

terra/contracts/token-bridge/src/state.rs
```diff
@@ -231,30 +231,72 @@ impl TransferInfo {
...
+            self.sender_address.to_vec(),
+            self.payload.clone(),
+        ]
+        .concat()
     }
-    pub fn get_payload(data: &Vec<u8>) -> Vec<u8> {
-        return data[132..].to_vec();
+
+    pub fn get_payload(data: &[u8]) -> Vec<u8> {
+        data[132..].to_vec()
```
jynnantonix, line 286: Unrelated but if we ever merge #1001, we can _borrow_ from the original data and avoid many unnecessary heap allocations.

jynnantonix (approved):

terra/contracts/token-bridge/src/state.rs
kcsongor: thanks -- I'll keep these in mind in the future, but won't implement for now so we can move forward with merging

kev1n-peters (approved):

PR #1655, opened 2022-09-27
Reisen/sdk solana

guibescos:
2 bug fixes :
- Make PostVAAData visible for crate users.
- Fix `post_vaa` (two variables had the same name)

Feature:
- Add `verify_signatures` instructions


hendrikhofstadt (changes requested):
Pls make linters and tests pass

Issue #1660, opened 2022-09-28
As an integrator, I want to get signed EVM messages at and earlier than finality


evan-gray:
## Expected Behavior

For chains which feature finality but not instant finality (Ethereum, Moonbeam, Polygon, etc), I should have the option to accept the risk of emitting a message _prior_ to finality. For chains that have explicit block statuses, e.g. how post-merge Ethereum has Unfinalized and Unfinalized (Safe), I should be able to specify at which status I want my message emitted. 

Furthermore, the default `postMessage` behavior should be to wait for finality and not longer. It doesn't make sense to wait for 1 block _after_ a finalized block. This would make explainer pages like https://book.wormhole.com/reference/contracts.html#blockchain-finality-recommendations only necessary for advanced integrators who want to go faster, into unsafe territory.

## Current Behavior

The EVM watcher puts all messages into a pending queue, which only releases messages after their block + consistency has been finalized. Even on the lowest consistency, 1, chains in which blocks are final require waiting until the _following_ final block.

## Possible Solution

Given that the existing EVM spec has `consistency` as a uint8, a 'magic number' could be specified to represent these special state exceptions for chains. For example:

- `200` = Unfinalized
- `201` = Unfinalized (Safe), only applicable to Ethereum

Then, when the watcher receives a message with the `consistency` set to `200`, it could instantly publish it instead of adding it to the pending queue. To handle achieving faster finality without unnecessarily waiting additional blocks, for any other values the watcher could use `minConfirmations` (and probably rename that `finalityConfirmations` or something). This could be set to 0 for all chains except for BSC which 15 is recommended. 

That leaves the only odd case of `201` on Ethereum. Since every epoch should yield a new final batch _and_ a safe batch, the watcher could leave the "Safe" number `201` which could be checked in the pending loop on new blocks and fall into a new, special path which uses `getBlock("safe")`.

tbjump:
I propose allocating the entire range of 200-255 (inclusive) of consistency_level for magic numbers and have the guardians not process any messages with consistency_level in that range where a magic number has not been allocated. (on all chains).
We should also check if there are any xApps who are actively using values in that range (I hope not).

tbjump:
What do you think of adding a special magic value (maybe zero) that is valid across all chains and would default to the Wormhole recommendations for ensured consensus specific to each chain, i.e. the values here: https://book.wormhole.com/reference/contracts.html#blockchain-finality-recommendations. 

[...]

PR #1664, opened 2022-09-29
node: eth publish immediately

bruce-riley:
Change-Id: I242f02d0ed5bcced5ed366a512c07d5757b300e8


evan-gray:
the "immediate" publish part of #1660's proposal

evan-gray (dismissed):

node/pkg/watchers/evm/watcher.go
```diff
@@ -314,6 +318,19 @@ func (w *Watcher) Run(ctx context.Context) error {
 				}
 
 				for _, msg := range msgs {
+					if msg.ConsistencyLevel == magicNumberPublishImmediately {
+						logger.Info("re-observed message publication transaction, publishing it immediately",
```
hendrikhofstadt, line 322: maybe use the same log message as below and add a new field for consistency level?

node/pkg/watchers/evm/watcher.go
```diff
@@ -392,6 +409,23 @@ func (w *Watcher) Run(ctx context.Context) error {
 					ConsistencyLevel: ev.ConsistencyLevel,
 				}
 
+				ethMessagesObserved.WithLabelValues(w.networkName).Inc()
+
+				if message.ConsistencyLevel == magicNumberPublishImmediately {
+					logger.Info("found new message publication transaction, publishing it immediately",
```
hendrikhofstadt, line 415: same here on the log message

hendrikhofstadt (approved):
Only reviewed the watcher part; looks good, small logging nit

node/pkg/watchers/evm/watcher.go
```diff
@@ -27,6 +27,10 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
+const (
+	magicNumberPublishImmediately = uint8(200)
```
jynnantonix, line 31: Can we please put constants like this in the sdk rather than having them spread out in various files?

hendrikhofstadt (dismissed):

node/pkg/watchers/evm/watcher.go
bruce-riley: I actually like having a different message that I can grep for, so I would like to keep this as is.

node/pkg/watchers/evm/watcher.go
bruce-riley: Done.

evan-gray (approved):

tbjump:
So with this change the eth watcher will still only process finalized blocks, correct? https://github.com/wormhole-foundation/wormhole/blob/2552e31b9b7fd11e5caa0621eb264759c382a84f/node/pkg/watchers/evm/connectors/poller.go#L161

tbjump (changes requested):
So with this change the eth watcher will still only process finalized blocks, correct? https://github.com/wormhole-foundation/wormhole/blob/2552e31b9b7fd11e5caa0621eb264759c382a84f/node/pkg/watchers/evm/connectors/poller.go#L157-L164

bruce-riley:
Correct. This change does not impact finality for eth.

tbjump (approved):
lgtm

PR #1668, opened 2022-09-30
Rust CI improvements

jynnantonix:
This PR enables rustfmt and clippy lints in CI for our rust crates.  It also enables the lints and tests for cosmwasm.


.github/workflows/build.yml
```diff
@@ -244,33 +244,57 @@ jobs:
...
           - sdk/rust/Cargo.toml
     steps:
       - name: Check out source
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
-      - name: Install stable rust toolchain
-        uses: actions-rs/toolchain@v1
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
```
kcsongor, line 260: I'm curious, does the upstream action now support the stable toolchain?
https://github.com/dtolnay/rust-toolchain/commit/84be2718c89917863ad106851300e80ecf0800f7 seems to only add stable as the default, but `actions-rs/toolchain` seems to support it as an additional flag: https://github.com/marketplace/actions/rust-toolchain#components

kcsongor (dismissed):
this is great 🎉 
I'm not a big fan of the way imports are formatted with the stable toolchain, but agree that formatting preferences shouldn't warrant using nightlies

.github/workflows/build.yml
jynnantonix: I switched because actions-rs is [unmaintained](https://github.com/actions-rs/toolchain/issues/216).

In classic dtolnay fashion, there is a clever solution for non-stable toolchains.  If you want a specific toolchain, you simply check out the rust-toolchain action at the same branch.  For example, for the 1.64.0 release I would use `rust-toolchain@1.64.0`, which has [this code](https://github.com/dtolnay/rust-toolchain/blob/1.64.0/action.yml#L8-L12) instead:

```yaml
inputs:
  toolchain:
    description: Rust toolchain specification -- see https://rust-lang.github.io/rustup/concepts/toolchains.html#toolchain-specification
    required: false
    default: 1.64.0
```

For nightly, we have to use `rust-toolchain@master` and then explicitly list the toolchain version as one of the inputs.  This is what we're currently doing for the solana CI action.

sdk/rust/core/src/vaa.rs
```diff
@@ -10,46 +10,24 @@
...
-    InvalidGovernanceChain,
-    InvalidGovernanceModule,
-};
-use crate::{
-    require,
-    Chain,
-    WormholeError,
+    InvalidGovernanceAction, InvalidGovernanceChain, InvalidGovernanceModule,
 };
+use crate::{require, Chain, WormholeError};
```
hendrikhofstadt, line 23: can't we group these?

sdk/rust/rustfmt.toml
```diff
@@ -1,18 +1,18 @@
 # Merge similar crates together to avoid multiple use statements.
-imports_granularity = "Module"
+# imports_granularity = "Module"
```
hendrikhofstadt, line 2: let's add a not on unstable features here as well

hendrikhofstadt (dismissed):

.github/workflows/build.yml
hendrikhofstadt: We might want to pin this to a hash for security purposes given this is an external dependency

jynnantonix: Done

sdk/rust/core/src/vaa.rs
jynnantonix: Done.

sdk/rust/rustfmt.toml
jynnantonix: Done.

jynnantonix:
* dev.v2
  * **PR #1668** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1668" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1668?utm_source=stack-comment).

hendrikhofstadt (approved):

kcsongor (approved):

PR #1669, opened 2022-09-30
Configurable private keys

JoeHowarth:


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "relayer/spy_relayer/xRaydium"]
```
evan-gray, line 1: Gitmodules? Was this intentional / is it necessary?

evan-gray (commented):
Could you elaborate on the collective purpose of the 39 commits? Maybe it could be rebased into a cleaner story. 🙂

PR #1670, opened 2022-09-30
node: poller timeout

bruce-riley:
The provider we are using for Arbitrum testnet limits the number of queries we can do. This was causing the poller connector to hang while calling eth_getBlockByNumber(). This PR adds a timeout to those calls so that the runnable will die and get restarted.

Change-Id: Ia324f1ac482fa9c5bea2b501970f0b22b16e67ce

### DIFF
--- a/node/pkg/watchers/evm/connectors/poller.go
+++ b/node/pkg/watchers/evm/connectors/poller.go
@@ -74,12 +74,15 @@ func (b *BlockPollConnector) run(ctx context.Context) error {
 }
 
 func (b *BlockPollConnector) pollBlocks(ctx context.Context, logger *zap.Logger, lastBlock *NewBlock) (lastPublishedBlock *NewBlock, retErr error) {
+	timeout, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
+
 	lastPublishedBlock = lastBlock
 
 	// Fetch the latest block on the chain
 	// We could do this on every iteration such that if a new block is created while this function is being executed,
 	// it would automatically fetch new blocks but in order to reduce API load this will be done on the next iteration.
-	latestBlock, err := b.getBlock(ctx, logger, nil)
+	latestBlock, err := b.getBlock(timeout, logger, nil)
 	if err != nil {
 		logger.Error("failed to look up latest block",
 			zap.Uint64("lastSeenBlock", lastBlock.Number.Uint64()), zap.Error(err))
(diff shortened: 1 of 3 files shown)


node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -74,12 +74,15 @@ func (b *BlockPollConnector) run(ctx context.Context) error {
 }
 
 func (b *BlockPollConnector) pollBlocks(ctx context.Context, logger *zap.Logger, lastBlock *NewBlock) (lastPublishedBlock *NewBlock, retErr error) {
+	timeout, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
```
jynnantonix, line 78: Are we concerned about any state that gets lost when we time out and the whole runnable is killed?  Would it be possible to rate limit on our end instead?  For example, we could buffer requests in a channel and have a separate goroutine send them out in a rate-limited manner.

node/pkg/watchers/evm/watcher.go
```diff
@@ -634,6 +634,7 @@ func (w *Watcher) Run(ctx context.Context) error {
 		}
 	}()
 
+	readiness.SetReady(w.readiness)
```
jynnantonix, line 637: Unrelated change?

jynnantonix (commented):
It would be nice to put the PR description in the commit as well so that it shows up in the git history.

node/pkg/watchers/evm/watcher.go
bruce-riley: Unrelated but related. When it can take a long time to reach finality (such as Arbitrum in testnet), it takes a long time for the runable to be marked ready, which causes tilt initialization to hang. This lets initialization continue sooner.

jynnantonix: Ok, then please add a comment here explaining that.

node/pkg/watchers/evm/connectors/poller.go
hendrikhofstadt: Agree, maybe we should be more graceful in L69 like the original version of the poller (pre my refactoring PR), allow X timeouts with some kind of backoff. This would automatically retry blocks that were missed.

bruce-riley: This change is being made to address an issue with Arbitrum in testnet because the provider we are using keeps disconnecting us. This should not happen in mainnet, and if it does, I feel it is reasonable for the runable to die. If we see the runable die in mainnet, we need to address the underlying cause, rather than preemptively code around the issue.

node/pkg/watchers/evm/watcher.go
bruce-riley: Added a comment.

node/pkg/watchers/evm/connectors/poller.go
jynnantonix: Ack.  I'm fine with leaving this as is and we can revisit if it becomes a problem in the future but please add a comment here explaining the situation for future readers who don't have / forgot the context.

jynnantonix (approved):

node/pkg/watchers/evm/connectors/poller.go
bruce-riley: Added a comment.

bruce-riley:
Added a couple of retries. Seems to help with Arbitrum testnet.

evan-gray (approved):

jynnantonix (approved):

PR #1672, opened 2022-10-02
tob-worm-4-panics: using options rather than unwrap & tob-worm-9: cosmwasm test build

kii-dot:
4. Use of panics on invalid inputs
Description In several places, the code panics when an arithmetic overflow or underflow occurs. Panics should be reserved for programmer errors (e.g., assertion violations). Panicking on user errors dilutes the utility of the panic operation. An example appears in figure 4.1. Thecomplete_nativefunction useschecked_subto verify that the transfer includes sufficient funds to cover the fee. However, complete_nativepanics if there are insufficient funds.Thus, the function could panic on either a programmer error or a user error.

Also added 9. cw20-wrapped tests do not build
**for tob-9**
Removed the arguments and ran cargo test. Works seamlessly

### DIFF
--- a/cosmwasm/contracts/token-bridge/src/contract.rs
+++ b/cosmwasm/contracts/token-bridge/src/contract.rs
@@ -924,7 +924,7 @@ fn handle_complete_transfer_token(
 
     amount = amount
         .checked_sub(fee)
-        .ok_or_else(|| StdError::generic_err("Insufficient funds"))?;
+        .ok_or(StdError::generic_err("Insufficient funds"))?;
 
     // Check high 128 bit of amount value to be empty
     if not_supported_amount != 0 || not_supported_fee != 0 {
@@ -1072,7 +1072,7 @@ fn handle_complete_transfer_token_native(
 
     amount = amount
         .checked_sub(fee)
-        .ok_or_else(|| StdError::generic_err("Insufficient funds"))?;
+        .ok_or(StdError::generic_err("Insufficient funds"))?;
 
     // Check high 128 bit of amount value to be empty
     if not_supported_amount != 0 || not_supported_fee != 0 {
@@ -1272,14 +1272,14 @@ fn handle_initiate_transfer_token(
                 .checked_rem(multiplier)
                 .and_then(|rem| amount.u128().checked_sub(rem))
                 .map(Uint128::new)
-                .ok_or_else(|| StdError::generic_err("Insufficient funds"))?;
+                .ok_or(StdError::generic_err("Insufficient funds"))?;
 
             fee = fee
                 .u128()
                 .checked_rem(multiplier)
                 .and_then(|rem| fee.u128().checked_sub(rem))
                 .map(Uint128::new)
-                .ok_or_else(|| StdError::generic_err("Invalid fee"))?;
+                .ok_or(StdError::generic_err("Invalid fee"))?;
 
             // This is a regular asset, transfer its balance
             submessages.push(SubMsg::reply_on_success(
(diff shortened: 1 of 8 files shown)


cosmwasm/contracts/token-bridge/src/contract.rs
```diff
@@ -872,7 +872,10 @@ fn handle_complete_transfer_token(
     let (not_supported_amount, mut amount) = transfer_info.amount;
     let (not_supported_fee, mut fee) = transfer_info.fee;
 
-    amount = amount.checked_sub(fee).unwrap();
+    amount = match amount.checked_sub(fee) {
+        None => return Result::Err(cosmwasm_std::StdError::generic_err("Insufficient funds")),
+        Some(value) => value
+    };
```
jynnantonix, line 878: style nit: I think `ok_or_else` is the more idiomatic way to do this:

```suggestion
    amount = amount
        .checked_sub(fee)
        .ok_or_else(|| StdError::generic_err("Insufficient funds"))?;
```

same for the other places where we do this.

cosmwasm/contracts/token-bridge/src/contract.rs
```diff
@@ -1208,16 +1214,22 @@ fn handle_initiate_transfer_token(
...
+                {
+                    None => return Result::Err(cosmwasm_std::StdError::generic_err("Insufficient Funds")),
+                    Some(value) => value
+                }
             );
 
             fee = Uint128::new(
-                fee.u128()
+                match fee.u128()
                     .checked_sub(fee.u128().checked_rem(multiplier).unwrap())
```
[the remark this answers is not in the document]
jynnantonix: I think we probably don't need checked math for div and rem operations because those can only fail when the right hand side is 0, which is never the case in the code so far.  We can still leave them in case we change the multiplier in the future.  If we want to make this more ergonomic we can do something like:

```rust
    fee = fee
        .u128()
        .checked_rem(multiplier)
        .and_then(|rem| fee.u128().checked_sub(rem))
        .map(Uint128::new)
        .ok_or_else(|| StdError::generic_err("Invalid fee"))?;
```

jynnantonix (commented):
Overall looks good to me but it does overlap a bit with  #1668.  I just have some style related suggestions.

solana/migration/src/api/add_liquidity.rs
```diff
@@ -97,11 +97,11 @@ pub fn add_liquidity(
     let share_amount = if accs.from_mint.decimals > accs.to_mint.decimals {
         data.amount
             .checked_mul(10u64.pow((accs.from_mint.decimals - accs.to_mint.decimals) as u32))
-            .unwrap()
+            .ok_or_else(|| Result::Err(SolitaireError::InsufficientFunds))?
```
jynnantonix, line 100: The return type for `ok_or_else` should be just the error type, not `Result<T, E>`.  Also there's no need for lazy evaluation here so we can use `ok_or` instead:

```suggestion
            .ok_or(SolitaireError::InsufficientFunds)?
```

jynnantonix (commented):
Please fix the compile errors as well as the formatting errors in the solana code.  You can test locally by running `make fmt`, `make test`, and `make clippy` in the solana directory.

jynnantonix (approved):

jynnantonix (dismissed):
Not sure if github will let you merge or if a committer needs to do it but please squash into a single commit when merging (instead of rebasing).

jynnantonix:
Ack.  Though we need one more approval before we can merge.

@tbjump @a5-pickle @kev1n-peters

hendrikhofstadt (approved):

hendrikhofstadt:
Great one!

evan-gray:
rebasing because tests were stalled and there was a ci test fix

evan-gray (approved):
not sure why @jynnantonix's review got dismissed on rebase through GH, but @hendrikhofstadt's didn't 🤔

PR #1510, opened 2022-09-02
wormchain: add wasmd module

conorpp:

PR #1674, opened 2022-10-03
Wormhole chain wasmd

conorpp:
Re-opening https://github.com/wormhole-foundation/wormhole/pull/1510, which had been reviewed by @jynnantonix and @hendrikhofstadt.

I added a small change to the wormchain genesis.

This adds wasmd to wormchain which will be used for accounting and btc bridge.


scripts/lint.sh
```diff
@@ -43,7 +43,7 @@ format(){
     fi
 
     # Use -exec because of pitfall #1 in http://mywiki.wooledge.org/BashPitfalls
-    GOFMT_OUTPUT="$(find "./sdk" "./node" "./event_database" "./wormhole_chain" -type f -name '*.go' -not -path './node/pkg/proto/*' -print0 | xargs -r -0 goimports $GOIMPORTS_ARGS 2>&1)"
+    GOFMT_OUTPUT="$(find "./sdk" "./node" "./event_database" "./wormhole_chain" -type f -name '*.go' -not -path './node/pkg/proto/*' -not -path '*.pb.go' -print0 | xargs -r -0 goimports $GOIMPORTS_ARGS 2>&1)"
```
jynnantonix, line 46: I think we can drop the `-not -path './node/pkg/proto'` part if we have a glob for all `*.pb.go` files.

wormhole_chain/testutil/keeper/wormhole.go
```diff
@@ -1,40 +1,136 @@
...
+	stateStore.MountStoreWithDB(keys[wasmtypes.StoreKey], sdk.StoreTypeIAVL, db)
+	stateStore.MountStoreWithDB(memKeys[types.MemStoreKey], sdk.StoreTypeMemory, nil)
+	stateStore.MountStoreWithDB(tkeys[paramstypes.TStoreKey], sdk.StoreTypeTransient, nil)
 	require.NoError(t, stateStore.LoadLatestVersion())
 
-	registry := codectypes.NewInterfaceRegistry()
+	encodingConfig := cosmoscmd.MakeEncodingConfig(app.ModuleBasics)
+	appCodec := encodingConfig.Marshaler
+	amino := encodingConfig.Amino
+	// txDecoder := encodingConfig.TxConfig.TxDecoder()
```
jynnantonix, line 59: nit: leftover comment

wormhole_chain/x/wormhole/client/cli/tx_wasmd.go
```diff
@@ -0,0 +1,153 @@
...
+				return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
+			} else {
+				hash_base := make([]byte, 8)
+				binary.BigEndian.PutUint64(hash_base, msg.CodeID)
+				hash_base = append(hash_base, []byte(msg.Label)...)
+				hash_base = append(hash_base, []byte(msg.Msg)...)
+				var hash [32]byte
+				keccak := sha3.NewLegacyKeccak256()
+				keccak.Write(hash_base)
+				keccak.Sum(hash[:0])
```
jynnantonix, line 142: `hash.Hash` implements `io.Write` so there's no need to build an array first:

```suggestion
				var hash [32]byte
				keccak := sha3.NewLegacyKeccak256()
				binary.Write(keccak, binary.BigEndian, msg.CodeID)
				keccak.Write([]byte(msg.Label))
				keccak.Write([]byte(msg.Msg))
				keccak.Sum(hash[:0])
```

wormhole_chain/x/wormhole/keeper/msg_server_wasmd.go
```diff
@@ -0,0 +1,129 @@
...
+	// The vaa governance payload must contain this hash.
+	hash_base := make([]byte, 8)
+	binary.BigEndian.PutUint64(hash_base, msg.CodeID)
+	hash_base = append(hash_base, []byte(msg.Label)...)
+	hash_base = append(hash_base, msg.Msg...)
+
+	var expected_hash [32]byte
+	keccak := sha3.NewLegacyKeccak256()
+	keccak.Write(hash_base)
+	keccak.Sum(expected_hash[:0])
```
jynnantonix, line 105: Same comment here about not needing to build an array first.  Maybe we can refactor this out into a helper function.

wormhole_chain/x/wormhole/keeper/msg_server_wasmd_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/tendermint/crypto/sha3"
+	keepertest "github.com/wormhole-foundation/wormhole-chain/testutil/keeper"
+	"github.com/wormhole-foundation/wormhole-chain/x/wormhole/keeper"
+	"github.com/wormhole-foundation/wormhole-chain/x/wormhole/types"
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+)
+
+var GOVERNANCE_EMITTER = [32]byte{00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 04}
+var GOVERNANCE_CHAIN = 1
+var WH_CHAIN_ID = 3104
```
jynnantonix, line 19: `vaa.ChainIDWormchain`?

wormhole_chain/x/wormhole/keeper/msg_server_wasmd_test.go
```diff
@@ -0,0 +1,236 @@
...
+	keccak.Sum(expected_hash[:0])
+
+	payload := []byte{}
+	payload = append(payload, keeper.WasmdModule[:]...)
+	payload = append(payload, byte(keeper.ActionInstantiateContract))
+	chain_bz := [2]byte{}
+	binary.BigEndian.PutUint16(chain_bz[:], uint16(WH_CHAIN_ID))
+	payload = append(payload, chain_bz[:]...)
+	// custom payload
+	payload = append(payload, expected_hash[:]...)
```
jynnantonix, line 53: nit: The conventional way to do this is to use a `bytes.Buffer`.

wormhole_chain/x/wormhole/client/cli/genesis.go
```diff
@@ -0,0 +1,447 @@
...
+const FLAG_PAYLOAD = "payload"
+
+// governance vaa flags
+const FLAG_MODULE = "module"
+const FLAG_ACTION = "action"
+const FLAG_CHAIN = "chain"
+const FLAG_PUBLIC_KEY = "public-key"
+const FLAG_NEXT_INDEX = "next-index"
+
+// GetTxCmd returns the transaction commands for this module
```
jynnantonix, line 43: "GetGenesisCmd"

wormhole_chain/x/wormhole/client/cli/genesis.go
```diff
@@ -0,0 +1,447 @@
...
+}
+
+func CmdDecodeAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "decode-address [address]",
+		Short: "Decode an address from either account, validator, or evm format",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			addrString := args[0]
+			if strings.Contains(addrString, sdk.GetConfig().GetBech32AccountAddrPrefix()) {
```
jynnantonix, line 104: `strings.HasPrefix`?

wormhole_chain/x/wormhole/client/cli/genesis.go
```diff
@@ -0,0 +1,447 @@
...
+				addr, err := sdk.AccAddressFromBech32(addrString)
+				if err != nil {
+					return nil
+				}
+				fmt.Println(base64.StdEncoding.EncodeToString(addr))
+			} else if strings.Contains(addrString, sdk.GetConfig().GetBech32ValidatorAddrPrefix()) {
```
jynnantonix, line 110: same here

wormhole_chain/x/wormhole/client/cli/genesis.go
```diff
@@ -0,0 +1,447 @@
...
+				addr, err := sdk.AccAddressFromBech32(addrString)
+				if err != nil {
+					return nil
+				}
+				fmt.Println(base64.StdEncoding.EncodeToString(addr))
+			} else {
+				// treat as hex
+				addrString = strings.Replace(addrString, "0x", "", 1)
+				addr, err := hex.DecodeString(addrString)
```
jynnantonix, line 119:
```suggestion
				addr, err := hex.DecodeString(strings.TrimPrefix(addrString, "0x"))
```

wormhole_chain/x/wormhole/client/cli/genesis.go
```diff
@@ -0,0 +1,447 @@
...
+				privateKeys = append(privateKeys, priv)
+			}
+			v, err := parseVaaFromFlags(cmd)
+			if err != nil {
+				return err
+			}
+			for i, key := range privateKeys {
+				v.AddSignature(key, uint8(i))
+				// address := crypto.PubkeyToAddress(privateKeys[0].PublicKey)
+				// fmt.Println("signed using ", hex.EncodeToString(address[:]))
```
jynnantonix, line 254: nit: leftover debugging lines

wormhole_chain/x/wormhole/client/cli/tx_register_account_as_guardian.go
```diff
@@ -31,6 +31,7 @@ func CmdRegisterAccountAsGuardian() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("malformed signature: %w", err)
 			}
+			fmt.Println("PUBKEY:", hex.EncodeToString(argGuardianPubkey.Key))
```
jynnantonix, line 34: nit: debugging line

jynnantonix (approved):
I know the cosmos convention is to check in generated proto files but do we want to switch to generating them at build time instead?  I think it would cut down on a lot of review noise and also remove the risk that someone forgets to update the generated code after changing a .proto file.

Otherwise lgtm % nits

conorpp:
> 

I'd like to do this.  I think what we'd loose is the ability for something else to add wormchain as a go dependency.  I'll add a commit trialing this and we can decide.

wormhole_chain/x/wormhole/keeper/msg_server_wasmd_test.go
```diff
@@ -34,34 +31,29 @@ func createWasmInstantiatePayload(code_id uint64, label string, json_msg string)
...
+	keccak.Write([]byte(json_msg))
 	keccak.Sum(expected_hash[:0])
 
-	payload := []byte{}
-	payload = append(payload, keeper.WasmdModule[:]...)
-	payload = append(payload, byte(keeper.ActionInstantiateContract))
-	chain_bz := [2]byte{}
-	binary.BigEndian.PutUint16(chain_bz[:], uint16(WH_CHAIN_ID))
-	payload = append(payload, chain_bz[:]...)
+	payload := bytes.NewBuffer(keeper.WasmdModule[:])
```
jynnantonix, line 41: The [documentation](https://pkg.go.dev/bytes#NewBuffer) for `bytes.NewBuffer` says:

> The new Buffer takes ownership of buf, and the caller should not use buf after this call.

So I think we need to make a copy here.  Or do something like:

```suggestion
	var payload bytes.Buffer
	payload.Write(keeper.WasmdModule[:])
```

jynnantonix (dismissed):

jynnantonix:
> > I know the cosmos convention is to check in generated proto files but do we want to switch to generating them at build time instead? I think it would cut down on a lot of review noise and also remove the risk that someone forgets to update the generated code after changing a .proto file.
> 
> I'd like to do this. I think what we'd loose is the ability for something else to add wormchain as a go dependency. I'll add a commit trialing this and we can decide.

This is a good point that I hadn't considered.  I think I still prefer leaving them out.  We can always add them back in when there is a use case that comes up.  

Or another option is that we keep the bindings checked in and add a CI check to generate fresh bindings every time and error if there is a diff between the fresh bindings and the ones in the PR.

wormhole_chain/x/wormhole/keeper/msg_server_wasmd_test.go
jynnantonix: Hmm, I wonder if committing the suggestion will dismiss the PR approval.

wormhole_chain/app/app.go
```diff
@@ -353,10 +373,11 @@ func New(
 	// Create Transfer Keepers
 	app.TransferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec, keys[ibctransfertypes.StoreKey], app.GetSubspace(ibctransfertypes.ModuleName),
-		app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
+		app.IBCKeeper.ChannelKeeper, app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
```
hendrikhofstadt, line 376: why is the keeper added multiple times?

wormhole_chain/x/wormhole/keeper/msg_server_wasmd_test.go
conorpp: it did :)

wormhole_chain/app/app.go
conorpp: The IBC transfer keeper splits up the required keepers/methods it wants into multiple interfaces, which the IBC Keeper implements both.

```
// ICS4Wrapper defines the expected ICS4Wrapper for middleware
type ICS4Wrapper interface {
	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
}

// ChannelKeeper defines the expected IBC channel keeper
type ChannelKeeper interface {
	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
	GetNextSequenceSend(ctx sdk.Context, portID, channelID string) (uint64, bool)
}
```

Terra does the same thing:
https://github.com/terra-money/core/blob/main/app/app.go#L408

conorpp:
> > > I know the cosmos convention is to check in generated proto files but do we want to switch to generating them at build time instead? I think it would cut down on a lot of review noise and also remove the risk that someone forgets to update the generated code after changing a .proto file.
> > 
> > 
> > I'd like to do this. I think what we'd loose is the ability for something else to add wormchain as a go dependency. I'll add a commit trialing this and we can decide.
> 
> This is a good point that I hadn't considered. I think I still prefer leaving them out. We can always add them back in when there is a use case that comes up.
> 
> Or another option is that we keep the bindings checked in and add a CI check to generate fresh bindings every time and error if there is a diff between the fresh bindings and the ones in the PR.

Okay let's leave them out until we need them, it will be easy to add them back.

jynnantonix (dismissed):

hendrikhofstadt (approved):

conorpp:
We can ignore this build error as it's no longer relevant:
https://github.com/wormhole-foundation/wormhole/actions/runs/3230577869/jobs/5289189424#step:6:1602

hendrikhofstadt:
The `tilt` tests still don't seem to compile the protos

jynnantonix (approved):

hendrikhofstadt (approved):

Issue #1679, opened 2022-10-04
Batch VAAs


evan-gray:
# Batch VAAs

## Overview

As an integrator who composes multiple Wormhole-using contracts, I want to receive an attestation which contains all of the messages emitted in a transaction. Messages could opt-out of a batch by setting a nonce of `0`.

## Current Behavior

The only way to compose multiple messages is to submit separate VAAs to the target contract in the same call. Each of these messages would need to be verified separately, adding prohibitive costs. This is also inconvenient to query for / collect from the guardians.

### Guardian

All messages will continue to produce a "payload 1" VAA. The guardian will collect messages with a non-zero nonce within the same transaction and produce a Batch VAA (payload 2).

### Smart Contract

(These notes are very EVM implementation specific.)

The Wormhole core contract will parse the Batch VAA into a new struct (VM2) and compute the batch hash which is the hash-of-hashes of each observation. The core contract will then verify the signatures against the batch hash. The calling contract may decide if they want to cache the hashes in the core contract state, which will allow for cheaper verification of individual messages in the batch later in the transaction.

The Headless VAAs within the batch will be parsed into the existing VM struct (version 3) while leaving the signatures and guardian set `null`. To verify these Headless VAAs the hash must be cached by the Wormhole core contract during batch signature verification.

The contract that verified the batch can clear the cached Headless VAA hashes from the core contract state after processing. This should be done atomically to reduce the gas costs.

### Finer Grain Batches

I may want to create multiple groups of these messages within the same transaction, which can be controlled by setting the nonce field.

### Partial Batches

Allow the redeemer to submit partial batches by adding an index-Headless VAA pair which will allow the core contract to verify each observation's hash against the array of hashes in the batch. The purpose of this is to further reduce gas usage, e.g. if not all messages in a batch are relevant to every chain being relayed to. The structure of the batch signed by the guardian would remain the same - this is solely a off-chain relayer function.

## Tracker

- [x] EVM Core Contracts to parse and verify Batch VAAs and Headless VAAs, update interfaces
- [x] #1857
- [x] #1893
- [x] #1891
- [ ] SDK parsing functions
- [ ] Book section with examples

PR #1314, opened 2022-06-27
WIP Batch VAAs

kcsongor:
---
This change is [Reviewable](https://reviewable.io/reviews/certusone/wormhole/1314)

PR #1680, opened 2022-10-04
evm: Batch VAA Implementation 

gator-boi:
The purpose of this PR is to add batch messaging (see #1679) to the EVM Wormhole core contracts. Batching messages allows for efficient verification of multiple messages (the batch only carries one set of guardian signatures) and makes delivering multiple messages to a target chain more convenient. 

There are four new endpoints being added to the Wormhole core contracts that need to be reviewed:
* `parseAndVerifyBatchVM`
* `verifyBatchVM`
* `parseBatchVM`
* `clearBatchCache`

This PR relies on [changes to the guardian](https://github.com/wormhole-foundation/wormhole/pull/1522) and should not be merged before these changes have been review and deployed.


evan-gray:
Part of https://github.com/wormhole-foundation/wormhole/issues/1679

hendrikhofstadt:
How does this relate to #1314 ?

gator-boi:
@hendrikhofstadt partial batch VAAs have been removed from this implementation. We are going to close 1314.

evan-gray (changes requested):
Why the sub module? I believe @kcsongor  has set up forge to avoid this

kcsongor:
> Why the sub module? I believe @kcsongor has set up forge to avoid this

yes, specifically set up the Makefile to avoid this, unfortunately there's currently (at least the last time I checked) no native way to get forge to behave as a conventional package manager, so the deps are managed by make: `make forge-test`.

ethereum/contracts/Getters.sol
```diff
@@ -53,4 +57,4 @@ contract Getters is State {
     function nextSequence(address emitter) public view returns (uint64) {
         return _state.sequences[emitter];
     }
-}
\ No newline at end of file
+}
```
evan-gray, line 60: Please configure your editor to insert a new line at the end of files for consistent diffs. e.g. https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

kcsongor: sometimes I wonder why this isn't the default in VS code 🤔

njkumr:
> Why the sub module? I believe @kcsongor has set up forge to avoid this

oops did something hacky to get forge working. Will remove the submodules and shift to make

ethereum/contracts/Getters.sol
evan-gray: Perhaps we should consider committing some vscode settings. Maybe after we do prettier config.

gator-boi: I believe the new files are adding that new line (that's the diff you see). Should we remove the new line to make the diffs consistent?

evan-gray: Ah, perhaps that’s it (can’t easily see from my phone 😛). But no, having them in there is the way.

kcsongor: Yeah we should add the newline at the end otherwise the last line is not a real line! (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

Some tools consider files with a missing newline at the end corrupted, though modern tools tend to just ignore it

justinschuldt:
I rebased this branch on `dev.v2` in order to make testing easier, I'm going to push my rebase.

calebrate (approved):

PR #1681, opened 2022-10-05
disable automatic re-observation requests

tbjump:
There is a bug with automatic re-observation requests leading to the governor having thousands of duplicate VAAs in the queue and eating up daily governor limits. This change is to temporarily disable automatic re-observations until we have time to identify and fix the root cause of this issue.


jynnantonix (approved):

evan-gray (changes requested):
I disagree with disabling automatic observations if the feature is working but Governor is bugged. The Governor should not enqueue duplicate hashes. If the hash was already signed it should allow it through, if the hash is already enqueued it should drop it. Does that sound right?

evan-gray:
Closing per the above comment.

PR #1682, opened 2022-10-05
Revert "node: governor listen to quorum gossip"

tbjump:
Reverts wormhole-foundation/wormhole#1487 until security review is complete.


jynnantonix (approved):

evan-gray (approved):
Approving, as I would like to see an issue that runs through the effective and philosophical difference this change (or the idea of the change) makes. The Governor feels less like a decentralized, trust-minimized security measure if guardians take other guardians’ approvals into account.

PR #1699, opened 2022-10-10
node: Arbitrum watcher fix

bruce-riley:
On Arbitrum we are unable to get blocks by transaction hash using the go-ethereum library because it fails with "transaction type not supported". However, calling the underlying eth_getBlockByHash directly works. This PR creates an Arbitrum specific connector that implements TimeOfBlockByHash using the raw connection.

### DIFF
--- a/node/pkg/watchers/evm/connectors/arbitrum.go
+++ b/node/pkg/watchers/evm/connectors/arbitrum.go
@@ -1,0 +1,42 @@
+// On Arbitrum we are unable to get blocks by transaction hash using the go-ethereum library
+// because it fails with "transaction type not supported". However, calling the underlying
+// eth_getBlockByHash directly works. The sole function of this connector is to implement
+// TimeOfBlockByHash using the raw connection.
+
+package connectors
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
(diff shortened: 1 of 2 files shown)


evan-gray (approved):

node/pkg/watchers/evm/connectors/arbitrum.go
```diff
@@ -0,0 +1,42 @@
...
+
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+type ArbitrumConnector struct {
+	Connector
+}
+
+func NewArbitrumConnector(ctx context.Context, baseConnector Connector) (*ArbitrumConnector, error) {
+	connector := &ArbitrumConnector{Connector: baseConnector}
```
hendrikhofstadt, line 21: nit: this doesn't need to return an error

hendrikhofstadt (approved):

node/pkg/watchers/evm/connectors/arbitrum.go
bruce-riley: True, but all of the other new connector methods do, so I would like to keep the interface standard.

Besides, who knows what might be added later.

evan-gray (approved):

PR #1701, opened 2022-10-10
Sui Core Bridge and Token Bridge Initial Implementations

optke3:
- Sui-Move core bridge and token bridge initial implementations
- Sui-Move unit tests
- Bash scripts for deployment and initialization


sui/token_bridge/sources/bridge_state.move
```diff
@@ -0,0 +1,113 @@
...
+      token_chain: U16,
+      token_address: ExternalAddress,
+   }
+
+   struct BridgeState {
+      governance_chain_id: U16,
+      governance_contract: ExternalAddress,
+
+      /// Set of consumed VAA hashes
+      consumed_vaas: VecSet<vector<u8>>,
```
kcsongor, line 33: for now this is fine, but we will have to consider a more efficient data structure here before the contracts go live. O(n) operations will become prohibitively expensive quickly

kcsongor (approved):

sui/token_bridge/sources/bridge_state.move
optke3: agree

PR #1707, opened 2022-10-11
Aptos WH SDK

heyitaki:
This PR extends the WH SDK to Aptos

Next steps:
- add e2e integrations tests
- documentation / examples in README

### DIFF
--- a/aptos/token_bridge/sources/transfer_tokens.move
+++ b/aptos/token_bridge/sources/transfer_tokens.move
@@ -17,7 +17,7 @@ module token_bridge::transfer_tokens {
 
     const E_TOO_MUCH_RELAYER_FEE: u64 = 0;
 
-    struct State has key {
+    struct EmitterCapabilityStore has key {
         emitter_cap: EmitterCapability,
     }
 
@@ -76,15 +76,20 @@ module token_bridge::transfer_tokens {
         wormhole_fee: u64,
         nonce: u64,
         payload: vector<u8>
-        ): u64 acquires State {
+    ): u64 acquires EmitterCapabilityStore {
+        if (!exists<EmitterCapabilityStore>(signer::address_of(sender))) {
+            let emitter_cap = register_emitter();
+            move_to<EmitterCapabilityStore>(
+                sender, 
+                EmitterCapabilityStore { emitter_cap: emitter_cap }
+            );
+        };
+        let emitter_cap = (
+            &mut borrow_global_mut<EmitterCapabilityStore>(signer::address_of(sender)).emitter_cap
+        );
+
         let coins = coin::withdraw<CoinType>(sender, amount);
         let wormhole_fee_coins = coin::withdraw<AptosCoin>(sender, wormhole_fee);
-
-        if (!exists<State>(signer::address_of(sender))) {
-            let emitter_cap = register_emitter();
-            move_to<State>(sender, State { emitter_cap: emitter_cap });
-        };
-        let emitter_cap = (&mut borrow_global_mut<State>(signer::address_of(sender)).emitter_cap);
         transfer_tokens_with_payload<CoinType>(
             emitter_cap,
             coins,
--- a/aptos/token_bridge/sources/transfer_tokens.move
+++ b/aptos/token_bridge/sources/transfer_tokens.move
@@ -1,10 +1,12 @@
 module token_bridge::transfer_tokens {
     use aptos_framework::aptos_coin::{AptosCoin};
     use aptos_framework::coin::{Self, Coin};
+    use aptos_framework::signer;
 
     use wormhole::u16::{Self, U16};
     use wormhole::external_address::{Self, ExternalAddress};
     use wormhole::emitter::{Self, EmitterCapability};
+    use wormhole::wormhole::{register_emitter};
 
     use token_bridge::state;
     use token_bridge::transfer;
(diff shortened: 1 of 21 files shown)


aptos/token_bridge/sources/transfer_tokens.move
```diff
@@ -62,6 +68,34 @@ module token_bridge::transfer_tokens {
...
+        nonce: u64,
+        payload: vector<u8>
+        ): u64 acquires State {
+        let coins = coin::withdraw<CoinType>(sender, amount);
+        let wormhole_fee_coins = coin::withdraw<AptosCoin>(sender, wormhole_fee);
+
+        if (!exists<State>(signer::address_of(sender))) {
+            let emitter_cap = register_emitter();
+            move_to<State>(sender, State { emitter_cap: emitter_cap });
+        };
```
jumpsiegel, line 86: I really wish there was another way to do these emitters... this emitter is going to be 110% meaningless to anybody else on the planet.  More importantly, it is going to be difficult to know exactly what your emitter address is until after you send your first message.  Maybe we need a specific API to potentially assign a emitter and return its value?

jumpsiegel (approved):

jumpsiegel:
Do we have any interest in having the API's return the txn's pre-signed and breaking the sign/send/wait functionality out?  We ended up having to do this for Near and for Algorand

heyitaki:
> Do we have any interest in having the API's return the txn's pre-signed and breaking the sign/send/wait functionality out? We ended up having to do this for Near and for Algorand

What's the benefit of this? @jumpsiegel

aptos/token_bridge/sources/transfer_tokens.move
```diff
@@ -15,6 +17,10 @@ module token_bridge::transfer_tokens {
 
     const E_TOO_MUCH_RELAYER_FEE: u64 = 0;
 
+    struct State has key {
```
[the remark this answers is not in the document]
kcsongor: `EmitterCapabilityStore` is a good name for this

sdk/js/src/token_bridge/getIsTransferCompleted.ts
```diff
@@ -209,3 +198,30 @@ export async function getIsTransferCompletedNear(
...
+  const signedVAAHash = await getSignedVAAHash(signedVAA);
+  try {
+    // when accessing Set<T>, key is type T and value is 0
+    await client.getTableItem(handle, {
+      key_type: "vector<u8>",
+      value_type: "u8",
+      key: signedVAAHash,
+    });
+    return true;
+  } catch {
```
kcsongor, line 224: can we be a little more specific here about what specific exception this handles? if it's not the "table key missing" one, then I think the exception should be bubbled up instead of silently swallowed

sdk/js/src/token_bridge/getIsWrappedAsset.ts
```diff
@@ -92,3 +94,19 @@ export function getIsWrappedAssetNear(
...
+  client: AptosClient,
+  tokenBridgeAddress: string,
+  assetAddress: string,
+): Promise<boolean> {
+  assetAddress = ensureHexPrefix(assetAddress);
+  try {
+    // get origin info from asset address
+    await client.getAccountResource(assetAddress, `${tokenBridgeAddress}::state::OriginInfo`);
+    return true;
+  } catch {
```
kcsongor, line 109: same here, catch the precise error

sdk/js/src/token_bridge/getOriginalAsset.ts
```diff
@@ -234,3 +238,32 @@ export async function getOriginalAssetNear(
 
   return retVal;
 }
+
+export async function getOriginalAssetAptos(
+  client: AptosClient,
+  tokenBridgeAddress: string,
+  assetAddress: string,
+): Promise<WormholeWrappedInfo> {
```
kcsongor, line 246: `WormholeWrappedInfo` sounds like a misnomer, given that it also classifies native assets

aptos/token_bridge/sources/transfer_tokens.move
```diff
@@ -62,6 +68,39 @@ module token_bridge::transfer_tokens {
...
+    public entry fun transfer_tokens_with_payload_with_signer<CoinType>(
+        sender: &signer,
+        amount: u64,
+        recipient_chain: u64,
+        recipient: vector<u8>,
+        wormhole_fee: u64,
+        nonce: u64,
+        payload: vector<u8>
+    ): u64 acquires EmitterCapabilityStore {
+        if (!exists<EmitterCapabilityStore>(signer::address_of(sender))) {
```
kcsongor, line 80: it might be nicer if we lifted this out to a separate function like `get_emitter_cap_signer` (or something like that) that creates one if and moves it to the signer if it doesn't exist yet.

aptos/token_bridge/sources/transfer_tokens.move
```diff
@@ -15,6 +17,10 @@ module token_bridge::transfer_tokens {
 
     const E_TOO_MUCH_RELAYER_FEE: u64 = 0;
 
+    struct EmitterCapabilityStore has key {
```
kcsongor, line 20: This is nice! We should move it to `emitter.move` instead though, and document that there are two ways of getting an emitter capability.
1) use your signer and then it gets stored for you (advantage: easier if you have signer; downside: when calling form a contract, you need a resource account which is a ton of incurred complexity)
2) don't use a signer and you get back the capability directly (advantage: no signer needed, very easy to set up; downside: you have to manage the emitter capability directly)

evan-gray:
Closing in favor of #1736

PR #1715, opened 2022-10-12
cosmwasm: support natives over 8 decimals

evan-gray:
Previously, the cosmwasm contracts only had to handle Terra Classic or Terra 2, both of which natives have 6 decimals. But with more cosmwasm chains on the horizon (like XPLA) with larger decimal sizes (18 in the [axpla case](https://docs.xpla.io/docs/develop/core-modules/#conventions)), the code needs to handle normalization to 8 decimals on the way out and in.

### DIFF
--- a/cosmwasm/contracts/token-bridge/src/contract.rs
+++ b/cosmwasm/contracts/token-bridge/src/contract.rs
@@ -135,50 +135,50 @@ pub enum TransferType<A> {
 /// Ok(Response::default())
 /// ```
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
-    // This migration adds a new field to the [`ConfigInfo`] struct. The
-    // state stored on chain has the old version, so we first parse it as
-    // [`ConfigInfoLegacy`], then add the new fields, and write it back as [`ConfigInfo`].
-    // Since the only place the contract with the legacy state is deployed is
-    // terra2, we just hardcode the new value here for that chain.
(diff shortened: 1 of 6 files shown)


evan-gray:
@kcsongor @jynnantonix I'm switching this to draft as I didn't get a chance to test this tonight and I can't confirm the hash of the deployed code ID on Terra 2 to know if the migration is appropriate (though the prior migration was deleted so I guess that means it was done 🤷)

Would greatly appreciate ideas, tests, reassurance, etc - and hope to try it for real in testnet tomorrow. 🙏

cosmwasm/contracts/token-bridge/src/contract.rs
```diff
@@ -1378,6 +1428,27 @@ fn handle_initiate_transfer_native_token(
...
+    let decimals = cfg.native_decimals;
+    let multiplier = 10u128.pow((max(decimals, 8u8) - 8u8) as u32);
+
+    // chop off dust
+    amount = Uint128::new(
+        amount
+            .u128()
+            .checked_sub(amount.u128().checked_rem(multiplier).unwrap())
+            .unwrap(),
+    );
```
kcsongor, line 1440: might be a little cleaner like this:
```suggestion
  amount = amount - amount % Uint128::new(multiplier);
```
`amount` is of type `UInt128` so its operations are all checked (e.g. https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/math/uint128.rs#L331-L341) and the error messages are handled better than what `unwrap()` gives. In any case, the rest of the code looks like this too, so for now we can just use this version and make a note of doing a refactoring round later.

kcsongor (approved):

kev1n-peters (dismissed):

kev1n-peters (approved):

bruce-riley (approved):

PR #1726, opened 2022-10-13
Script changes for acala upgrade

bruce-riley:


kcsongor (changes requested):
I think we should move away from the `deploy-*-only` pattern as it abuses the migration logic of truffle. We already have implementation deploy scripts for all of the contracts here: https://github.com/wormhole-foundation/wormhole/tree/dev.v2/ethereum/scripts, which can be run by eg
```
npx truffle scripts/deploy_core_bridge.js --network ethereum
```
so I think we should just use those and remove all of the `-implementation-only` migrations.
(FWIW the https://github.com/wormhole-foundation/wormhole/blob/dev.v2/ethereum/upgrade script does this under the hood. Not super well documented, but it can be used to deploy a contract by running `./upgrade mainnet TokenBridge ethereum`. I want to document it better and add a usage string etc.)

bruce-riley:
Removed the new deploy scripts

kcsongor (dismissed):

ckeun (dismissed):

ckeun (approved):

panoel (dismissed):

panoel (approved):

panoel (approved):

ckeun (approved):

PR #1727, opened 2022-10-13
Node: Support for EVM watcher safe blocks

bruce-riley:
This adds the ability for the EVM watcher to listen for safe blocks in addition to finalized blocks. If a message has a consistency level of 201, it will wait for the safe block rather than the finalized one.

This feature is currently only enabled for Ethereum.

### DIFF
--- a/node/pkg/watchers/evm/connectors/poller.go
+++ b/node/pkg/watchers/evm/connectors/poller.go
@@ -23,19 +23,21 @@ type PollFinalizer interface {
 // finalizer which will be used to only return finalized blocks on subscriptions.
 type BlockPollConnector struct {
 	Connector
-	Delay        time.Duration
-	useFinalized bool
-	finalizer    PollFinalizer
-	blockFeed    ethEvent.Feed
-	errFeed      ethEvent.Feed
+	Delay             time.Duration
+	useFinalized      bool
+	publishSafeBlocks bool
+	finalizer         PollFinalizer
+	blockFeed         ethEvent.Feed
+	errFeed           ethEvent.Feed
 }
 
-func NewBlockPollConnector(ctx context.Context, baseConnector Connector, finalizer PollFinalizer, delay time.Duration, useFinalized bool) (*BlockPollConnector, error) {
+func NewBlockPollConnector(ctx context.Context, baseConnector Connector, finalizer PollFinalizer, delay time.Duration, useFinalized bool, publishSafeBlocks bool) (*BlockPollConnector, error) {
 	connector := &BlockPollConnector{
-		Connector:    baseConnector,
-		Delay:        delay,
-		useFinalized: useFinalized,
-		finalizer:    finalizer,
+		Connector:         baseConnector,
+		Delay:             delay,
+		useFinalized:      useFinalized,
+		publishSafeBlocks: publishSafeBlocks,
+		finalizer:         finalizer,
 	}
 	err := supervisor.Run(ctx, "blockPoller", connector.run)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


evan-gray:
Relates to #1660

bruce-riley:
Fixes #1660

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -164,17 +187,21 @@ func (b *BlockPollConnector) SubscribeForBlocks(ctx context.Context, sink chan<-
...
 
 // getBlock is a free function that can be called from other connectors to get a single block.
-func getBlock(ctx context.Context, logger *zap.Logger, conn Connector, number *big.Int, useFinalized bool) (*NewBlock, error) {
+func getBlock(ctx context.Context, logger *zap.Logger, conn Connector, number *big.Int, useFinalized bool, safe bool) (*NewBlock, error) {
 	var numStr string
 	if number != nil {
 		numStr = ethHexUtils.EncodeBig(number)
 	} else if useFinalized {
-		numStr = "finalized"
+		if safe {
```
panoel, line 200: useFinalized and safe are not mutually exclusive.  useFinalized must be true to get safe mode.  Is that by design?

bruce-riley: Yes. Safe only makes sense with finalized. I can add a comment.

panoel (approved):

panoel (dismissed):

node/pkg/watchers/evm/watcher.go
```diff
@@ -574,12 +617,24 @@ func (w *Watcher) Run(ctx context.Context) error {
...
+					// If it's not safe, only process messages wanting finalized.
+					if safeBlocksSupported {
+						if ev.Safe != (pLock.message.ConsistencyLevel == vaa.ConsistencyLevelSafe) {
+							continue
+						}
+					}
+
 					var expectedConfirmations uint64
-					if w.waitForConfirmations {
+					if w.waitForConfirmations && !ev.Safe {
```
evan-gray, line 637: Not sure I understand why the change to this check is needed. In fact, It may not make sense to allow `waitForConfirmations` to be set when `useFinalized` is true (and hence the only case where `Safe` could be true). Had to do some heavy thonking on this one though and I haven't finished my coffee, so I'm open to thoughts.

evan-gray (approved):
Looks like a decently surgical change and I'm excited to have safe support!

node/pkg/watchers/evm/watcher.go
bruce-riley: We definitely don't want to add the consistency level if we are processing a message requesting safe (since adding 201 would make no sense). Technically the check may not be necessary due to the implied relationship you mentioned, but I think I would like to leave the check just to be safe / clear.

tbjump:
Is the plan to eventually also add polling for "latest", corresponding to `vaa.ConsistencyLevelPublishImmediately`? In that case, it may be better to start handling the various consistency states as an array instead of adding additional variables? 

With all these special cases the code is getting increasingly difficult to read. Could we simplify some of the if/else cases or documenting the state machine and decision trees somewhere?

hendrikhofstadt (approved):

hendrikhofstadt:
LGTM, agree with tobi that there is more generalization potential for the different states especially considering we might want to add more in the future.

bruce-riley:
@tbjump `vaa.ConsistencyLevelPublishImmediately` is already implemented, and is not dependent on latest. That flag means to publish immediately, without waiting for any blocks.

tbjump:
@bruce-riley that makes sense then. Thanks for pointing that out. In that case I think there won't be any more states on the currently supported evm chains and no need to generalize the code.

PR #1744, opened 2022-10-17
tilt: shorten terra block time

hendrikhofstadt:
Reducing terra block times helps tilt startup times as transactions are sent block by block.

I intentionally did not set `skip_timeout_commit` to `true` as this would create blocks as fast as possible and `create_empty_blocks = false` is not an option as the ABCI apphash changes every block on terra which forces block creation.


evan-gray (approved):
Excited for faster times. 🐎

kcsongor (approved):

PR #1746, opened 2022-10-17
node: optimism watcher

bruce-riley:


node/pkg/watchers/evm/finalizers/optimism.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// f.logger.Info("got rollup info", zap.String("eth_network", f.connector.NetworkName()),
+	// 	zap.String("mode", m.Mode),
+	// 	zap.Uint64("l1_blockNumber", m.EthContext.BlockNumber),
+	// 	zap.Uint64("l2_blockNumber", m.RollupContext.Index),
+	// 	zap.Uint64("verified_index", m.RollupContext.VerifiedIndex),
+	// 	zap.Stringer("desired_block", block.Number),
+	// )
+
+	return block.Number.Uint64() <= m.RollupContext.Index, nil
```
panoel, line 55: Placeholder:  I'm not sure this is sufficient.

panoel (dismissed):

node/cmd/guardiand/node.go
```diff
@@ -631,6 +644,13 @@ func runNode(cmd *cobra.Command, args []string) {
 		} else if *arbitrumContract != "" {
 			logger.Fatal("If --arbitrumContract is specified, then --arbitrumRPC is required")
 		}
+		if *optimismRPC != "" {
+			if *optimismContract == "" {
+				logger.Fatal("If --optimismRPC is specified, then --optimismContract is required")
+			}
+		} else if *optimismContract != "" {
+			logger.Fatal("If --optimismContract is specified, then --optimismRPC is required")
+		}
```
jynnantonix, line 653: minor nit: sadly go doesn't have boolean xors but we can make this a single check by doing

```suggestion
		if *optimismRPC == "" != *optimismContract == "" {
			logger.Fatal("Both --optimismContract and --optimismRPC must be set together or both unset")
		} 
```

node/pkg/watchers/evm/finalizers/optimism.go
```diff
@@ -0,0 +1,72 @@
...
+		return false, err
+	}
+
+	// f.logger.Info("got rollup info", zap.String("eth_network", f.connector.NetworkName()),
+	// 	zap.String("mode", m.Mode),
+	// 	zap.Uint64("l1_blockNumber", m.EthContext.BlockNumber),
+	// 	zap.Uint64("l2_blockNumber", m.RollupContext.Index),
+	// 	zap.Uint64("verified_index", m.RollupContext.VerifiedIndex),
+	// 	zap.Stringer("desired_block", block.Number),
+	// )
```
jynnantonix, line 53: nit: leftover debugging lines

node/pkg/watchers/evm/finalizers/optimism.go
```diff
@@ -0,0 +1,72 @@
...
+		"syncing":false,
+		"ethContext":{
+			"blockNumber":7763392,"timestamp":1665680949 // This is a few blocks behind the latest block on goerli.
+			},
+		"rollupContext":{
+			"index":1952690,"queueIndex":13285,"verifiedIndex":0 // This is a few blocks behind the latest block on optimism.
+		}
+	}
+}
+*/
```
jynnantonix, line 72: same here

node/pkg/watchers/evm/finalizers/optimism.go
jynnantonix: Agreed.  Reading [How Optimism Works](https://community.optimism.io/docs/how-optimism-works/#fault-proofs), it says:

> If a proposed state commitment goes unchallenged for the duration of the challenge window (currently set to 7 days), then it is considered final. Once a commitment is considered final, smart contracts on Ethereum can safely accept proofs about the state of Optimism based on that commitment.

So I think we'd first need to wait for the Eth block to be finalized and then wait another 7 days for the Optimism block to be considered finalized before we can release the VAA.  Since the latency is so bad I assume we'll also need to provide an opt-in mechanism for integrators who want their VAAs faster at the risk of the block being rolled back (basically the equivalent of the consistency level 200 thing we did for Eth).

jynnantonix (commented):
The code itself looks good to me but we should figure out what we want to do for finality here.

evan-gray:
@jynnantonix I recall a discussion with @hendrikhofstadt about not needing to wait to full 7 days if the sequencer has published the transaction and your running node matches. See this section about fault detection https://github.com/smartcontracts/simple-optimism-node/#fault-detector. But totally agree I would like many minds to be pretty certain of this 😁

node/pkg/watchers/evm/finalizers/optimism.go
bruce-riley: Yes, keeping for now.

node/pkg/watchers/evm/finalizers/optimism.go
bruce-riley: Useful documentation. Keeping.

node/pkg/watchers/evm/finalizers/optimism.go
hendrikhofstadt: As Evan pointed out: This is true for on-chain anchoring of the L2. Given the optimistic nature the sequencer can submit incorrect new state roots to the root L1 and therefore emit forged messages. This is why there is the dispute period for people to challenge the *execution* results. However the sequence of transactions is finalized (as long as the L1 does not roll back) as that is fully maintained by the L1 SC CanonicalTransactionChain (CTC).
While executing all state transitions (txs) on the L1 would defeat the purpose of the L2, Wormhole guardians can run "full-nodes" that execute the txs in the CTC and are able to derive events and the new state root. Instead of the optimistic model where consensus on the view of the system is achieved by someone proposing the correct result and others raising disputes, wormhole achieves consensus using a MultiSig / majority consensus and achieve "instant finality" on the state root. This means that even if the L2 sequencer misbehaves and an incorrect state root is submitted on-chain, guardian full-nodes that are replaying the CTC would see the correct view to which the on-chain contract would also converge post dispute-resolultion (as long as there is not a bug in the node software).
All that Wormhole needs to check in addition is that the L1 is unlikely to roll back the transaction ordering before a message is confirmed.
The [Fault Detector](https://github.com/smartcontracts/simple-optimism-node/#fault-detector) piece that @evan-gray pointed out might still be helpful as as the L2 enters into dispute resolution mode unexpected issues might occur. This adapter might halt block finalization if the fault detector triggers as a precaution.

jynnantonix: So your description makes sense to me but the thing I'm having difficulty with is understanding the exact order in which the different components interact with each other in the wormhole context.  I _think_ this is what's happening:

1. The Optimism sequencer publishes blocks to the CTC on Eth.
1. Guardians run full nodes that download the blocks from the sequencer and then replay the execution, performing the relevant state transitions and emitting events.
1. guardiand connects to the guardian full node (via `NewEthereumConnector`) and listens for events from the wormhole contract.
1. Since each guardian trusts their own full node, the VAA has "instant finality" and so can be published immediately.

Is this correct?  At what step do we wait for ETH finality?  Does each full node wait for finality before replaying the L2 block execution (step 2) or is guardiand supposed to wait for ETH finality (step 4)?

Regardless of the answer here I think this interaction is sufficiently complex that it really needs to be documented alongside the actual code.

hendrikhofstadt: Yes I think your understanding is correct. The wormhole guardian should verify that the ETH block the CTC was submitted on for a certain block is finalized on ETH (which from what I can see is not happening here yet)

jumpsiegel (approved):

panoel:
Finality discussion:

The proposal is to have the watcher use Optimism's rollup_getInfo function as defined:
rollup_getInfo
Returns useful L2-specific information about the current node.
Object

mode: STRING - "sequencer" or "verifier" depending on the node's mode of operation
syncing: BOOLEAN - true if the node is currently syncing, false otherwise
ethContext: OBJECT
blockNumber: QUANTITY - Block number of the latest known L1 block
timestamp: QUANTITY - Timestamp of the latest known L1 block
rollupContext: OBJECT
queueIndex: QUANTITY - Index within the CTC of the last L1 to L2 message ingested
index: QUANTITY - Index of the last L2 tx processed
verifiedIndex: QUANTITY - Index of the last tx that was ingested from a batch that was posted to L1

In the rollup context we can use the verifiedIndex to see which was the last L2 block to be posted to the L1.
In the ethContext we can grab the block number to see which was the last L1 block known to the L2.
With this information, we can reason that the verified index was in an L1 block <= ethContext::blockNumber.
If we wait for finality on that L1 block, we should be reasonably confident that the L2 was finalized.

Comments?

panoel (dismissed):

node/pkg/watchers/evm/finalizers/optimism.go
```diff
@@ -0,0 +1,142 @@
...
+				isFinalized = true
+				f.logger.Info("found in finalizerMapping", zap.Uint64("L2", myBlockNumber))
+			}
+			// Only prune up to the L2 passed in, which may be different than what we looked at in the array
+			if entry.l2Block == myBlockNumber {
+				pruneIdx = idx
+			}
+			break
+		} else {
+			pruneIdx = idx
```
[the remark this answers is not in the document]
jynnantonix: The correctness of this function should not depend on implementation details in some other part of the program (especially when those assumptions are not documented in the code).  So whatever is happening in the rest of guardiand is not really relevant here.  See also: [action at a distance](https://en.wikipedia.org/wiki/Action_at_a_distance_(computer_programming))

Also, your example shows the case where everything goes right but my point is that there still exist scenarios in which the current implementation will return the incorrect result.

* We call IsBlockFinalized for block N, rollup_getInfo() returns N with ETH block M.  ETH block M is finalized so we return true and also, because of lines 96-99 we will prune the returned info:
```go
			// Only prune up to the L2 passed in, which may be different than what we looked at in the array
			if entry.l2Block == myBlockNumber {
				pruneIdx = idx
			}
```
* Now we call IsBlockFinalized for block N-100 and this time rollup_getInfo() returns N+10 with ETH block M+10.  If ETH block M+10 is not finalized, this function will incorrectly return false because we only return true in the case that `finalizedL1Block >= entry.l1Block` (lines 92-95).


Basically, it seems like there is an assumption in the pruning code that IsBlockFinalized will never be called for older blocks.  I am just pointing out that this is not the case and that we still need to store *some* information from the pruned entries.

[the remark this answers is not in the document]
panoel: This watcher code is ONLY used by the poller.  The poller is guaranteed to only move forward.  It will never move backward.  This code is NEVER used for re-observations.  But even if the poller had a hiccup where a lower number block was used, this code would still work and that block would eventually be identified as finalized.  I believe that based on this context, this code is reasonable and correct.  This code is not meant to be used as a general purpose library function.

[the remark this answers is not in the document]
jynnantonix: Everything you said is true today but is not actually documented anywhere.  Someone might start using the PollFinalizer interface in the future for something else and may not realize that the correctness of this function in a completely different part of the codebase depends on the input parameter monotonically increasing.  This kind of undocumented inter-dependence is very hard to debug when things go wrong.

I don't think it's unreasonable to ask that this function should work correctly independently of the input parameters.  The code change is not that large and I even provided a code snippet that does most of the rework already.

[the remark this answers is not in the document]
hendrikhofstadt: I agree we should design individual components to work indepentently of being implicitly coupled to any other component.

node/pkg/watchers/evm/finalizers/optimism.go
panoel: The ETH block is being checked for finality.

node/pkg/watchers/evm/finalizers/optimism.go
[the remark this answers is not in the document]
panoel: Added code to keep the latestFinalizedL2Block

jynnantonix (approved):

PR #1747, opened 2022-10-17
Node: karura/acala watcher use finalized blocks

bruce-riley:
This requires a new RPC endpoint, so should not be merged until that is deployed by the guardians.


panoel (dismissed):

node/pkg/watchers/evm/watcher.go
```diff
@@ -151,9 +152,15 @@ func NewEthWatcher(
 func (w *Watcher) Run(ctx context.Context) error {
 	logger := supervisor.Logger(ctx)
 
+	useFinalizedBlocks := (w.chainID == vaa.ChainIDEthereum && (!w.unsafeDevMode))
 	if w.shouldCheckSafeMode {
 		if err := w.checkForSafeMode(ctx); err != nil {
-			return err
+			if !strings.Contains(err.Error(), "is not using safe mode") {
```
jynnantonix, line 158: Since `checkForSafeMode` is a function we control I think we should have it return `(bool, error)` rather than returning the value in-band.

node/pkg/watchers/evm/watcher.go
```diff
@@ -151,9 +152,15 @@ func NewEthWatcher(
...
+				return err
+			}
+
+			// If the endpoint is not in safe mode, assume it is the version that supports finalized blocks.
```
jynnantonix, line 162: Should this be reversed?  Generally we should fall back to the safe option unless explicitly told otherwise.

panoel (dismissed):

node/pkg/watchers/evm/watcher.go
bruce-riley: Changed.

node/pkg/watchers/evm/watcher.go
bruce-riley: No, in this case "safe mode" refers to connecting to an additional Karura / Acala RPC layer that only publishes finalized blocks. That was a special layer they created for us. Now they want to deprecate it because their official RPC will now support finalized blocks. Once that is deployed by all the guardians, we can remove this check.

jynnantonix: That's kind of my point though.  If we're not using the "safe mode" endpoints then we're assuming that the endpoint supports finalized blocks.  This is a dangerous assumption if the endpoint doesn't actually support finalized blocks.

Instead we should be querying the endpoint to see if it has finalized block support and fall back to the "safe mode" endpoint if it doesn't.  Is there a version number check or something similar we can do to query the endpoint?

panoel (approved):

node/pkg/watchers/evm/watcher.go
```diff
@@ -133,28 +133,33 @@ func NewEthWatcher(
...
 }
 
 func (w *Watcher) Run(ctx context.Context) error {
 	logger := supervisor.Logger(ctx)
 
-	if w.shouldCheckSafeMode {
-		if err := w.checkForSafeMode(ctx); err != nil {
+	useFinalizedBlocks := (w.chainID == vaa.ChainIDEthereum && (!w.unsafeDevMode))
+	if (w.chainID == vaa.ChainIDKarura || w.chainID == vaa.ChainIDAcala) && (!w.unsafeDevMode) {
+		ufb, err := w.getAcalaMode(ctx)
```
evan-gray, line 155: go semantics make me twitch sometimes

evan-gray (approved):
looks clean

PR #1748, opened 2022-10-17
scripts: uppercase digest in governance generator

kcsongor:
the injection command prints out in this format, so it's easier to read for guardians this way


SEJeff (approved):
Awesome, thanks!

evan-gray (approved):

PR #1754, opened 2022-10-19
cloud_functions: Added Aptos and Xpla support

ckeun:
Xpla doesn't currently have coingecko support but still added xpla to allowlist assuming cgid will be xpla. 😬 
Do we want to use native address for aptos or is hex rep okay?


kev1n-peters (approved):

evan-gray (approved):
Deployed and working 👍

PR #1755, opened 2022-10-19
node: Added gov gossip support to p2p

ckeun:
Trying to do something with the gov cfg & status msgs instead of throwing them in the trash.


hendrikhofstadt:
What are we planning to do with the information (expose via admin rpc?).

ckeun:
> i think the idea was to use these msgs to create api endpoints for projects to be able to hit and know what the current gov limits are and what available notional exists for transfers

bruce-riley (approved):

evan-gray:
> What are we planning to do with the information (expose via admin rpc?).

One use is for the fly from https://github.com/wormhole-foundation/wormhole-explorer which will use this to provide a unified view of guardian limits.

PR #1774, opened 2022-10-20
ci: unscoped package check robustness improvements

tbjump:
This change tries to improve the previous bash script by
1) supporting spaces and other special characters in file paths
2) using jq instead of a simple grep to parse the json
3) being more restrictive in the regexes on what to allow to pass through

Would like to get the feedback from the bash wiz though

### DIFF
--- a/scripts/check-docker-pin.sh
+++ b/scripts/check-docker-pin.sh
@@ -11,7 +11,10 @@
 #   - We ignore scratch because it's literally the docker base image
 #   - We ignore solana AS (builder|ci_tests) because it's a relative reference to another FROM call
 #
-git ls-files | grep "Dockerfile*" | xargs grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos AS (builder|ci_tests|tests)'
+# original command:
+# git ls-files | grep "Dockerfile*" | xargs grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos AS (builder|ci_tests|tests)'
+# update (works with spaces in directories but includes files outside of 'git ls-files':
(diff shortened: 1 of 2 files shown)


evan-gray (changes requested):
Should this be different from the dockerfile check? It was a copy paste of that.

tbjump:
> Should this be different from the dockerfile check? It was a copy paste of that.

I changed it in `./scripts/check-docker-pin.sh` as well.

evan-gray (changes requested):
The docker check changed recently, can you please resolve merge conflicts.

scripts/check-docker-pin.sh
```diff
@@ -11,7 +11,10 @@
 #   - We ignore scratch because it's literally the docker base image
 #   - We ignore solana AS (builder|ci_tests) because it's a relative reference to another FROM call
 #
-git ls-files | grep "Dockerfile*" | xargs grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos AS (builder|ci_tests|tests)'
+# original command:
+# git ls-files | grep "Dockerfile*" | xargs grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos AS (builder|ci_tests|tests)'
+# update (works with spaces in directories but includes files outside of 'git ls-files':
+find . -type f -name 'Dockerfile*' -print0 | xargs -r -0 grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos AS (builder|ci_tests|tests)'
```
kcsongor, line 17: why change from `git ls-files`? `find` is much slower and will show false positives (when run locally, in the presence of build artifacts and testing stuff)

PR #1673, opened 2022-10-03
node: NEAR watcher rewrite

tbjump:
Please see README.md for the approach here. 

The objective of this rewrite is to 
1. add unit tests
2. make the code more readable
3. make it more robust to various error conditions
4. report errors and performance metrics appropriately 

Looking for high-level feedback on the approach and if you think its good: a line-by-line review

Jeff, please take a look at the logging and performance monitoring approach.

PR #1777, opened 2022-10-21
WH-1776: rewrite of near watcher to address finality concerns

jumpsiegel:


jumpsiegel:
I am assuming after 120 blocks, a block is final...    I could do the behavior of walking forward and looking back

tbjump:
I also addressed the same issues in https://github.com/wormhole-foundation/wormhole/pull/1673. Please take a look and let me know your feedback on that PR.

Issue #1764, opened 2022-10-20
Refactor validation logic into VAA for better reuse.


gabzim:
## Expected Behavior

I'm trying to validate a VAA off-chain. While the VAA offers a `VerifySignatures(guardianSetKeys)` method, this only checks for signatures as the name implies, and I see that the `Processor#handleInboundSignedVAAWithQuorum` (in `node/pkg/processor/observation.go`) performs additional checks like reaching quorum and that the VAA has len(signatures) > 0. 

## Current Behavior

You need to remember to copy/paste those checks if you want to be sure your VAA is valid in a different project that leverages the node's code to connect to the gossip network.

## Possible Solution

Maybe we can either export a top level function or add a method to the VAA struct that performs this extended validation? It would need to receive the guardianset and the quorum int (or the guardianset could calculate its quorum instead of the `CalculateQuorum` being in the `processor` package).

## Steps to Reproduce

N/A

## Context (Environment)

Building an indexer for the wormhole ecosystem.

panoel:
This issue was fixed by PR #1779

PR #1779, opened 2022-10-21
sdk/vaa: add verify function to vaa

panoel:
This addresses issue https://github.com/wormhole-foundation/wormhole/issues/1764
The relevant pieces of node/pkg/processor/observation.go::handleInboundSignedVAAWithQuorum() were pulled into
sdk/vaa/structs.go:Verify()

Downside:  CalculateQuorum() was copied from node/pkg/processor/quorum.go to sdk/vaa/structs.go.  Should be a better way than copying functions.

Resolution:
Moving CalculateQuorum() to a new file in the sdk and making sure test run correctly.

### DIFF
--- a/node/pkg/processor/observation.go
+++ b/node/pkg/processor/observation.go
@@ -192,7 +192,7 @@ func (p *Processor) handleObservation(ctx context.Context, m *gossipv1.SignedObs
 		// We have made this observation on chain!
 
 		// 2/3+ majority required for VAA to be valid - wait until we have quorum to submit VAA.
-		quorum := CalculateQuorum(len(gs.Keys))
+		quorum := vaa.CalculateQuorum(len(gs.Keys))
 
 		p.logger.Info("aggregation state for observation",
 			zap.String("digest", hash),
@@ -247,36 +247,8 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(ctx context.Context, m *gos
 		return
 	}
 
-	// Check if VAA doesn't have any signatures
-	if len(v.Signatures) == 0 {
-		p.logger.Warn("received SignedVAAWithQuorum message with no VAA signatures",
-			zap.String("digest", hash),
-			zap.Any("message", m),
-			zap.Any("vaa", v),
-		)
-		return
-	}
-
-	// Verify VAA has enough signatures for quorum
-	quorum := CalculateQuorum(len(p.gs.Keys))
-	if len(v.Signatures) < quorum {
-		p.logger.Warn("received SignedVAAWithQuorum message without quorum",
-			zap.String("digest", hash),
-			zap.Any("message", m),
-			zap.Any("vaa", v),
-			zap.Int("wanted_sigs", quorum),
-			zap.Int("got_sigs", len(v.Signatures)),
-		)
-		return
-	}
-
-	// Verify VAA signatures to prevent a DoS attack on our local store.
-	if !v.VerifySignatures(p.gs.Keys) {
-		p.logger.Warn("received SignedVAAWithQuorum message with invalid VAA signatures",
-			zap.String("digest", hash),
-			zap.Any("message", m),
-			zap.Any("vaa", v),
-		)
+	if err := v.Verify(p.gs.Keys); err != nil {
+		p.logger.Warn("dropping SignedVAAWithQuorum message because it failed verification: " + err.Error())
 		return
 	}
 
--- a/sdk/vaa/structs.go
+++ b/sdk/vaa/structs.go
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"math/big"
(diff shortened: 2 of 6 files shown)


sdk/vaa/structs.go
```diff
@@ -438,6 +438,33 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
...
+
+	// Verify VAA signatures to prevent a DoS attack on our local store.
+	if !v.VerifySignatures(addresses) {
+		return false
+	}
+
+	return true
+}
+
+func CalculateQuorum(numGuardians int) int {
```
bruce-riley, line 465: Are you going to make the processor code use these routines from the sdk? Otherwise we are going to have the quorum calculation in multiple places, which I don't think is good.

panoel: Fixed.

sdk/vaa/structs.go
```diff
@@ -438,6 +438,37 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
...
+	}
+
+	return true
+}
+
+// CalculateQuorum returns the minimum number of guardians that need to sign a VAA for a given guardian set.
+//
+// The canonical source is the calculation in the contracts (solana/bridge/src/processor.rs and
+// ethereum/contracts/Wormhole.sol), and this needs to match the implementation in the contracts.
+func CalculateQuorum(numGuardians int) int {
```
panoel, line 469: Move this function to its own file/package.

sdk/vaa/structs.go
```diff
@@ -438,6 +438,37 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
 
 	return true
 }
+func (v *VAA) Verify(addresses []common.Address) bool {
```
gabzim, line 441: The only problem is that if you return `bool` instead of `error`, we can't log or show what the issue was with the `vaa`, which may be ok, I don't know if it's necessary to be so explicit.

node/pkg/processor/observation.go
```diff
@@ -192,7 +192,7 @@ func (p *Processor) handleObservation(ctx context.Context, m *gossipv1.SignedObs
 		// We have made this observation on chain!
 
 		// 2/3+ majority required for VAA to be valid - wait until we have quorum to submit VAA.
-		quorum := CalculateQuorum(len(gs.Keys))
+		quorum := vaa.CalculateQuorum(len(gs.Keys))
```
gabzim, line 195: Maybe you can use the new `verify` function so the logic is not repeated here and we only update it there?

evan-gray (changes requested):
Agree with @gabzim ’s comments that we should return an error (as is a little more go) and use the new method back in the guardian. Also looks like lints need to pass 🙂

sdk/vaa/structs.go
panoel: It is in its own file in the vaa directory/package

sdk/vaa/structs.go
panoel: Now, returning err

node/pkg/processor/observation.go
panoel: I don't think we have a VAA here.

node/pkg/processor/observation.go
```diff
@@ -247,36 +247,10 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(ctx context.Context, m *gos
...
-
-	// Verify VAA signatures to prevent a DoS attack on our local store.
-	if !v.VerifySignatures(p.gs.Keys) {
-		p.logger.Warn("received SignedVAAWithQuorum message with invalid VAA signatures",
-			zap.String("digest", hash),
-			zap.Any("message", m),
-			zap.Any("vaa", v),
-		)
+	vError := v.Verify(p.gs.Keys)
+	if vError != nil {
```
jynnantonix, line 251: This error variable is `vError` but we end up logging `err` on line 253.  Let's stick with the go convention of always calling the error `err`.  We can also combine the verification and error check into a single line:

```suggestion
	if err := v.Verify(p.gs.Keys); err != nil {
```

jynnantonix (approved):
lgtm with nit

node/pkg/processor/observation.go
panoel: Fixed

gabzim (dismissed):

bruce-riley (dismissed):

node/pkg/processor/observation.go
```diff
@@ -247,36 +247,8 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(ctx context.Context, m *gos
...
-
-	// Verify VAA signatures to prevent a DoS attack on our local store.
-	if !v.VerifySignatures(p.gs.Keys) {
-		p.logger.Warn("received SignedVAAWithQuorum message with invalid VAA signatures",
-			zap.String("digest", hash),
-			zap.Any("message", m),
-			zap.Any("vaa", v),
-		)
+	if err := v.Verify(p.gs.Keys); err != nil {
+		p.logger.Warn("dropping SignedVAAWithQuorum message because it failed verification: " + err.Error())
```
jynnantonix, line 251: This is fine too but there is a [`zap.Error()`](https://pkg.go.dev/go.uber.org/zap#Error) function for logging errors.

jynnantonix (approved):

node/pkg/processor/observation.go
panoel: I couldn't get that to work and it was trying to import a new zap package.  So, I left it out for simplicity.

hendrikhofstadt (approved):

sdk/vaa/structs.go
```diff
@@ -438,6 +439,29 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
 
 	return true
 }
+func (v *VAA) Verify(addresses []common.Address) error {
```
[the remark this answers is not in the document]
panoel: Added

sdk/vaa/structs.go
```diff
@@ -444,6 +444,8 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
 
 	return true
 }
+
+// Takes the guardian set keys and checks this VAA for signatures and quorum
```
hendrikhofstadt, line 448: Being super nit-picky here but let's make sure we follow to godoc guidelines: https://go.dev/blog/godoc

This means comments on functions and structs should start with the name of the func or struct being described.

`Verify takes the guardian set keys and checks this VAA for signatures and quorum`

Also let's be a little more specific in what it actually does. Does it check for signatures or verify the signatures given the provided guardian set.

E.g. `Verify verifies the signatures on the VAA using the provided guardian set keys and checks for quorum. It returns an error if any verification step fails.`

hendrikhofstadt (changes requested):

bruce-riley (approved):

sdk/vaa/structs.go
panoel: Correct go documentation attempted.

bruce-riley (approved):

hendrikhofstadt (approved):

evan-gray (approved):

panoel:
Fixes #1764

PR #1780, opened 2022-10-21
worm: speedup

kcsongor:
the startup time is now ~300ms (from 2-3s before), which is a lot more acceptable

### DIFF
--- a/clients/js/main.ts
+++ b/clients/js/main.ts
@@ -11,7 +11,7 @@
 // drop that particular message...
 // </sigh>
 const info = console.info;
-console.info = function (x) {
+console.info = function(x: string) {
   if (x != "secp256k1 unavailable, reverting to browser version") {
     info(x);
   }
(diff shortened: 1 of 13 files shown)


jumpsiegel (approved):

evan-gray (approved):
I have a need for speed.

clients/js/main.ts
```diff
@@ -22,48 +22,14 @@ import yargs from "yargs";
...
-} from "@certusone/wormhole-sdk";
 import { ethers } from "ethers";
 import { NETWORKS } from "./networks";
 import base58 from "bs58";
-import { execute_algorand } from "./algorand";
-import { execute_injective } from "./injective";
-import { execute_xpla } from "./xpla";
 import { isOutdated } from "./cmds/update";
+import { setDefaultWasm } from "@certusone/wormhole-sdk/lib/cjs/solana/wasm";
+import { assertChain, assertEVMChain, ChainName, CHAINS, CONTRACTS as SDK_CONTRACTS, isCosmWasmChain, isEVMChain, isTerraChain, toChainId, toChainName } from "@certusone/wormhole-sdk/lib/cjs/utils/consts";
```
panoel, line 32: Prettier will probably change this line the next time someone saves this file.  Can you run prettier on this file?

panoel (approved):

clients/js/main.ts
kcsongor: ah drat, I missed this. since we have the approvals now, I'd rather merge it in and sort out prettier stuff when it's mandatory in CI

kcsongor: no, turns out I didn't miss it, it's just that the auto-formatter I have enabled by default (`tsfmt`) keeps it this way.

PR #1781, opened 2022-10-21
Rename wormchain

conorpp:
We have inconsistent naming for wormchain: `wormholechain`, `wormhole-chain`, and `wormchain`.  `wormchain` is definitely the name I've heard everyone use and just wanted to make it consistent while we still can.


SEJeff (commented):
hero

clients/js/main.ts
```diff
@@ -883,8 +883,9 @@ yargs(hideBin(process.argv))
           argv["contract-address"],
           argv["rpc"]
         );
-      } else if (chain === "wormholechain") {
-        throw Error("Wormhole Chain is not supported yet");
+      } else if (chain == "wormholechain" || (chain+"") === "wormchain") {
```
jynnantonix, line 886: javascript lol

jynnantonix (dismissed):

hendrikhofstadt (approved):
@evan-gray tilt fails on this because it does not see the SDK updates. I assume we can merge this and the cut a release / import that one?

conorpp:
I think we should merge and update to a new release of the js sdk.  I believe most places have backwards compat, the only spot that is breaking currently is the clients/js/main.ts tool.

kev1n-peters (approved):

evan-gray:
Overriding on the assurances of cutting an sdk release and fixing up the client with it. One day we'll switch to [Lerna](https://lerna.js.org/)

PR #1784, opened 2022-10-24
Node: update governor limits for Moonbeam

bruce-riley:
Moonbeam limits are exhausted daily and now that multiple DEXs are accepting wrapped assets from chains with $5m notional limits, it would be keen to up the Moonbeam limits to match.


tbjump:
Makes sense to me, but let's give the other guardians ~2 days to chime in here as well before we change the reference config.

tbjump (approved):
lgtm

evan-gray (approved):
lgtm, haven't received any opposition

PR #1788, opened 2022-10-25
Aptos SDK documentation

heyitaki:
- Added documentation to Aptos SDK: docstrings + surfacing some design decisions
- Minor API changes (use fully qualified types instead of module addresses where it makes sense)

### DIFF
--- a/sdk/js/src/token_bridge/getForeignAsset.ts
+++ b/sdk/js/src/token_bridge/getForeignAsset.ts
@@ -14,8 +14,8 @@ import {
   ChainName,
   CHAIN_ID_ALGORAND,
   coalesceChainId,
-  ensureHexPrefix,
-  getForeignAssetAddress,
+  getAssetFullyQualifiedType,
+  coalesceModuleAddress,
 } from "../utils";
 import { Provider } from "near-api-js/lib/providers";
 import { LCDClient as XplaLCDClient } from "@xpla/xpla.js";
@@ -192,12 +192,12 @@ export async function getForeignAssetNear(
 }
 
 /**
- * Get native module address of asset given its origin info.
+ * Get qualified type of asset on Aptos given its origin info.
  * @param client Client used to transfer data to/from Aptos node
  * @param tokenBridgeAddress Address of token bridge
  * @param originChain Chain ID of chain asset is originally from
  * @param originAddress Asset address on origin chain
- * @returns Asset module address on Aptos
+ * @returns Fully qualified type of asset on Aptos
  */
 export async function getForeignAssetAptos(
   client: AptosClient,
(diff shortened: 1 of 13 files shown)


sdk/js/src/token_bridge/getForeignAsset.ts
```diff
@@ -191,22 +191,37 @@ export async function getForeignAssetNear(
...
 
   try {
     // check if asset exists and throw if it doesn't
-    await client.getAccountResource(assetAddress, `0x1::coin::CoinInfo<${ensureHexPrefix(assetAddress)}::coin::T>`);
-    return assetAddress;
+    await client.getAccountResource(
+      getModuleAddressFromFullyQualifiedType(assetFullyQualifiedType),
+      `0x1::coin::CoinInfo<${assetFullyQualifiedType}>`
+    );
+    return assetFullyQualifiedType;
```
evan-gray, line 224: This is a subtle but breaking change for integrators, no? Is there a reason this change is recommended over what was working?

sdk/js/src/token_bridge/getIsWrappedAsset.ts
```diff
@@ -116,16 +121,25 @@ export function getIsWrappedAssetNear(
...
   tokenBridgeAddress: string,
-  assetAddress: string,
+  assetFullyQualifiedType: string
 ): Promise<boolean> {
-  assetAddress = ensureHexPrefix(assetAddress);
+  assetFullyQualifiedType = ensureHexPrefix(assetFullyQualifiedType);
   try {
     // get origin info from asset address
-    await client.getAccountResource(assetAddress, `${tokenBridgeAddress}::state::OriginInfo`);
+    await client.getAccountResource(
```
evan-gray, line 139: Same comment. It doesn’t seem unreasonable to know the address from a type in Aptos.

evan-gray (changes requested):
Great to have the comments, though I’m not sure if I agree with the breaking changes.

sdk/js/src/token_bridge/getIsWrappedAsset.ts
evan-gray: Actually, this seems backwards compatible since all it’s doing is a split.

sdk/js/src/token_bridge/getForeignAsset.ts
evan-gray: Discussed and this feels right, so we'll minor version bump. @heyitaki can you update the CHANGELOG and package.json to reflect this?

sdk/js/src/utils/aptos.ts
```diff
@@ -193,3 +229,44 @@ export async function getTypeFromExternalAddress(
     return null;
   }
 }
+
+export const getModuleAddressFromFullyQualifiedType = (
```
evan-gray, line 233: discussed and we can change this to "coalesce" and remove the `isValidAptosType` requirement

sdk/js/src/token_bridge/getForeignAsset.ts
heyitaki: done

evan-gray (changes requested):
Changelog entry, pls

kev1n-peters (approved):

evan-gray (approved):

PR #1798, opened 2022-10-26
Node: Arbitrum watcher redesign

bruce-riley:
This PR redesigns the Arbitrum finalizer to not consider an L2 block complete until the L1 block containing it is finalized on Ethereum. This involves adding a getter to the EthWatcher so the Arbitrum finalizer can access the latest finalized block on Ethereum. If you have a better way to do this, please let me know.


panoel (approved):

hendrikhofstadt (approved):
Generally this seems a bit hacky but I don't think there's a clean way to solve this without some large-scale refactoring of the root watcher.

evan-gray (changes requested):
I may be missing something, but how is this anything other than waiting some blocks on Eth. It doesn't seem to actually add to the validity of the Arbitrum message as it doesn't check to ensure that the batch remains on Eth after the wait.

hendrikhofstadt (dismissed):

panoel (approved):

evan-gray (approved):
Discussed with @bruce-riley and @hendrikhofstadt, while there is an alternative of something like 
```
GetL1Confirmations(arbitrumBlockHash) >= ethWatcher.getLatest - ethWatcher.getFinalized
```
there's still a possible desync issue between the connected Eth node and L2->L1. The current approach should cause the block ticking to wait for Eth finality then, crucially (as the EVM watcher fundamentally does), check that the block is still included on-chain. As far as I can tell, Arbitrum should be handling rollbacks and would have removed the block if this occurred.

Ultimately, `GetL1Confirmations` was an appropriate call for finality pre-Merge, and this should be updated for the current era with something like `GetIsL1Finalized`. I intend to open an issue against nitro for this, as there does not appear to be a better way to guarantee finalized block inclusion from the watcher.

In light of this, approving as is, no changes required at this time.

node/cmd/guardiand/node.go
```diff
@@ -924,12 +924,14 @@ func runNode(cmd *cobra.Command, args []string) {
 		//
 		// NOTE:  The "none" is a special indicator to disable a watcher until it is desirable to turn it back on.
 
+		var ethWatcher *evm.Watcher
 		if shouldStart(ethRPC) {
 			logger.Info("Starting Ethereum watcher")
 			readiness.RegisterComponent(common.ReadinessEthSyncing)
 			chainObsvReqC[vaa.ChainIDEthereum] = make(chan *gossipv1.ObservationRequest, observationRequestBufferSize)
+			ethWatcher = evm.NewEthWatcher(*ethRPC, ethContractAddr, "eth", common.ReadinessEthSyncing, vaa.ChainIDEthereum, lockC, setC, 1, chainObsvReqC[vaa.ChainIDEthereum], false, nil)
```
evan-gray, line 932: Tests are failing because `*unsafeDevMode` isn't getting passed anymore and it's trying to call for finalized blocks. I think @kev1n-peters is trying to update ganache to support this!

evan-gray (changes requested):

node/cmd/guardiand/node.go
bruce-riley: Oops! Fixed.

evan-gray (approved):
approving now that tests pass

PR #1813, opened 2022-10-27
wormchain: add back protobuf

conorpp:
We trialed removing the `*.pb.go` files to make PR's easier to read but I think we should add them back, mainly because it complicates the build process for guardian and node providers (e.g. need to install ignite which may require updating glibc, etc...).  Also, if we ever need to have wormchain as a dependency for an sdk later, we'd need to do this anyways.  

To verify this PR, run the following steps.

```
cd wormchain
find . -name *.pb.go -delete
curl https://get.ignite.com/cli@v0.23.0 | bash
./ignite generate proto-go
git diff
```

This deletes all the .pb.go files and regenerates them.  There shouldn't be any diff generated after.

Thanks to @gearlbrace for the feedback!


evan-gray:
@conorpp are these generated from committed protos? if so, I was planning on committing generated proto files for node and having a CI action to validate them. Perhaps once that is in place we can share 🤝

jynnantonix (approved):

hendrikhofstadt (approved):

conorpp:
> @conorpp are these generated from committed protos? if so, I was planning on committing generated proto files for node and having a CI action to validate them. Perhaps once that is in place we can share 🤝

Yes we can make a CI step to validate -- probably we can use ignite still for that or switch to something like [buf](https://docs.buf.build/introduction) to generate.

PR #1814, opened 2022-10-28
sui/watcher: full node now up in tilt and a watcher stubbed out

jumpsiegel:


Tiltfile
```diff
@@ -187,6 +187,16 @@ def build_node_yaml():
                     "0xde0036a9600559e295d5f6802ef6f3f802f510366e0c23912b0655d972166017::state::WormholeMessageHandle",
                 ]
 
+            if sui:
+                container["command"] += [
+                    "--suiRPC",
+                    "http://sui:9002",
+                    "--suiAccount",
```
kcsongor, line 194: can you put a note here saying that the mainnet flags will need the additional argument? whatever's here is going to be used as a template for the mainnet instructions and it's better to catch that early

jumpsiegel: How about this?

kcsongor: perfect

kcsongor (approved):

PR #1819, opened 2022-10-28
SDK: Remove support for Ropsten

bruce-riley:
Remove Ropsten support from the SDK. 

Once this PR has been merged and the SDK released, a separate PR will be used to remove support from client/js and the relayer.

Additionally, PR 1818 is removing support from the node / guardian code.

### DIFF
--- a/sdk/js/src/utils/consts.ts
+++ b/sdk/js/src/utils/consts.ts
@@ -27,7 +27,6 @@ export const CHAINS = {
   gnosis: 25,
   pythnet: 26,
   xpla: 28,
-  ropsten: 10001,
   wormchain: 3104,
 } as const;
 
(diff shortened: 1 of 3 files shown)


kev1n-peters (approved):

sdk/js/src/utils/consts.ts
```diff
@@ -27,7 +27,6 @@ export const CHAINS = {
   gnosis: 25,
   pythnet: 26,
   xpla: 28,
-  ropsten: 10001,
```
kev1n-peters, line 30: I originally thought about leaving this as a reserved chain id to prevent it from being reused, but now I don't think it matters much, because this is a testnet chain

evan-gray (changes requested):
0.9.0 actually hasn't released yet - can you undo the version bump, maybe we can get this into that breaking change as well.

[see versions](https://www.npmjs.com/package/@certusone/wormhole-sdk)

evan-gray (approved):
Thanks!

kev1n-peters (approved):

PR #1820, opened 2022-10-28
ethereum: deploy test token nonce race condition fix

kev1n-peters:
Noticed this error happening in tilt sometimes:

[tests] Error: Returned error: VM Exception while processing transaction: the tx doesn't have the correct nonce. account has nonce of: 17 tx has nonce of: 16

It's not safe to submit txs in parallel, because the nonce can get out of sync. Instead we should submit them serially.


evan-gray (approved):
🐛 🔫

kcsongor (approved):

PR #1828, opened 2022-10-29
solana/tilt: run registrations in parallel (fixes #1826)

### LINKED ISSUE #1826: Make Solana ready faster by improving chain registration submissions
## Expected Behavior

The containers in tilt should become usable as fast as reasonably possible. If there are obvious optimizations to be made with minimal complexity, they should be made.

## Current Behavior

Every chain registration is performed in series by running `worm submit $VAA --chain solana --network devnet`

https://github.com/wormhole-foundation/wormhole/blob/3877f6f3c3f6f097ae7b3630bd271b8620f3ab6b/solana/devnet_setup.sh#L107

## Possible Solution

`worm submit` performs multiple transactions per VAA and uses a confirmed connection to submit them. Perhaps there are easy devnet-specific improvements to make there. Alternatively, could the registration accounts be generated / dumped and restored? The only downside to that approach would be that the registration flow of the contracts and `worm submit` wouldn't be tested in this way anymore.

## Context (Environment)

Tilt

kcsongor:
Running registrations in parallel reduces the time the process takes from ~90s to ~9s, which is a 10x improvement. Closes #1826

### DIFF
--- a/solana/Dockerfile
+++ b/solana/Dockerfile
@@ -9,7 +9,12 @@ RUN if [ -e /certs/cert.pem ]; then cp /certs/cert.pem /etc/ssl/certs/ca-certifi
 # Add bridge contract sources
 WORKDIR /usr/src/bridge
 
-COPY . .
+COPY bridge bridge
+COPY modules modules
+COPY migration migration
+COPY Cargo.toml Cargo.toml
+COPY Cargo.lock Cargo.lock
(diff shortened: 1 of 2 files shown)


bruce-riley (approved):

evan-gray (approved):
🏎️Much speed
🔨Such bash
🙀Wow

PR #1829, opened 2022-10-29
Node: Disable more PythNet logging

bruce-riley:
This reduces the amount of logging for Pyth messages even more. If it makes it into release v2.13.1, cool, but if it waits until the next release, that's cool too.


jynnantonix (commented):
I'm personally not a fan of adding a bunch of `if s.chainID != vaa.ChainIDPythNet` conditions everywhere.  What if other chains start having pythnet levels of activity?  It also adds a bunch of clutter and it's very easy to forget when adding new log lines.  

I think we have a couple of options:

1. Implement the [`zap.Core`](https://pkg.go.dev/go.uber.org/zap@v1.23.0/zapcore#Core) interface to filter out logs for pythnet.  See [this issue](https://github.com/uber-go/zap/issues/453) for more details but basically we would filter based on whether the log has a field with `emitter_chain == pythnet` and has a message that matches one of the ones we want to filter out.  When constructing the pythnet logger we would call [`logger.WithOptions`](https://pkg.go.dev/go.uber.org/zap#Logger.WithOptions) with [`zap.WrapCore`](https://pkg.go.dev/go.uber.org/zap#WrapCore).
1. Change the log levels of the messages we want to filter to Debug instead of Info.  Then when constructing the pythnet logger we would do something like:
```
infoLogger := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
	return lvl >= zapcore.InfoLevel
})

pythnetLogger := logger.WithOptions(zap.IncreaseLevel(infoLogger))
```

hendrikhofstadt (changes requested):
+1 for trying to implement this as a log filter on the emitter_chain field.

bruce-riley:
I took a slightly different approach which achieves the same result.

jumpsiegel (dismissed):

hendrikhofstadt (changes requested):
I generally think log level specific logic should not leak into the watcher which limits this functionality to very specific cases.
I'd strongly vouch for going for one of @jynnantonix 's proposed paths.
A single line clearly encoding a filter on pythnet seems more future-proof than implicitly asking for all new log statements here to use this particular filter level while potentially losing granularity.

panoel (dismissed):

hendrikhofstadt:
@jynnantonix 's 2nd option is a 2 line change changing the logger that is passed in while this leaks more deeply into the codebase

bruce-riley:
Chatted with @hendrikhofstadt offline and we agreed these messages could be suppressed for Solana as well, so just changed the log level to debug.

panoel (approved):

hendrikhofstadt (approved):

hendrikhofstadt (approved):

PR #1830, opened 2022-10-29
solana: build worm in the dockerfile

kcsongor:
Instead of devnet-setup.sh. It's more compositional this way as the devnet-setup script no longer builds another tool. Also better caching in docker


kcsongor:
closing because the CI cache seems to be broken, and resetting it now isn't worth it, there are better ways to improve the performance here anyway (build worm once)

PR #1836, opened 2022-10-31
Node: Arbitrum poller failling without details

bruce-riley:
The Arbitrum poller is occasionally failing, but the error text is not getting logged when the runnable dies. 

Additionally, the poller retries three times before dying, but we are not seeing the log error from any of them. Adding a 250 ms delay between retries.

Change-Id: Id067d0be18a7c391ef90a82cc78f0684c72ab2ad


panoel (dismissed):

hendrikhofstadt (changes requested):
What's the exact motivation behind the delay? Please add a comment explaining it as well.

hendrikhofstadt (approved):

panoel (approved):

PR #1845, opened 2022-11-01
client/js: Remove support for Ropsten

bruce-riley:
This PR updates the client/js code to remove support for Ropsten since it is deprecated. 

Note that it removes a call to ixFromRust() because it no longer exists in the SDK. I did a test submit on Solana and it seemed to work.

This PR partially fixes issue 1770.

Change-Id: I314c2299fb00be0c1647729dc3016c9dd34beb3a

### DIFF
--- a/clients/js/networks.ts
+++ b/clients/js/networks.ts
@@ -139,10 +139,6 @@ const MAINNET = {
     rpc: "https://rpc.gnosischain.com/",
     key: get_env_var("ETH_KEY"),
   },
-  ropsten: {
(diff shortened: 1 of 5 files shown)


panoel (approved):

bruce-riley:
This PR requires upgrading the sdk version to 0.9.0, which breaks the client because of wasm changes. Therefore I am closing this PR until the client can be reworked to not rely on wasm. Leaving issue 1770 open as a placeholder. so we don't forget to go back later an clean up client/js.

PR #1849, opened 2022-11-02
Node: Rework min/max confs for EVM watchers

bruce-riley:
This PR changes the way the EVM watchers handle determining when a transaction can be published and when it should be declared abandoned.

The existing functionality uses a single minConfirmations parameter and multiplies it by four to get the max confirmations. This has broken down over time because the watchers for most EVM chains work off of finalized blocks, so there is no need to wait for additional blocks, but setting minConfirmations  to zero does not work to compute the maximum number of blocks to wait.

This PR changes the EVM watcher to take two parameters:

waitForConfirmations - bool indicating if we need to wait the number of additional confirmations specified by the consistency level. It will be set to false for chains that have true finality.

maxWaitConfirmations - this is added to the transaction block and the watcher declares the transaction abandoned at that block.

By default the watcher assumes waitForConfirmations is false and maxWaitConfirmations is four. (@evan-gray and @bruce-riley came with these values). If a given chain requires different values, they can be set by calling SetConfirmationParams() before calling Run().

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -917,7 +917,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Info("Starting Ethereum watcher")
 			readiness.RegisterComponent(common.ReadinessEthSyncing)
 			chainObsvReqC[vaa.ChainIDEthereum] = make(chan *gossipv1.ObservationRequest, observationRequestBufferSize)
-			ethWatcher = evm.NewEthWatcher(*ethRPC, ethContractAddr, "eth", common.ReadinessEthSyncing, vaa.ChainIDEthereum, lockC, setC, 1, chainObsvReqC[vaa.ChainIDEthereum], *unsafeDevMode)
+			ethWatcher = evm.NewEthWatcher(*ethRPC, ethContractAddr, "eth", common.ReadinessEthSyncing, vaa.ChainIDEthereum, lockC, setC, chainObsvReqC[vaa.ChainIDEthereum], *unsafeDevMode)
 			if err := supervisor.Run(ctx, "ethwatch",
 				ethWatcher.Run); err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


node/cmd/guardiand/node.go
```diff
@@ -903,7 +903,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Info("Starting Ethereum watcher")
 			readiness.RegisterComponent(common.ReadinessEthSyncing)
 			chainObsvReqC[vaa.ChainIDEthereum] = make(chan *gossipv1.ObservationRequest, observationRequestBufferSize)
-			ethWatcher = evm.NewEthWatcher(*ethRPC, ethContractAddr, "eth", common.ReadinessEthSyncing, vaa.ChainIDEthereum, lockC, setC, 1, chainObsvReqC[vaa.ChainIDEthereum], *unsafeDevMode, nil)
+			ethWatcher = evm.NewEthWatcher(*ethRPC, ethContractAddr, "eth", common.ReadinessEthSyncing, vaa.ChainIDEthereum, lockC, setC, 0, 100, chainObsvReqC[vaa.ChainIDEthereum], *unsafeDevMode, nil)
```
evan-gray, line 906: I think all the maxes should be 4 (effectively what the value that's running today is) except BSC and Polygon (see below)

node/pkg/watchers/evm/watcher.go
```diff
@@ -531,13 +536,8 @@ func (w *Watcher) Run(ctx context.Context) error {
...
 
 				for key, pLock := range w.pending {
-					expectedConfirmations := uint64(pLock.message.ConsistencyLevel)
-					if expectedConfirmations < w.minConfirmations {
-						expectedConfirmations = w.minConfirmations
-					}
-
 					// Transaction was dropped and never picked up again
-					if pLock.height+4*uint64(expectedConfirmations) <= blockNumberU {
+					if pLock.height+w.maxConfirmations <= blockNumberU {
```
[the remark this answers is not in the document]
evan-gray: Actually, for chains that can still have custom consistencyLevels (i.e. BSC) this may still need to be `expected * 4` or `expected + max`

node/pkg/watchers/evm/watcher.go
```diff
@@ -124,6 +127,9 @@ type (
 	}
 )
 
+// DefMaxConfs is default value for the maximum number of confirmations to wait before declaring a transaction abandoned.
+const DefMaxConfs = 32
```
evan-gray, line 131: can we just spell out default? because I missed this multiple times on reading through.

evan-gray (changes requested):

bruce-riley:
This PR assumes PR 1858 has been merged!

node/pkg/watchers/evm/watcher.go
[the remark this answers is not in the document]
bruce-riley: Renamed the parameter, also am waiting expected + max.

node/pkg/watchers/evm/watcher.go
```diff
@@ -360,10 +368,7 @@ func (w *Watcher) Run(ctx context.Context) error {
 						continue
 					}
 
-					expectedConfirmations := uint64(msg.ConsistencyLevel)
-					if expectedConfirmations < w.minConfirmations {
-						expectedConfirmations = w.minConfirmations
-					}
+					expectedConfirmations := uint64(msg.ConsistencyLevel) + w.additionalConfirmations
```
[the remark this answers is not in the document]
bruce-riley: I think I captured this in the rework. . .

node/cmd/guardiand/node.go
bruce-riley: Done.

node/pkg/watchers/evm/watcher.go
bruce-riley: Got rid of this constant.

bruce-riley:
Fixes #1900

hendrikhofstadt (approved):
Looks good to me.
Also what's the reasoning behind picking 4 for polygon and 60 for bsc?
Requesting additional review from Sec @tbjump @claudijd

bruce-riley:
Now that the Polygon checkpointing PR (#1858) is merged, we can treat Polygon in mainnet the same as the other chains that wait for finalized blocks, so I have changed that.

My understanding (and @evan-gray can correct me if I'm wrong) is that BSC does not have true finality, so we need to wait the specified number of confirmations.

Also, there's no particular reason to maxWaitConfirmations only be four, so I have raised it to 60, which should be suitable for everything, including BSC. That number is sort of arbitrary anyway, and it is better to wait too long before dropping a message than not waiting long enough.

evan-gray (approved):
This setter mechanism is kinda neat. Just to _confirm_, Polygon testnet will only wait the consistencyLevel in the contract now? (Assuming that's because Polygon testnet checkpoints are super slow) 🤔 Though I do wonder if the discrepancy will cause any integrator confusion. Overall though, big improvement for the rest of the chains I expect.

evan-gray:
@claudijd @tbjump you were called out by @hendrikhofstadt - would be good to have some extra reviews on this one.

jynnantonix (approved):

PR #1851, opened 2022-11-03
njkumr/hash check in verify vm

njkumr:
verifyVM currently doesn't verify if the hash field provided in the vm matches against the hash of the body of the vm. This is presently not very well documented on why it wouldn't be a good idea for wormhole integrators to use verifyVM on a vm they haven't securely parsed. Adding a check to verify for the same to make verifyVM safer if any one is using it directly along with tests for the same.

### DIFF
--- a/ethereum/contracts/Messages.sol
+++ b/ethereum/contracts/Messages.sol
@@ -15,6 +15,7 @@ contract Messages is Getters {
     /// @dev parseAndVerifyVM serves to parse an encodedVM and wholy validate it for consumption
     function parseAndVerifyVM(bytes calldata encodedVM) public view returns (Structs.VM memory vm, bool valid, string memory reason) {
         vm = parseVM(encodedVM);
+        /// setting checkHash to false as we can trust the hash field in this case given that parseVM computes and then sets the hash field above
         (valid, reason) = verifyVMInternal(vm, false);
     }
 
@@ -33,6 +34,8 @@ contract Messages is Getters {
     /**
     * @dev `verifyVMInternal` serves to validate an arbitrary vm against a valid Guardian set
     * if checkHash is set then the hash field of the vm is verified against the hash of its contents
+    * in the case that the vm is securely parsed and the hash field can be trusted, checkHash can be set to false
+    * as the check would be redundant
     */
     function verifyVMInternal(Structs.VM memory vm, bool checkHash) internal view returns (bool valid, string memory reason) {
         /// @dev Obtain the current guardianSet for the guardianSetIndex provided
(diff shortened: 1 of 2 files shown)


ethereum/contracts/Messages.sol
```diff
@@ -26,9 +26,45 @@ contract Messages is Getters {
     *  - it aims to verify the signatures provided against the guardianSet
     */
     function verifyVM(Structs.VM memory vm) public view returns (bool valid, string memory reason) {
+        (valid, reason) = verifyVMInternal(vm, true);    
+    }
+
+    /**
+    * @dev `verifyVMInternal` serves to validate an arbitrary vm against a valid Guardian set
+    *  - it is similar to verifyVM but drops the verification of the hash as it is called on the result of parseVM whose vm.hash field can be trusted.
```
kcsongor, line 34: this is outdated now because of the argument

ethereum/contracts/Messages.sol
```diff
@@ -26,9 +26,45 @@ contract Messages is Getters {
...
+    *  - it aims to ensure the guardianSet is not expired
+    *  - it aims to ensure the VM has reached quorum
+    *  - it aims to verify the signatures provided against the guardianSet
+    */
+    function verifyVMInternal(Structs.VM memory vm, bool checkHash) internal view returns (bool valid, string memory reason) {
         /// @dev Obtain the current guardianSet for the guardianSetIndex provided
         Structs.GuardianSet memory guardianSet = getGuardianSet(vm.guardianSetIndex);
 
+        /**
+         * @dev Verify that the hash field in the vm matches with the hash of the contents of the vm if checkHash is set
```
kcsongor, line 45: I think natspec comments aren't interpreted in function bodies, so we can drop the "@dev" annotation here

evan-gray:
This is a pretty core code change, let’s make sure there are lots of eyes and consensus on this. Personally, I feel this is a good example of something that should have been opened as an issue for discussion first.

ethereum/contracts/Messages.sol
njkumr: updated the comments now

ethereum/contracts/Messages.sol
```diff
@@ -24,25 +24,22 @@ contract Messages is Getters {
...
     }
 
     /**
     * @dev `verifyVMInternal` serves to validate an arbitrary vm against a valid Guardian set
-    *  - it is similar to verifyVM but drops the verification of the hash as it is called on the result of parseVM whose vm.hash field can be trusted.
-    *  - it aims to make sure the VM is for a known guardianSet
-    *  - it aims to ensure the guardianSet is not expired
-    *  - it aims to ensure the VM has reached quorum
-    *  - it aims to verify the signatures provided against the guardianSet
+    * if checkHash is set then the hash field of the vm is verified against the hash of its contents
```
kcsongor, line 35: maybe explain here why this is not something we always want to do? adding a comment in parseAndVerifyVM where we call this function with false explaining why would also be nice

njkumr: makes sense, added

kcsongor (dismissed):
looks good

kcsongor (approved):

a5-pickle (approved):

gator-boi (approved):
Looks good to me.

PR #1858, opened 2022-11-07
Node: Polygon watcher redesign

bruce-riley:
This PR changes the Polygon watcher to use checkpointing on Ethereum to determine when Polygon blocks are final. For details on Polygon finality, please see the prologue of node/pkg/watchers/evm/connectors/polygon.go.

To enable this feature, the guardians must specify two additional config parameters. If those parameters are not present, the Polygon watcher works as it previously did. It is likely that checkpointing will not be enabled in testnet because it can take a very long time for a transaction to be confirmed.

Note that this PR adds a new parameter to the EVM Watcher, extraParams. This is meant to be a generalized way to pass chain specific parameters into the watcher without having to keep adding additional parameters. If you have a better way to do this, please let me know.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -75,8 +75,10 @@ var (
 	bscRPC      *string
 	bscContract *string
 
-	polygonRPC      *string
-	polygonContract *string
+	polygonRPC                      *string
+	polygonContract                 *string
+	polygonRootChainRpc             *string
+	polygonRootChainContractAddress *string
 
 	auroraRPC      *string
 	auroraContract *string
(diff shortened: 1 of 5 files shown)


panoel (approved):

node/cmd/guardiand/node.go
```diff
@@ -915,26 +919,29 @@ func runNode(cmd *cobra.Command, args []string) {
...
 			}
 		}
 
 		if shouldStart(polygonRPC) {
-			polygonMinConfirmations := uint64(512)
-			if *testnetMode {
-				polygonMinConfirmations = 64
+			var extraParams map[string]interface{}
+			var minConfirmations uint64 = 512
+			if *polygonRootChainRpc != "" {
```
evan-gray, line 930: Checkpointing should probably be required on mainnet.

evan-gray (changes requested):

jynnantonix (commented):
It's a bit of a bigger refactor but one way to deal with the additional parameters is to switch to the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).  This way we only need to specify the relevant options for each invocation of `evm.NewWatcher` rather than requiring *all* the parameters at every invocation.

node/cmd/guardiand/node.go
bruce-riley: Fixed.

bruce-riley:
The functional options pattern sounds interesting, but I'm not sure I want to make that switch as part of this PR. If we want to consider that, we can open a github issue so the rework can be scheduled.

I did rework the code so passing the root chain parameters is not so messy.

jynnantonix (approved):

panoel (dismissed):

panoel (approved):

jynnantonix (approved):

panoel (approved):

evan-gray (approved):

PR #1859, opened 2022-11-07
sui: introduce emitter cap for core and token bridge

optke3:
- remove hash map sequence store in favor of emitter capability pattern

### DIFF
--- a/sui/wormhole/sources/emitter.move
+++ b/sui/wormhole/sources/emitter.move
@@ -1,0 +1,85 @@
+module wormhole::emitter {
+
+    use wormhole::serialize;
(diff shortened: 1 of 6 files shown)


sui/wormhole/sources/emitter.move
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public(friend) fun use_sequence(emitter_cap: &mut EmitterCapability): u64 {
+        let sequence = emitter_cap.sequence;
+        emitter_cap.sequence = sequence + 1;
+        sequence
+    }
+}
+
+// #[test_only]
```
optke3, line 68: TODO - make these tests Sui-style

PR #1860, opened 2022-11-07
sdk/js: Injective refactor

kev1n-peters:
Updated injectivelabs packages, refactored code, added utility functions

### DIFF
--- a/sdk/js/package.json
+++ b/sdk/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@certusone/wormhole-sdk",
-  "version": "0.9.0",
+  "version": "0.9.1",
   "description": "SDK for interacting with Wormhole",
   "homepage": "https://wormhole.com",
   "main": "./lib/cjs/index.js",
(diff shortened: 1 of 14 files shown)


evan-gray (approved):

sdk/js/package.json
```diff
@@ -63,7 +63,7 @@
   "dependencies": {
     "@certusone/wormhole-sdk-proto-web": "^0.0.5",
     "@certusone/wormhole-sdk-wasm": "^0.0.1",
-    "@injectivelabs/sdk-ts": "^1.0.75",
+    "@injectivelabs/sdk-ts": "^1.0.211",
```
panoel, line 66: Do we want to make sure that the client is also on this set of injective releases?  Should that be a clickup task?

panoel (approved):

sdk/js/package.json
kev1n-peters: When the wormhole-sdk package is published to npm and the client updates to that version, we'll need to make sure the injective packages match

PR #1875, opened 2022-11-09
sui/0.15: make publish_message work again and fix up the watcher

jumpsiegel:


kcsongor (approved):

node/pkg/watchers/sui/watcher.go
```diff
@@ -188,17 +205,61 @@ func (e *Watcher) Run(ctx context.Context) error {
...
 			logger.Info("Received obsv request")
 
-		case <-timer.C:
-			if e.unsafeDevMode {
-				logger.Info("tick unsafe")
-			} else {
-				logger.Info("tick safe")
+			id := base64.StdEncoding.EncodeToString(r.TxHash)
+
+			buf := fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "sui_getEvents", "params": [{"Transaction": "%s"}, null, 10, true]}`, id)
```
hendrikhofstadt, line 210: unsanitized formatting here could maybe open up an attack vector

node/pkg/watchers/sui/watcher.go
```diff
@@ -188,17 +205,61 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+
+			logger.Info(buf)
```
hendrikhofstadt, line 212: debug logging

node/pkg/watchers/sui/watcher.go
```diff
@@ -188,17 +205,61 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+					logger.Error(e.suiRPC, zap.Error(err))
+				}
+			}
+
+		case <-timer.C:
 			for {
 				var msg = make([]byte, 10000)
 				var n int
-				ws.SetReadDeadline(time.Now().Local().Add(100_000_000))
+				err := ws.SetReadDeadline(time.Now().Local().Add(100_000_000))
```
hendrikhofstadt, line 254: not sure we need the local here

hendrikhofstadt: @jynnantonix you may know better how deadlines in the ws library are handled and whether this is correct handling

jynnantonix: I agree the `Local()` is not necessary here.  Otherwise as far as I can tell, this call eventually gets forwarded to `setsockopt` on the underlying socket so this should be fine.

node/pkg/watchers/sui/watcher.go
```diff
@@ -5,6 +5,9 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"strings"
 	"time"
 
 	"encoding/base64"
```
jynnantonix, line 13: We're using "github.com/gorilla/websocket" in all the other watchers so we should stay consistent.  I think the API is basically the same.

PR #1886, opened 2022-11-10
sui/emitter: start at 1

jumpsiegel:


kcsongor (approved):

kcsongor:
this is in line with the aptos implementation

PR #1897, opened 2022-11-11
add: safety check against signatory being address(0)

Rassska:
- Comes from Immunefi report №13286.

### Description
- I've spent a lot of time on finding the way to forge signatures due to the lack of `address(0)` check after `verifySignatures()::ecrecover()`, but didn't successed, since there is no a way to manipulate with guardians. If we had an opportunity to include `guardians.keys` along with signatures in calldata, we simply could pass zero addresses along with some random signatures and `ecrecover()` will return `address(0)` and the check performed here would be easily bypassed: `if(ecrecover(hash, sig.v, sig.r, sig.s) != guardianSet.keys[sig.guardianIndex])`

- Well, this report is about preventing such possibilities(in the future) by simply adding the check against signatory being `address(0)`.

### Impact
- Bugs that are likely to occur in future stages of development but do not manifest themselves yet

### DIFF
--- a/wormchain/Makefile
+++ b/wormchain/Makefile
@@ -11,7 +11,8 @@ ldflags = \
     -X github.com/cosmos/cosmos-sdk/version.Name=wormchain\
 	-X github.com/cosmos/cosmos-sdk/version.ServerName=wormchaind\
 	-X github.com/cosmos/cosmos-sdk/version.Version=$(VERSION) \
-	-X github.com/cosmos/cosmos-sdk/version.Commit=$(COMMIT) 
+	-X github.com/cosmos/cosmos-sdk/version.Commit=$(COMMIT) \
+	-X "github.com/cosmos/cosmos-sdk/version.BuildTags=ledger"
 BUILD_FLAGS := -ldflags '$(ldflags)'
 
 .PHONY: all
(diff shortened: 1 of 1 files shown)


jumpsiegel (dismissed):

evan-gray (changes requested):
Requesting changes to hold while a couple more EVM folks take a look. My initial thought - Is there any reason this couldn't be assured against by a test instead?

kcsongor:
The check is sensible, as `ecrecover`, being the beautifully designed primitive it is, returns 0 when the parameters are invalid. Which could lead to a problem if a guardian pubkey was ever set to 0. That's not the case currently, but there's no check in the code to ensure it never will, so agreed this is a potential refactoring hazard. Doing the extra check does no harm

ethereum/contracts/Messages.sol
```diff
@@ -88,9 +89,10 @@ contract Messages is Getters {
             /// introduced accidentally later due to the nontrivial storage
             /// semantics of solidity.
             require(sig.guardianIndex < guardianCount, "guardian index out of bounds");
+            require(signatory != address(0));
```
kcsongor, line 92: any reason why this check is not on line 80 (right after the variable declaration)?

Rassska: We can move it up and that's fair enough!

gator-boi, line 92: I agree that this check is sensible, however, it might be redundant given that the guardian set is stored using this [method](https://github.com/wormhole-foundation/wormhole/blob/6de97de242e57be4a0a8a0840f5a16bf4cd55bda/ethereum/contracts/Setters.sol#L17).

hendrikhofstadt:
Agree the check should be moved up after the `ecrecover`
Generally I think this encodes a valid constraint i.e. signature is not invalid so IMO we should add this.
Gas cost impact seems close to 0 to me.

PR #1899, opened 2022-11-11
Node: Neon Watcher Redesign

### LINKED ISSUE #1898: Node: Neon Watcher Redesign
Guardian Neon watcher should wait for finality, rather than waiting an arbitrary number of blocks.

bruce-riley:
This PR changes the Neon watcher to wait for finality on Solana before publishing Neon blocks.

Fixes #1898


jumpsiegel (approved):
One of the feedbacks we have gotten from auditors is maybe we should wrap the standard arguments for these watchers into a "object" to reduce the total number of arguments in the function.    If you think about how you would actually then set all those arguments, it would be just as ugly...

panoel (approved):

PR #1905, opened 2022-11-14
WH-1903:  basic BTC node support in Tilt

jumpsiegel:
This is intended for the development work of BTC support in wormhole


evan-gray:
Seems like something @conorpp would be interested in.

evan-gray (approved):
Confirmed the k8s resource comes up green ✅ not sure how to evaluate much more than that, other than it seems reasonable, since it isn't used anywhere yet

conorpp (approved):

conorpp:
Awesome!  Looks set up perfectly.

PR #1916, opened 2022-11-15
Sui/tests

jumpsiegel:


node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+				err := e.inspectBody(logger, chunk)
+				if err != nil {
+					logger.Error(e.suiRPC, zap.Error(err))
+				}
+			}
+
+		case <-timer.C:
+			for {
+				var msg = make([]byte, 10000)
+				err := ws.SetReadDeadline(time.Now().Local().Add(100_000_000))
```
jynnantonix, line 275: I think the .Local() here is unnecessary.

node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+						return errors.New(errVal.String())
+					}
+
+					id := parsedMsg.Get("id")
+					if id.Exists() {
+						if id.Int() == e.subId {
+							e.subscribed = true
+						}
+						continue
+					}
```
jynnantonix, line 311: Rather than doing this within the loop can we handle the subscription response beforehand?  IIUC, we should receive exactly one message like this right at the beginning.

node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+			}
+
+			resp, err := http.Post(e.suiRPC, "application/json", strings.NewReader(`{"jsonrpc":"2.0", "id": 1, "method": "sui_getCommitteeInfo", "params": []}`))
+			if err != nil {
+				logger.Error(fmt.Sprintf("sui_getCommitteeInfo: %s", err.Error()))
+				p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDSui, 1)
+				break
+
+			}
+			body, err := io.ReadAll(resp.Body)
```
jynnantonix, line 322: `resp.Body.Close()`

jynnantonix (commented):
I only reviewed the node/ directory.  In general it looks fine but I have a preference for parsing into structured data rather than manually pulling out fields with gjson.

Otherwise, this PR is quite large and it's going to be hard to get a meaningful review for ~6000 lines of changes.  Would be nice if we could split them up into smaller, independent, more easily reviewable PRs.  The graphite CLI makes it relatively easy to do this.

If we're not going to split this branch into smaller PRs then are we planning on squashing the whole branch before committing?  I suspect that there isn't much value in indefinitely preserving the development history of this branch.  For reference, we did the same thing for wormchain where we squashed the dev history down into a couple of commits.

evan-gray:
@jumpsiegel did you intend to PR this against dev.v2?

node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"strings"
+	"time"
+
+	"encoding/base64"
+
+	"golang.org/x/net/websocket"
```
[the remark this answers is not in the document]
hendrikhofstadt: My fault here for recommending the x/net library. Think we should make the switch here as long as it's not too painful.

node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+		unsafeDevMode: unsafeDevMode,
+		msgChan:       messageEvents,
+		obsvReqC:      obsvReqC,
+		subId:         0,
+		subscribed:    false,
+	}
+}
+
+func (e *Watcher) inspectBody(logger *zap.Logger, body gjson.Result) error {
+	txDigest := body.Get("txDigest")
```
hendrikhofstadt, line 88: Given this has a static schema I'd recommend using the struct decoder. 

You can turn json into a struct here: https://mholt.github.io/json-to-go/

jumpsiegel:
As a side note, this was actually a test branch to see if the tests passed against the current sui/integration branch with no intention of actually merging it.

BUT the feedback is all valid so will update the watcheqr

evan-gray:
> As a side note, this was actually a test branch to see if the tests passed against the current sui/integration branch with no intention of actually merging it.
> 
> BUT the feedback is all valid so will update the watcher

@jumpsiegel you should make it against the sui/integration and keep it draft then 😁 will convert it for you for now

node/pkg/watchers/sui/watcher.go
jumpsiegel: We could break this out into its own loop but the response does not happen instantly but occures at some point on the websocket.  It would not make the code any clearer

node/pkg/watchers/sui/watcher.go
```diff
@@ -0,0 +1,355 @@
...
+			}
+			logger.Info(string(body))
+
+			if !gjson.Valid(string(body)) {
+				logger.Error("InvalidJson: " + string(body))
+				p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDAptos, 1)
+				break
+			}
+
+			outcomes := gjson.ParseBytes(body).Get("result.data")
```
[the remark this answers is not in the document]
jumpsiegel: I disagree.  I think you are significantly adding to the size of the code with a implied reduction in code readability

PR #1917, opened 2022-11-15
cosmwasm: Add wormhole-bindings package

jynnantonix:
This package defines the bindings into the native wormhole module on
wormchain.

### DIFF
--- a/cosmwasm/packages/wormhole-bindings/src/query.rs
+++ b/cosmwasm/packages/wormhole-bindings/src/query.rs
@@ -1,0 +1,42 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Binary, CustomQuery, Empty};
+
+#[cw_serde]
+pub struct Signature {
+    /// The index of the guardian in the guardian set.
+    pub index: u8,
+
+    /// The signature, which should be exactly 65 bytes with the following layout:
+    ///
+    /// ```markdown
+    /// 0  .. 64: Signature   (ECDSA)
+    /// 64 .. 65: Recovery ID (ECDSA)
+    /// ```
+    pub signature: Binary,
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum WormholeQuery {
+    /// Verifies that `data` has been signed by a quorum of guardians from `guardian_set_index`.
+    #[returns(Empty)]
+    VerifyQuorum {
+        data: Binary,
+        guardian_set_index: u32,
+        signatures: Vec<Signature>,
+    },
+
+    /// Verifies that `data` has been signed by a guardian from `guardian_set_index`.
+    #[returns(Empty)]
+    VerifySignature {
+        data: Binary,
+        guardian_set_index: u32,
+        signature: Signature,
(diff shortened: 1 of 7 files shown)


jynnantonix:
* dev.v2
  * **PR #1917** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1917" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1917?utm_source=stack-comment).

cosmwasm/packages/wormhole-bindings/src/query.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum WormholeQuery {
+    // Verifies that `data` has been signed by a quorum of observers from `observer_set_index`.
+    #[returns(Empty)]
+    VerifyQuorum {
+        data: Binary,
+        observer_set_index: u32,
```
kcsongor, line 20: Any reason why we're not calling this `guardian_set_index`?

cosmwasm/packages/wormhole-bindings/src/query.rs
```diff
@@ -0,0 +1,42 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Binary, CustomQuery, Empty};
+
+#[cw_serde]
+pub struct Signature {
+    // The index of the observer in the observer set.
+    pub index: u16,
+
+    // The signature.
+    pub signature: Binary,
```
kcsongor, line 10: Given that this is expected to have a specific structure, might be worth making the type more precise. Unfortunately I've seen different libraries/ecosystems use different permutations of the 65 byte (v,r,s) triple, so the signature bytes can't be treated abstract in a portable way.

kcsongor (dismissed):

cosmwasm/packages/wormhole-bindings/src/query.rs
conorpp: The 65 byte (v,r,s) serialization is the most common I've seen.  Unless we're commonly parsing out v,r,s instead of passing it transparently to a library, I think keeping the blob would be the most simple.

conorpp (approved):

cosmwasm/packages/wormhole-bindings/src/query.rs
jynnantonix: Done.

cosmwasm/packages/wormhole-bindings/src/query.rs
jynnantonix: Done.

conorpp (dismissed):

kcsongor (dismissed):

jynnantonix (commented):
I changed the index field in the Signature struct to a u8 so your reviews were dismissed.  Please take another look.

hendrikhofstadt (approved):

conorpp (approved):

PR #1918, opened 2022-11-15
cosmwasm: Add core accounting package

jynnantonix:
The accounting package implements the reusable, chain agnostic part of
tokenbridge accounting.

Part of #1880

PR #1920, opened 2022-11-15
cosmwasm: Add wormchain-accounting contract

jynnantonix:
This PR adds the wormchain accounting contract.  Changes in other directories can be ignored as they are a result of merging in #1917 and #1918 and will be removed once those PRs are merged and this one is rebased.


jynnantonix:
* dev.v2
  * **PR #1919** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1919" target="_blank">Graphite</a> 
    * **PR #1920** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1920" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1920?utm_source=stack-comment).

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,491 @@
...
+                data: msg.instantiate.clone(),
+                observer_set_index: msg.observer_set_index,
+                signatures,
+                block_time: env.block.height,
+            }
+            .into(),
+        )
+        .context(ContractError::VerifyQuorum)?;
+
+    let init: Instantiate =
```
conorpp, line 59: Is this instantiate message similar or the same type of VAA as for deploying other contracts?

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,491 @@
...
+
+    Ok(Response::new()
+        .add_attribute("action", "instantiate")
+        .add_attribute("owner", info.sender)
+        .add_event(event))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(_deps: DepsMut<WormholeQuery>, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
+    Ok(Response::default())
```
conorpp, line 82: I believe as is the cosmos account owner would be freely available to change the contract.

My proposal would be to require a VAA assertion before allowing the contract to upgrade.

conorpp: I believe `migrate` can be freely executed by the owner of the contract which we don't want.  Maybe move the VAA verification currently in `upgrade_contract` to `migrate`?

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,491 @@
...
+            token_chain,
+            recipient_chain,
+            ..
+        } => transfer::Data {
+            amount: Uint256::from_be_bytes(amount.0),
+            token_address: TokenAddress::new(token_address.0),
+            token_chain: token_chain.into(),
+            recipient_chain: recipient_chain.into(),
+        },
+        _ => bail!("Unknown tokenbridge payload"),
```
conorpp, line 240: It might be nice to return this error immediately rather than collecting 13 signatures for it and then failing.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,491 @@
...
+
+    if data.signatures().len() < quorum {
+        // Still need more signatures so just save the pending transfer data and exit.
+        key.save(deps.storage, &pending)
+            .context("failed to save pending transfers")?;
+
+        return Ok(None);
+    }
+
+    let (msg, _) = serde_wormhole::from_slice_with_payload(&o.payload)
```
conorpp, line 218: similarly, I think it'd be a little more optimal to return this failure before saving signatures.

conorpp:
Will approve pending `migrate` comment, looks really nice.  I like the separation with the accounting package.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: No, it's a completely different message (defined in src/msg.rs).

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: I believe what we do for other cosmwasm contracts is to set the owner to itself.  Then upgrades can only happen through governance and no one else should be able to call `migrate`.

If we can't set the contract owner to itself then the `UpgradeContract` message would be useless since the contract would not be allowed to call migrate.  This actually caused some issues during testing where I had to do a gross hack to be able to set the contract owner to itself (see proper_instantiate() in tests/helpers/mod.rs).

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: It's basically a question of whether we want to parse the payload every time or only once we might reach quorum.  

The value type for `PENDING_TRANSFERS` is a `Vec<transfer::Data>` so we can continue to add arbitrarily many different observations for the same key until one of them reaches quorum.  So in the case where we fail to parse the payload, we can fix the issue offline and then resubmit new observations for the same key.  Once the new observation reaches quorum it will get committed and everything else will be deleted.

If we make the assumption that an observation that reaches quorum is highly unlikely to have an invalid payload, which doesn't seem *that* unreasonable, then we can optimize for that case and only parse it once at the cost of delaying the error case.  This seems like a worthwhile tradeoff to me but I don't feel that strongly about it so I'm happy to change it if you still prefer that we check the payload every time.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: Same comment as above.  This is an optimization with the assumption that observations that reach quorum are highly unlikely to have invalid or unsupported payloads.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
conorpp: Ah okay that makes sense, thanks for explaining.

conorpp (approved):

cosmwasm/packages/wormhole-bindings/src/query.rs
```diff
@@ -0,0 +1,42 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Binary, CustomQuery, Empty};
+
+#[cw_serde]
+pub struct Signature {
+    /// The index of the guardian in the guardian set.
+    pub index: u16,
```
nik-suri, line 7: the sdk defines the `Signature` type with a `uint8` -- https://github.com/wormhole-foundation/wormhole/blob/dev.v2/sdk/vaa/structs.go#L77-L82

Any reason to use `u16` here instead of `u8` for the guardian index?

jynnantonix (commented):
I had to move the dockerfile to the parent directory to pull in the rust sdk.  Also updated the index type from u16 -> u8.  PTAL.

nik-suri (approved):

conorpp (approved):

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,469 @@
...
+                chain: emitter_chain,
+            },
+        )
+        .context("failed to query chain registration")?;
+    ensure!(
+        *registered_emitter == **o.key.emitter_address(),
+        "unknown emitter address"
+    );
+
+    accounting::commit_transfer(
```
a5-pickle, line 244: Should the event from this method be propagated to the `Event` down below with `add_event` like in `instantiate`? Some of the same info from the `commit_transfer` `Event` is shared with this method's event.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -0,0 +1,469 @@
...
+    let event = Event::new("Modification")
+        .add_attribute("sequence", msg.sequence.to_string())
+        .add_attribute("chain_id", msg.chain_id.to_string())
+        .add_attribute("token_chain", msg.token_chain.to_string())
+        .add_attribute("token_address", msg.token_address.to_string())
+        .add_attribute("kind", msg.kind.to_string())
+        .add_attribute("amount", hex::encode(msg.amount.to_be_bytes()))
+        .add_attribute("reason", msg.reason.clone());
+
+    accounting::modify_balance(deps, msg).context("failed to modify account balance")?;
```
a5-pickle, line 296: Should the event from this method be propagated to the `Event` down below with `add_event` like in `instantiate`? All of the same info from `modify_balance` is shared with this method's event.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: I'm inclined to say no because the information is repeated, as you mentioned.  Someone listening for this event can get all the details by parsing out the `payload` field.

That said, I'm also ok with adding both events.  I don't feel that strongly either way.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: Done.  I dropped this event in favor of the one from `modify_balance`.  Also added a test for the event.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
a5-pickle: I'm cool the way it is. Someone else integrating w/ the accounting package can choose to use this event or their own custom event. In this case, the wormchain accounting smart contract chooses to use its own Event.

a5-pickle (approved):
Looks good

justinschuldt:
Hey @jynnantonix, I found when that deploying this contract to wormchain fails because there are two codepaths that produce Float64s, which do not exist in the cosmwasm VM.

Looks like `serde::de:Unexpected` sometimes creates a Float64 when it is formatted to a string.

Here are steps to reproduce:

# prereqs

wasm artifacts built from cosmwasm contracts

    DOCKER_BUILDKIT=1 docker build --target artifacts -o cosmwasm/artifacts -f Dockerfile.cosmwasm .

wormchaind build

    # from /wormchain
    make client

# minimal reproducer

start wormchain

    # from /wormchain
    make run

upload a wasm artifact, to test the successful path

    ./build/wormchaind tx wasm store ../cosmwasm/artifacts/cw20_base.wasm --from tiltGuardian --gas auto --home build/

upload the wormchain_accounting.wasm artifact, which whill fail

    ./build/wormchaind tx wasm store ../cosmwasm/artifacts/wormchain_accounting.wasm --from tiltGuardian --gas auto --home build/

the resulting error will be:

    Error: rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: Error calling the VM: Error compiling Wasm: Could not compile: WebAssembly translation error: Error in middleware Gatekeeper: Float operator detected: F64Load { memarg: MemoryImmediate { align: 3, offset: 8, memory: 0 } }. The use of floats is not supported.: create wasm contract failed [wormhole-foundation/wasmd@v0.28.0-wormhole-2/x/wasm/keeper/keeper.go:189] With gas wanted: '0' and gas used: '2576224' : unknown request

# debugging

decode `cosmwasm/artifacts/wormchain_accounting.wasm` to see where the Float64 is coming from.

get [WABT: The WebAssembly Binary Toolkit](https://github.com/WebAssembly/wabt)

    git clone --recursive https://github.com/WebAssembly/wabt
    cd wabt
    git submodule update --init

    # build the WABT tools
    mkdir build
    cd build
    cmake ..
    cmake --build .

in order for WABT to produce anything meaningful, we need to run it on a debug build of wormchain_accounting's wasm artifact. build a debug wasm artifact:

    # from wormhole/cosmwasm
    cargo build --target wasm32-unknown-unknown

Now run WABT's wasm2wat on the debug wasm build of wormchain_accounting.wasm

    # note: my path for cloned repos is ~/code, replace it with your path.
    ~/code/wabt/build/wasm2wat ~/code/wormhole/cosmwasm/target/wasm32-unknown-unknown/debug/wormchain_accounting.wasm --generate-names -f -v -o ~/code/wormhole/cosmwasm/wormchain_accounting.wat

open the newly created `cosmwasm/wormchain_accounting.wat` file and find `f64.load`. The two results are the callstacks that throw errors when deploying the contract.

# references

- https://twitter.com/ibossptk/status/1575435046076358656
- https://medium.com/cosmwasm/debugging-floating-point-generation-in-rust-wasm-smart-contract-f47d833b5fba

hendrikhofstadt (approved):

hendrikhofstadt (approved):

PR #1923, opened 2022-11-15
Protect token implementation for arbitrum&optimism

kcsongor:
This PR protects from a temporary DoS risk arising from the `initialize()` function only partially defining the `tokenContract` variable, combined with the fact that `initialize()` is public. This PR fixes the immediate issue, will follow up with a proposal to eliminate this class of bugs in due course.

The EVM upgrade test is expected to fail for this one, because it simulates an upgrade against ethereum, which in this case is expected to revert (since this upgrade targets arbitrum and optimism exclusively).

The token implementations in question are:
https://arbiscan.io/address/0x53b56de645b9de6e5a40ace047d1c74e8b42eccb
https://optimistic.etherscan.io/address/0xb91e3638f82a1facb28690b37e3aae45d2c33808

### DIFF
--- a/ethereum/contracts/bridge/BridgeImplementation.sol
+++ b/ethereum/contracts/bridge/BridgeImplementation.sol
@@ -18,43 +18,26 @@ contract BridgeImplementation is Bridge {
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
         address tokenContract;
-        uint256 evmChainId;
-        uint16 chain = chainId();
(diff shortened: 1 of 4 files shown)


njkumr (approved):
lgtm!

bruce-riley (dismissed):

njkumr (approved):

ckeun (approved):

bruce-riley (approved):

PR #1929, opened 2022-11-15
Sui/more watcher updates

jumpsiegel:


node/pkg/watchers/sui/watcher.go
```diff
@@ -84,62 +133,57 @@ func NewWatcher(
...
 	}
 
-	if !txDigest.Exists() || !timestamp.Exists() || !packageId.Exists() || !account.Exists() || !consistency_level.Exists() || !nonce.Exists() || !sender.Exists() || !sequence.Exists() {
-		return errors.New("Missing event fields")
+	if moveEvent.Fields == nil {
+		return nil
+	}
+	fields := *moveEvent.Fields
+	if (fields.ConsistencyLevel == nil) || (fields.Nonce == nil) || (fields.Payload == nil) || (fields.Sender == nil) || (fields.Sequence == nil) {
+		return nil
```
jynnantonix, line 153: Should this be an error since some fields are missing?

jynnantonix (approved):

node/pkg/watchers/sui/watcher.go
jumpsiegel: yes but...  move generates a lot of automatic events as well.  For example, whenever you mutate an account, create a new account, etc.  I was thinking I could add another field into the event saying "THIS is the droid you are looking for" and error if THAT event is missing fields as apposed to how it currently works where it looks for any event that contains all the fields we are looking for.   This is what I did in the near contract.   I will make an issue to modify the contracts and then update the watcher to match

PR #1932, opened 2022-11-16
cosmwasm: Add backfill method for accounting

### LINKED ISSUE #1883: Bootstrap mechanism for initial on-chain state
There have been a lot of transfers since the inception of the portal bridge and the initial state for accounting is going to be much larger than the maximum transaction size for cosmwasm contracts.  So we can't use the normal `instantiate` method for initializing the state.

Instead we need to add a "bootstrap" ExecuteMsg to build up the on-chain state in "max transaction size" chunks.  In the initial `instantiate` message, we will send a list of the hashes of the bootstrap messages.  Then we will send each bootstrap message in separate transactions.  While the contract is in its bootstrapping state, no other execute methods will be allowed.  Once the contract has received all the bootstrap messages from the `instantiate` message list, it will permanently switch over into the activated state.  At this point, only non-boostrap execute messages will be allowed.

We have a couple of choices for the permanent switch:

* Keep track of it as a value on-chain and check it for every execute message.
* Use a completely different contract for the initial bootstrap that uses the same on-chain data structures and then upgrade to the actual contract once all the on-chain state has been initialized.

My preference is for the latter since it more closely matches the intent and is much cleaner in the code itself.

### LINKED ISSUE #1884: Method for backfilling transfers in accounting
It's possible that some transfers might not end up in the accounting contract before a VAA is produced for them.  This is going to happen for transfers between the time that we take a snapshot to produce the initial state and the time that the guardians switch over to checking all token transfers before signing them.

To deal with this we should have a "backfill" method to update the on-chain state with the missing transfers.  To avoid creating extra work we can reuse the signed VAA as the input parameter for this method.  The contract would then just check if the transfer was already committed in its internal state and if not, then add it to the state.

jynnantonix:
Add a mechanism to backfill missing transfer messages.  This will also
be used to initialize the on-chain state as there is too much data to
initialize the contract via the normal `instantiate` mechanism.

Fixes #1883, fixes #1884.

### DIFF
--- a/cosmwasm/packages/wormhole-bindings/src/fake.rs
+++ b/cosmwasm/packages/wormhole-bindings/src/fake.rs
@@ -1,17 +1,14 @@
 use std::{cell::RefCell, collections::BTreeSet, fmt::Debug, rc::Rc};
 
-use anyhow::{anyhow, bail, ensure};
+use anyhow::{anyhow, bail, ensure, Context};
 use cosmwasm_std::{to_binary, Addr, Api, Binary, BlockInfo, CustomQuery, Empty, Querier, Storage};
 use cw_multi_test::{AppResponse, CosmosRouter, Module};
-use k256::ecdsa::{
-    self,
-    signature::{Signature as SigT, Signer, Verifier},
-    SigningKey,
-};
+use k256::ecdsa::{recoverable, signature::Signer, SigningKey};
 use schemars::JsonSchema;
 use serde::de::DeserializeOwned;
+use wormhole::vaa::{digest, Signature};
 
-use crate::{Signature, WormholeQuery};
+use crate::WormholeQuery;
 
 #[derive(Debug)]
 struct Inner {
@@ -70,15 +67,14 @@ impl WormholeKeeper {
     }
 
     pub fn sign(&self, msg: &[u8]) -> Vec<Signature> {
+        let d = digest(msg).unwrap();
         self.0
             .borrow()
             .guardians
             .iter()
             .map(|g| {
-                <SigningKey as Signer<ecdsa::Signature>>::sign(g, msg)
-                    .as_bytes()
-                    .to_vec()
-                    .into()
+                let sig: recoverable::Signature = g.sign(&d.hash[..]);
+                sig.as_ref().try_into().unwrap()
             })
             .enumerate()
             .map(|(idx, sig)| Signature {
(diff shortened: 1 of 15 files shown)


jynnantonix:
* dev.v2
  * **PR #1919** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1919" target="_blank">Graphite</a> 
    * **PR #1920** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1920" target="_blank">Graphite</a> 
      * **PR #1932** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1932" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1932?utm_source=stack-comment).

conorpp (dismissed):

hendrikhofstadt (dismissed):

a5-pickle (approved):

jynnantonix:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/1932?utm_source=stack-comment).


Job Reference: amKr7GjyAVe9WLuUBSnW

cosmwasm/packages/wormhole-bindings/src/fake.rs
```diff
@@ -122,12 +118,18 @@ impl WormholeKeeper {
...
         if let Some(g) = this.guardians.get(sig.index as usize) {
-            let s = ecdsa::Signature::try_from(&*sig.signature).unwrap();
-            g.verifying_key()
-                .verify(data, &s)
-                .map(|()| Empty {})
-                .map_err(From::from)
+            let s = recoverable::Signature::try_from(&sig.signature[..])
+                .context("failed to decode signature")?;
+            let verifying_key = s
+                .recover_verify_key_from_digest_bytes(&d.secp256k_hash.into())
```
hendrikhofstadt, line 126: why did this become more complex?

hendrikhofstadt (approved):

a5-pickle (approved):

cosmwasm/packages/wormhole-bindings/src/fake.rs
jynnantonix: Going to mark this resolved so that github will let me merge but please let me know if you'd like me to make any changes to this in a follow-up PR.

PR #1935, opened 2022-11-16
sdk/js: added vaa repair script to utils

ckeun:
repairVaa.ts attempts to reconstruct a valid vaa from a vaa with an expired GS. First, it checks if the vaa is from the current guardian set. if not, it checks which signatures are still valid and tries to reconstruct the vaa with the remaining signatures. if unsuccessful, it returns the original vaa.


sdk/js/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,137 @@
...
+const ETHEREUM_CORE_BRIDGE = CONTRACTS["MAINNET"].ethereum.core;
+
+function hex(x: string): string {
+  return ethers.utils.hexlify(x, { allowMissingPrefix: true });
+}
+
+export async function getGuardianSets(
+  provider: ethers.providers.JsonRpcProvider
+) {
+  let result: any = {};
```
kcsongor, line 17: can we be a little more precise about the return type here? looks like it's actually a well-specified object

sdk/js/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export async function repairVaa(vaaHex: string, rpcUrl: string) {
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl);
+  const guardianSetsInfo = await getGuardianSets(provider);
+  const currentGuardianSetIndex = guardianSetsInfo.currentGuardianSetIndex;
+  const guardianSets = guardianSetsInfo?.guardianSet;
+  const guardianSetIndexes = Object.keys(guardianSets);
+  const vaaGuardianSetIndex = parseInt(vaaHex.slice(2, 10));
+  if (vaaGuardianSetIndex === parseInt(currentGuardianSetIndex)) {
+    console.log("Vaa has current guardian set index. no repair needed");
```
kcsongor, line 37: I think we should not have console logs in sdk functions

sdk/js/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,137 @@
...
+    (result.currentGuardianSetIndex = await core.getCurrentGuardianSetIndex());
+  result.guardianSet = {};
+  for (let i of Array(result.currentGuardianSetIndex + 1).keys()) {
+    let guardian_set = await core.getGuardianSet(i);
+    result.guardianSet[i] = { keys: guardian_set[0], expiry: guardian_set[1] };
+  }
+  return result;
+}
+
+export async function repairVaa(vaaHex: string, rpcUrl: string) {
```
kcsongor, line 29: would be good to have a pure version of this function that doesn't hit the rpc and instead just takes in the guardian set as an argument. Then another version can wrap the pure one and do the actual RPC query. Simpler to understand in separation

sdk/js/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,137 @@
...
+                "hex"
+              )}`
+            );
+            validSignatures.push(signature);
+          } else {
+            console.error(
+              `did not find a match for gs index=${currentIndex}, vaaGuardianPublicKey=${vaaGuardianPublicKey}, current public key=${currentGuardianPublicKey}`
+            );
+          }
+        } catch (err) {
```
kcsongor, line 88: maybe make the try block narrower? there's too much stuff happening inside to easily follow what exactly we're catching. Also, the exception there looks like a genuine error in the VAA, so might be best to not catch and just bubble it up to the caller - they're in a better position to handle an invalid VAA than this function

sdk/js/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,137 @@
...
+          .padStart(8, "0")}${numRepairedSignatures
+          .toString(16)
+          .padStart(2, "0")}${repairedSignatures}`;
+
+        const repairedVaa = `${newSignatureBody}${vaaHex.slice(
+          12 + numSignatures * 132
+        )}`;
+        return repairedVaa;
+      }
+    } catch (e) {
```
kcsongor, line 129: same as above.
1) the try is too big
2) not sure if worth try-catching anyway. just let the exception propagate

sdk/js/src/utils/repairVaa.ts
ckeun: so it might not be a genuine error in the VAA if that guardian at that index is no longer valid with the guardian set that was provided. and it looks like the recoverAddress function will throw an error if it can't find a valid address... so i kept the try/catch there so it can loop through the rest of the signatures.

kcsongor (dismissed):

sdk/js/src/utils/repairVaa.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  expect(() => {
+    repairVaa(vaa, guardianSetData);
+  }).toThrow("Attempt to access memory outside buffer bounds");
+});
+
+test("Use current guardian set", async () => {
```
kev1n-peters, line 59: Can we do `test.skip` here and the test below so we don't hit the public ankr RPC in the CI tests?

kcsongor (dismissed):

kcsongor (approved):

kev1n-peters (approved):

PR #1939, opened 2022-11-16
initial wormchain mainnet information

conorpp:
This is the genesis and initial network configuration guardians may use to join the network.  We will update the `persistent_peers` later after more guardians have joined.

I'd like to host this in the repo for people to download easily.

### DIFF
--- a/wormchain/mainnet/genesis.json
+++ b/wormchain/mainnet/genesis.json
@@ -1,0 +1,351 @@
+{
+  "genesis_time": "2022-09-27T22:08:43.044644477Z",
+  "chain_id": "wormchain",
+  "initial_height": "1",
+  "consensus_params": {
+    "block": {
+      "max_bytes": "22020096",
+      "max_gas": "-1",
+      "time_iota_ms": "1000"
+    },
+    "evidence": {
+      "max_age_num_blocks": "100000",
+      "max_age_duration": "172800000000000",
+      "max_bytes": "1048576"
+    },
+    "validator": {
+      "pub_key_types": [
+        "ed25519"
+      ]
+    },
+    "version": {}
+  },
+  "app_hash": "",
+  "app_state": {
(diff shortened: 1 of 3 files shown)


jynnantonix (approved):

wormchain/mainnet/genesis.json
```diff
@@ -0,0 +1,351 @@
...
+    },
+    "genutil": {
+      "gen_txs": [
+        {
+          "body": {
+            "messages": [
+              {
+                "@type": "/cosmos.staking.v1beta1.MsgCreateValidator",
+                "description": {
+                  "moniker": "jumpcrypto",
```
evan-gray, line 144: Is this "jumpcrypto" indicative of the expectation that they are / were the first guardian to stand up the chain?

wormchain/mainnet/genesis.json
```diff
@@ -0,0 +1,351 @@
...
+          ]
+        }
+      ],
+      "supply": [],
+      "denom_metadata": [
+        {
+          "description": "Wormchain's native test asset",
+          "denom_units": [
+            {
+              "denom": "utest",
```
evan-gray, line 65: should there be a test token?

evan-gray (approved):
nothing immediately jumped out at me, but left a couple comments

wormchain/mainnet/genesis.json
conorpp: Yes we will start with one node and vote the other guardians in, allowing them to join async

wormchain/mainnet/genesis.json
conorpp: Yes I'd like to use the test token to create the guardian accounts when guardians generate their validator addresses. I.e.:
1. guardian generates validator address.
2. we send them a test token as a quick way to create the account for their address.

PR #1943, opened 2022-11-17
ethereum: Set impl initialised during setup

### LINKED ISSUE #1930: Remove `initialize()` undefined behaviour
## Background & problem

The `initialize()` family of functions is a constant source of confusion for newcomers and bug reporters alike. The discussion in this post applies to all three of the core bridge, token bridge, and NFT bridge contracts, so I will use the core bridge for illustration purposes, but the points generalise to the other two accordingly. The final proposed solution to this problem is very simple, but requires a fair bit of exposition to understand the motivation, so I include some discussion for posterity's sake.

The core bridge uses the upgradeable proxy pattern (https://docs.openzeppelin.com/contracts/4.x/api/proxy) to overcome the fact that EVM doesn't natively support upgradeable contracts. In this setup, the user interacts with a proxy contract, which is a thin layer that delegates all function calls to some underlying implementation contract. The implementation contract can be dynamically swapped out to a different one, effectively allowing upgrading the behaviour of the contract (while the proxy keeps its address).

A common problem with this pattern is that the implementation contract needs to handle initial setup logic for when the contract is first deployed, and extra care must be taken that this initialisation step is not replayable after subsequent upgrades. The way the wormhole contracts tackle this issue is by using a two-phase deployment method, where the initial implementation is a single-use setup contract that performs the initial setup:
https://github.com/wormhole-foundation/wormhole/blob/12f578c4bc1af10e966b5805886b523c56b28231/ethereum/contracts/Setup.sol#L11-L39
finally, after setting the state variables, the setup upgrades itself to the _actual_ implementation:
https://github.com/wormhole-foundation/wormhole/blob/12f578c4bc1af10e966b5805886b523c56b28231/ethereum/contracts/Setup.sol#L37
The actual implementation contract has no code paths that handle this initial setup (as they have just been performed by `Setup`). This separation of concerns between the setup contract and the implementation contract eliminates a whole class of bugs by construction.

Performing a contract upgrade however may require some one-off code to be executed during the upgrade process. This may include things such as state migration, or initialising storage required for newly added functionality. This one-off logic is added to the `initialize()` method, abusing the name somewhat. As of writing, it contains the migration code for setting the EVM chain ids for each chain
https://github.com/wormhole-foundation/wormhole/blob/12f578c4bc1af10e966b5805886b523c56b28231/ethereum/contracts/Implementation.sol#L33-L60
this function is never meant to be invoked by hand. Instead, it's delegatecalled by the upgrade logic upon each upgrade
https://github.com/wormhole-foundation/wormhole/blob/12f578c4bc1af10e966b5805886b523c56b28231/ethereum/contracts/Governance.sol#L180

To recap, the state machine is:

```
                                  +---------+
                                  |         | upgrade()
+-------+  setup()  +-------------v--+      | => initialize()
| Setup |---------> | Implementation | -----+
+-------+           +----------------+

```

The `initializer` modifier on `initialize()` ensures that it can only be called at most once, so when the upgrade logic calls it on a new implementation, it won't be callable again by anyone. However, it's not invoked on the first deployment, so technically anyone can call `initialize()` on a freshly deployed contract. Traditionally, the `initialize()` function is empty when deploying a new contract, but this is only a social contract, and not enforced. Deploying a contract with a non-empty `initialize()` method leads to undefined behaviour, which is undesirable and dangerous. https://github.com/wormhole-foundation/wormhole/pull/1923 fixes one such concrete scenario.

## Proposed solution

In `setup()` we call the `setInitialized(address)` setter to mark the newly deployed implementation contract as initialised. This will make it impossible for anyone to call the `initialize()` method on this implementation, only for subsequent upgrades. In those cases, it will have been called by the upgrade code itself already.

## Alternative solution

Ideally we'd like to make `initialize()` hidden from external callers so it can only be called by the contract itself. A natural thing to attempt would be to make this function `internal` instead of `public`, but that doesn't work. The upgrade code path of the *old* contract must be able to call the `initialize()` method of the *new* contract. `internal` functions do not get assigned a 4 byte identifier and an entry in the jump table of the contract's entry point, and instead are compiled to direct jumps to the appropriate location within the bytecode. The old contract has to treat this as an external function call, so the function must be public. Instead, we can add a check to the `initialize()` function
```solidity
require(msg.sender == address(this))
```
which ensures that the function can only be called by the contract itself. However, this won't readily work, because `initialize()` is *delegatecall*ed during upgrade, which retains the `msg.sender` of the original caller, i.e. `initialize()` won't see the implementation contract as the `msg.sender` but whoever called `upgrade()` in the first place. To fix this, we need to change the `delegatecall` above to
```solidity
        (bool success, bytes memory reason) = address(this).call(abi.encodeWithSignature("initialize()"));
```

### Pros
- Easier to understand who calls `initialize()` just by looking at it.

### Cons
- Awkward to roll out. Changing to this scheme requires two upgrades: in the first, changing the `delegatecall` to `call`, then in the second, adding the `msg.sender == address(this)` check. These can't be done in the same upgrade, because `initialize()` is *always* called by the *old* implementation's code, which will at first still perform a delegatecall, making the the `msg.sender` check fail.

## Conclusion

With this problem solved, we may finally lay all the remaining `initialize()` issues to rest. As a bonus, we could consider renaming `initialize()` to `migrate()` to further clarify the role of this function.

kcsongor:
See #1930 for more context on this change.
One small step for man, one giant leap for the bug bountry program.

### DIFF
--- a/ethereum/contracts/Setup.sol
+++ b/ethereum/contracts/Setup.sol
@@ -35,5 +35,9 @@ contract Setup is Setters, ERC1967Upgrade {
         setEvmChainId(evmChainId);
 
         _upgradeTo(implementation);
+
+        // See https://github.com/wormhole-foundation/wormhole/issues/1930 for
+        // why we set this here
+        setInitialized(implementation);
     }
 }
--- a/ethereum/contracts/bridge/BridgeSetup.sol
+++ b/ethereum/contracts/bridge/BridgeSetup.sol
@@ -36,5 +36,9 @@ contract BridgeSetup is BridgeSetters, ERC1967Upgrade {
         setEvmChainId(evmChainId);
 
         _upgradeTo(implementation);
+
+        // See https://github.com/wormhole-foundation/wormhole/issues/1930 for
+        // why we set this here
+        setInitialized(implementation);
     }
 }
--- a/ethereum/contracts/nft/NFTBridgeSetup.sol
+++ b/ethereum/contracts/nft/NFTBridgeSetup.sol
@@ -33,5 +33,9 @@ contract NFTBridgeSetup is NFTBridgeSetters, ERC1967Upgrade {
         setEvmChainId(evmChainId);
 
         _upgradeTo(implementation);
+
+        // See https://github.com/wormhole-foundation/wormhole/issues/1930 for
+        // why we set this here
+        setInitialized(implementation);
     }
 }


njkumr (dismissed):

bruce-riley (dismissed):

bruce-riley (approved):

kcsongor:
@njkumr can you please re-review? Your review got dismissed

njkumr (approved):

PR #1951, opened 2022-11-17
Basic Sui watcher support

jumpsiegel:


kcsongor:
Shouldn't we merge the core contract in first? That way, it can be deployed in tilt and the watcher actually connected to it

jumpsiegel:
No, this can be deployed and even run without the contracts deployed.  It doesn't cause an error

jynnantonix (approved):

kcsongor (approved):

PR #1952, opened 2022-11-17
Sui: Update deploy.sh to register emitter and init_and_share core and token bridge states

optke3:
- augment `deploy.sh` to also `init_and_share` core and token bridge states
- for core bridge, we set the`chain_id`, `governance_chain_id`, `governance_contract`, `initial_guardians` via init_and_share
- register emitter for token bridge before calling `init_and_share`

With the above changes, the core and token bridge state objects become shared, so they can be accessed by anyone.


sui/token_bridge/Move.toml
```diff
@@ -7,4 +7,4 @@ Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-fram
 Wormhole = { local = "../wormhole" }
 
 [addresses]
-token_bridge="0x0"
\ No newline at end of file
+token_bridge = "0xf8fdb4bacdec4faa9416d6ebdedd2231a09cc60a"
```
jumpsiegel, line 10: We sure you want to check this in?  WIth a address?

optke3: my bad...still editing haha

jumpsiegel (approved):

PR #1956, opened 2022-11-18
node: gov: separate governor config and status message prefix

### LINKED ISSUE #1957: node: gov: Governor status and governor config p2p messages use the same message prefix in hash computation
Governor status and governor config p2p messages use the same message prefix in hash computation: https://github.com/wormhole-foundation/wormhole/blob/a6156ed1c8bf30e5f6b9f3a9ba748d1ee2218f5c/node/pkg/governor/governor_monitoring.go#L501
https://github.com/wormhole-foundation/wormhole/blob/a6156ed1c8bf30e5f6b9f3a9ba748d1ee2218f5c/node/pkg/governor/governor_monitoring.go#L580

## Expected Behavior

Every message type should have its own message prefix such that they cannot be confused. Message confusion could happen e.g. when a guardian signs a governor config message and an attacker causes that message to be consumed somewhere as a governor status message, leading to undefined and potentially undesired behavior. 

## Possible Solution
Define a new message prefix for gov config/status messages respectively. See https://github.com/wormhole-foundation/wormhole/pull/1956. 
The drawback is that this would break any system currently verifying the signatures of said messages with the old message prefixes. Not sure if implementing a deprecation schedule and migration path would be worth the effort though. Better to fix this sooner than later before other integrators start depending on this format.

tbjump:
fixes #1957

### DIFF
--- a/node/pkg/governor/governor_monitoring.go
+++ b/node/pkg/governor/governor_monitoring.go
@@ -461,7 +461,8 @@ func (gov *ChainGovernor) CollectMetrics(hb *gossipv1.Heartbeat, sendC chan []by
 	}
 }
 
-var governorMessagePrefix = []byte("governor|")
+var governorMessagePrefixConfig = []byte("governor_config_000000000000000000|")
+var governorMessagePrefixStatus = []byte("governor_status_000000000000000000|")
 
 func (gov *ChainGovernor) publishConfig(hb *gossipv1.Heartbeat, sendC chan []byte, gk *ecdsa.PrivateKey, ourAddr ethCommon.Address) {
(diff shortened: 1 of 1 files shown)


tbjump:
marking as draft because I want your feedback on this approach first @bruce-riley

bruce-riley:
Need to check with @ckeun that this change won't break the fly.

ckeun:
> Need to check with @ckeun that this change won't break the fly.

i don't think this change will impact the fly...

bruce-riley (approved):

bruce-riley (approved):

claudijd (approved):

PR #1958, opened 2022-11-18
node: enforce minimum heartbeat message length

tbjump:
Enforce a minimum heartbeat message length such that signed observations, which have the format `signature(hash(hash(observation))` cannot be replayed as heartbeat messages, which have the format `signature(hash("heartbeat|..."))` in case where `hash(observation) = "heartbeat|...". 

I can't think of any concrete security risk with the current implementation, but it's better to tidy things up.

### DIFF
--- a/node/pkg/p2p/p2p.go
+++ b/node/pkg/p2p/p2p.go
@@ -443,6 +443,11 @@ func processSignedHeartbeat(from peer.ID, s *gossipv1.SignedHeartbeat, gs *node_
 
 	digest := heartbeatDigest(s.Heartbeat)
 
+	// SECURITY: see whitepapers/0009_guardian_key.md
+	if len(heartbeatMessagePrefix)+len(s.Heartbeat) < 34 {
+		return nil, fmt.Errorf("invalid message: too short")
+	}
+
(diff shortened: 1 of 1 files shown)


node/pkg/p2p/p2p.go
```diff
@@ -443,6 +443,11 @@ func processSignedHeartbeat(from peer.ID, s *gossipv1.SignedHeartbeat, gs *node_
 
 	digest := heartbeatDigest(s.Heartbeat)
 
+	// SECURITY: see whitepapers/0009_guardian_key.md
```
bruce-riley, line 446: I don't see this whitepaper in GH. I assume it will be coming soon?

bruce-riley (approved):

node/pkg/p2p/p2p.go
tbjump: yes, PR: https://github.com/wormhole-foundation/wormhole/pull/1960

jynnantonix (approved):

jynnantonix (approved):

hendrikhofstadt (approved):

PR #1959, opened 2022-11-18
node: p2p: change message type string for signed observation

tbjump:
same vain as https://github.com/wormhole-foundation/wormhole/pull/1958 but in this case I think it's better to just update the message type since I'd assume nobody external is parsing these kinds of messages from the p2p network. 

fixes https://github.com/wormhole-foundation/wormhole/issues/1991

### DIFF
--- a/node/pkg/p2p/p2p.go
+++ b/node/pkg/p2p/p2p.go
@@ -443,6 +443,11 @@ func processSignedHeartbeat(from peer.ID, s *gossipv1.SignedHeartbeat, gs *node_
 
 	digest := heartbeatDigest(s.Heartbeat)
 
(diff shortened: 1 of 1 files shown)


evan-gray:
@tbjump by "no one external" do you mean no one who isn't using `p2p.go`? The [wormhole-explorer](https://github.com/wormhole-foundation/wormhole-explorer/blob/main/fly/main.go) certainly uses that as may anyone.

tbjump:
I believe the answer is "yes"...
Created an alternative solution: https://github.com/wormhole-foundation/wormhole/pull/1992
Also created an issue to discuss further: https://github.com/wormhole-foundation/wormhole/issues/1991

PR #1961, opened 2022-11-18
sui/0_16_images: update to 0.16

jumpsiegel:


kcsongor:
Thanks! the Sui test is failing, but it looks like a contract error so we can just fix that separately

kcsongor (approved):

PR #1976, opened 2022-11-22
sui: create wrapped token

optke3:
## Updates
The old create wrapped workflow no longer works, because of changes to Sui. We experimented with several approaches for creating a wrapped token (some new, some old), including
- transferable one-time witness (TOTW)
- several variants of one-time witness (OTW) (question was where to call `coin::create_currency` )

One variant of the OTW approach worked, allowing us to deploy a wrapped asset and register it with the token bridge.

## Learnings

We learned that a one-time witness has to be created by the Sui runtime in an `init` function. In addition, the `init` function cannot take any arguments (such as token bridge and core bridge objects). As a consequence, we must call `coin::create_currency<COIN_WITNESS>(...vaa...)` in the `init` function in `coin.move`. We paste the token attestation VAA directly in the init function (hardcoded). `Init` obtains a treasury cap, and finally return it to the user. Then we let the user call `token_bridge::create_wrapped<COIN_WITNESS>(...treasury_cap...) `to finally give the treasury cap the to token bridge.

## Deployment Steps
Navigate to the `wormhole/sui `folder. Make sure that in all Move.toml files, the address under [addresses] is set to "0x0". Then
1. `./scripts/deploy.sh`
2. `./scripts/deploy_coin.sh`
3. `./scripts/create_wrapped.sh`


sui/coin/sources/coin_witness.move
```diff
@@ -1,15 +1,25 @@
...
     struct COIN_WITNESS has drop {}
 
-    // publish this module to call into token_bridge and create a wrapped coin
-    fun init(_ctx: &mut TxContext) {
-        // paste the VAA below
-        let _vaa = x"deadbeef00001231231231";
-        // create_wrapped_coin(vaa, COIN_WITNESS {})
+    fun init(coin_witness: COIN_WITNESS, ctx: &mut TxContext) {
+        // Step 1. Paste token attestation VAA below
+        let vaa_bytes = x"0100000000010080366065746148420220f25a6275097370e8db40984529a6676b7a5fc9feb11755ec49ca626b858ddfde88d15601f85ab7683c5f161413b0412143241c700aff010000000100000001000200000000000000000000000000000000000000000000000000000000deadbeef000000000150eb23000200000000000000000000000000000000000000000000000000000000beefface00020c424545460000000000000000000000000000000000000000000000000000000042656566206661636520546f6b656e0000000000000000000000000000000000";
```
heyitaki, line 14: we should be able to init witness without having to manually paste VAA

heyitaki (approved):

hq
```diff
@@ -0,0 +1,21 @@
+  aptos/integration[m
```
kcsongor, line 1: what's this file?

sui/token_bridge/sources/wrapped.move
```diff
@@ -1,85 +1,122 @@
...
+        let t_cap_store = treasury::create_treasury_cap_store<CoinType>(treasury_cap, ctx);
+        let origin_chain = asset_meta::get_token_chain(&metadata);
+        let external_address = asset_meta::get_token_address(&metadata);
+        let wrapped_asset_info = bridge_state::create_wrapped_asset_info(origin_chain, external_address, ctx);
+
+        bridge_state::store_treasury_cap<CoinType>(bridge_state, t_cap_store);
+        transfer::transfer(wrapped_asset_info, tx_context::sender(ctx));
+    }
+
+    public entry fun create_wrapped_coin_test_5<CoinType>(
```
kcsongor, line 65: should these test functions be removed?

PR #1983, opened 2022-11-22
sui: byte addressing for Sui CoinTypes

optke3:
Introduce byte-addressing for CoinTypes on-chain. The byte addresses are taken from an increasing sequence of numbers starting from 0.

- `assign_coin_type_bytes_address`
- `get_coin_type_bytes_address`

Write tests for these as well.

### DIFF
--- a/sui/token_bridge/sources/attest_token.move
+++ b/sui/token_bridge/sources/attest_token.move
@@ -4,7 +4,8 @@ module token_bridge::attest_token {
     use sui::transfer::transfer;
     use sui::tx_context::{Self, TxContext};
 
-    use token_bridge::bridge_state::{BridgeState};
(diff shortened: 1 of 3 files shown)


kcsongor:
let's start indexing at 1. solidity can't tell the difference between 0 and missing data, and better to safeguard us against any potential issues down the line

kcsongor (approved):
did some refactoring and fixed the bug where the VAA was not verified at all.

The key change I made is removing the separate object storing the coin's id, and just used the (already existing) `token_address` field in `NativeAssetInfo`. Now there's no separate function to call to register the id, `register_native_asset` takes care of it. Updated the tests accordingly

kcsongor:
closes #2027

PR #1985, opened 2022-11-23
Build wormchaind with Ledger support by default

ruuda:
We needed to apply these changes to be able to generate the operator address with a Ledger.

If you want to make it optional, you might take inspiration from[ Osmosis’ makefile](https://github.com/osmosis-labs/osmosis/blob/9e178a631f91ffc91c51f3665ed915c9f15e1807/Makefile#L6).


conorpp (approved):

jynnantonix (approved):

conorpp:
Thanks @ruuda !

PR #1987, opened 2022-11-23
sui/token_bridge: Safer wrapped asset creation

kcsongor:
With this setup we no longer have to trust (or verify) that the OTW initialiser doesn't do anything funny, as it's guaranteed by the type system.


heyitaki (approved):
nice! i like the fact that we bundle `vaa_bytes` with the `treasury_cap` now so `create_wrapped_coin` can no longer be called with diff bytes than were used to create the capability

PR #1960, opened 2022-11-18
whitepaper: add whitepaper explaining guardian key usage

tbjump:

PR #1988, opened 2022-11-23
node: remove race condition detector

tbjump:
race condition detector should not be enabled on a dev build because it can lead to ballooning memory consumption. We can think about adding it back in into a separate test build

### DIFF
--- a/Dockerfile.node
+++ b/Dockerfile.node
@@ -19,7 +19,7 @@ RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
 COPY node node
 COPY sdk sdk
 
-ARG GO_BUILD_ARGS=-race
+ARG GO_BUILD_ARGS=
 
 RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
   cd node && \


kcsongor:
please follow https://github.com/wormhole-foundation/wormhole/blob/23bdee60352243b3dd3883961d81f214f9acd933/CONTRIBUTING.md#L69-L95

kcsongor (changes requested):
please follow https://github.com/wormhole-foundation/wormhole/blob/23bdee60352243b3dd3883961d81f214f9acd933/CONTRIBUTING.md#L69-L95

tbjump:
@kev1n-peters We haven't been able to precisely pin down all the conditions in which the memory consumption explodes but it seems to be getting worse the more goroutines there are and also getting worse with golang 1.19. 

With golang 1.17 we saw the guardiand crash every ~20m and with golang 1.19 every ~1m. There is something seemingly related to the race condition checker that causes memory consumption to explode instantly and resulting in an oomkill.

tbjump:
@kev1n-peters this fixes e.g. the tilt tests in https://github.com/wormhole-foundation/wormhole/pull/1960 and many other PRs

kcsongor (approved):

Issue #1881, opened 2022-11-10
Add bindings for cosmwasm contracts to query the native cosmos wormhole module


jynnantonix:
Proposed API:

```rust
#[cw_serde]
pub struct Signature {
    // The index of the observer in the observer set.
    pub index: u16,

    // The signature.
    pub signature: Binary,
}

#[cw_serde]
#[derive(QueryResponses)]
pub enum WormholeQuery {
    // Verifies that `data` has been signed by a quorum of observers from `observer_set_index`.
    #[returns(Empty)]
    VerifyQuorum {
        data: Binary,
        guardian_set_index: u32,
        signatures: Vec<Signature>,
    },

    // Verifies that `data` has been signed by an observer from `observer_set_index`.
    #[returns(Empty)]
    VerifySignature {
        data: Binary,
        guardian_set_index: u32,
        signature: Signature,
    },

    // Returns the number of signatures necessary for quorum for the given observer set index.
    #[returns(u32)]
    CalculateQuorum { guardian_set_index: u32 },
}

```

jynnantonix:
@kcsongor @hendrikhofstadt @conorpp for API review

kcsongor:
Overall looks good, what does the `block_time` mean in the `VerifySignature` query?

jynnantonix:
`block_time` is the value of [`env.block.height`](https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.BlockInfo.html#structfield.height).  Maybe it would be better if it was renamed to `block_height`?

kcsongor:
`block_height` sounds a little better. Is the idea there to verify if a VAA was valid at a given time in the past (to account for guardian set upgrades)?

jynnantonix:
The block height is used to check if the guardian set has expired.  If there's a way for the native module to independently fetch the current block height while handling this query then we can remove this field completely.

kcsongor:
The handler in the native module will have access to the `Context` struct https://github.com/wormhole-foundation/cosmos-sdk/blob/main/types/context.go#L25-L42 which in turn contains the `BlockHeight` https://github.com/wormhole-foundation/cosmos-sdk/blob/main/types/context.go#L50.

conorpp:
This looks okay to me.

[...]

PR #1994, opened 2022-11-23
wormchain: add accounting contract bindings

nik-suri:
Adds custom wormhole query handler for integration with cosmwasm contracts such as accounting.

Should not be merged before https://github.com/wormhole-foundation/wormhole/pull/1920

This PR will be updated with contract integration tests once https://github.com/wormhole-foundation/wormhole/pull/1920 is merged.


wormchain/x/wormhole/keeper/query_plugins.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			return []byte{}, nil
+		}
+		if wormholeQuery.VerifySignature != nil {
+			// handle the verify signature query
+			err := keeper.VerifySignature(ctx, wormholeQuery.VerifySignature.Data, wormholeQuery.VerifySignature.GuardianSetIndex, wormholeQuery.VerifySignature.Signature)
+			if err != nil {
+				return nil, err
+			}
+			return []byte{}, nil
```
conorpp, line 63: There is no marshaled type to return?

conorpp (approved):

conorpp:
nice work!

wormchain/x/wormhole/keeper/query_plugins.go
nik-suri: In the proposed API https://github.com/wormhole-foundation/wormhole/issues/1881 the return types for these queries are `cosmwasm_std::Empty` -- https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Empty.html.

Not sure exactly what they serialize to but empty byte array seems to be a reasonable placeholder until https://github.com/wormhole-foundation/wormhole/pull/1920 is merged and I can test the contract e2e.

jynnantonix (dismissed):

conorpp (approved):

jynnantonix (dismissed):

bruce-riley (approved):

wormchain/x/wormhole/keeper/query_plugins.go
```diff
@@ -0,0 +1,78 @@
...
+type verifyQuorumParams struct {
+	Data             []byte           `json:"data"`
+	GuardianSetIndex uint32           `json:"guardian_set_index"`
+	Signatures       []*vaa.Signature `json:"signatures"`
+}
+
+type verifySignatureParams struct {
+	Data             []byte         `json:"data"`
+	GuardianSetIndex uint32         `json:"guardian_set_index"`
+	Signature        *vaa.Signature `json:"signature"`
```
jynnantonix, line 34: I think this is currently causing an encoding mismatch.  This type has a `SignatureData` field, which implements `MarshalJSON` but not `UnmarshalJSON`.  The `MarshalJSON` encodes the signature as a hex string but since there is no `UnmarshalJSON` it falls back to expecting an array of bytes.

Meanwhile, the rust bindings use `cosmwasm_std::Binary` for the same field, which encodes the signature into a base64 string.

As part of #1932 I changed this field to always encode as an array of bytes but given the existing `MarshalJSON` method for `vaa.SignatureData` that does something different we should define our own `Signature` type here for encoding purposes and then provide a conversion function to convert to `vaa.Signature`.

PR #1996, opened 2022-11-24
sui: simplify storage model

kcsongor:
This PR introduces a new module, `dynamic_set`, which acts like the storage model of traditional move. That is, objects are identified by their type, and there's at most a single one of them. Implemented as dynamic object fields, so rather than belonging to an account, the elements belong to another object (the bridge state in this case). This module allows cleaning up the dummy fields, and overall a little more concise.


heyitaki (approved):

optke3 (approved):
nice abstraction

PR #2001, opened 2022-11-24
devnet: add memory limits

### LINKED ISSUE #1984: Tilt/devnet: containers have no memory limit
Containers in devnet/ have no [memory resource limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) defined. If some component starts consuming much more memory, tests may continue to pass depending on the environment where they are being executed. 

## Expected Behavior

I propose to add memory resource limits based on current memory usage plus some buffer. This will allow catching changes that significantly increase memory usage.

tbjump:
fixes #1984 (see the issue for the rationale here)

I picked the memory limits based on running everything locally for 1h and picking the max memory consumption plus some buffer in some cases, see devnet/README.md. Everything together added up to more than 16GiB, which is the limit in `start-recommended-minikube`, so I move some services to be jobs such that the peak memory consumption stays under 16GiB. 

I'm no expert here, so would appreciate feedback on this approach.

### DIFF
--- a/devnet/README.md
+++ b/devnet/README.md
@@ -1,0 +1,43 @@
+# Memory allocation
+| container          | measured VmHWM | set limit | service/job           |
+|--------------------|----------------|-----------|-----------------------|
+| guardiand          | 2777Mi         | 3300Mi    | service               |
+| ganache            | 471Mi          | 500Mi     | service               |
+| eth-deploy         | ?              | 1Gi       | job                   |
+| mine               | 271Mi          | 300Mi     | service               |
+| spy                | 116Mi          | 150Mi     | service               |
+| algorand-postgres  | ?              | 80Mi      | service               |
+| algorand-algod     | 644Mi          | 1000Mi    | service               |
+| algorand-indexer   | ?              | 100Mi     | service               |
+| algorand-contracts | ?              | 200Mi     | service, could be job |
+| aptos-node         | 143Mi          | 500Mi     | service               |
+| aptos-contracts    | ?              | 300Mi     | service, could be job |
+| btc-node           | 310Mi          | 350Mi     | service               |
+| near-node          | 639Mi          | 700Mi     | service               |
+| near-deploy        | 462Mi          | 500Mi     | job                   |
+| solana-devnet      | 1769Mi         | 2000Mi    | service               |
+| solana-setup       | ?              | 750Mi     | service, could be job |
+| spy-listener       | ?              | 150Mi     | service               |
+| spy-relayer        | 76Mi           | 200Mi     | service               |
+| spy-wallet-monitor | 81Mi           | 100Mi     | service               |
+| spy                | 102Mi          | 120Mi     | service               |
+| terra-terrad       | 343Mi          | 400Mi     | service               |
+| terra-contracts    | ?              | 200Mi     | could be job          |
+| fcd-postgres       | ?              | 50Mi      | service               |
+| fcd-collector      | ?              | 500Mi     | service               |
+| fcd-api            | ?              | 200Mi     | service               |
+| wormchaind         | 559Mi          | 1000Mi    | service               |
+| sdk-ci-tests       | ?              | 2500Mi    | job                   |
+| spydk-ci-tests     | ?              | 1000Mi    | job                   |
+
+## Debugging
+* Detecting oomkill:
+    * The symptom of an oomkill is the container being killed.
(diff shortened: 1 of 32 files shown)


jynnantonix (commented):
Change itself looks fine but please reword your commits to follow the [naming convention](https://github.com/wormhole-foundation/wormhole/blob/dev.v2/CONTRIBUTING.md#commit-naming-convention).

I think the initial limits seems reasonable and we can always adjust later if necessary.

devnet/README.md
```diff
@@ -0,0 +1,40 @@
...
+| terra-contracts    | ?              | 200Mi     | could be job |
+| fcd-postgres       | ?              | 50Mi      | service      |
+| fcd-collector      | ?              | 500Mi     | service      |
+| fcd-api            | ?              | 200Mi     | service      |
+| wormchaind         | 559Mi          | 1000Mi    | service      |
+| sdk-ci-tests       | ?              | 2500Mi    | job          |
+| spydk-ci-tests     | ?              | 1000Mi    | job          |
+
+## Debugging
+* oomkill messages should be in `/var/log/messages`.
```
jynnantonix, line 34: <i class='graphite__hidden'>[Re: lines 34 to 34]<br/></i>

Would be nice to also include an example of what this message looks like.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2001?utm_source=unchanged-line-comment">Graphite</a>.</span>

devnet/README.md
```diff
@@ -0,0 +1,40 @@
+# Memory allocation
+| container          | measured VmHWM | set limit | service/job  |
```
jumpsiegel, line 2: While I think this is a great idea,  the error message when we bust a limit is non-existent.  How will we diagnose why the PR is failing?

tbjump: Maybe we could get this to get the error message: 
cmd: `kubectl get pod -o jsonpath='{.items[*].status.containerStatuses[*].lastState.terminated.exitCode},{.items[*].status.containerStatuses[*].lastState.terminated.reason}'` 
output: `137,OOMKilled`

tbjump: added more debugging instructions.

tbjump:
These changes increase the runtime of the tilt tests from ~8m to ~11m, but it is not the longest running github action, so overall runtime remains unchanged.

sdk/js/package.json
```diff
@@ -17,7 +17,7 @@
     "build-lib": "tsc -p tsconfig.json && tsc -p tsconfig-cjs.json && node scripts/copyEthersTypes.js",
     "build-all": "npm run build-deps && npm run build-lib",
     "test": "jest --config jestconfig.json --verbose",
-    "test-ci": "NEAR_NO_LOGS=true jest --config jestconfig.json --verbose --setupFiles ./ci-config.js --forceExit",
+    "test-ci": "NEAR_NO_LOGS=true jest --config jestconfig.json --verbose --setupFiles ./ci-config.js --forceExit --runInBand",
```
evan-gray, line 20: `--runInBand` is going to have a significant performance impact. I would strongly prefer to get all the sdk tests to be async and parallel safe (outside of what is purposefully in a describe block). Understandable as long as this is just set for now to get some better consistency in the tests and removed later.

tbjump: yes, the goal is to get all tests to run reliably first and then we will tackle performance improvements of the tests

node/pkg/watchers/near/finalizer.go
```diff
@@ -20,7 +20,7 @@ type Finalizer struct {
 }
 
 func newFinalizer(eventChan chan eventType, nearAPI nearapi.NearApi, mainnet bool) Finalizer {
-	finalizedBlocksCache, _ := lru.New(workerCountTxProcessing * quequeSize)
+	finalizedBlocksCache, _ := lru.New(100_000)
```
evan-gray, line 23: Why is this changing in this PR?

evan-gray (changes requested):
I would appreciate 3 separate commits following the naming conventions. One for `devnet:`, another for `sdk/js:`, and third for `node:` (though I'm not sure I understand why the change should be there)

node/pkg/watchers/near/finalizer.go
tbjump: Good catch. That shouldn't be in this PR. Will remove it.

tbjump:
verified that tilt ci tests now pass locally with only 16GB RAM.

jynnantonix (dismissed):

evan-gray:
Can we factor this out into multiple PRs to discuss. Notes from review
1. I don't think the SDK changes are necessary anymore.
2. The changes to the deploy / test scripts mean they no longer work outside tilt.
3. The switch to Jobs may have unintended consequences for resetting chains.

evan-gray:
Closing per the above comment, recognizing the discussion https://github.com/wormhole-foundation/wormhole/discussions/2226

PR #2012, opened 2022-11-28
sui/token_bridge: fix complete_transfer

kcsongor:
This PR fixes two issues in the sui token bridge `complete_transfer` function. Both of these fixes are lifted from the aptos contract, with small modifications.

1. Properly pays the fee recipient.
2. Fixes a critical security bug where the `CoinType` was not verified to match the information in the VAA.

Since 2. is a critical safety check, and we have yet to implement `complete_transfer_with_payload` which will require the same check, I decided to introduce a foolproof mechanism that guarantees that this check is performed. `mint` and `withdraw` now require a witness that ensures that `CoinType` has been verified. See the documentation for `VerifiedCoinType`.

### DIFF
--- a/sui/token_bridge/sources/bridge_state.move
+++ b/sui/token_bridge/sources/bridge_state.move
@@ -24,6 +24,8 @@ module token_bridge::bridge_state {
     const E_IS_NOT_REGISTERED_NATIVE_ASSET: u64 = 1;
     const E_COIN_TYPE_HAS_NO_REGISTERED_INTEGER_ADDRESS: u64 = 2;
     const E_COIN_TYPE_HAS_REGISTERED_INTEGER_ADDRESS: u64 = 3;
+    const E_ORIGIN_CHAIN_MISMATCH: u64 = 4;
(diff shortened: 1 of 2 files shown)


optke3 (approved):
lgtm

sui/token_bridge/sources/bridge_state.move
```diff
@@ -256,6 +253,43 @@ module token_bridge::bridge_state {
...
+    /// See the documentation for `VerifiedCoinType` above.
+    public fun verify_coin_type<CoinType>(
+        bridge_state: &BridgeState,
+        token_chain: U16,
+        token_address: ExternalAddress
+    ): VerifiedCoinType<CoinType> {
+        let coin_origin = origin_info<CoinType>(bridge_state);
+        assert!(coin_origin.token_chain == token_chain, E_ORIGIN_CHAIN_MISMATCH);
+        assert!(coin_origin.token_address == token_address, E_ORIGIN_ADDRESS_MISMATCH);
+        VerifiedCoinType {}
```
optke3, line 290: why not return  VerifiedCoinType<CoinType> {}? It automatically casts?

PR #2016, opened 2022-11-28
sui: token bridge register chain tests

optke3:


sui/token_bridge/sources/register_chain.move
```diff
@@ -55,11 +55,10 @@ module token_bridge::register_chain {
     }
 
     public entry fun submit_vaa(wormhole_state: &mut WormholeState, bridge_state: &mut BridgeState, vaa: vector<u8>, ctx: &mut TxContext) {
-        let vaa = token_bridge_vaa::parse_verify_and_replay_protect(wormhole_state, bridge_state, vaa, ctx);
+        let vaa = corevaa::parse_and_verify(wormhole_state, vaa, ctx);
```
kcsongor, line 58: why change from `parse_verify_and_replay_protect`?

sui/wormhole/sources/state.move
```diff
@@ -230,8 +230,8 @@ module wormhole::test_state{
             let deployer = take_from_address<DeployerCapability>(&test, admin);
             state::init_and_share_state(
                 deployer,
-                2, // chain id
-                22, // governance chain
+                2, // chain id (arbitrarily picked)
```
kcsongor, line 233: let's put the actual sui id here

sui/token_bridge/sources/register_chain.move
optke3: it was broken up into `parse_and_verify` and then `replay_protect`. One has to call `corevaa::parse_and_verify` instead of `parse_and_verify`, because the latter checks if the vaa chain is registered, so a vaa for registering a chain would never pass the check

kcsongor: ah right, I see that's how we did in aptos too

kcsongor (approved):

heyitaki (approved):

PR #2017, opened 2022-11-28
wormchain: add recipe to enable ledger support

conorpp:


jynnantonix (approved):
This seems fine to me but #1985 is doing the same thing.

PR #2021, opened 2022-11-29
add: safety check against signatory being address(0)

hendrikhofstadt:
Reopening #1897 by @Rassska which got auto closed (and couldn't be reopened) when I pushed the wrong branch to the remote.

### DIFF
--- a/ethereum/contracts/Messages.sol
+++ b/ethereum/contracts/Messages.sol
@@ -76,6 +76,10 @@ contract Messages is Getters {
         uint256 guardianCount = guardianSet.keys.length;
         for (uint i = 0; i < signatures.length; i++) {
             Structs.Signature memory sig = signatures[i];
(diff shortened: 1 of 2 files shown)


kcsongor (dismissed):
great patch

jumpsiegel:
This breaks the ethereum test...

Rassska:
Actually, I was expecting that some test cases with revert messages might fail if we move it up.

The question is, is this case 100% covered with another 2 requires? If it is, this additional check is redundant.

a5-pickle (dismissed):
I pushed a commit that fixes the forge test. It doesn't feel right to approve these changes since I modified the PR, so leaving it for the other reviewers to look everything over.

jumpsiegel (approved):

kcsongor (approved):

PR #2028, opened 2022-11-29
sdk/js: Added .prettierrc.json file

kev1n-peters:
This is the same config used in the example UI https://github.com/wormhole-foundation/example-token-bridge-ui/blob/main/.prettierrc.json.

### DIFF
--- a/sdk/js/.prettierrc.json
+++ b/sdk/js/.prettierrc.json
@@ -1,0 +1,13 @@
+{
+  "printWidth": 80,
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": false,
+  "quoteProps": "as-needed",
+  "jsxSingleQuote": false,
+  "trailingComma": "es5",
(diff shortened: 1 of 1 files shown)


heyitaki:
Agree with most of these config settings, I don't have a strong opinion but I'd prefer the following:
```
{
  "printWidth": 100,
  "singleQuote": true,
}
```

An argument for longer line length: https://lkml.org/lkml/2020/5/29/1038
Single quote because I personally follow Airbnb [style guide](https://github.com/airbnb/javascript#strings--quotes), but again not a strong opinion on either of these things

heyitaki (approved):

kev1n-peters:
Almost all of the SDK code currently uses the formatting specified in this file currently and changing it would make future diffs harder to read with formatting changes. For this reason I would vote to leave it as is

heyitaki:
Didn't realize formatting already conformed to this config. Agreed in that case

PR #2030, opened 2022-11-30
aptos: Build Worm CLI in Tilt

### LINKED ISSUE #2026: worm CLI shouldn't be packaged in Aptos Docker image
As the title suggests, worm CLI should be built in CI as opposed to pre-built in the base image to propagate updates to CLI without requiring another publish of the base image. 

This issue is blocking #1990 because it changes the airdrop amount in `worm aptos faucet`, but the change is not being picked up in CI as the base image does not have the `worm` update.

heyitaki:
This PR moves the Worm CLI build layers from the pre-built base Docker image to the default one so that updates to the Worm CLI properly propagate to the Aptos image.

Fixes #2026

### DIFF
--- a/aptos/Dockerfile
+++ b/aptos/Dockerfile
@@ -1,3 +1,9 @@
+FROM node:lts-alpine@sha256:2ae9624a39ce437e7f58931a5747fdc60224c6e40f8980db90728de58e22af7c as worm-cli
+
+COPY clients/js  /tmp/clients/js
+WORKDIR /tmp/clients/js
+RUN make install
+
 FROM ghcr.io/wormhole-foundation/aptos:1.1.0@sha256:9240135c950207e0bbe70d188999343b59d50d506dcc67c1b5586c1dd57b443c as aptos
 
 # Support additional root CAs
--- a/aptos/Dockerfile
+++ b/aptos/Dockerfile
@@ -25,7 +25,7 @@ COPY examples/ examples
 COPY Makefile Makefile
 COPY scripts scripts
 # In this step we optionally copy the .env file too
-COPY start_node.sh .env* .
+COPY start_node.sh .env* ./
 
 FROM aptos AS tests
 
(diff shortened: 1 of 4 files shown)


jumpsiegel:
This will make bringing up aptos in tilt a great deal slower... are you sure you really want to do this?

heyitaki:
Let me look into caching the build, a one time cost seems worth the convenience of building the CLI

aptos/Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM node:lts-alpine@sha256:2ae9624a39ce437e7f58931a5747fdc60224c6e40f8980db90728de58e22af7c as worm-cli
+
+COPY clients/js  /tmp/clients/js
+WORKDIR /tmp/clients/js
+RUN make install
```
kcsongor, line 5: Your code will work, but docker won't cache it too well. Docker's caching works by looking at the files you copy, and if nothing has changed, then the following commands are not re-evaluated, just brought in from cache. However, if a single file changes, then the whole cache gets invalidated. In concrete terms, that means that if any of the typescript files change in `clients/js`, then `make install` will start over with no cache, and will reinstall all the npm dependencies again.

A better way is to make the file copy a little more granular, start with `package.json` and `package-lock.json`, and run `npm ci` first, and only then invoke `make install`. This will ensure that dependencies are only installed when the package manifest files actually change (then `make` is clever enough to not run `npm ci` again). Also, when running `npm ci` we can specify a cache mount, which basically tells docker to keep that folder from previous invocations of this command.
```suggestion
WORKDIR /tmp/clients/js
COPY clients/js/package.json clients/js/package-lock.json ./
RUN --mount=type=cache,target=/home/node/.npm \
    npm ci
COPY clients/js ./
RUN make install
```

jumpsiegel:
As long as we can make sure this hits the cache 99% of the time in the ci tests, I have no objections.  My concern really is slowing down every PR test going forward which does have a dramatic effect on productivity

heyitaki:
I've accepted @kcsongor's suggestions but a couple points of discussion before we merge this in:

1. I noticed that a previous PR (#1850) fixed a GH runner heisenbug by removing similar caching, could that issue also be present here?
2. Does `npm ci` need to install `devDependencies`? Can we use `npm ci --production` instead (or is this already set by `NODE_ENV` perhaps)?

Also did a little bit of cleanup with `clients/js` npm dependencies.

evan-gray:
Let's see if we can fix instead by tackling https://github.com/wormhole-foundation/wormhole/issues/2157

PR #1990, opened 2022-11-23
Update chain registration script to include Sui

heyitaki:
This PR adds Sui to registration script and sneaks in a cleanup change (removing hardcoded Aptos consts in clients/js)

PR #2031, opened 2022-11-30
Increase Aptos faucet default airdrop amount

heyitaki:
This PR increases the default airdrop amount of `worm aptos faucet`, since the deployer is currently running out of funds on branches that have added 1+ new chain registrations. Also a bit of cleanup RE: sdk consts.

Along with #2030, fixes #1990.

### DIFF
--- a/clients/js/cmds/aptos.ts
+++ b/clients/js/cmds/aptos.ts
@@ -321,7 +321,7 @@ exports.builder = function(y: typeof yargs) {
         let NODE_URL = "http://0.0.0.0:8080/v1";
         let FAUCET_URL = "http://0.0.0.0:8081";
         let account = "0x277fa055b6a73c42c0662d5236c65c864ccbf2d4abd21f174a30c8b786eab84b";
-        let amount = 20000000;
+        let amount = 40000000;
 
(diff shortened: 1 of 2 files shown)


jumpsiegel (approved):

kcsongor (approved):

kev1n-peters (approved):

clients/js/cmds/aptos.ts
```diff
@@ -55,24 +55,17 @@ function assertNetwork(n: string): asserts n is Network {
...
-          alias: "a",
-          required: true,
-          describe: "Address where the wormhole module is deployed",
-          type: "string",
-        })
     }, async (argv) => {
       const network = argv.network.toUpperCase();
       assertNetwork(network);
-      const contract_address = evm_address(argv["contract-address"]);
+      const contract_address = evm_address(CONTRACTS[network].aptos.core);
```
kcsongor, line 68: these slipped past code review it seems, but this now only works for the core contract -- instead, these commands should take the module as an argument. I'll revert this for now for the nft bridge

heyitaki: Ahh good catch, will fix

PR #2036, opened 2022-11-30
sui: create wrapped, complete transfer unit tests

optke3:
## Unit Tests
- `creating_wrapped_coin`, `register_wrapped_coin`
- `complete_transfer` native coin
- `complete_transfer` wrapped coin


sui/coin/sources/coin_witness.move
```diff
@@ -1,13 +1,13 @@
-module coin::coin_witness {
+module coin::coin {
```
heyitaki, line 1: Should we rename this file `coin.move` to match this module name change?

kcsongor (commented):
since `token_bridge::native_coin_witness` and `token_bridge::coin_witness` are used only in tests, let's put them in `sui/token_bridge/sources/testing/` and make the modules `#[test_only]` so they don't get deployed as part of the package

sui/token_bridge/sources/complete_transfer.move
```diff
@@ -84,11 +110,165 @@ module token_bridge::complete_transfer {
...
         // take out fee from the recipient's coins. `extract` will revert
         // if fee > amount
         let fee_coins = coin::split(&mut recipient_coins, fee_amount, ctx);
         transfer_object::transfer(recipient_coins, recipient);
         transfer_object::transfer(fee_coins, fee_recipient);
     }
 }
+
+#[test_only]
+module token_bridge::test_complete_transfer {
```
kcsongor, line 122: in most other places we use the `complete_transfer_test` naming convention, though I'm now seeing a couple exceptions to that rule, which we should fix up too (low priority)

sui/token_bridge/sources/structs/transfer.move
```diff
@@ -16,8 +16,8 @@ module token_bridge::transfer {
 
     friend token_bridge::transfer_tokens;
 
-    //#[test_only]
-    //friend token_bridge::complete_transfer_test;
+    #[test_only]
+    friend token_bridge::test_complete_transfer;
```
heyitaki, line 20: Nit: can we either prefix or suffix `test` the same way in all test methods

sui/token_bridge/sources/complete_transfer.move
optke3: I think in most places it is `module token_bridge::test_something` rather than of something_test. Which do you think is better?

heyitaki (approved):

sui/coin/sources/coin_witness.move
optke3: good idea

sui/token_bridge/sources/structs/transfer.move
optke3: ok, I'm going to put `test` in front everywhere as a prefix

sui/token_bridge/sources/complete_transfer.move
heyitaki: We should just stick to whatever is used more in the rest of the codebase (probably `..._test`) although I personally like (`test_...`) because I can tell that its a test faster.

sui/token_bridge/sources/structs/transfer.move
optke3: actually maybe suffix

sui/token_bridge/sources/complete_transfer.move
kcsongor: a quick grep in the sui folder:
```
module token_bridge::test_register_chain {
module wormhole::deserialize_test {
module wormhole::external_address_test {
module wormhole::test_state{
module token_bridge::string32_test {
module token_bridge::normalized_amount_test {
// module wormhole::emitter_test {
module wormhole::vaa_test {
module wormhole::test_serialize {
module wormhole::guardian_pubkey_test {
module wormhole::guardian_set_upgrade_test {
module token_bridge::test_bridge_state{
module token_bridge::token_bridge_vaa_test{
module token_bridge::transfer_test {
```

which means 10 `*_test` and 4 `test_*`.

kcsongor (approved):

PR #2057, opened 2022-11-30
node: inject wormchain cosmwasm governance messages

nik-suri:
This PR allows guardians to inject wormchain cosmwasm governance VAAs for uploading & instantiating contracts. It adds two new `GovernanceMessage` types: `WormchainStoreCode` and `WormchainInstantiateContract`.

`WormchainStoreCode` takes a single parameter: `WasmHash`, which is a sha3 256 hash of the wasm bytecode that should be uploaded. Guardians can locally verify the hash by building the cosmwasm contract themselves. The payload of this governance VAA is the wasmhash.

`WormchainInstantiateContract` takes the same parameters required to instantiate a cosmwasm contract: `CodeId`, `Label`, and `InstantiationMsg`, which is a json-encoded message to be passed to the contract on instantiation. Once a cosmwasm contract is uploaded to wormchain, guardians can verify these params by querying wormchain for information on the cosmwasm contracts. The payload of this governance VAA is `keccak256 hash(hash(hash(BigEndian(CodeID)), Label), Msg)`.


conorpp (approved):

conorpp:
The new messages to govern uploading contracts to wormchain look correct to me.

kcsongor (approved):

node/cmd/guardiand/adminserver.go
```diff
@@ -162,6 +162,43 @@ func tokenBridgeUpgradeContract(req *nodev1.BridgeUpgradeContract, timestamp tim
 	return v, nil
 }
 
+// wormchainStoreCode converts a nodev1.WormchainStoreCode to its canonical VAA representation
+// Returns an error if the data is invalid
+func wormchainStoreCode(req *nodev1.WormchainStoreCode, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+	// validate the length of the hex passed in
+	b, err := hex.DecodeString(req.WasmHash)
+	if err != nil {
+		return nil, errors.New("invalid cosmwasm bytecode hash (expected hex)")
```
jynnantonix, line 171: nit: we should wrap and return the actual error:

```suggestion
		return nil, fmt.Errorf("invalid cosmwasm bytecode hash (expected hex): %w", err)
```

node/cmd/guardiand/adminserver.go
```diff
@@ -162,6 +162,43 @@ func tokenBridgeUpgradeContract(req *nodev1.BridgeUpgradeContract, timestamp tim
...
+	}
+
+	if len(b) != 32 {
+		return nil, errors.New("invalid cosmwasm bytecode hash (expected 32 bytes)")
```
jynnantonix, line 175: I suggest also including `len(b)` in the error message.

proto/node/v1/node.proto
```diff
@@ -142,6 +147,22 @@ message BridgeUpgradeContract {
   string new_contract = 3;
 }
 
+message WormchainStoreCode {
+  // payload is the hex string of the sha3 256 hash of the wasm binary being uploaded
+  string wasm_hash = 1;
```
jynnantonix, line 152: Protobufs have a `bytes` type so why not use the bytes directly rather than the hex?

sdk/vaa/governance.go
```diff
@@ -23,3 +26,31 @@ func CreateGovernanceVAA(timestamp time.Time, nonce uint32, sequence uint64, gua
 
 	return vaa
 }
+
+// Compute the hash for cosmwasm contract instatiation params.
+// The hash is keccak256 hash(hash(hash(BigEndian(CodeID)), Label), Msg).
+// We compute the nested hash so there is no chance of bytes leaking between CodeID, Label, and Msg.
```
jynnantonix, line 32: The PR description says this is `sha3.Sum(BigEndian(CodeID) || Label || Msg)`, which doesn't match this comment description.  Also I don't quite understand what "bytes leaking" means in this context.

sdk/vaa/payloads.go
```diff
@@ -73,47 +83,39 @@ func (b BodyGuardianSetUpdate) Serialize() []byte {
...
-	// Write action ID
-	MustWrite(buf, binary.BigEndian, uint8(1))
-	// Write target chain (0 = universal)
-	MustWrite(buf, binary.BigEndian, uint16(0))
-	// Write chain to be registered
-	MustWrite(buf, binary.BigEndian, r.ChainID)
-	// Write emitter address of chain to be registered
-	buf.Write(r.EmitterAddress[:])
+func (r BodyWormchainStoreCode) Serialize() []byte {
+	return serializeBridgeGovernanceVaa("WasmdModule", 1, 3104, r.WasmHash)
```
jynnantonix, line 94: please use `ChainIDWormchain` instead of hardcoding the constant.

sdk/vaa/payloads.go
```diff
@@ -73,47 +83,39 @@ func (b BodyGuardianSetUpdate) Serialize() []byte {
...
+}
 
-	return buf.Bytes()
+func (r BodyWormchainInstantiateContract) Serialize() []byte {
+	return serializeBridgeGovernanceVaa("WasmdModule", 2, 3104, r.InstantiationParamsHash)
```
jynnantonix, line 98: Same here

jynnantonix (commented):
Please fix up the commit descriptions to follow the [naming convention](https://github.com/wormhole-foundation/wormhole/blob/main/CONTRIBUTING.md#commit-naming-convention).  This might require splitting up the sdk, node, and wormchain changes into separate commits.

Also I don't really understand how nesting hashes is supposed to give us better security, especially if we're using the same algorithm every time.  We double hash VAAs to reduce gas costs, not for any security benefit.

bruce-riley (dismissed):

proto/node/v1/node.proto
nik-suri: My thought here was it'd be easier for a user to pass in the string hex than binary when constructing a prototxt message.

sdk/vaa/governance.go
nik-suri: The PR description is wrong - I just updated it.

@kcsongor brought up an interesting point that if `label = AAA` and `msg = BBB`, under the `keccak 256 hash(BigEndian(CodeID) || Label || Msg)`, an attacker could submit an instantiation message with `label = AA` and `msg = ABBB` and it would still hash to the same thing.

It's likely not a major security vulnerability since `msg` is meant to be a json blob, but using the hash of hashes is a bit cleaner and avoids this issue entirely.

jynnantonix: Ack.  Another option could be to inject a field separator (like "/" or ":") but this is fine as well.

jynnantonix (approved):

sdk/vaa/payloads.go
```diff
@@ -73,47 +83,39 @@ func (b BodyGuardianSetUpdate) Serialize() []byte {
...
-	// Write action ID
-	MustWrite(buf, binary.BigEndian, uint8(1))
-	// Write target chain (0 = universal)
-	MustWrite(buf, binary.BigEndian, uint16(0))
-	// Write chain to be registered
-	MustWrite(buf, binary.BigEndian, r.ChainID)
-	// Write emitter address of chain to be registered
-	buf.Write(r.EmitterAddress[:])
+func (r BodyWormchainStoreCode) Serialize() []byte {
+	return serializeBridgeGovernanceVaa("WasmdModule", 1, ChainIDWormchain, r.WasmHash)
```
bruce-riley, line 94: Are these action IDs documented somewhere? Should they be consts?

bruce-riley (approved):

kcsongor (dismissed):

sdk/vaa/payloads.go
nik-suri: The action IDs and the "WasmdModule" string are currently defined in the wormchain `msg_server_wasmd` - https://github.com/wormhole-foundation/wormhole/blob/main/wormchain/x/wormhole/keeper/msg_server_wasmd.go#L15-L21

Maybe worth moving them all over to the sdk?

sdk/vaa/governance.go
```diff
@@ -23,3 +27,33 @@ func CreateGovernanceVAA(timestamp time.Time, nonce uint32, sequence uint64, gua
...
+func CreateInstatiateCosmwasmContractHash(codeId uint64, label string, msg []byte) [32]byte {
+	var expected_hash [32]byte
+
+	// hash(BigEndian(CodeID))
+	var codeId_hash [32]byte
+	keccak := sha3.NewLegacyKeccak256()
+	if err := binary.Write(keccak, binary.BigEndian, codeId); err != nil {
+		panic(fmt.Errorf("failed to write binary data: %d", codeId).Error())
```
jynnantonix, line 41: We should include the actual error in the panic:

```suggestion
		panic(fmt.Sprintf("failed to write binary data (%d): %v", codeId, err))
```

jynnantonix (approved):

hendrikhofstadt (approved):

PR #2060, opened 2022-12-01
ethereum: token verification instructions

### LINKED ISSUE #2058: Add EVM token verification instructions
Verifying token contracts on EVM explorers is a common task that integrators want to perform, so we should include instructions in the repo. There are a couple of difficulties:
1. The source files that need to be uploaded depend on the time of deployment of the contract (since the `main` branch may contain yet to be deployed changes)
2. The constructor arguments need to be derived dynamically.

To tackle 1., my plan is to create a CI action that's triggered whenever a tag is pushed that matches a certain pattern (maybe `evm-*` or such), and it flattens all the solidity source files in the repo and pushes them into an artifact using https://github.com/actions/upload-artifact. This way, the files will be available without having to flatten the contract locally. Then, another CI action that runs on each PR (or maybe each push to `main`?) that warns if there are undeployed changes in the EVM contracts.

For 2., we already have the https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/scripts/TokenABI.s.sol script, so this just needs some documentation around usage.

kcsongor:
Fixes #2058. 

### Tasks
- [x] Makefile target to flatten contracts
- [x] CI action to push flattened artifacts on specific tags
- [x] Instructions document (markdown) explaining how to perform verification


a5-pickle (dismissed):

ethereum/docs/token_verification.md
```diff
@@ -0,0 +1,147 @@
...
+
+First, install the dependencies by running the `make dependencies` in the
+`ethereum` folder:
+
+```sh
+wormhole/ethereum $ make dependencies
+```
+
+Next, we will run a `forge` script to compute the constructor arguments. For
+this, we will need to pieces of data. First, the address of the Portal token
```
gator-boi, line 107: `to` should be `two` on line 107

gator-boi (dismissed):

ethereum/docs/token_verification.md
kcsongor: good catch!

gator-boi (approved):
Great work!

hendrikhofstadt (approved):
ack .github/workflows

PR #2067, opened 2022-12-01
evm: Added Core and NFT Bridge EVM chain ID migrations

kev1n-peters:
These migrations are still needed for the core and nft bridge contracts except on ethereum which has been fully migrated.


ethereum/contracts/Implementation.sol
```diff
@@ -32,6 +32,31 @@ contract Implementation is Governance {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
+        uint256 evmChainId;
+        uint16 chain = chainId();
+
+        // Wormhole chain ids explicitly enumerated
```
kcsongor, line 38: we should wrap this whole thing in a
```
if (evmChainId == 0) {...}
```
so that we can keep this code around for longer and run it for newer deployments if needed (if those need to be updated too for some other reason).

ethereum/contracts/nft/NFTBridgeImplementation.sol
```diff
@@ -17,6 +17,31 @@ contract NFTBridgeImplementation is NFTBridge {
 
     function initialize() initializer public virtual {
         // this function needs to be exposed for an upgrade to pass
+        uint256 evmChainId;
+        uint16 chain = chainId();
+
+        // Wormhole chain ids explicitly enumerated
```
kcsongor, line 23: same here

ethereum/contracts/Implementation.sol
kev1n-peters: changed

ethereum/contracts/nft/NFTBridgeImplementation.sol
kev1n-peters: changed

jumpsiegel (approved):

a5-pickle (approved):

PR #2076, opened 2022-12-02
sdk/go: Add unmarshal for vaa.Address

bruce-riley:
As part of the guardian changes for accounting, we need to start persisting common.MessagePublication objects in the badger db. It would be nice to be able to use JSON for that. To make that work, we need a custom unmarshal method for vaa.Address, since it has a custom marshal. (I don't want to change the existing marshal because that may cause unexpected breakage.)

Change-Id: I1beb99f82673d1fc3225a8c6628a0019648d7e01

### DIFF
--- a/sdk/vaa/structs.go
+++ b/sdk/vaa/structs.go
@@ -126,7 +126,7 @@ func (a Address) MarshalJSON() ([]byte, error) {
 // Standard marshal stores the Address like this: "[0,0,0,0,0,0,0,0,0,0,0,0,2,144,251,22,114,8,175,69,91,177,55,120,1,99,183,183,169,161,12,22]"
 // The above MarshalJSON stores it like this (66 bytes): ""0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16""
 func (a *Address) UnmarshalJSON(data []byte) error {
-	addr, err := StringToAddress("0x" + string(data[1:65]))
+	addr, err := StringToAddress(strings.Trim(string(data), `"`))
 	if err != nil {
 		return err
 	}
--- a/sdk/vaa/structs.go
+++ b/sdk/vaa/structs.go
@@ -123,6 +123,17 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	return []byte(fmt.Sprintf(`"%s"`, a)), nil
 }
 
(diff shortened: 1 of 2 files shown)


sdk/vaa/structs.go
```diff
@@ -123,6 +123,17 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	return []byte(fmt.Sprintf(`"%s"`, a)), nil
 }
 
+// Standard marshal stores the Address like this: "[0,0,0,0,0,0,0,0,0,0,0,0,2,144,251,22,114,8,175,69,91,177,55,120,1,99,183,183,169,161,12,22]"
+// The above MarshalJSON stores it like this (66 bytes): ""0000000000000000000000000290fb167208af455bb137780163b7b7a9a10c16""
+func (a *Address) UnmarshalJSON(data []byte) error {
+	addr, err := StringToAddress("0x" + string(data[1:65]))
```
jynnantonix, line 129: If we're going to sub-slice the data then we should do a length check first.  Otherwise we could use `strings.Trim` to trim out the quotes.  Also, the "0x" prefix is optional for `StringToAddress`.

```suggestion
	addr, err := StringToAddress(strings.Trim(data, `"`))
```

sdk/vaa/structs_test.go
```diff
@@ -108,6 +109,32 @@ func TestAddress_MarshalJSON(t *testing.T) {
...
+	b, err := json.Marshal(addr)
+	require.NoError(t, err)
+
+	var unmarshalAddr Address
+	err = json.Unmarshal(b, &unmarshalAddr)
+	require.NoError(t, err)
+
+	assert.Equal(t, addr, unmarshalAddr)
+}
+
```
jynnantonix, line 137: Would also be good to add tests for cases where the string is too short or too long.

sdk/vaa/structs.go
bruce-riley: Fixed.

sdk/vaa/structs_test.go
bruce-riley: Added tests.

justinschuldt (approved):

jynnantonix (approved):

hendrikhofstadt (approved):

PR #2078, opened 2022-12-03
cosmwasm: Added injective mainnet verify script

kev1n-peters:


cosmwasm/verify
```diff
@@ -148,7 +154,11 @@ echo $hash1
 echo "$wasm hash:"
 echo $hash2
 
-if [ "$hash1" == "$hash2" ]; then
+if [[ "$chain" == "injective" ]]; then
+  shopt -s nocasematch
+fi
```
jynnantonix, line 159: I'd prefer if we just unconditionally converted the hashes to lowercase (or uppercase):

```bash
hash1 = "${hash1,,}" # lowercase
```

or

```bash
hash2 = "${hash2^^}" # uppercase
```

If we care about compatibility with bash versions older than 4 (bash on mac os is currently 3.2) then we can use `tr` instead:

```bash
hash1 = "$(echo "${hash1}" | tr '[:upper:]' '[:lower:]')"
```

jumpsiegel: I was this days old when I discovered this cool way in bash to upper and lower case.. (I always used tr)

panoel (approved):

cosmwasm/verify
kev1n-peters: Thanks for teaching me about the `tr` utility, I decided to use that for bash 3.2 compatibility.

panoel (approved):

jynnantonix (approved):

PR #2079, opened 2022-12-05
sui: wormhole test guardian set expiry

optke3:
test that:
- old guardian set does not expire immediately after a guardian set upgrade
- but does so after 3 epochs have passed

In general, what should we set delta epochs to be? Right now it is set to 2 epochs. 

Note: the token bridge tests do not pass right now, because of the update to the CoinMetadata, which is fixed on a different branch

### DIFF
--- a/sui/wormhole/sources/state.move
+++ b/sui/wormhole/sources/state.move
@@ -86,7 +86,7 @@ module wormhole::state {
             governance_contract: external_address::from_bytes(governance_contract),
             guardian_set_index: u32::from_u64(0),
             guardian_sets: vec_map::empty<U32, GuardianSet>(),
-            guardian_set_expiry: u32::from_u64(0),
+            guardian_set_expiry: u32::from_u64(2), // TODO - what is the right #epochs to set this to?
             consumed_governance_actions: set::new(ctx),
             emitter_registry: emitter::init_emitter_registry(),
             message_fee: 0,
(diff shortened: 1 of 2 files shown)


sui/wormhole/sources/state.move
```diff
@@ -86,7 +86,7 @@ module wormhole::state {
             governance_contract: external_address::from_bytes(governance_contract),
             guardian_set_index: u32::from_u64(0),
             guardian_sets: vec_map::empty<U32, GuardianSet>(),
-            guardian_set_expiry: u32::from_u64(0),
+            guardian_set_expiry: u32::from_u64(2), // TODO - what is the right #epochs to set this to?
```
kcsongor, line 89: we should aim for at least 24h, it's not immediately obvious reading on sui's website what the epoch lengths are, they're seemingly decided at genesis so I guess technically we don't know yet until mainnet launches. Setting it to 2 for now makes sense

kcsongor (approved):
Looks good - let's merge the coin metadata stuff in first, so we can get the tests to pass on this one

sui/wormhole/sources/state.move
jumpsiegel: I don't think the epoch's are updating right now anyhow.  At least last time I looked the RPC's were always returning 0

optke3: at least the increment_epoch function works

PR #2093, opened 2022-12-08
tilt: remove wormchain gentx folder to make deployment robust

nik-suri:
Running `make validators` in wormchain will fail if there already exists a `wormchain/build/config/gentx/` folder:
```
Error: failed to write signed gen tx: open build/config/gentx/gentx-<uuid>.json: file exists`
```

This change ensures that starting tilt with wormchain doesn't fail with this error by remove the `gentx/` from the wormchain directory that's copied over to the docker container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,5 @@ target
 bin
 **/target
 **/node_modules
+
+wormchain/build/config/gentx/


Dockerfile.wormchain
```diff
@@ -13,6 +13,7 @@ COPY ./sdk /sdk
 RUN go mod download
 
 COPY ./wormchain .
+RUN rm -rf ./build/config/gentx/
```
jynnantonix, line 16: Rather than removing the directory here can we add it to the `.dockerignore` file?

hendrikhofstadt (approved):

PR #2095, opened 2022-12-12
sdk/js: Add instruction decoder to solana

guibescos:
The current sdk/js for solana is a little bit lackluster :
- There is no encoder for PostMessage
- There is no intruction decoder

I need both things for cross-chain governance tools. I decided to copy what anchor does for encoding/decoding.
I updated WormholeInstructionCoder to parse the layouts of the instructions from wormhole.json (wormhole's IDL). I then use these layouts in WormholeInstructionCoder to encode/decode instructions. This means we no longer need to hand-serialize/deserialize instructions. To add a new instructions you just need to add it to wormhole.json.

### DIFF
--- a/sdk/js/src/solana/wormhole/coder/idl.ts
+++ b/sdk/js/src/solana/wormhole/coder/idl.ts
@@ -1,0 +1,112 @@
+// Borrowed from coral-xyz/anchor
+//
+// https://github.com/coral-xyz/anchor/blob/master/ts/packages/anchor/src/coder/borsh/idl.ts
+
+import * as borsh from "@coral-xyz/borsh"
+import { Layout } from "buffer-layout";
+import { IdlField, IdlTypeDef } from "../../anchor";
+import { camelCase } from "lodash";
+
+export class IdlCoder {
+    public static fieldLayout(
+      field: { name?: string } & Pick<IdlField, "type">,
+      types?: IdlTypeDef[]
+    ): Layout {
+      const fieldName =
+        field.name !== undefined ? camelCase(field.name) : undefined;
+      switch (field.type) {
+        case "bool": {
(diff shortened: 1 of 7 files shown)


a5-pickle (dismissed):
These changes are awesome 💪

sdk/js/src/solana/wormhole/coder/instruction.ts
```diff
@@ -1,41 +1,88 @@
-import { Idl, InstructionCoder } from "@project-serum/anchor";
-import { ETHEREUM_KEY_LENGTH } from "../../utils";
+import { Idl, Instruction, InstructionCoder } from "@project-serum/anchor";
+import { bs58 } from "@project-serum/anchor/dist/cjs/utils/bytes";
```
a5-pickle, line 2: I didn't catch this the first time I reviewed it. Is this going to cause any issues with esm?

sdk/js/src/solana/wormhole/coder/idl.ts
```diff
@@ -0,0 +1,112 @@
+// Borrowed from coral-xyz/anchor
```
a5-pickle, line 1: Just making a comment here as a note that we would probably like this for Token Bridge and NFT Bridge program decoding.

a5-pickle (changes requested):

sdk/js/src/solana/wormhole/coder/instruction.ts
guibescos: Ok, so I just solved this by explicitly naming bs58 in `package.json`. bs58 was already in package-lock.json so I don't think this increases the size of the package.

a5-pickle (approved):

kev1n-peters (approved):

PR #2097, opened 2022-12-12
[SDK] Adds parsing for v2 and v3 vaas

gabzim:
Missing from the PR:

- [ ] Tests
- [ ] The parseVaa function now returns a union type, is this breaking change ok?

This is a draft, I'm not a contributor so I can't mark it as such but I'm looking for feedback in terms of the coding style and approach taking for parsing batch vaas.


evan-gray:
Per @gabzim closing as this isn't holding up https://github.com/wormhole-foundation/trustless-generic-relayer since the switch to single vaas

PR #2102, opened 2022-12-13
aptos: NFT bridge

kcsongor:


aptos/nft_bridge/sources/contract_upgrade.move
```diff
@@ -0,0 +1,217 @@
...
+    public entry fun submit_vaa_entry(vaa: vector<u8>) acquires UpgradeAuthorized {
+        submit_vaa(vaa);
+    }
+
+    fun authorize_upgrade(hash: &Hash) acquires UpgradeAuthorized {
+        let nft_bridge = state::nft_bridge_signer();
+        if (exists<UpgradeAuthorized>(@nft_bridge)) {
+            // TODO(csongor): here we're dropping the upgrade hash, in case an
+            // upgrade fails for some reason. Should we emit a log or something?
+            let UpgradeAuthorized { hash: _ } = move_from<UpgradeAuthorized>(@nft_bridge);
```
jumpsiegel, line 98: You mean since the previous upgrade hash would now fail?

aptos/nft_bridge/sources/contract_upgrade.move
```diff
@@ -0,0 +1,217 @@
...
+        };
+        move_to(&nft_bridge, UpgradeAuthorized { hash: hash.hash });
+    }
+
+    #[test_only]
+    public fun authorized_hash(): vector<u8> acquires UpgradeAuthorized {
```
jumpsiegel, line 104: Is there a reason this is test_only?

aptos/nft_bridge/sources/contract_upgrade.move
kcsongor: yes, only the most recent one is allowed to pass, this is actually the right behaviour I think (and it's identical to how the token bridge works). probably best to remove the todos

aptos/nft_bridge/sources/register_chain.move
```diff
@@ -0,0 +1,166 @@
...
+
+    /// "NFTBridge" (left padded)
+    const NFT_BRIDGE: vector<u8> = x"00000000000000000000000000000000000000000000004e4654427269646765";
+
+    const E_INVALID_MODULE: u64 = 0;
+    const E_INVALID_ACTION: u64 = 1;
+    const E_INVALID_TARGET: u64 = 2;
+
+    struct RegisterChain has copy, drop {
+        /// Chain ID
```
jumpsiegel, line 20: Since RegisterChain has drop, does this mean you can re-register the same chain multiple times since it will just drop the previous registration?

kcsongor: the registrations are stored in a table, which we upsert to so it should allow re-registration (but not because of the drop here)

aptos/nft_bridge/sources/contract_upgrade.move
jumpsiegel: yup.. I agree 100%

aptos/nft_bridge/sources/register_chain.move
jumpsiegel: Don't we normally not allow re-registration?

aptos/nft_bridge/sources/contract_upgrade.move
jumpsiegel: Just feels like a generally useful function for confirming when a upgrade vaa was successfully submitted...

kcsongor: these functions can't be run off chain anyway, we can just look up the storage directly and query the auth object

aptos/nft_bridge/sources/register_chain.move
kcsongor: yeah but it's not specified in the whitepaper, and this seemed like better behaviour

njkumr (approved):

evan-gray (approved):
Approving based on prior security approval.

PR #2119, opened 2022-12-14
Aptos/nft bridge fixes

optke3:
## Fix Complete Transfer 
- create_tokendata and setting property keys

## Unit Test for Transfer NFT
- in `transfer_nft.move`

## Edge Cases
- should a wrapped token be burnable by creator or owner AND creator?


kcsongor (approved):

jumpsiegel (approved):

heyitaki (approved):

kcsongor:
should only be burnable by creator, not owner

optke3:
> should only be burnable by creator, not owner

agree, made this the case. It's interesting how a creator can burn an NFT in someone else's account

PR #2126, opened 2022-12-15
Aptos/nft bug fix + unit tests

optke3:
## Bug Fix
- In the following line in `complete_transfer`, it should say `!=` instead of `==`
-  `let is_wrapped_asset: bool = token_chain == wormhole_state::get_chain_id();`

## Complete Transfer
- added test `test_complete_transfer_wrapped_asset`, which completes a transfer of an nft from a wrapped nft collection
- added test `test_complete_transfer_native_token_full_loop`, which transfers an Aptos-native nft to the nft bridge, then calls complete_transfer to complete the full loop 

## Transfer NFT
- added test `test_transfer_native_nft`, which performs a native nft transfer (in effect locks up the nft with the nft bridge)


jumpsiegel (approved):

kcsongor (changes requested):

aptos/nft_bridge/sources/complete_transfer.move
```diff
@@ -227,4 +227,4 @@ module nft_bridge::complete_transfer_test {
     }
 
     // TODO - failure test cases
-}
```
heyitaki: RE commit conventions, is the idea here that commits like this are fine because you anticipate NFT bridge to be squashed into main?

heyitaki (approved):

PR #2121, opened 2022-12-15
Add support for chain registration to accounting

jynnantonix:
This PR adds support for handling tokenbridge chain registration VAAs to the accounting contract.  This will let us deploy accounting without first having to deploy the tokenbridge and all its dependencies.

Currently the governance chain and governance emitter are hardcoded but this can be made configurable, if necessary.

Fixes #2103.

PR #2120, opened 2022-12-15
cosmwasm: accounting: Fix chain registration query return type

jynnantonix:
The tokenbridge chain registration query returns a
`ChainRegistrationResponse` struct and not a `Vec<u8>`.  Use the proper
return type when sending the query.

PR #2131, opened 2022-12-16
Store message digests on chain for accounting

### LINKED ISSUE #2146: Check for mismatched digests when handing duplicate observations
It's possible for accounting to receive duplicate observations for transfers that are already committed.  One way this can happen is if a node re-observes a transfer and follows the usual flow of sending it to accounting first.

When this happens we want to be able to differentiate the case where the observation data exactly matches the transfer that was committed and when it doesn't.  The former can happen during normal operation (re-observations) but the latter implies that something very bad has happened because transfer details should not change once they have been committed.

In order to avoid having to store *all* information about a transfer, only store the digest of the details on-chain.  This digest should be the same thing that the guardians sign to produce the VAA.  This requires some changes to the `Observation` message to add in additional fields that are part of the VAA.

jynnantonix:
Fixes #2146


jynnantonix:
* main
  * **PR #2120** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2120" target="_blank">Graphite</a> 
    * **PR #2121** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2121" target="_blank">Graphite</a> 
      * **PR #2131** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2131?utm_source=stack-comment).

jynnantonix:
This PR includes the digest change as well as some cleanup afterwards (removing `InstantiateMsg` and returning the digest from queries).  

Please let me know if you want me to split this up into separate PRs.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -339,18 +340,48 @@ fn handle_vaa(deps: DepsMut<WormholeQuery>, vaa: Binary) -> anyhow::Result<Event
...
+    };
+
+    digest_key
+        .save(deps.storage, &digest)
+        .context("failed to save message digest")?;
+
+    Ok(evt)
 }
 
 fn handle_governance_vaa(
```
a5-pickle, line 387: I think we should save the governance VAAs in `DIGESTS` like how you were saving the hashes in the previous `GOVERNANCE_VAAS` storage so none of the previous governance actions can be replayed.

a5-pickle (changes requested):
Only found one issue (commented above)

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: This was moved to lines 380-382 so that saving the digests is shared between the `handle_governance_vaa` and `handle_tokenbridge_vaa` methods.  So they are still being saved.  There is also a test for this: https://github.com/wormhole-foundation/wormhole/blob/acc3ec14d8264075a6c8d084d007538d3def395d/cosmwasm/contracts/wormchain-accounting/tests/chain_registration.rs#L122-L135

a5-pickle: Doh! That's right, looks good

a5-pickle (approved):

hendrikhofstadt (approved):

PR #2148, opened 2022-12-19
ci: Fix flaky solana tests and improve CI time

hendrikhofstadt:
This removes the `common::sync` function which was not necessary and causing flakiness.

It also aggressively caches cargo dependencies and build artifacts. The binaries will still get rebuilt when there are changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,12 +122,16 @@ jobs:
           toolchain: ${{ steps.toolchain.outputs.version }}
           components: "clippy,rustfmt"
 
-      - name: Cache rust packages
+      - name: Cache rust packages / build cache
         uses: actions/cache@v3
         env:
           cache-name: solana-rust-packages
         with:
-          path: ~/.cargo/registry
+          path: |
+            ~/.cargo/bin
+            ~/.cargo/registry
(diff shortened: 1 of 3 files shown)


evan-gray (approved):

jumpsiegel (approved):

.github/workflows/build.yml
```diff
@@ -122,12 +122,16 @@ jobs:
...
         uses: actions/cache@v3
         env:
           cache-name: solana-rust-packages
         with:
-          path: ~/.cargo/registry
+          path: |
+            ~/.cargo/bin
+            ~/.cargo/registry
+            ~/.cargo/git/db
+            solana/target
```
jynnantonix, line 134: Due to the way `actions/cache` works, the new directories won't get picked up until the cache key changes or we delete the current cache entry with this key from the repo.

jynnantonix (approved):

.github/workflows/build.yml
jynnantonix: This also appears to significantly increase the size of the cache from 390MB -> 1.7GB.  See https://github.com/wormhole-foundation/wormhole/actions/caches

PR #2160, opened 2022-12-21
docker: reorganize dockerfiles

### LINKED ISSUE #2156: Move Dockerfiles into subdirectories
The repo root is currently cluttered with several Dockerfiles. Previously, the logic was to store Dockerfiles in the same context they were being run and as several of these Dockerfiles copy shared resources, they were being run in the root of the repo. We can reduce clutter in the root by moving Dockerfiles into their respective subdirectories and only have those Dockerfiles that *build* shared resources in the root of the repo as opposed to those that also copy these resources. Contexts for the docker builds are also clearly specified in `Tiltfile`.

heyitaki:
Declutter root by moving Dockerfiles to respective subdirectories. They are still executed with the same context as before.

Resolves #2156


bruce-riley (approved):

jumpsiegel (approved):

kcsongor (approved):
looks good -- did you search the repo for these files to make sure all mentions are caught? looks like you did, just checking

heyitaki:
Yessir

heyitaki:
Yes, dockerignore uses root of context, which is defined in Tiltfile and has been kept constant here

PR #2170, opened 2022-12-21
sdk/js: Use BN.toArrayLike for compatibility with browserify and similar tools.

kev1n-peters:
Fixes this error when calling `deriveWrappedMintKey` in the browser:
TypeError: (intermediate value).toBuffer is not a function


evan-gray (dismissed):
Equal parts sad and angry.

a5-pickle (approved):

evan-gray (approved):

Issue #2002, opened 2022-11-25
Re-observations for transfers stuck in global accountant


jynnantonix:
Since accounting happens before a VAA is broadcasted over the p2p network we cannot use the existing re-observation infrastructure to unblock transfers that get stuck in accounting waiting for quorum.

However, since we know exactly which signatures are missing we also don't need to spam the entire p2p network for re-observations.  Instead we can have each guardian query the accounting contract for any pending transfers for which it doesn't have an observation from that guardian.  The guardian can then re-observe only those txs and send the observations once it has them.

The proposal is to add a new query method to the accounting contract:

```rust

#[cw_serde]
pub enum QueryMsg {
    /// Query the contract for any pending observations for which it doesn't have an observation
    /// from the provided guardian index.
    #[returns(QueryMissingObservationsResponse)]
    MissingObservations {
        /// The index of the guardian whose signatures we want to check.
        index: u8,
    },
}

#[cw_serde]
pub struct QueryMissingObservationsResponse {
    /// A list of transfers for which the accounting contract does not have an observation from the
    /// requested guardian, along with a list of possible tx_hashes where that transfer may have
    /// occurred.
   missing_observations: Vec<transfer::Key, Vec<Binary>>,
}
```

guardiand would then send this query every (5? 7? 11?) minutes and try to re-observe any missing observations.

bruce-riley:
Here is a proposed design for handling transfers that are stuck in accounting.

# Guardian Accounting Audit Design
The audit will run periodically, say every ten minutes. It will consist of two phases performed back-to-back.

## Phase 1: Transfers the contract has in the pending state
- Issue the **AllPendingTransfers** query to the contract.
- Loop through the results.
  - If the transfer is in our pending map:
    - Validate that the digest matches. If not, peg a critical error and skip it. **Not sure how to recover from this!**
    - Update our timestamp.
    - If we are not marked as having signed it, resend the observation to the contract.
    - Otherwise continue waiting for it to reach quorum.
  - If the transfer is not in our pending map, submit a reobservation request to the **obsvReqC** channel.

## Phase 2: Transfers that we have in the pending state but the contract does not
Phase 2 covers a couple of infrequent scenarios:
- We previously failed to send an observation request to the contract.
- We missed the WasmTransfer event published by the contract.

Phase two involves looping through our pending map and doing the following:
  - If the timestamp indicates it has been idle too long (also meaning it wasn't handled in phase 1):
    - Update our timestamp.
[...]

PR #2171, opened 2022-12-22
cosmwasm: accounting: Add query for missing observations

jynnantonix:
Add a query so that guardians can directly check with the on-chain accounting state for transfers that are missing observations from that guardian.  This should make it easier for those guardians to do targeted re-observations.

Part of #2002.

### DIFF
--- a/cosmwasm/contracts/wormchain-accounting/src/contract.rs
+++ b/cosmwasm/contracts/wormchain-accounting/src/contract.rs
@@ -27,8 +27,9 @@ use crate::{
     error::{AnyError, ContractError},
     msg::{
         AllAccountsResponse, AllModificationsResponse, AllPendingTransfersResponse,
-        AllTransfersResponse, ChainRegistrationResponse, ExecuteMsg, MigrateMsg, Observation,
-        QueryMsg, TransferResponse, Upgrade,
+        AllTransfersResponse, ChainRegistrationResponse, ExecuteMsg, MigrateMsg,
+        MissingObservation, MissingObservationsResponse, Observation, QueryMsg, TransferResponse,
+        Upgrade,
     },
     state::{self, Data, PendingTransfer, CHAIN_REGISTRATIONS, DIGESTS, PENDING_TRANSFERS},
 };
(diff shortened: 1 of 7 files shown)


jynnantonix:
* main
  * **PR #2171** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2171?utm_source=stack-comment).

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -651,3 +655,24 @@ fn query_chain_registration(
...
         .map(|address| ChainRegistrationResponse { address })
 }
+
+fn query_missing_observations(
+    deps: Deps<WormholeQuery>,
+    index: u8,
+) -> StdResult<MissingObservationsResponse> {
+    let mut missing = Vec::new();
+    for pending in PENDING_TRANSFERS.range(deps.storage, None, None, Order::Ascending) {
+        let (_, v) = pending?;
```
kcsongor, line 665: under what circumstances would `pending` be an error here? would it make sense to just ignore it instead of failing the whole query?

kcsongor (dismissed):

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: It would be an error if it failed to deserialize the key or value from the underlying data, which should only happen if there is corrupted data in the store.

Data corruption implies bigger issues so I don't know if we should just silently ignore it.  That said, I don't think we'll actually hit this error in practice.

kcsongor: that makes sense, thanks

hendrikhofstadt (changes requested):

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -651,3 +655,24 @@ fn query_chain_registration(
         .load(deps.storage, chain)
         .map(|address| ChainRegistrationResponse { address })
 }
+
+fn query_missing_observations(
```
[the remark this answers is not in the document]
jynnantonix: Good catch.  Done.

hendrikhofstadt (approved):

a5-pickle (approved):

PR #2175, opened 2022-12-24
sdk/js: injective package bump

bangjelkoski:
Hey guys, once more again - can we please bump Injective to the latest npm SDK packages? I've included the changes in this PR and I've also allowed the `sdk-ts` version to take the latest version when installing the `wormhole-sdk` in your (and our) projects so we don't install 2 `sdk-ts` versions as there will be typescript issues between the two. 

This package bump also includes a fix for the issue we had with the `MsgExecuteContract` and signing the message using Ethereum Native Wallets on the `wormhole-sdk` level. We still need to do a chain upgrade on our side but this will prepare wormhole's SDK to be ready once the chain upgrade is completed.

Thank you for taking the time to review, please ping me if there is a need to change something and when you guys republish the packages so I can update on my end.


evan-gray:
Converting to draft while we wait to hear back on how @bangjelkoski 's testing went.

bangjelkoski:
I've updated the PR to include a new message we've introduced `MsgExecuteContractCompat` which allows us to generate EIP712 typed data for the message which then can be signed on Ethereum native wallets like Metamask. 

This is important to have as it allows support for Ledger to be used on Keplr as well.

evan-gray (approved):

kev1n-peters:
Successfully tested that we can sign injective messages with Keplr and a Ledger device.

kev1n-peters (approved):

PR #2190, opened 2023-01-04
node: algorand recover

jumpsiegel:


node/pkg/watchers/algorand/watcher.go
```diff
@@ -141,7 +142,20 @@ func lookAtTxn(e *Watcher, t types.SignedTxnInBlock, b types.Block, logger *zap.
 	}
 }
 
-func (e *Watcher) Run(ctx context.Context) error {
+func (e *Watcher) Run(ctx context.Context) (result error) {
+	defer func() {
```
bruce-riley, line 146: Should this be a function in supervisor or something so it doesn't have to be replicated in all the watchers. We could do something like this in all the watchers:
`defer result = supervisor.Recover("Algorand watcher")`

PR #2207, opened 2023-01-06
cosmwasm: Add cw_transcode crate

jynnantonix:
The cw_transcode crate provides a way to transcode any arbitrary rust
struct into a `cosmwasm_std::Event` via that struct's `Serialize` impl,
ensuring that the event attribute values are encoded as proper json.

This will make it easier for client code to parse the event back into
structured data without having to write custom parsing code for each
individual event type.

### DIFF
--- a/cosmwasm/Cargo.lock
+++ b/cosmwasm/Cargo.lock
@@ -595,6 +595,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "cw_transcode"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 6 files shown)


hendrikhofstadt (dismissed):
This is awesome

kcsongor (dismissed):
nice!

bruce-riley (approved):

kcsongor (approved):

bruce-riley (approved):

PR #2209, opened 2023-01-06
whitepaper: clarify transfer completion caveat in 0003_token_bridge.md

### LINKED ISSUE #2208: Revise liveness properties in token bridge (& NFT bridge) whitepaper
The following two documents
https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0003_token_bridge.md#caveats
https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0006_nft_bridge.md#caveats
claim that transactions can get stuck indefinitely due to guardian set changes. While this statement is true in general, there do exist recovery mechanisms for many common scenarios, so we should consider updating the documents to include those.
Some of these, off the top of my head:
1. The subset of guardians that persist between guardian set changes is still enough to reach quorum, and they can simply reobserve the transaction.
2. The new guardians replayed all transactions from genesis (or at least from some state that's older than the missing transaction) and they can reobserve the message.
3. With https://github.com/wormhole-foundation/wormhole/pull/2183 a new guardian can sign a VAA even if it hasn't seen it, by trusting an earlier guardian set (and if at most one signature is missing to reach quorum again).

tbjump:
Fixes #2208 

Please review for accuracy and see if I missed anything. Also if there are any cli tools or more specific instructions available, it would be good to link them. 

Once everything looks good, I'll propose the same changes to [0006_nft_bridge.md#caveats](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0006_nft_bridge.md#caveats)


whitepapers/0003_token_bridge.md
```diff
@@ -236,10 +236,16 @@ NewContract [32]uint8
...
 ## Caveats
 
-There is no guarantee for completion of transfers. If a user initiates a transfer and doesn't call `completeTransfer`
-on the target chain, a transfer might not be completed. In case a guardian set change happens in-between and the
-original signer guardian set expires, the transfer will be stuck indefinitely.
+### Transfer completion
+A user who initiated a transfer should call `completeTransfer` within 24 hours. Guardian Sets are guaranteed to be valid for at least 24 hours. If the user waits longer, it could be that the Guardian Set has changed between the time where the transfer was initiated and the the time the user attempts to redeem the VAA. Let's call the Guardian Set at the time of signing `setA` and the Guardian Set at the time of redeeming on the target chain `setB`.
 
+If `setA != setB` and more than 24 hours have passed, there are multiple options for still redeeming the VAA on the target chain:
+1. The Guardians that signed the VAA may still be part of `setB`. In this case, the VAA meerly needs to be modified to have the new Guardian Set Index and it will be a valid VAA that can be redeemed.
```
SEJeff, line 243:
```suggestion
1. The Guardians that signed the VAA may still be part of `setB`. In this case, the VAA merely needs to be modified to have the new Guardian Set Index along with any `setA` only guardian signatures removed to make a valid VAA. The updated VAA can then be be redeemed. The typescript sdk includes a `repairVaa()` function to perform this automatically.
```

If a redeem fails due to a `guardian set expired` error, the only process to repair it is:

* Determine what guardians were removed: `to_remove = setA.difference(setB)`
* Ensure that >= quorum + 1 guardians signed the VAA that are not only in `setA` (13 out of 19)
* Fix the raw VAA so it is valid:
  * Remove the old Guardian Set Index only signatures from the VAA in `to_remove`
  * Update the Guardian Set Index to the latest
* Manually redeem the VAA (e.g. the advanced redeem workflow in portal where you paste the raw bytes or a direct smart contract call)

These two functions are what do it:
https://github.com/wormhole-foundation/wormhole/blob/f41eebe0b305f2d5fc07e73694197561a9deb114/sdk/js/src/utils/repairVaa.ts#L44-L115

Perhaps at some point one of the wormhole contributors will hook at sdk function up to the example ui or front ends such as portal bridge so that it can be automated for guardian set expired errors.

whitepapers/0003_token_bridge.md
```diff
@@ -236,10 +236,16 @@ NewContract [32]uint8
...
+2. The intersection between `setA` and `setB` is greater than 2/3 of `setB`, but not all signatures of the VAA are from Guardians in `setB`. Then it may be possible to gather signatures from the other Guardians from historic gossip network observations, and replace now-invalid signatures with those.
```
SEJeff, line 244:
```suggestion
```

I'm not fully sure about this, but think 1 and 2 here are the same thing. Also, I'm not sure how a guardian would gather historic gossip network observations, via a pcap or something? This seems incorrect.

whitepapers/0003_token_bridge.md
```diff
@@ -236,10 +236,16 @@ NewContract [32]uint8
...
+3. Guardians may elect to do the above automatically and make those updated VAAs available for users over the public RPC.
```
SEJeff, line 245: What exactly does this mean? All VAAs are available via public RPC. This is how front-ends can tell a VAA is available to be redeemed. I don't think anyone currently does this automatically so this entry should be removed.
```suggestion
```

whitepapers/0003_token_bridge.md
```diff
@@ -236,10 +236,16 @@ NewContract [32]uint8
...
+4. A Guardian may send a re-observation request to all other Guardians, causing the old message to be re-observed and a new VAA to be generated.
```
SEJeff, line 246:
```suggestion
3. A Guardian may send a signed re-observation request to the network using the `send-observation-request` admin command. A new valid VAA with an updated Guardian Set Index is generated once enough Guardians have re-observed the old message.
```

tbjump:
Please correct me if I'm wrong, but I thought that in the current implementation a VAA has a maximum of 13 signatures, hence if any one of those signatures is from an old Guardian, the remaining signatures won't make up a valid VAA, unless the Guardian Set shrunk in size.

whitepapers/0003_token_bridge.md
tbjump: Guardians today aren't doing that, but I think the wormhole explorer for example will.

whitepapers/0003_token_bridge.md
```diff
@@ -236,10 +236,16 @@ NewContract [32]uint8
...
-on the target chain, a transfer might not be completed. In case a guardian set change happens in-between and the
-original signer guardian set expires, the transfer will be stuck indefinitely.
+### Transfer completion
+A user who initiated a transfer should call `completeTransfer` within 24 hours. Guardian Sets are guaranteed to be valid for at least 24 hours. If the user waits longer, it could be that the Guardian Set has changed between the time where the transfer was initiated and the the time the user attempts to redeem the VAA. Let's call the Guardian Set at the time of signing `setA` and the Guardian Set at the time of redeeming on the target chain `setB`.
 
+If `setA != setB` and more than 24 hours have passed, there are multiple options for still redeeming the VAA on the target chain:
+1. The Guardians that signed the VAA may still be part of `setB`. In this case, the VAA merely needs to be modified to have the new Guardian Set Index along with any `setA` only guardian signatures removed to make a valid VAA. The updated VAA can then be be redeemed. The typescript sdk includes a [`repairVaa()`](../sdk/js/src/utils/repairVaa.ts) function to perform this automatically.
```
hendrikhofstadt, line 243: could be clearer in that they still need to have a quorum

hendrikhofstadt (changes requested):

SEJeff:
> Please correct me if I'm wrong, but I thought that in the current implementation a VAA has a maximum of 13 signatures, hence if any one of those signatures is from an old Guardian, the remaining signatures won't make up a valid VAA, unless the Guardian Set shrunk in size.

That is wrong :)

I've literally watched @ckeun do this on VAAs with 14 or more signatures and was able to give the bytes to the user who was then able to redeem (without any reobservations)

whitepapers/0003_token_bridge.md
SEJeff: Yeah I guess so. It just seems a bit hand wavy and superfluous. I like keeping that kind of stuff out of design docs, but your call.

whitepapers/0003_token_bridge.md
SEJeff: What exactly is the wormhole explorer going to be doing here? Taking VAAs with an expired guardian set index and presenting them somehow to the user (as raw bytes I'd assume) so they can redeem them?

Apologies if I'm being dense here, but this doesn't make sense to me at all.

tbjump: I added the reference to the wormholescan API which now exists. That should clarify it.

whitepapers/0003_token_bridge.md
tbjump: Removed this point since it was decided to not implement this feature

whitepapers/0003_token_bridge.md
tbjump: clarified.

tbjump:
addressed review comments

SEJeff (approved):

whitepapers/0006_nft_bridge.md
```diff
@@ -130,14 +130,19 @@ NewContract [32]uint8
...
 
+If `setA != setB` and more than 24 hours have passed, there are multiple options for still redeeming the VAA on the target chain:
+1. The Guardians that signed the VAA may still be part of `setB`. In this case, the VAA merely needs to be modified to have the new Guardian Set Index along with any `setA` only guardian signatures removed to make a valid VAA. The updated VAA can then be be redeemed. The typescript sdk includes a [`repairVaa()`](../sdk/js/src/utils/repairVaa.ts) function to perform this automatically.
+2. The intersection between `setA` and `setB` is greater than 2/3 of `setB`, but not all signatures of the VAA are from Guardians in `setB`. Then it may be possible to gather signatures from the other Guardians from other sources. E.g. some Guardians may store additional signatures and make them available in the future (although this is not implemented today), or other infrastructure like the Wormhole Explorer could make this data available.
+3. Guardians may elect to do the above automatically and make those updated VAAs available for users over the public RPC.
```
evan-gray, line 139: option 3 here is not in the list on the token bridge.

whitepapers/0006_nft_bridge.md
```diff
@@ -130,14 +130,19 @@ NewContract [32]uint8
...
 
+If `setA != setB` and more than 24 hours have passed, there are multiple options for still redeeming the VAA on the target chain:
+1. The Guardians that signed the VAA may still be part of `setB`. In this case, the VAA merely needs to be modified to have the new Guardian Set Index along with any `setA` only guardian signatures removed to make a valid VAA. The updated VAA can then be be redeemed. The typescript sdk includes a [`repairVaa()`](../sdk/js/src/utils/repairVaa.ts) function to perform this automatically.
+2. The intersection between `setA` and `setB` is greater than 2/3 of `setB`, but not all signatures of the VAA are from Guardians in `setB`. Then it may be possible to gather signatures from the other Guardians from other sources. E.g. some Guardians may store additional signatures and make them available in the future (although this is not implemented today), or other infrastructure like the Wormhole Explorer could make this data available.
```
evan-gray, line 138: This differs from the description in the above doc

evan-gray (changes requested):
Another option, though I'm not sure if it's worth calling out, is that a new guardian in the current set who replaced a guardian in the previous set (with the same index) whose signature replacement would allow the VAA to have quorum, can re-sign the VAA with the `sign-existing-vaa` or `sign-existing-vaas-csv` admin command.

whitepapers/0006_nft_bridge.md
tbjump: fixed

whitepapers/0006_nft_bridge.md
tbjump: fixed

tbjump:
>Another option, though I'm not sure if it's worth calling out, is that a new guardian in the current set who replaced a guardian in the previous set (with the same index) whose signature replacement would allow the VAA to have quorum, can re-sign the VAA with the sign-existing-vaa or sign-existing-vaas-csv admin command.

Good call-out. There is currently no defined process or software for propagating the output of those two admin commands, so I think it's not worth including here (yet).

evan-gray (approved):

bruce-riley (approved):

PR #2211, opened 2023-01-09
evm gas profile setup

njk-64:
Using foundry to help profile gas costs for parseAndVerifyVM() across arbitary length VAAs, VAAs of length as expected by the Token bridge and Pyth contracts. A first step towards #1870


aadam-10:
this PR needs a new owner to complete.

PR #2215, opened 2023-01-09
tilt: add ibc relayer which connects wormchain and terra2 to dev envi…

nik-suri:
…ronment

https://github.com/wormhole-foundation/wormhole/issues/2197

This PR adds an IBC relayer to the dev tilt environment. The IBC relayer waits for wormchain and terra to startup and then connects them using IBC.

This also adds new accounts (and their corresponding mnemonics) to the genesis files for wormchain and terra2. These accounts are for the relayer to post messages to each chain.


cosmwasm/devnet/config/config.toml
```diff
@@ -370,7 +370,7 @@ timeout_precommit_delta = "500ms"
 # How long we wait after committing a block, before starting on the new
 # height (this gives us a chance to receive some more precommits, even
 # though we already have +2/3).
-timeout_commit = "0.5s"
+timeout_commit = "1s"
```
jynnantonix, line 373: Why are we making this change?  I don't see anything in the commit or PR description that explains the reason.

cosmwasm/devnet/config/genesis.json
```diff
@@ -403,7 +403,7 @@
     },
     "staking": {
       "params": {
-        "unbonding_time": "600s",
+        "unbonding_time": "1814400s",
```
jynnantonix, line 406: Same here.

wormchain/ibc-relayer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM docker.io/golang:1.19.0@sha256:4c00329e17be6fedd8bd4412df454a205348da00f9e0e5d763380a29eb096b75
+
+#used for a readiness probe
+RUN apt-get update
+RUN apt install -y netcat
+RUN apt install -y jq
+
+WORKDIR /app
+
+RUN git clone https://github.com/nik-suri/relayer.git
```
jynnantonix, line 10: What's the long-term plan here?

Also, cloning a git repo isn't really deterministic.  At the very least this should be a shallow clone: `git clone --depth 1 --branch link-then-start-fix https://github.com/nik-suri/relayer.git`

cosmwasm/devnet/config/config.toml
nik-suri: This change is required to sync terra2 and wormchain using IBC. A `timeout_commit` of 0.5s is too fast and leads to Terra2's clock going into the future. Once the clocks on each chain differ by > 10 minutes, we can no longer sync IBC headers.

wormchain/ibc-relayer/Dockerfile
nik-suri: The `link-then-start` relayer command was broken and I fixed it in this PR - https://github.com/cosmos/relayer/pull/1067

The long-term plan is to use the actual relayer codebase when they make a new release.

cosmwasm/devnet/config/genesis.json
nik-suri: This is also required to sync terra2 and wormchain using IBC. An `unbonding_time` of 10 minutes translates into a trusting period equal to 0, which isn't allowed. `unbonding_time = 1814400s` is the default value which translates to a valid trusting period.

cosmwasm/devnet/config/config.toml
jynnantonix: Thanks for the explanation.  This is useful context that should be documented somewhere, either in the comments in this config or in the commit message.

cosmwasm/devnet/config/genesis.json
jynnantonix: Ack.  Please document this either in this config file or in the commit message.

wormchain/ibc-relayer/Dockerfile
jynnantonix: The point of shallow cloning is to not clone the entire history so the current clone + checkout will only work as long as your commit is the head commit of the main branch.

If we know the commit hash we can fetch a tarball at that commit and not bother with any of the git history: `curl -sSL https://github.com/cosmos/relayer/archive/c30b8d9e0ba6b45f8a576f768cad0777de545cb3.tar.gz`

conorpp (approved):

jynnantonix (approved):

wormchain/ibc-relayer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+#used for a readiness probe
+RUN apt-get update
+RUN apt install -y netcat
+RUN apt install -y jq
+
+WORKDIR /app
+
+RUN curl -OL --ssl https://github.com/cosmos/relayer/archive/c30b8d9e0ba6b45f8a576f768cad0777de545cb3.tar.gz
+RUN tar xzf c30b8d9e0ba6b45f8a576f768cad0777de545cb3.tar.gz
```
hendrikhofstadt, line 11: we'd ideally also verify the hash

hendrikhofstadt (approved):

hendrikhofstadt (approved):

jynnantonix (approved):

PR #2219, opened 2023-01-10
node/pythnet: websocket subscription

jumpsiegel:
First, I switched all the go funcs over to using the scissor thread wrappers that give us consistent behavior for unexpected errors

Second, The websocket library used is "nhooyr.io/websocket".  This seems to be the only websocket library still in active development.  gorilla has been made read-only.  x/net/websockets is also deprecated and they are recommending nhooyr.io/websocket.    Finally, the nhooyr library is the only one that was correctly handling the framing of pyth messages that were larger than a MTU on the wire.

If you do not supply the WS URL, it continues to use the old path.. ie, there is no flow changes for solana or for pythnet unless people actively switch to using the websocket subscription.

Another subject to think about is how I am timing out websocket subscriptions after 5 minutes.  This is fine for pyth since data flows in far faster but if we were to use this against standard solana, i suspect we would be timing out far more often


Tiltfile
```diff
@@ -206,6 +208,18 @@ def build_node_yaml():
...
 
+            if pythnet:
+                container["command"] += [
+                    "--pythnetRPC",
+#                    "http://solana-devnet:8899",
+                     "http://pythnet.rpcpool.com",
+                    "--pythnetWS",
+#                   "ws://solana-devnet:8900",
+                    "wss://pythnet.rpcpool.com",
+                    "--pythnetContract",
```
jumpsiegel, line 219: I'll switch these back to the local tilt once the first round of feedback is done

node/go.mod
```diff
@@ -233,7 +233,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
-	nhooyr.io/websocket v1.8.6 // indirect
+	nhooyr.io/websocket v1.8.7 // indirect
 )
```
jumpsiegel, line 237: notice how this was already being used...

node/pkg/watchers/solana/client.go
```diff
@@ -189,31 +346,37 @@ func (s *SolanaWatcher) Run(ctx context.Context) error {
...
+						zap.String("commitment", string(s.commitment)),
+						zap.Uint64("slot", slot),
+						zap.Uint64("lastSlot", lastSlot),
+						zap.Uint64("pendingSlots", slot-lastSlot),
+						zap.Uint64("from", rangeStart), zap.Uint64("to", rangeEnd),
+						zap.Duration("took", time.Since(start)))
+
+					// Requesting each slot
+					for slot := rangeStart; slot <= rangeEnd; slot++ {
+						_slot := slot
```
jumpsiegel, line 363: race detector works! It figured out that I needed to copy slot..

node/pkg/watchers/solana/client.go
```diff
@@ -135,6 +199,80 @@ func NewSolanaWatcher(
...
+			default:
+				rCtx, cancel := context.WithTimeout(ctx, time.Second*300) // 5 minute
+				defer cancel()
+
+				if _, msg, err := ws.Read(rCtx); err != nil {
+					if strings.HasSuffix(err.Error(), "context deadline exceeded") {
+						// When a websocket context times out, it closes the websocket...  This means we have to re-subscribe
+						ws.Close(websocket.StatusNormalClosure, "")
+
+						ws, _, err = websocket.Dial(ctx, *s.wsUrl, nil)
```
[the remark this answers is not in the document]
jumpsiegel: This better?

bruce-riley (approved):

node/pkg/watchers/solana/client.go
```diff
@@ -548,6 +718,59 @@ func (s *SolanaWatcher) fetchMessageAccount(ctx context.Context, logger *zap.Log
...
+	data, err = base64.StdEncoding.DecodeString(value.Account.Data[0])
+	if err != nil {
+		logger.Error(*s.wsUrl, zap.Error(err))
+		p2p.DefaultRegistry.AddErrorCount(s.chainID, 1)
+		return err
+	}
+
+	// Other accounts owned by the wormhole contract seem to send updates...
+	switch string(data[:3]) {
+	case accountPrefixReliable:
```
jynnantonix, line 763: Go case blocks don't fall through by default so this is basically ignoring the `accountPrefixReliable` case.  Is that intentional?

If we actually do want to process the message in the `accountPrefixReliable` case then we either need to add an explicit fallthrough or check for both values in a single case:

```go
switch string(data[:3]) {
case accountPrefixReliable, accountPrefixUnreliable:
	// Handle message
}
```

If we don't want to process the message then we can remove that line completely.

jumpsiegel: good catch

node/pkg/watchers/solana/client.go
```diff
@@ -135,6 +199,77 @@ func NewSolanaWatcher(
...
+						ws.Close(websocket.StatusNormalClosure, "")
+						err, ws = s.SetupSubscription(ctx)
+						if err != nil {
+							return err
+						}
+						continue
+					}
+
+					logger.Error(fmt.Sprintf("ReadMessage: '%s'", err.Error()))
+					if strings.HasSuffix(err.Error(), "EOF") {
```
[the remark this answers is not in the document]
jumpsiegel: ty.. didn't know this

felix-jump (dismissed):

bruce-riley (approved):

hendrikhofstadt (commented):
Is there a reason this is not using [Solana-Go](https://github.com/gagliardetto/solana-go#index--ws-subscriptions--programsubscribe) which we're already using for normal JSON-RPC.

I think the current flow with functions of switching between WS and JSONRPC mode is not very intuitive for a reader of the code. I'd recommend that we build 2 adapters with shared code for shared paths to keep things clean.

hendrikhofstadt (approved):

felix-jump (approved):

PR #2221, opened 2023-01-11
sdk/rust: Add serde_wormhole::RawMessage

jynnantonix:
Add a RawMessage type that can be used to defer parsing parts of a
payload, similar to the `json.RawMessage` from Go.  The implementation
is inspired by `serde_json::RawValue`, which does a similar thing.

When serializing, RawMessage will serialize to a base64-encoded string
if it detects that the data format is human readable (like JSON).
Otherwise it will simply forward the raw bytes to the serializer.

RawMessage has both borrowed and boxed versions.  The borrowed version
is the most efficient as it enables zero-copy handling of the input data
but also requires that the input data already contains raw bytes and is
not suitable when dealing with human-readable formats like JSON.

The boxed version is more flexible as it supports byte slices, base64-
encoded strings, and byte sequences but is slightly less efficient as it
requires copying or decoding the input data.

### DIFF
--- a/cosmwasm/Cargo.lock
+++ b/cosmwasm/Cargo.lock
@@ -1649,8 +1649,10 @@ dependencies = [
 name = "serde_wormhole"
 version = "0.1.0"
 dependencies = [
+ "base64",
  "itoa",
  "serde",
+ "serde_bytes",
  "thiserror",
 ]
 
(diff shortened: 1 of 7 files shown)


jynnantonix:
* main
  * **PR #2221** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2221" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2221?utm_source=stack-comment).

jynnantonix:
I think the RawMessage type is flexible enough that we'll be able to replace all current usage of the `*_with_payload` functions with it.  I plan to do that in a follow-up change.

hendrikhofstadt (dismissed):

hendrikhofstadt (approved):

a5-pickle (approved):

PR #2225, opened 2023-01-11
sdk/js: Use BN.toArrayLike in `tokenIdToMint` function for browserify compatibility

kev1n-peters:


evan-gray (approved):
Hope we searched the SDK for all `.toBuffer`s

a5-pickle (approved):

PR #2150, opened 2022-12-19
wormchain - contract deploy + test

justinschuldt:
This adds cosmwasm contract deploy for wormchain in tilt, for local development, and as a standalone script for CI testing.

Couple things here:
- addresses and such added to `devnet-consts.json`
- added some code to copy `devnet-consts.json` similar to how `.env` files are added to chain directories
- happy path tests for contract interacton
- had to add libwasmvm.x86_64.so to the wormchain docker image to get it build

PR #2228, opened 2023-01-12
wormchain - two nodes for devnet

justinschuldt:
this allows 1 or 2 nodes in the validator set. seems like a good stopping point before making this dynamic for _n_ nodes, like the guardian.

Quick overview of how this works:
- the tilt `--num` flag that controls guardian replicas and the initial guardian set is passed to the wormchain Dockerfile. A  shell script updates `genesis.json` to include guardianSet & validator info that matches the rest of devnet.
- Tiltfile modifies the `devnet/wormchain.yaml` to have a `--num` replicas in the StatefulSet. A Kubernetes Service is created for each pod in the wormchain set, so the individual wormchain instances can resolve each other by hostname for p2p communication.
- a shell script [wormchain/devnet/create-config.sh](https://github.com/wormhole-foundation/wormhole/blob/1a0957ac8846bf3591d4b1fe255712d15c62a7d4/wormchain/devnet/create-config.sh) is executed at runtime so it can read the hostname of the container and apply the appropriate config for that instance (name, keys, p2p networking).
- `wormchain-0` collects and disseminates p2p networking info. It acts as a "Seed Node", an all the other instances use it as their source of truth for p2p addresses. This setup allows a much simpler config than supplying each node with all the addresses of the others. [Relevant reading](https://kb.certus.one/peers.html#tendermint-p2p-layer).

If/when we want to run more than 2 instances, we'll just have to create keys and add them to `wormchain/devnet/`, generate accounts & gentxs, and save the `guardianKey`/`validatorAddr` somewhere so it can be read at build time. 
Ultimately I think we should put devnet wormchain keys & data in [scripts/devnet-consts.json](https://github.com/wormhole-foundation/wormhole/blob/main/scripts/devnet-consts.json#L238-L247), so they can be read by whatever needs it (devnet guardian instances, etc.). I'll circle back after this [config update lands](https://github.com/wormhole-foundation/wormhole/pull/2150/files#diff-dca0ff62e8fbc9411c0896d795c847245c9f5f14adf03833eb2de0e53deee014).


jumpsiegel (approved):

Tiltfile
```diff
@@ -721,14 +728,52 @@ if wormchain:
...
+            if obj["kind"] == "Service" and obj["metadata"]["name"] == "wormchain-0":
+
+                # make a Service for each replica so we can resolve it by name from other pods.
+                # copy wormchain-0's Service then set the name and selector for the instance.
+                for instance_num in list(range(1, num_instances)):
+                    instance_name = 'wormchain-%s' % (instance_num)
+
+                    # Copy the Service's properties to a new dict, by value, three levels deep.
+                    # tl;dr - if the value is a dict, use a comprehension to copy it immutably.
+                    service = { k: ({ k2: ({ k3:v3
```
conorpp, line 754: maybe use deepcopy?
```
import copy
# ...
service = copy.deepcopy(obj)
```

wormchain/devnet/base/config/config.toml
```diff
@@ -93,7 +93,7 @@ laddr = "tcp://0.0.0.0:26657"
 # A list of origins a cross-domain request can be executed from
 # Default value '[]' disables cors support
 # Use '["*"]' to allow any origin
-cors_allowed_origins = ["*", ]
+cors_allowed_origins = []
```
conorpp, line 96: i think we should keep cors enabled for development, will be needed for any web or explorer like development

wormchain/devnet/client/config/config.toml
```diff
@@ -93,7 +93,7 @@ laddr = "tcp://0.0.0.0:26657"
 # A list of origins a cross-domain request can be executed from
 # Default value '[]' disables cors support
 # Use '["*"]' to allow any origin
-cors_allowed_origins = ["*", ]
+cors_allowed_origins = []
```
conorpp, line 96: reenable cors here

wormchain/devnet/create-config.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# update the moniker
+sed -i "s/moniker = \"wormchain\"/moniker = \"$hostname\"/g" ${home_path}/config/config.toml
+
+# set the external address so wormchain peers can resolve each other
+sed -i "s/external_address = \"\"/external_address = \"${hostname}:26656\"/g" ${home_path}/config/config.toml
+
+if [ $instance -eq 0 ] && [ $NUM_GUARDIANS -ge 2 ]; then
+  echo "$hostname: enabling seed mode in config.toml."
+  sed -i "s/pex = false/pex = true/g" ${home_path}/config/config.toml
```
conorpp, line 46: why not always have `pex = true` in config.toml?  I think it's the default?

conorpp:
nice work!  good idea to use the seed node & 1 config file.

conorpp:
I almost wonder if we want to revisit with >2 validators, we could create multiple genesis files for each case too.

conorpp (dismissed):

Tiltfile
justinschuldt: I tried _real_ hard to import copy, I could not get it working with tilt `v0.30.5`. Does importing like that work for you?

I looked into starlark docs and it seems like it should just be there, you don't have to import standard libs? They are just in the global namespace it seems (not sure though). I'm not sure if Tilt is not exposing them, or what. I'm open to whatever thoughts here.

wormchain/devnet/create-config.sh
justinschuldt: no reason in particular beyond just being explicit about who is doing what. I just checked and `pex` is `true` by default. I'll flip it on by default.

wormchain/devnet/client/config/config.toml
justinschuldt: done

wormchain/devnet/base/config/config.toml
justinschuldt: kk, done

conorpp (approved):

Tiltfile
```diff
@@ -256,12 +263,12 @@ def build_node_yaml():
             if wormchain:
                 container["command"] += [
                     "--wormchainWS",
-                    "ws://guardian-validator:26657/websocket",
+                    "ws://wormchain:26657/websocket",
                     "--wormchainLCD",
-                    "http://guardian-validator:1317"
+                    "http://wormchain:1317"
```
nik-suri, line 268: Let's update this so that each guardian connects to the corresponding wormchain node? For example, `guardian-0` connects to `wormchain-0` and `guardian-1` connects to `wormchain-1`

justinschuldt: These configs apply to the StatefulSet (all pods). To apply configs per instance we would ether need to do it with an InitContainer and ConfigMap, or at runtime like the guardian key (quicker & easier). We need to do this for the supplying wormchain keys anyway, adding unique RPC addresses will be trivial once the key-loading is in place.

I captured this in an issue https://github.com/wormhole-foundation/wormhole/issues/2246 and added it to the milestone.

nik-suri (approved):

hendrikhofstadt (changes requested):
Does this actually work with a single guardian. The `genesis.json` has 2 gentxs registering validators so there should be 2 active.
The genesis generation script takes values `> 2` but is only built to support 1 or 2 (and I'm not sure it actually properly registers them as validators, it seems it just whitelists them as guardians leaving them to do the validator registration).

justinschuldt:
> Does this actually work with a single guardian. The `genesis.json` has 2 gentxs registering validators so there should be 2 active. The genesis generation script takes values `> 2` but is only built to support 1 or 2 (and I'm not sure it actually properly registers them as validators, it seems it just whitelists them as guardians leaving them to do the validator registration).

@hendrikhofstadt Yes, this works with a single guardian. The gentxs in `genesis.json` register node keys as cosmos validators. Consensus can/will be determined by the validators that are up and communicating via p2p. There is a distinction between being a cosmos validaor (gentx message of type `/cosmos.staking.v1beta1.MsgCreateValidator`), and being a wormhole validator (genesis.json `app_state.wormhole.guardianValidatorList`).

As far is registering wormhole validators, the [guardian set and validator list](https://github.com/wormhole-foundation/wormhole/blob/a88b3ccc6e232b15c948e5e8d60be22a4af01b9e/wormchain/devnet/base/config/genesis.json#L417-L431) in `genesis.json` is what is determines the "wormchain validators" (ie. nodes with validator keys that have been signed+registered on behalf of a guardian). Setting the keys of `guardianSetList` and `guardianValidatorList` sets the chain state to be as if the guardians signed validator keys and then registered them, and then were elected to the guardian set.

This change will make the validator set 1 or 2 guardians. We're going to support `n` wormchain validators in the future, but decided to do two for now to unblock wormchain testing.

I verified this all works by running the tests from the [wormchain contracts deploy + test PR](https://github.com/wormhole-foundation/wormhole/pull/2150), with 1 and 2 validators. Not just the two nodes connecting, but interacting with contracts that take/verify VAAs and verify signatures against the active guardian set, etc.

hendrikhofstadt (approved):
Thanks for clarifying Justin!

evan-gray (approved):
Approving based on scope (devnet only), prior approvals, and CI passing.

PR #2240, opened 2023-01-13
sdk/js: run tests in sequence instead of in parallel (temporary fix)

tbjump:


jumpsiegel:
do we still need this?  the tests have been passing pretty reliably now

tbjump:
just a draft to test some inconsistency with my local tilt and the github action tilt.

evan-gray (changes requested):
Agree with @jumpsiegel , I don’t think this should be necessary.

Issue #2188, opened 2023-01-03
Guardian accounting interface should batch requests


bruce-riley:
+1, if multiple tx's are submitted in parallel, there will be sequence number collisions.

However, only submitting 1 message per 1 tx per 1 block I think will run into a bottleneck pretty quickly.

I think we'll need to submit a variable number of messages per tx.  So the new goroutine should consume up to ~100-200 messages at a time from the channel and submit them in a single tx per block.
Multiple observations/messages can be included here in the TX:
https://github.com/wormhole-foundation/wormhole/blob/9fe03b07b27f894d95d4e8cbefb0660046272964/node/pkg/wormconn/send_tx.go#L37

This could make error handling more tricky.  If one of the messages gets an error, I believe the whole tx will be rejected.

_Originally posted by @conorpp in https://github.com/wormhole-foundation/wormhole/pull/2181#discussion_r1060718754_

conorpp:
Just to update -- there are multiple observations submitted in a single tx, so I think the current 1 tx per block is fine.  The optimization could just be:
1. Read all messages from channel. 
2. Create new message containing observations of input messages, and submit the new message instead.

PR #2216, opened 2023-01-10
cosmwasm: accounting: Use cw_transcode for events

jynnantonix:
Use cw_transcode to ensure that event attribute values are always
encoded as proper json, making it easier for clients to parse them back
into structured data.

This also lets us reuse the input messages for the events, reducing the
number of different structs that we need to track.

PR #2241, opened 2023-01-13
cosmwasm: accounting: Return transfer status for observations

jynnantonix:
When submitting a batch of observations, we don't want an observation
for an already committed transfer to fail the entire batch.  This leads
to more complexity in the guardian and also delays all the legitimate
observations by at least one more block (~5 seconds).

Fix this by returning the transfer status of each observation as part
of the response data.  Observations for committed transfers will get
a `TransferStatus::Committed` response without failing the tx as long
as the digest of the observation matches the digest of the committed
transfer.  Digest mismatches are still an error and will fail the entire
batch.

Part of #2188

### DIFF
--- a/cosmwasm/contracts/wormchain-accounting/src/contract.rs
+++ b/cosmwasm/contracts/wormchain-accounting/src/contract.rs
@@ -3,7 +3,7 @@ use std::marker::PhantomData;
 use accounting::{
     query_balance, query_modification,
     state::{account, transfer, Modification, TokenAddress, Transfer},
-    validate_transfer, TransferError,
+    validate_transfer,
 };
 use anyhow::{ensure, Context};
 #[cfg(not(feature = "library"))]
@@ -29,8 +29,9 @@ use crate::{
     msg::{
         AllAccountsResponse, AllModificationsResponse, AllPendingTransfersResponse,
         AllTransfersResponse, BatchTransferStatusResponse, ChainRegistrationResponse, ExecuteMsg,
-        MigrateMsg, MissingObservation, MissingObservationsResponse, Observation, QueryMsg,
-        TransferDetails, TransferStatus, Upgrade,
+        MigrateMsg, MissingObservation, MissingObservationsResponse, Observation, ObservationError,
+        ObservationStatus, QueryMsg, SubmitObservationResponse, TransferDetails, TransferStatus,
+        Upgrade,
     },
     state::{Data, PendingTransfer, CHAIN_REGISTRATIONS, DIGESTS, PENDING_TRANSFERS},
 };
@@ -118,20 +119,39 @@ fn submit_observations(
     let observations: Vec<Observation> =
         from_binary(&observations).context("failed to parse `Observations`")?;
 
-    let events = observations
-        .into_iter()
-        .map(|o| {
-            let key = transfer::Key::new(o.emitter_chain, o.emitter_address.into(), o.sequence);
-            handle_observation(deps.branch(), o, guardian_set_index, quorum, signature)
-                .with_context(|| format!("failed to handle observation for key {key}"))
-        })
-        .filter_map(Result::transpose)
-        .collect::<anyhow::Result<Vec<_>>>()
-        .context("failed to handle `Observation`")?;
+    let mut responses = Vec::with_capacity(observations.len());
+    let mut events = Vec::with_capacity(observations.len());
+    for o in observations {
+        let key = transfer::Key::new(o.emitter_chain, o.emitter_address.into(), o.sequence);
+        match handle_observation(deps.branch(), o, guardian_set_index, quorum, signature) {
+            Ok((status, event)) => {
+                responses.push(SubmitObservationResponse { key, status });
+                if let Some(evt) = event {
+                    events.push(evt);
+                }
+            }
+            Err(e) => {
+                let err = ObservationError {
+                    key,
+                    error: format!("{e:#}"),
+                };
+                let evt = cw_transcode::to_event(&err)
+                    .context("failed to transcode observation error")?;
+                events.push(evt);
+                responses.push(SubmitObservationResponse {
+                    key: err.key,
+                    status: ObservationStatus::Error(err.error),
+                });
+            }
+        }
+    }
+
+    let data = to_binary(&responses).context("failed to serialize transfer details")?;
 
     Ok(Response::new()
         .add_attribute("action", "submit_observations")
         .add_attribute("owner", info.sender)
+        .set_data(data)
         .add_events(events))
 }
 
@@ -141,8 +161,10 @@ fn handle_observation(
     guardian_set_index: u32,
     quorum: usize,
     sig: Signature,
-) -> anyhow::Result<Option<Event>> {
+) -> anyhow::Result<(ObservationStatus, Option<Event>)> {
     let digest_key = DIGESTS.key((o.emitter_chain, o.emitter_address.to_vec(), o.sequence));
+    let tx_key = transfer::Key::new(o.emitter_chain, o.emitter_address.into(), o.sequence);
+
     if let Some(saved_digest) = digest_key
         .may_load(deps.storage)
         .context("failed to load transfer digest")?
@@ -152,10 +174,8 @@ fn handle_observation(
             bail!(ContractError::DigestMismatch);
         }
 
-        bail!(TransferError::DuplicateTransfer);
+        return Ok((ObservationStatus::Committed, None));
     }
-
-    let tx_key = transfer::Key::new(o.emitter_chain, o.emitter_address.into(), o.sequence);
 
     let key = PENDING_TRANSFERS.key(tx_key.clone());
     let mut pending = key
@@ -181,7 +201,7 @@ fn handle_observation(
         key.save(deps.storage, &pending)
             .context("failed to save pending transfers")?;
 
-        return Ok(None);
+        return Ok((ObservationStatus::Pending, None));
     }
 
     let msg = serde_wormhole::from_slice::<Message<&RawMessage>>(&o.payload)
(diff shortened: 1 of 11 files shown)


jynnantonix:
* main
  * **PR #2207** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2207" target="_blank">Graphite</a> 
    * **PR #2216** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2216" target="_blank">Graphite</a> 
      * **PR #2241** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2241" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2241?utm_source=stack-comment).

cosmwasm/contracts/wormchain-accounting/tests/submit_observations.rs
```diff
@@ -798,9 +836,9 @@ fn emit_event_with_quorum() {
         )
         .add_attribute("payload", serde_json_wasm::to_string(&o.payload).unwrap());
 
-    assert_eq!(responses.len(), quorum);
+    assert_eq!(responses.len(), num_guardians);
     for (i, r) in responses.into_iter().enumerate() {
-        if i < quorum - 1 {
+        if i < quorum - 1 || i >= quorum {
```
conorpp, line 841: nit: I believe this is always true

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -117,20 +117,28 @@ fn submit_observations(
...
+    for o in observations {
+        let key = transfer::Key::new(o.emitter_chain, o.emitter_address.into(), o.sequence);
+        let (status, event) =
             handle_observation(deps.branch(), o, guardian_set_index, quorum, signature)
-                .with_context(|| format!("failed to handle observation for key {key}"))
-        })
-        .filter_map(Result::transpose)
-        .collect::<anyhow::Result<Vec<_>>>()
-        .context("failed to handle `Observation`")?;
+                .with_context(|| format!("failed to handle observation for key {key}"))?;
```
conorpp, line 126: If there's an error with an observation, should we emit an error event or something instead of returning via `?` operator?  I'm not sure how likely it is, but if it happens, it will block a guardian from submitting all observations I think.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -117,20 +117,28 @@ fn submit_observations(
...
+            events.push(evt);
+        }
+    }
+
+    let data = to_binary(&details).context("failed to serialize transfer details")?;
 
     Ok(Response::new()
         .add_attribute("action", "submit_observations")
         .add_attribute("owner", info.sender)
+        .set_data(data)
```
hendrikhofstadt, line 141: TransferDetails can be a pretty large object. Have we considered the impact of this on the transaction result store of validators / gas costs in the wasm environment?

bruce-riley: Related to this, is there a suggested limit to the size of a batch that the guardian should submit?

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: Now that duplicate signatures are not treated as errors, any errors we do encounter would be more serious ("insufficient balance", "digest mismatch", etc) so I don't think it's unreasonable to fail the entire batch.

That said, I'm also ok if we want to add a `TransferStatus::Error` variant so that we can continue processing the rest of the batch.

@bruce-riley @hendrikhofstadt thoughts?

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: @hendrikhofstadt I did consider this when making this change.  I think we can make the response much smaller by not returning any data with the `TransferStatus::Pending` variant.  Then the only data would be the actual transfer details + digest for committed transfers. It's not clear if the on-chain state for pending transfers is that useful outside of testing anyway.

@bruce-riley The contract itself has no limits.  We're only limited by the max transaction size and the gas limit for the tx.

cosmwasm/contracts/wormchain-accounting/tests/submit_observations.rs
jynnantonix: It is false when `i == quorum - 1`.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: Actually while thinking about this some more I found a couple of places where we can reduce the size of the on-chain state, which would also reduce the amount of data returned in a TransferStatus (see [1](https://github.com/wormhole-foundation/wormhole/pull/2241/commits/8b442fc3c4429c1baeebc8cabf8969cd84242ecb) and [2](https://github.com/wormhole-foundation/wormhole/pull/2241/commits/80e29d835dbcfef8828973d7c832ad887194f6c0)). 

We can still drop the data returned in the `TransferStatus::Pending` variant if we feel like it's still too much data in the response.

bruce-riley: I don't think the guardian needs any additional info with the pending ones. Actually, I'm not even sure it needs anything with the completed ones (A digest mismatch would throw an error, right?).

jynnantonix: Yes, a digest mismatch would throw an error in the completed case.

Basically it just boils down to whether we want to re-use the types from the queries or whether we want a new type that's basically just `TransferStatus` but without the data.

jynnantonix:
I can split out the 2 follow up commits that reduce the size of the on-chain state into separate PRs if preferred.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
```diff
@@ -107,80 +107,93 @@ fn submit_observations(
...
         .map(Option::unwrap_or_default)
         .context("failed to load `PendingTransfer`")?;
-    let data = match pending
-        .iter_mut()
-        .find(|d| d.guardian_set_index() == guardian_set_index && d.observation() == &o)
-    {
+    let data = match pending.iter_mut().find(|d| {
+        d.guardian_set_index() == guardian_set_index
+            && d.digest() == &digest
+            && d.tx_hash() == &o.tx_hash
```
hendrikhofstadt, line 174: any chance you're not also matching for emitter chain here?

hendrikhofstadt (dismissed):

a5-pickle (approved):
Maybe I missed some documentation somewhere, but is the plan to have a hard cap of number of guardians be 127?

I'm fine with the type of `signatures` being `u128` since there are only 19 guardians currently. But if the plan still is to allow up to 255 (guardian set upgrades only allow for encoding `u8` currently), we will have to change this `signatures` type down the road.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: I've changed the enum here to `ObservationStatus` where the `Pending` and `Committed` variants don't have any data associated with them.  It also includes an `Error` variant with the error string.

This should significantly reduce the amount of data returned by the contract and also let us keep processing the rest of the batch if an observation fails.  Not failing on error also means we can now emit an event that guardians can listen for when an observation fails.

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: The emitter chain is included as part of the digest calculation so I didn't feel like it was necessary to check it again (whereas the tx_hash is not part of the digest).

cosmwasm/contracts/wormchain-accounting/src/contract.rs
jynnantonix: I've changed this so that we return an error status and emit an event whenever we fail to handle the observation.  This will not fail the other observations in the batch.

jynnantonix:
> Maybe I missed some documentation somewhere, but is the plan to have a hard cap of number of guardians be 128?
> 
> I'm fine with the type of `signatures` being `u128` since there are only 19 guardians currently. But if the plan still is to allow up to 255 (guardian set upgrades only allow for encoding `u8` currently), we will have to change this `signatures` type down the road.

I mentioned this in the commit message (but not the PR description) but I chose u128 only because it's the largest integer supported by rust.  If we ever have more than 128 guardians we'll need to change this code and migrate the contract but I'm ok with punting that problem to later.

a5-pickle (approved):

bruce-riley:
I really like these latest changes! They have made the guardian code much more straight forward.

I did notice one thing. I'm not sure if it can be fixed or not. The go code is having to treat the EmitterAddress in both the observation response and wasm-ObservationError event as []byte rather than vaa.Address, like in the observation request and the wasm-Observation event. If this is a big deal to change, then don't worry about it.

jynnantonix:
> I did notice one thing. I'm not sure if it can be fixed or not. The go code is having to treat the EmitterAddress in both the observation response and wasm-ObservationError event as []byte rather than vaa.Address, like in the observation request and the wasm-Observation event. If this is a big deal to change, then don't worry about it.

I changed the emitter address to use hex encoding instead of base64.

bruce-riley:
I verified that the emitter address changes work with the guardian.

bruce-riley (dismissed):
I just verified the message changes.

hendrikhofstadt (dismissed):
Could also be good to test that the error events are published

jynnantonix:
> Could also be good to test that the error events are published

Good point.  I will add this test in a follow-up PR so that this one doesn't get blocked.

bruce-riley:
FWIW, I did confirm that I see the error event in tilt when I force an digest error.

evan-gray:
@jynnantonix can you resolve conflicts on this one please.

a5-pickle (dismissed):

jynnantonix:
> > Could also be good to test that the error events are published
> 
> Good point. I will add this test in a follow-up PR so that this one doesn't get blocked.

Had to rebase this PR so I also added this test.

hendrikhofstadt (approved):

a5-pickle (approved):

PR #2248, opened 2023-01-14
node: publicweb: add logging of body

tbjump:


jynnantonix (commented):
What's the reason for this change?

node/cmd/guardiand/publicweb.go
```diff
@@ -76,10 +77,16 @@ func publicwebServiceRunnable(
...
+			dumped, err := httputil.DumpRequest(req, true)
+			if err != nil {
+				dumped = []byte("error dumping")
+			}
+
 			logger.Info("public rpc request",
 				zap.String("method", req.Method),
 				zap.String("path", req.URL.Path),
 				zap.String("remote_addr", req.RemoteAddr),
+				zap.ByteString("body", dumped),
```
bruce-riley, line 89: I'm assuming there is no concern about the amount of output here?

bruce-riley (approved):

evan-gray:
Discussed with @hendrikhofstadt and as @bruce-riley noted above we're uncomfortable with the amount of data that could be thrown at this. I propose anyone looking to log the body (or more information related to RPC / web request) add this a layer up, say in an nginx reverse proxy.

PR #2249, opened 2023-01-16
sdk/rust: Remove `*_with_payload` methods

jynnantonix:
The RawMessage type provides a more flexible way to handle trailing
payloads so replace all usage of the `*_with_payload` functions to use
`RawMessage` instead.

There should be no functional change.

PR #2250, opened 2023-01-16
sdk/rust: Properly support tokenbridge payload3 messages

jynnantonix:
Add the payload as an explicit field to the `TransferWithPayload` enum
variant.  This is a generic parameter that defaults to `Box<RawMessage>`
for maximum flexibility (and to avoid leaking lifetimes higher up the
stack) but users are encouraged to replace this default type parameter
with an explicit `&RawMessage` in places where the serde_wormhole data
format is used.

The main benefit of this change is that the payload is now included as
part of the actual message and no longer requires callers to awkwardly
append it after serialization.  This is especially useful in human-
readable formats like JSON (see the `transfer_with_payload` test in
token.rs for an example of this simplification).

The main downside is that this now requires explicit type annotations
when using the non-payload3 variants so that the compiler will pick up
the default generic parameter.  This is a relatively minor inconvenience
and the benefit appears to be worth the cost.

There should be no functional change.

### DIFF
--- a/cosmwasm/contracts/wormchain-accounting/src/contract.rs
+++ b/cosmwasm/contracts/wormchain-accounting/src/contract.rs
@@ -184,7 +184,7 @@ fn handle_observation(
         return Ok(None);
     }
 
-    let (msg, _) = serde_wormhole::from_slice::<(Message, &RawMessage)>(&o.payload)
+    let msg = serde_wormhole::from_slice::<Message<&RawMessage>>(&o.payload)
         .context("failed to parse observation payload")?;
     let tx_data = match msg {
         Message::Transfer {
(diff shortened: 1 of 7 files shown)


jynnantonix:
* main
  * **PR #2221** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2221" target="_blank">Graphite</a> 
    * **PR #2249** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2249" target="_blank">Graphite</a> 
      * **PR #2250** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2250" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2250?utm_source=stack-comment).

hendrikhofstadt (approved):

hendrikhofstadt (approved):
Just making sure I understand this correctly. Previously the digest_with_payload function would have appended the payload twice if present in the struct, correct? Is there a test with hardcoded digests that would catch if there was a bug introduced?

a5-pickle (approved):

jynnantonix:
> Just making sure I understand this correctly. Previously the digest_with_payload function would have appended the payload twice if present in the struct, correct? Is there a test with hardcoded digests that would catch if there was a bug introduced?

There is no test with a hardcoded digest but there is a test that ensures that the digests produced by hashing the raw bytes, calling `Body::digest` with a typed payload, and calling `Body::digest` with a `&RawMessage` payload are all the same: https://github.com/wormhole-foundation/wormhole/blob/421a030dcad2254601b4a897b724551b50f00f87/sdk/rust/core/src/vaa.rs#L330-L357

hendrikhofstadt (approved):

jynnantonix:
I'm merging this for now but I will add a test with a hardcoded digest in a follow-up change.

PR #2256, opened 2023-01-17
worm: speed up `evm info` by improving concurrency

kcsongor:
Using `await` sequentially will force sequential blocking evaluation of code that could otherwise happily run concurrently. Instead of using `await`, this PR uses `Promise.all` to execute a bunch of independent computations in parallel.
Verified (by diffing) that the output matches the old version.

Before this PR, the following command
```
worm evm info -c ethereum -m NFTBridge -n mainnet
```
took around 10s to run on my machine. With this PR, it runs in ~1s.

### DIFF
--- a/clients/js/evm.ts
+++ b/clients/js/evm.ts
@@ -37,20 +37,43 @@ export async function query_contract_evm(
       const core = Implementation__factory.connect(contract_address, provider)
       result.address = contract_address
       result.currentGuardianSetIndex = await core.getCurrentGuardianSetIndex()
-      result.guardianSet = {}
-      for (let i of Array(result.currentGuardianSetIndex + 1).keys()) {
-        let guardian_set = await core.getGuardianSet(i)
-        result.guardianSet[i] = { keys: guardian_set[0], expiry: guardian_set[1] }
-      }
-      result.guardianSetExpiry = await core.getGuardianSetExpiry()
-      result.chainId = await core.chainId()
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):
Nice!

kev1n-peters (approved):

clients/js/evm.ts
```diff
@@ -59,23 +82,51 @@ export async function query_contract_evm(
...
+        tb.chainId(),
+        tb.finality(),
+        maybeUnsupported(tb.evmChainId()),
+        maybeUnsupported(tb.isFork()),
+        tb.governanceChainId(),
+        tb.governanceContract(),
+        tb.WETH(),
+        registrationsPromise
+      ])
+      result.wormhole = wormhole
```
heyitaki, line 115: Would be useful for legibility and correctness to type `result`. Could declare as `Partial<ResultType>` so these incremental assignments don't error and assert `ResultType` before returning.

heyitaki (commented):
Nice, definitely looks like this will speed things up. Would like to see return types

clients/js/evm.ts
kcsongor: good idea!

bruce-riley (approved):

clients/js/evm.ts
kcsongor: will merge this is now, I want to do a general cleanup of the client code in a separate PR

bruce-riley (approved):

kev1n-peters (approved):

PR #2267, opened 2023-01-19
sdk: Aptos NFT SDK

heyitaki:
This PR adds a JS SDK for the Aptos NFT bridge. Previously, Aptos token bridge tests used the account of the deployer wallet in the tests because it was pre-funded from Tilt setup. However, using the same wallet in the NFT tests caused transient errors when both the Aptos token bridge and NFT bridge tests ran simultaneously since the respective clients were submitting transactions with the same sequence numbers (for the same account). Therefore, the tests now use randomly generated accounts that are funded within the tests themselves.


heyitaki:
Formatting commit might seem random, but something in there was causing the `lint-and-tests` check to fail

sdk/js/src/utils/consts.ts
```diff
@@ -328,7 +329,8 @@ const TESTNET = {
     core: "0x5bc11445584a763c1fa7ed39081f1b920954da14e04b32440cba863d03e19625",
     token_bridge:
       "0x576410486a2da45eee6c949c995670112ddf2fbeedab20350d506328eefc9d4f",
-    nft_bridge: undefined,
+    nft_bridge:
+      "0x1bdffae984043833ed7fe223f7af7a3f8902d04129b14f801823e64827da7130",
```
kev1n-peters, line 333: I don't think this is actually deployed in testnet

heyitaki:
`getEmitterAddressAptos` is intentionally left unimplemented as there is no good way to fetch the emitter address from the Aptos core bridge, which uses sequential IDs given to each new registered emitter. Because the Aptos SDK doesn't return [types of values in account resources](https://aptos-labs.github.io/ts-sdk-doc/types/Types.MoveStructValue.html), there is no way to programmatically search for and find the emitter address, since we would not be able to tell it apart from another struct with the same fields.

sdk/js/src/nft_bridge/__tests__/aptos-integration.ts
```diff
@@ -0,0 +1,430 @@
...
+      await aptosClient.getEventsByEventHandle(
+        aptosAccount.address(),
+        "0x3::token::TokenStore",
+        "deposit_events",
+        { limit: 1 }
+      )
+    )[0] as DepositEvent;
+    const depositTokenId = event.data.id;
+
+    // transfer NFT from Aptos to Solana
```
kev1n-peters, line 239: Should be `to Ethereum`

sdk/js/src/utils/aptos.ts
```diff
@@ -239,6 +244,114 @@ export const coalesceModuleAddress = (str: string): string => {
   return str.split("::")[0];
 };
 
+/**
+ * The NFT bridge creates resource accounts, which in turn create a collection
+ * and mint a singl token for each transferred NFT. This method derives the
```
kev1n-peters, line 249: singl -> single

solana/keys/solana-devne2.json
```diff
@@ -0,0 +1 @@
+[118,84,4,83,83,183,31,184,20,172,95,146,7,107,141,183,124,196,66,246,215,243,54,61,118,188,239,237,168,108,227,169,93,119,180,216,9,169,30,4,167,235,188,51,70,24,181,227,189,59,163,161,252,219,17,105,197,241,19,66,205,188,232,131]
```
evan-gray, line 1: `devne` ? typo

evan-gray (changes requested):

kev1n-peters (dismissed):

sdk/js/src/nft_bridge/getForeignAsset.ts
```diff
@@ -99,3 +107,70 @@ export async function getForeignAssetSolana(
...
+ * @param originChain
+ * @param originAddress External address of token on origin chain, or token hash
+ * if origin chain is Aptos
+ * @returns Unique token identifier on Aptos
+ */
+export async function getForeignAssetAptos(
+  client: AptosClient,
+  nftBridgeAddress: string,
+  originChain: ChainId | ChainName,
+  originAddress: Uint8Array
```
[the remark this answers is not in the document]
heyitaki: I think you're right, we should be taking in token id and iterate through token creation events until `name === tokenId`; will push fix and a test shortly

evan-gray (approved):
Approving based on a cursory review and testing. Definitely some complexity here that would benefit from future unit tests.

evan-gray (approved):

a5-pickle (approved):
Reviewed changes in _solana_ subdirectory.

a5-pickle (dismissed):

evan-gray (approved):

kev1n-peters (approved):

a5-pickle (approved):

PR #2272, opened 2023-01-19
wormchain - peer-exchange reactor off by default

justinschuldt:
this fixes the incessant `wormchain │ {"level":"error","module":"p2p","err":"EOF","peer":{"Logger":{},"Data":{}},"time":"2023-01-19T01:56:18Z","message":"Stopping peer for error"}` log in CI

### DIFF
--- a/wormchain/devnet/base/config/config.toml
+++ b/wormchain/devnet/base/config/config.toml
@@ -249,7 +249,7 @@ send_rate = 5120000
 recv_rate = 5120000
 
 # Set true to enable the peer-exchange reactor
-pex = true
+pex = false
 
 # Seed mode, in which node constantly crawls the network and looks for
 # peers. If another node asks it for addresses, it responds and disconnects.
--- a/wormchain/devnet/client/config/config.toml
+++ b/wormchain/devnet/client/config/config.toml
@@ -249,7 +249,7 @@ send_rate = 5120000
 recv_rate = 5120000
 
 # Set true to enable the peer-exchange reactor
-pex = true
+pex = false
 
 # Seed mode, in which node constantly crawls the network and looks for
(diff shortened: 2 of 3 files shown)


nik-suri (approved):
lgtm

how does peering happen in the dev env if `pex` is off and `persistent_peers` is empty?

hendrikhofstadt:
Generally this should be on in prod. In the CI environment we’ll need it if we don’t want to add all peers to the persistent peers or explicitly connect them via API.

conorpp (approved):

justinschuldt:
> lgtm
> 
> how does peering happen in the dev env if `pex` is off and `persistent_peers` is empty?

@nik-suri `wormchain-0` has `pex = true`, and is a seed_node. See https://github.com/wormhole-foundation/wormhole/pull/2272/files#diff-d9cc7a9e9d3d83b62a024703da20f4d56d037937b2286be43d4cd0efdfdcff3bR46

justinschuldt:
> Generally this should be on in prod. In the CI environment we’ll need it if we don’t want to add all peers to the persistent peers or explicitly connect them via API.
> 
> If its just about the logs we could explicitly filter them in the config which afaik allows quite flexible log filtering.

@hendrikhofstadt this is for devnet (`/wormchain/devnet/base/config/config.toml`), prod config is `/wormchain/mainnet/config.toml`.

`wormchain-0` has `pex = true` and `seed_mode = true`. `wormchain-* >= 1` connect to `wormchain-0` to get the p2p address book.  having `pex = true` on all the nodes cripples performance, block times slow, transactions take seconds to execute, I assume because they're spinning looking for peers and failing to connect. 

supplying `persistant_peers` is only needed if you're not using using a `seed_node`, or you want a robust setup for prod, etc, which is not a concern for devnet.

PR #2278, opened 2023-01-22
CI: debug: skip near in required tests

evan-gray:
Testing for impoveded consistency / reduced flakiness

PR #2279, opened 2023-01-23
CI: split lint-and-tests

evan-gray:
Testing for impoveded consistency / reduced flakiness

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -218,7 +218,6 @@ jobs:
       - name: Run tests via docker
         run: make test-docker
 
-
   terra:
     runs-on: ubuntu-20.04
     steps:
@@ -277,8 +276,8 @@ jobs:
           go-version: "1.19.3"
       - run: cd sdk/vaa && go test
 
-  # Run Go linters, Go tests and other outside-of-Tilt things.
-  lint-and-tests:
+  # Run Go linters
+  node-lint:
     # The linter is slow enough that we want to run it on the self-hosted runner
     runs-on: tilt-kube-public
     concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -300,16 +299,31 @@ jobs:
...
-        run: cd node && go test -v -race -ldflags '-extldflags "-Wl,--allow-multiple-definition" ' ./...
       - name: Ensure generated proto matches
         run: |
           rm -rf node/pkg/proto
           docker build --target go-export -f Dockerfile.proto -o type=local,dest=node .
           git diff --name-only --exit-code && echo "✅ Generated proto matches committed proto" || (echo "❌ Generated proto differs from committed proto, run \`rm -rf node/pkg/proto && docker build --target go-export -f Dockerfile.proto -o type=local,dest=node .\` and commit the result" >&2 && exit 1)
 
+  # Run Go tests
+  node-tests:
+    # The tests are slow enough that we want to run it on the self-hosted runner
```
kcsongor, line 310: are they though? we might be able to get away without using the self-hosted one here

evan-gray: Yes, left this comment in the other PR
https://github.com/wormhole-foundation/wormhole/pull/2278#issuecomment-1399630069
> baseline
>
> - self-hosted
>   - lints 15s
>   - tests 5s
> - cloud
>   - lints 4m 50s
>   - tests 12m 50s

evan-gray:
Going to try pushing up a couple test commits to see if I can get this to fail (like with the tests back where they were)

evan-gray:
Ugh, evidence that the full test fails https://github.com/wormhole-foundation/wormhole/actions/runs/3982537866/jobs/6827087640

evan-gray:
Chatted with @hendrikhofstadt, we'd like to see the lint and tests split out and leave the near tests failing sporadically until we can debug and resolve.

evan-gray:
If this gets consensus, I will switch the required checks.

jynnantonix (approved):
Is the intention here to split these into two so that the `node-lint` job continues to be required whereas the `node-tests` job becomes optional?

In that case I'd prefer we move just the near tests into a separate job and skip them in the `lint-and-tests` job.  The testing package has a [`SkipNow`](https://pkg.go.dev/testing#T.SkipNow) method we can use for this:

```go
func TestFoo(t *testing.T) {
	if os.Getenv("SKIP_NEAR_TESTS") != "" {
		t.SkipNow()
	}

	// Run the test.
}
```

Then we set the `SKIP_NEAR_TESTS` env variable for the `lint-and-tests` job and add a new `near-tests` job that just does `go test -v -race node/pkg/watchers/near/...`.  This way we keep running the near tests in CI but they don't block unrelated PRs and we continue requiring that all the *other* node tests pass.


If both jobs are going to continue being required then you can ignore everything above.  I don't have any opinions about the lint/test split.  Preemptively approving in this case.

hendrikhofstadt (approved):

evan-gray:
> Is the intention here to split these into two so that the `node-lint` job continues to be required whereas the `node-tests` job becomes optional?
> 
> In that case I'd prefer we move just the near tests into a separate job and skip them in the `lint-and-tests` job. The testing package has a [`SkipNow`](https://pkg.go.dev/testing#T.SkipNow) method we can use for this:
> 
> ```go
> func TestFoo(t *testing.T) {
> 	if os.Getenv("SKIP_NEAR_TESTS") != "" {
> 		t.SkipNow()
> 	}
> 
> 	// Run the test.
> }
> ```
> 
> Then we set the `SKIP_NEAR_TESTS` env variable for the `lint-and-tests` job and add a new `near-tests` job that just does `go test -v -race node/pkg/watchers/near/...`. This way we keep running the near tests in CI but they don't block unrelated PRs and we continue requiring that all the _other_ node tests pass.
> 
> If both jobs are going to continue being required then you can ignore everything above. I don't have any opinions about the lint/test split. Preemptively approving in this case.

Thanks! I had written up a skip case as well, but we decided against skipping (and hope to just fix them). Will require both tests after this merges 🙂

PR #2285, opened 2023-01-23
sui: worm cli initialize core and token bridge

optke3:
## 1. Sui Worm CLI Updates

In this PR, we add support for initializing the Sui core and token bridges in the Worm CLI.

- [**initialization**] in Worm CLI, support initializing and sharing core bridge and token bridge state 
- [**object auto-discovery**] use json-rpc call `getObjectsOwnedByAddress` to auto-detect `DeployerCapability` in deployer inventory using package id, instead of requiring user to pass in deployer capability object id in command line.
- [**print transaction effects**] necessary not only so we know whether tx succeeds or fails, but because we need to know object ids of created objects, e.g. package IDs, `wormhole_state` ID, `bridge_state` ID. We need to manually copy/record some of the created object IDs, which might be a pain point in the full deployment sequence.

## 2. Sui Contract Updates
- [**auto-create emitter capability**] in Sui tokenbridge `bridge_state.move` `init_and_share_state`, we pass in `worm_state` and create a new `emitter_capability` on-demand using the `worm_state`, instead of requiring the user pass in an emitter cap. 

## 3. Testing
1. publish wormhole core bridge
2. initialize wormhole core bridge
3. publish token bridge (dependent on core bridge being published, require package id to be set in Move.toml)
4. initialize token bridge (requires core bridge`worm_state` as an initialization arg)

### Sample output for initializing the token bridge state
`worm sui init-tokenbridge -n devnet --package-id 0xc90a2535fe8ad48b9f02d336173f8a33b944cec0 --worm_state 0x6cdd43fa444efdc8beb0bdae961e2cfff0b2f3ce
`
```
owner:  1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791
network:  DEVNET
rpc:  http://0.0.0.0:9000
package id:  0xc90a2535fe8ad48b9f02d336173f8a33b944cec0
deployer object id:  0x66537e46ec5fae565035433f522e6429a307a555
wormhole state object id:  0x6cdd43fa444efdc8beb0bdae961e2cfff0b2f3ce
network:  DEVNET
rpc:  http://0.0.0.0:9000
package object id:  0xc90a2535fe8ad48b9f02d336173f8a33b944cec0
module:  bridge_state
moveCallTxn:  {
  EffectsCert: {
    certificate: {
      transactionDigest: 'HZMUvLc5yLwozVa4RgDvzpedzV2C5Bw8wkurcLYZN2vb',
      data: [Object],
      txSignature: 'AJN4b4fuFX+HvXGmMZCQYAFwquDztHpSwIGzoXTn2V2s80EmRr7pUiDOS0i8qc33w2sPZDk1hOAtNzfdZHTZnQZ5FDb/qznQR8MjgJmjXcAOTVBD8CXqcozVI7+uYnQ/xQ==',
      authSignInfo: [Object]
    },
    effects: {
      transactionEffectsDigest: 'IxyqYQDW0wDgGVzrZeJ+7UUb6FPgx3Xv6exhf61otLY=',
      effects: [Object],
      authSignInfo: [Object]
    },
    confirmed_local_execution: true
  }
}
effects:  {
  transactionEffectsDigest: 'IxyqYQDW0wDgGVzrZeJ+7UUb6FPgx3Xv6exhf61otLY=',
  effects: {
    status: { status: 'success' },
    gasUsed: { computationCost: 474, storageCost: 67, storageRebate: 58 },
    sharedObjects: [
      {
        objectId: '0x6cdd43fa444efdc8beb0bdae961e2cfff0b2f3ce',
        version: 8,
        digest: '5YDelEHdpEM4AaqHj1qaGjgdwy/pj35VgSMiAvVX54U='
      }
    ],
    transactionDigest: 'HZMUvLc5yLwozVa4RgDvzpedzV2C5Bw8wkurcLYZN2vb',
    created: [
      {
        owner: { Shared: { initial_shared_version: 10 } },
        reference: {
          objectId: '0x6feafdc4122268d6d0cab71b61d6d1d3e7f7ffc9',
          version: 10,
          digest: 'OU3aq4kbj8h/ERLe7t/2y78JhyuAzdrQl9syeiOPouM='
        }
      }
    ],
    mutated: [
      {
        owner: { AddressOwner: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791' },
        reference: {
          objectId: '0x55f97b87d99b32264338b076c23145a2081dc8f2',
          version: 10,
          digest: '4TXQ0oUe3yfDYx/1A33UPPyKUdKGqwj5gw4a7QSElVc='
        }
      },
      {
        owner: { Shared: { initial_shared_version: 8 } },
        reference: {
          objectId: '0x6cdd43fa444efdc8beb0bdae961e2cfff0b2f3ce',
          version: 10,
          digest: 'oTaxXEfCy2l4/HRi5Sjn+FAy/vy/mrLQM0c0DUqNa0o='
        }
      }
    ],
    deleted: [
      {
        objectId: '0x66537e46ec5fae565035433f522e6429a307a555',
        version: 10,
        digest: 'Y2NjY2NjY2NjY2NjY2NjY2NjY2NjY2NjY2NjY2NjY2M='
      }
    ],
    gasObject: {
      owner: { AddressOwner: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791' },
      reference: {
        objectId: '0x55f97b87d99b32264338b076c23145a2081dc8f2',
        version: 10,
        digest: '4TXQ0oUe3yfDYx/1A33UPPyKUdKGqwj5gw4a7QSElVc='
      }
    },
    events: [
      {
        coinBalanceChange: {
          packageId: '0x0000000000000000000000000000000000000002',
          transactionModule: 'gas',
          sender: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791',
          changeType: 'Gas',
          owner: {
            AddressOwner: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791'
          },
          coinType: '0x2::sui::SUI',
          coinObjectId: '0x55f97b87d99b32264338b076c23145a2081dc8f2',
          version: 9,
          amount: -483
        }
      },
      {
        coinBalanceChange: {
          packageId: '0x0000000000000000000000000000000000000002',
          transactionModule: 'gas',
          sender: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791',
          changeType: 'Pay',
          owner: {
            AddressOwner: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791'
          },
          coinType: '0x2::sui::SUI',
          coinObjectId: '0x55f97b87d99b32264338b076c23145a2081dc8f2',
          version: 9,
          amount: -1
        }
      },
      {
        mutateObject: {
          packageId: '0xc90a2535fe8ad48b9f02d336173f8a33b944cec0',
          transactionModule: 'bridge_state',
          sender: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791',
          objectType: '0x7177d3db1f086074bb1d11c1a901be764d41890b::state::State',
          objectId: '0x6cdd43fa444efdc8beb0bdae961e2cfff0b2f3ce',
          version: 10
        }
      },
      {
        newObject: {
          packageId: '0xc90a2535fe8ad48b9f02d336173f8a33b944cec0',
          transactionModule: 'bridge_state',
          sender: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791',
          recipient: { Shared: { initial_shared_version: 10 } },
          objectType: '0xc90a2535fe8ad48b9f02d336173f8a33b944cec0::bridge_state::BridgeState',
          objectId: '0x6feafdc4122268d6d0cab71b61d6d1d3e7f7ffc9',
          version: 10
        }
      },
      {
        deleteObject: {
          packageId: '0xc90a2535fe8ad48b9f02d336173f8a33b944cec0',
          transactionModule: 'bridge_state',
          sender: '0x1a971b003ebf81bfcdf43c67e3bb4dceb4cbd791',
          objectId: '0x66537e46ec5fae565035433f522e6429a307a555',
          version: 10
        }
      }
    ],
    dependencies: [
      '6GnxypeR3WtN3zrGhqcVW2KbrXuFmpFhBxbT9gYG69Hu',
      '8BKCttaVjEUJgausZXvNqk8CXGAsYvxKKX1h5c9Q2utn'
    ]
  },
  authSignInfo: {
    epoch: 3,
    signature: 'Aah/ncU+WEhH5g8yKOMVBGMJRRFfJYQ03Lq3rYNH6GMGgIf/gQOcE30iNJB6IIpUTw==',
    signers_map: [
      58, 48, 0, 0, 1,  0, 0,
       0,  0, 0, 2, 0, 16, 0,
       0,  0, 0, 0, 1,  0, 2,
       0
    ]
  }
}

```


clients/js/cmds/sui.ts
```diff
@@ -74,64 +80,115 @@ exports.builder = function(y: typeof yargs) {
...
           describe: "Governance contract",
           type: "string",
-          default: "0x0000000000000000000000000000000000000000000000000000000000000004",
+          default: "0000000000000000000000000000000000000000000000000000000000000004",
           required: false
         })
         .option("initial-guardian", {
           alias: "ig",
           required: true,
+          default: "beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe",
```
kcsongor, line 104: I think we should not have a default value for this

clients/js/cmds/sui.ts
```diff
@@ -65,6 +65,12 @@ exports.builder = function(y: typeof yargs) {
       return yargs
         .option("network", network_options)
         .option("rpc", rpc_description)
+        .option("package-id", {
+          alias: "pid",
```
kcsongor, line 69: maybe drop this alias, `pid` is confusing as it conventionally means process id

clients/js/cmds/sui.ts
```diff
@@ -74,64 +80,115 @@ exports.builder = function(y: typeof yargs) {
...
+      return yargs
+        .option("network", network_options)
+        .option("rpc", rpc_description)
+        .option("package-id", {
+          alias: "pid",
+          describe: "Package/module ID",
+          required: true,
+          type: "string"
+        })
+        .option("deployer-capability", {
```
kcsongor, line 153: is there a way to discover this automatically? the sdk has a `getOwnedObjectsByOwner` function. Using that we should be able to find the deployer capability.

clients/js/cmds/sui.ts
```diff
@@ -74,64 +80,115 @@ exports.builder = function(y: typeof yargs) {
...
+          alias: "d",
+          describe: "Deployer capability object ID",
+          required: true,
+          type: "string",
+        })
+        .option("emitter-capability", {
```
kcsongor, line 159: I just realised that the emitter capability is an argument to the init state function. any reason we can't create this object in the `init_and_share_state` function directly?

I propose:
1. remove the `emitter_cap` argument from `init_and_share_state` and create the object in there directly
2. remove the `get_new_emitter` entry function

kcsongor (changes requested):

clients/js/cmds/sui.ts
kcsongor: also, won't discovering the deployer capability (from my other comment) tell us what the package id is?

clients/js/cmds/sui.ts
optke3: Actually, this doesn't save us an argument to the init state function, because we would have to pass in the `worm_state` to create a new emitter. Do you think one way is preferable to the other? I feel like they are about the same.

clients/js/cmds/sui.ts
optke3: I grepped this repo and did not find this function: https://github.com/MystenLabs/sui/tree/main/sdk/typescript/src 

Seems like you have to make a json-rpc call to get the owned objects by owner? Looks like it's this? https://docs.sui.io/sui-jsonrpc#sui_getObjectsOwnedByAddress

Maybe the sdk provides a simple way of making json-rpc calls.

clients/js/cmds/sui.ts
kcsongor: ah good point, we'd need `worm_state`.

I feel like that's still a little bit better, because we don't need a separate step to create the emitter capability 🤔

clients/js/cmds/sui.ts
kcsongor:
```
import { JsonRpcProvider } from '@mysten/sui.js';
const provider = new JsonRpcProvider('https://gateway.devnet.sui.io:443');
const objects = await provider.getOwnedObjectRefs(
  '0xbff6ccc8707aa517b4f1b95750a2a8c666012df3'
);
```
from http://typescript-sdk-docs.s3-website-us-east-1.amazonaws.com/
(^ that example calls a different function, but shows how to call the rpc)

kcsongor: yes it's `getObjectsOwnedByAddress` my bad!

clients/js/cmds/sui.ts
kcsongor: definitely have to find the one that corresponds to the `package-id` argument, but that should be simple

clients/js/cmds/sui.ts
optke3: this works, very convenient! working on using this for stuff like detecting deployer caps and package id

heyitaki:
@optke3 can you resolve these tests? I think rebasing main should fix Aptos, Sui has a version mismatch

clients/js/cmds/sui.ts
```diff
@@ -16,6 +17,31 @@ function assertNetwork(n: string): asserts n is Network {
...
+  have been granted upon publication of the package with package_id
+
+  The objects is in the format returned by a json-rpc call "provider.getObjectsOwnedByAddress"
+*/
+function findDeployerCapability(package_id: string, module_name: string, objects: Array<any>): string {
+      let deployer;
+      for (var obj of objects){
+        let t = obj.type;
+        // do a basic check and see if the type is as expected (has a normal package, module, type name separated by "::")
+        var count = [...t].filter(x => x === ":").length;
```
heyitaki, line 31: we have a method `isValidAptosType` in the SDK that I think would a better way to check if this is a valid fully qualified type

clients/js/cmds/sui.ts
```diff
@@ -16,6 +17,31 @@ function assertNetwork(n: string): asserts n is Network {
...
+        if (count==4){
+          let t_split = t.split("::") // load the package, module, type name in an array
+          if (t_split[0].toUpperCase() == package_id.toUpperCase()){
```
heyitaki, line 34: can we construct the fully qualified type given by `packageId::moduleName::DeployerCapability` and just string match? logic would be much cleaner than having to split the string and conditionally matching on tokens

heyitaki: maybe something like
```
function findDeployerCapability(packageId: string, moduleName: string, objects: any[]): string | null {
  const type = `${packageId}::${moduleName}::DeployerCapability`;
  return objects.find(o => o.type.toLowerCase() === type.toLowerCase())?.objectId ?? null;
}
```

clients/js/cmds/sui.ts
```diff
@@ -37,6 +63,28 @@ exports.command = 'sui';
...
+      const network = argv.network.toUpperCase()
+      const rpc = argv.rpc ?? NETWORKS[network]["sui"].rpc;
+      const provider = new JsonRpcProvider(rpc);
+      const owner = argv.owner;
+      const objects = await provider.getObjectsOwnedByAddress(
+        owner
+      );
+      console.log("network: ", network)
+      console.log("owner: ", owner)
+      console.log(objects)
```
heyitaki, line 86: Should this be JSON stringified so we can read nested objects?

clients/js/cmds/sui.ts
```diff
@@ -37,6 +63,28 @@ exports.command = 'sui';
...
+        .option("network", network_options)
+        .option("rpc", rpc_description)
+        .option("owner", {
+          alias: "o",
+          describe: "Owner address",
+          required: true,
+          type: "string",
+        })
+    }, async (argv)=>{
+      const network = argv.network.toUpperCase()
```
heyitaki, line 77: there is a nice util that @kcsongor wrote for the Aptos CLI that asserts the network, can you move that to a shared util file and use it here too? https://github.com/wormhole-foundation/wormhole/blob/4c7df41984b6f4947773846050bdd68efc3a8cff/clients/js/cmds/aptos.ts#L47

clients/js/cmds/sui.ts
```diff
@@ -59,79 +107,139 @@ exports.builder = function(y: typeof yargs) {
...
-          governance_contract,
-          [initial_guardian]
+          [...Buffer.from(governance_contract, "hex")],
+          [[...Buffer.from(initial_guardian, "hex")]]
         ],
     )
-      //await publishPackage(network, rpc, `${dir}/tokenbridge`);
     })
-    .command("init-coin", "Publish coin contract", (yargs) => {
+    .command("init-tokenbridge", "Init token bridge contract", (yargs) => {
```
heyitaki, line 193: let's change this to `init-token-bridge` to match aptos
https://github.com/wormhole-foundation/wormhole/blob/4c7df41984b6f4947773846050bdd68efc3a8cff/clients/js/cmds/aptos.ts#L64

clients/js/cmds/sui.ts
```diff
@@ -59,79 +107,139 @@ exports.builder = function(y: typeof yargs) {
...
           required: false,
-          type: "number"
+          type: "string"
         })
         .option("governance-chain-id", {
           alias: "gci",
           describe: "Governance chain ID",
-          default: 1,
-          type: "number",
+          default: "3",
```
heyitaki, line 131: can you explain why this is 3? i would expect it to be 1

clients/js/sui.ts
```diff
@@ -58,13 +62,18 @@ export async function callEntryFunc(
...
         function: func,
         typeArguments: type_args,
         arguments: args,
-        gasBudget: 20000,
-      });
-      console.log('moveCallTxn', moveCallTxn);
+        gasBudget: 50000,
+    });
+    console.log('moveCallTxn: ', moveCallTxn);
+    console.log("effects: ", util.inspect(moveCallTxn["EffectsCert"]["effects"], {showHidden: false, depth: null, colors: true}))
```
heyitaki, line 78: passing `depth: null` means this will recurse indefinitely if there is ever a circular reference in the object

i'd prefer to get rid of the util import and JSON stringify the obj isntead
```suggestion
    console.log("effects: ", JSON.stringify(moveCallTxn["EffectsCert"]["effects"], null, 2))
```

heyitaki (changes requested):

clients/js/cmds/sui.ts
optke3: yes, perhaps

clients/js/cmds/sui.ts
optke3: stringifying it displays all nested layers? I didn't know that :O

clients/js/cmds/sui.ts
optke3: very clean

bruce-riley (approved):

PR #2286, opened 2023-01-24
cosmwasm: accountant: Allow transfers on the same chain

jynnantonix:
While sending tokens to another address on the same chain via wormhole
is quite inefficient, it's not strictly disallowed and we do have some
VAAs on solana that do this.  Explicitly check for this case and allow
it.

There are no restrictions on native tokens sent in this way but wrapped
transfers are still subject to some checks: the wrapped account for
the token must exist and it must have a balance larger than the amount
being transferred.  If either of those checks fails then that means
that the sender acquired some wrapped tokens that did not go through
the accountant and so that transfer should be blocked and require manual
intervention.

### DIFF
--- a/cosmwasm/packages/accountant/src/contract.rs
+++ b/cosmwasm/packages/accountant/src/contract.rs
@@ -155,7 +155,10 @@ fn transfer<C: CustomQuery>(deps: Deps<C>, t: &Transfer) -> anyhow::Result<(Acco
         Account { key, balance }
     };
 
-    let mut dst = {
+    let mut dst = if t.key.emitter_chain() == t.data.recipient_chain {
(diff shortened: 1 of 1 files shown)


jynnantonix:
* main
  * **PR #2286** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2286" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2286?utm_source=stack-comment).

a5-pickle (dismissed):

hendrikhofstadt (approved):

a5-pickle (approved):

PR #2302, opened 2023-01-26
wormchain - config comments and defaults

justinschuldt:
Adds the cosmos default comments and options from cosmos `init` to `wormchain/build` config files.
It's helpful to have the comments describing what the files and options do. Also some config options were missing, so those got added back.


hendrikhofstadt (dismissed):

conorpp (approved):

evan-gray (approved):
Re-approving since @hendrikhofstadt 's approval was lost

PR #2305, opened 2023-01-27
wormchain: require addresses to be allowlisted to submit tx

conorpp:


conorpp:
I forgot to add CLI commands.  Will do that shortly.

conorpp:
This is ready for review.

wormchain/x/wormhole/ante/ante.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	verified_addresses := make(map[string]bool)
+
+	// For every message we should check:
+	// 1. There is an allowlist entry for the signer(s), OR
+	// 2. The signer(s) are validators in current or future guardian set.
+	// We cache the result for performance, since this is a ante handler that runs for everything.
+	for _, msg := range tx.GetMsgs() {
+		for _, signer := range msg.GetSigners() {
```
hendrikhofstadt, line 36: Shouldn't we be ok if one signer is allowlisted?

wormchain/x/wormhole/keeper/msg_server_allowlist_test.go
```diff
@@ -0,0 +1,249 @@
...
+	// still works with new guardian set
+	_, err = msgServer.CreateAllowlist(context, &types.MsgCreateAllowlistRequest{
+		Signer:  getSigner(&guardians[0]),
+		Address: getRandomAddress(),
+	})
+	assert.NoError(t, err)
+
+	// Anyone can remove stale allowlists
+	// (new_address list is now stale as it's validator is no longer in validator set)
+	_, err = msgServer.DeleteAllowlist(context, &types.MsgDeleteAllowlistRequest{
```
hendrikhofstadt, line 151: One more test case here that would make sense is: New validator can "overwrite" existing stale entry according to handler#L21

wormchain/x/wormhole/keeper/allowlist.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Checks if a given address is registered as a guardian validator and either:
+// * Is in the current guardian set, OR
+// * Is in a future guardian set
+func (k Keeper) IsAddressValidatorOrFutureValidator(ctx sdk.Context, addr string) bool {
+	currentIndex, _ := k.GetConsensusGuardianSetIndex(ctx)
+	validators := k.GetAllGuardianValidator(ctx)
+	addrBz, err := sdk.AccAddressFromBech32(addr)
+	var matchedValidator *types.GuardianValidator = nil
```
hendrikhofstadt, line 59: nit: = nil not required

wormchain/x/wormhole/keeper/allowlist.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return false
+	}
+	// look up the guardian validator
+	for _, val := range validators {
+		if bytes.Equal(val.ValidatorAddr, addrBz) {
+			matchedValidator = &val
```
hendrikhofstadt, line 66: note: using pointers to loop vars is dangerous when you don't break. Fine here

wormchain/x/wormhole/keeper/grpc_query_allowlist.go
```diff
@@ -0,0 +1,72 @@
...
+func (k Keeper) AllowlistAll(c context.Context, req *types.QueryAllValidatorAllowlist) (*types.QueryAllValidatorAllowlistResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	var allowedAddresses []*types.ValidatorAllowedAddress
+	ctx := sdk.UnwrapSDKContext(c)
+
+	store := ctx.KVStore(k.storeKey)
+	sequenceCounterStore := prefix.NewStore(store, types.KeyPrefix(types.ValidatorAllowlistKey))
```
hendrikhofstadt, line 23: sequenceCounterStore?

wormchain/x/wormhole/keeper/grpc_query_allowlist.go
```diff
@@ -0,0 +1,72 @@
...
+
+	pageRes, err := query.Paginate(sequenceCounterStore, req.Pagination, func(key []byte, value []byte) error {
+		var allowedAddress types.ValidatorAllowedAddress
+		if err := k.cdc.Unmarshal(value, &allowedAddress); err != nil {
+			return err
+		}
+		// this will cause a less then expected amount to be returned in the pagination,
+		// but the alternative is to rework how pagination works, which is complex.
+		// this lists are not expected to be long anyways and won't need pagination.
+		if allowedAddress.ValidatorAddress == req.ValidatorAddress {
```
hendrikhofstadt, line 61: I think we'll have to rework this in some way. Right now the moment the total count of allowlist entries exceeds the page size it could return empty pages despite there being results on the next page.

In case you think pagination is not required let's just drop it entirely

wormchain/x/wormhole/client/cli/tx_allowlist.go
```diff
@@ -0,0 +1,74 @@
...
+
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+// StoreCodeCmd will upload code to be reused.
+func CmdDeleteAllowedAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "delete-allowed-address [wormchain-address]",
+		Short:   "Remove an allowlist. The allowlist must be stale (only valid under old guardian set) or you must be the creator of the allowlist.",
```
hendrikhofstadt, line 51: remove an allowlist entry

hendrikhofstadt (changes requested):
See inline comments

hendrikhofstadt:
Rly nice one!

wormchain/proto/wormhole/tx.proto
```diff
@@ -8,9 +8,12 @@ option go_package = "github.com/wormhole-foundation/wormchain/x/wormhole/types";
 
 // Msg defines the Msg service.
 service Msg {
-      rpc ExecuteGovernanceVAA(MsgExecuteGovernanceVAA) returns (MsgExecuteGovernanceVAAResponse);
+  rpc ExecuteGovernanceVAA(MsgExecuteGovernanceVAA) returns (MsgExecuteGovernanceVAAResponse);
   rpc RegisterAccountAsGuardian(MsgRegisterAccountAsGuardian) returns (MsgRegisterAccountAsGuardianResponse);
 
+  rpc CreateAllowlist(MsgCreateAllowlistRequest) returns (MsgAllowlistResponse);
+  rpc DeleteAllowlist(MsgDeleteAllowlistRequest) returns (MsgAllowlistResponse);
```
nik-suri, line 15: nits: `CreateAllowlistEntry` and `DeleteAllowlistEntry`, since these don't create/delete entire validator allowlists

wormchain/x/wormhole/keeper/grpc_query_allowlist.go
nik-suri: can we use `query.FilteredPaginate` here?

wormchain/x/wormhole/keeper/allowlist.go
```diff
@@ -0,0 +1,86 @@
...
+		list = append(list, val)
+	}
+
+	return
+}
+
+// Checks if a given address is registered as a guardian validator and either:
+// * Is in the current guardian set, OR
+// * Is in a future guardian set
+func (k Keeper) IsAddressValidatorOrFutureValidator(ctx sdk.Context, addr string) bool {
```
nik-suri, line 55: general question on this: what if a guardian is registered as a validator, and then the guardian is removed from the guardian set?
This function guards against that case. Should there additionally be some mechanism in wormchain that removes a validator if they are no longer part of the guardian set? (or does this mechanism already exist)?

nik-suri (commented):
changes look good to me. Minor nit suggestion is changing `CreateAllowlist` and `DeleteAllowlist` function names everywhere to be `CreateAllowlistEntry` and `DeleteAllowlistEntry`, as these seem to be more descriptive of what is happening.

wormchain/testutil/keeper/wormhole.go
```diff
@@ -115,7 +115,8 @@ func WormholeKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 		supportedFeatures,
 	)
 	ctx := sdk.NewContext(stateStore, tmproto.Header{
-		Time: time.Now(),
+		Time:   time.Now(),
+		Height: 1,
```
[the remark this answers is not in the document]
conorpp: adding comment:
```
		// The height should be at least 1, because the allowlist antehandler
		// passes everything at height 0 for gen tx's.
```

wormchain/x/wormhole/keeper/grpc_query_allowlist.go
conorpp: TIL FilteredPaginate, but how to use it properly is not clear to me.  Opting to remove entirely.

wormchain/x/wormhole/keeper/allowlist.go
conorpp: It does not currently exist, right now the root of trust is being in the guardian set or not.

wormchain/x/wormhole/ante/ante.go
conorpp: actually you're correct, i was thinking messages were individually signed for some reason

wormchain/app/app.go
```diff
@@ -582,6 +584,20 @@ func New(
...
 
+// Wrap the standard cosmos-sdk antehandlers with our wormhole allowlist antehandler.
+func WrapAnteHandler(originalHandler sdk.AnteHandler, wormKeeper wormholemodulekeeper.Keeper) sdk.AnteHandler {
+	whHandler := wormholemoduleante.NewWormholeAllowlistDecorator(wormKeeper)
+	return func(ctx sdk.Context, tx sdk.Tx, simulate bool) (sdk.Context, error) {
+		newCtx, err := originalHandler(ctx, tx, simulate)
+		if err != nil {
+			return newCtx, err
+		}
+		return whHandler.AnteHandle(ctx, tx, simulate, func(ctx sdk.Context, tx sdk.Tx, simulate bool) (sdk.Context, error) {
```
hendrikhofstadt, line 595: do we need to pass the newCtx here?

hendrikhofstadt (approved):

wormchain/x/wormhole/keeper/allowlist.go
nik-suri: got it. So we're covered on all other paths here? Only other one I can think of is registering guardian as validator, but we're covered there

wormchain/app/app.go
conorpp: Since we're not relying on kvs changes from previous ante handlers I don't think this is an issue, but we should still use it.  I will update in subsequent PR.

PR #2311, opened 2023-01-27
ethereum/scripts: reduce register script

### LINKED ISSUE #2224: Tilt Eth registrations should dynamically read env vars
agreed the old scripts should be deleted. 
additionally, we should consider dynamically picking up the VAAs from the environment 
so adding new chains does not require updating all the deployment scripts. 
This is done for aptos as an example: https://github.com/wormhole-foundation/wormhole/blob/main/aptos/scripts/register_devnet#L13

and I think the other chains too

_Originally posted by @kcsongor in https://github.com/wormhole-foundation/wormhole/issues/2214#issuecomment-1377248236_

panoel:
Fixes #2224

### DIFF
--- a/ethereum/scripts/register_all_chains.js
+++ b/ethereum/scripts/register_all_chains.js
@@ -6,16 +6,19 @@ const NFTBridge = artifacts.require("NFTBridgeEntrypoint");
 const BridgeImplementationFullABI = jsonfile.readFileSync(
   "../build/contracts/BridgeImplementation.json"
 ).abi;
-const solTokenBridgeVAA = process.env.REGISTER_SOL_TOKEN_BRIDGE_VAA;
-const solNFTBridgeVAA = process.env.REGISTER_SOL_NFT_BRIDGE_VAA;
-const terraTokenBridgeVAA = process.env.REGISTER_TERRA_TOKEN_BRIDGE_VAA;
-const terraNFTBridgeVAA = process.env.REGISTER_TERRA_NFT_BRIDGE_VAA;
-const terra2TokenBridgeVAA = process.env.REGISTER_TERRA2_TOKEN_BRIDGE_VAA;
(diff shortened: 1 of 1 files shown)


evan-gray (approved):
nice! 🙌

ethereum/scripts/register_all_chains.js
```diff
@@ -6,16 +6,19 @@ const NFTBridge = artifacts.require("NFTBridgeEntrypoint");
...
-const wormchainTokenBridgeVAA = process.env.REGISTER_WORMCHAIN_TOKEN_BRIDGE_VAA;
-const aptosTokenBridgeVAA = process.env.REGISTER_APTOS_TOKEN_BRIDGE_VAA;
+
+// The input parameter is a RegExp
+// It returns an array of process.env variables satisfying the input RegExp
+function getFilteredEnvs(regexp) {
+  const filteredEnvs = [];
+  for (const [key, value] of Object.entries(process.env)) {
+    if (regexp.test(key) && value) {
+      console.log("getFilteredEnvs: pushing " + key);
```
kcsongor, line 16: I feel like we don't need this log here

kcsongor (approved):
left a minor comment, not a blocker 👍

ethereum/scripts/register_all_chains.js
evan-gray: Could go either way, was kinda nice to see which bridges were getting registered (though maybe change the log output). I’m 🤷‍♂️. Need this for tests though so I’ll merge and we can revisit in the next “make CI have less output” event. 🙃

evan-gray (approved):

ethereum/scripts/register_all_chains.js
kcsongor: agreed it's nice to see which chain is getting registered, but the log shouldn't be emitted by the helper function that grabs the env var

PR #2312, opened 2023-01-27
Node: Initialize scissors metrics

bruce-riley:
This PR initializes the "run with scissors" metrics so that they will show up even if there have not been any errors.

Change-Id: I018a663aa5b9ea902f934ab71c23b1786c8aeeb0

### DIFF
--- a/node/pkg/common/scissors.go
+++ b/node/pkg/common/scissors.go
@@ -20,6 +20,7 @@ var (
 // Start a go routine with recovering from any panic by sending an error to a error channel
 func RunWithScissors(ctx context.Context, errC chan error, name string, runnable supervisor.Runnable) {
 	go func() {
+		ScissorsErrors.WithLabelValues("scissors", name).Add(0)
 		defer func() {
 			if r := recover(); r != nil {
 				switch x := r.(type) {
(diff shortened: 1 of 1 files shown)


node/pkg/common/scissors.go
```diff
@@ -20,6 +20,7 @@ var (
 // Start a go routine with recovering from any panic by sending an error to a error channel
 func RunWithScissors(ctx context.Context, errC chan error, name string, runnable supervisor.Runnable) {
 	go func() {
+		ScissorsErrors.WithLabelValues("scissors", name).Add(0)
```
evan-gray, line 23: @hendrikhofstadt do you know if it'd be better to do this one line up?

node/pkg/common/scissors.go
```diff
@@ -41,6 +42,7 @@ func RunWithScissors(ctx context.Context, errC chan error, name string, runnable
 
 func WrapWithScissors(runnable supervisor.Runnable, name string) supervisor.Runnable {
 	return func(ctx context.Context) (result error) {
+		ScissorsErrors.WithLabelValues("scissors", name).Add(0)
```
evan-gray, line 45: ditto

evan-gray (approved):

node/pkg/common/scissors.go
SEJeff: Yeah I'd do them in the function (aka one line up), but outside of the goroutine. There's no inherent value and it just adds more code to the goroutine.

SEJeff (approved):

node/pkg/common/scissors.go
hendrikhofstadt: Agree

hendrikhofstadt (approved):

evan-gray (approved):

hendrikhofstadt (approved):

PR #2334, opened 2023-02-03
node: Fix defer

panoel:
It looks like the defer wasn't working as intended.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -246,26 +246,32 @@ func (s *SolanaWatcher) SetupWebSocket(ctx context.Context) error {
 			case <-ctx.Done():
 				return nil
 			default:
-				rCtx, cancel := context.WithTimeout(ctx, time.Second*300) // 5 minute
-				defer cancel()
+				if err := func() error {
 
-				if _, msg, err := ws.Read(rCtx); err != nil {
-					if errors.Is(err, context.DeadlineExceeded) {
-						// When a websocket context times out, it closes the websocket...  This means we have to re-subscribe
-						ws.Close(websocket.StatusNormalClosure, "")
-						err, ws = s.SetupSubscription(ctx)
-						if err != nil {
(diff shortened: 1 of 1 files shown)


node/pkg/watchers/solana/client.go
```diff
@@ -347,7 +355,7 @@ func (s *SolanaWatcher) Run(ctx context.Context) error {
 					ContractAddress: contractAddr,
 				})
 
-				if s.wsUrl == nil {
+				if useWs {
```
bruce-riley, line 358: Should be !useWs. I think that's why tests are failing.

node/pkg/watchers/solana/client.go
```diff
@@ -246,26 +246,32 @@ func (s *SolanaWatcher) SetupWebSocket(ctx context.Context) error {
...
-				defer cancel()
-
-				if _, msg, err := ws.Read(rCtx); err != nil {
-					if errors.Is(err, context.DeadlineExceeded) {
-						// When a websocket context times out, it closes the websocket...  This means we have to re-subscribe
-						ws.Close(websocket.StatusNormalClosure, "")
-						err, ws = s.SetupSubscription(ctx)
-						if err != nil {
-							return err
+				if err := func() error {
```
jynnantonix, line 249: nit: I think this code would be more readable if we moved this into a named function and called it here instead of inlining it like this.

jynnantonix (approved):
Good find!  Looks good to me once Bruce's comment is addressed.

bruce-riley (approved):

node/pkg/watchers/solana/client.go
panoel: Fixed

hendrikhofstadt (approved):
Agree with Chirantan

jumpsiegel (approved):

node/pkg/watchers/solana/client.go
panoel: Will do in a future PR

PR #2339, opened 2023-02-03
sui: run faucet in Tilt

heyitaki:
Quick PR to run Sui faucet in Tilt


evan-gray (commented):
What did the removed (now commented out) parts of this script do? Can you comment on why they are no longer needed?

PR #2329, opened 2023-02-02
wormchain: add migrating contracts with vaa

conorpp:
wasmd updated to latest v0.30.0 and disabled migration handlers here:
https://github.com/wormhole-foundation/wasmd/pull/9

PR #2340, opened 2023-02-03
Remove accountant upgrade and modification authentication

conorpp:
As of https://github.com/wormhole-foundation/wormhole/pull/2329, migrate in wasmd will be authenticated by VAA's, which means wormchain contracts do not need to worry about authenticating their upgrades.  

Also, accountant `Modification` messages can be included in the the `msg` arg of migrate and reuse that authenticated flow without having to add more types or commands to guardian node software.


evan-gray (commented):
Should we add this functionality to `instantiate` as well? Its message is `Empty` 🤷

cosmwasm/contracts/global-accountant/src/msg.rs
```diff
@@ -139,7 +116,17 @@ pub enum ExecuteMsg {
...
+pub struct MigrateMsg {
+    // Optional list modifications to apply
+    pub modifications: Vec<ModifyBalance>,
+}
+
+/// Modifies the balance of a single account.  Used to manually override the balance.
+#[cw_serde]
+pub struct ModifyBalance {
+    // A serialized `Modification` message.
+    pub modification: Binary,
```
jynnantonix, line 128: This was a serialized message because we needed to verify signatures but since that's taken care of we should be able to use `accountant::Modification` directly in the message body.

cosmwasm/contracts/global-accountant/tests/helpers/mod.rs
```diff
@@ -67,39 +69,33 @@ impl Contract {
...
             self.addr(),
-            &ExecuteMsg::ModifyBalance {
-                modification,
-                guardian_set_index,
-                signatures,
-            },
-            &[],
+            &migrate_msg,
+            self.code_id,
         )
```
jynnantonix, line 86: It feels a bit weird to me that the only way to modify an account's balance is to migrate the entire contract.  The two things are kind of orthogonal.

cosmwasm/contracts/global-accountant/tests/helpers/mod.rs
```diff
@@ -253,41 +249,32 @@ pub fn proper_instantiate() -> (fake::WormholeKeeper, Contract) {
...
+        ContractWrapper::new(
+            global_accountant::contract::execute,
+            global_accountant::contract::instantiate,
+            global_accountant::contract::query,
+        )
+        .with_migrate(global_accountant::contract::migrate),
+    ));
+
+    // We used to rely on the contract being able to upgrade itself,
+    // but then we decided it was easiest to rely on wrapped VAA autentication
```
jynnantonix, line 262: nit: "autentication" -> "authentication"

cosmwasm/contracts/global-accountant/Cargo.toml
```diff
@@ -37,3 +37,4 @@ anyhow = { version = "1", features = ["backtrace"] }
 cw-multi-test = "0.13.2"
 serde-json-wasm = "0.4"
 wormhole-bindings = { version = "0.1", features = ["fake"] }
+serde_json = "1.0"
```
jynnantonix, line 40: We already have serde-json-wasm in the dev-dependency list.  Can we reuse that?

conorpp: no longer need it after removing need to serialize `Modification `

cosmwasm/contracts/global-accountant/tests/helpers/mod.rs
conorpp: I'm open for further discussion on this.  Adding new messages for guardians to sign is sensitive & can add operationally overhead.  The migrate + instantiate are kind of secure channels we can reuse conveniently as a pattern, but I agree it seems weird.  At least in the migration governance message it will be clear that the code_id is staying the same, and it's the balances getting "migrated".

cosmwasm/contracts/global-accountant/tests/modify_balance.rs
```diff
@@ -18,48 +16,62 @@ fn simple_modify() {
...
-        .add_attribute("kind", serde_json_wasm::to_string(&m.kind).unwrap())
-        .add_attribute("amount", serde_json_wasm::to_string(&m.amount).unwrap())
-        .add_attribute("reason", serde_json_wasm::to_string(&m.reason).unwrap());
-
-    resp.assert_event(&evt);
-
-    let actual = contract.query_modification(m.sequence).unwrap();
-    assert_eq!(m, actual);
+    // Test modification via both instantiate + migrate channels.
+    let instantiate_msgs = vec![
```
jynnantonix, line 21: minor nit: this could also be an array instead of a vec (just remove the `vec!` prefix).

cosmwasm/contracts/global-accountant/tests/modify_balance.rs
```diff
@@ -18,48 +16,62 @@ fn simple_modify() {
...
+        None,
+        Some(InstantiateMsg {
+            modifications: vec![m.clone()],
+        }),
+    ];
+
+    for instantiate_msg in instantiate_msgs {
+        let (_wh, contract) = if instantiate_msg.is_some() {
```
jynnantonix, line 29: nit: use `if let`

```suggestion
        let (_wh, contract) = if let Some(instantiate) = instantiate_msg.as_ref() {
```

jynnantonix (approved):

hendrikhofstadt:
I think I agree with Chirantan here in that I'm not sure it's the best design to force modifications into the instantiate and migrate flows. E.g. in the future when there's real migrations things might break, become harder to follow and track i.e. find and parse all txs that modify balances. Surely there's the event but for tracking code change etc.  I think this will be pretty confusing.

Could you elaborate on the rationale behind this? I generally agree we shouldn't duplicate the authentication code too much and reduce attack surface but I think here we're mixing up concepts.

bruce-riley (approved):
Go code looks good. I did not review the rust.

conorpp:
Redid this PR in #2371

PR #2341, opened 2023-02-04
node/p2p: enforce freshness of heartbeats and correctness of guardianAddr

tbjump:
guardianAddr inside a heartbeat is currently a untrusted field. I don't see a reason to not make it trusted, and added an additional check. 

Similarly, the guardian node currently accepts heartbeats that are very old or far from the future. I added a check to make sure the difference to the local clock is no more than `heartbeatMaxTimeDifference = 15 minutes`, which should account for clock skew and network delay. 

Feedback appreciated.


node/pkg/p2p/p2p.go
```diff
@@ -60,6 +60,9 @@ var heartbeatMessagePrefix = []byte("heartbeat|")
 
 var signedObservationRequestPrefix = []byte("signed_observation_request|")
 
+// Heartbeats that are this old or this much into the future will be dropped. This value should encompass clock skew and network delay.
```
hendrikhofstadt, line 63: godoc strings should start with the variable name such as "heartbeatMaxTimeDifference indicates the time [...]"

hendrikhofstadt (approved):

node/pkg/p2p/p2p.go
```diff
@@ -254,29 +257,15 @@ func Run(
 					if gov != nil {
 						gov.CollectMetrics(heartbeat, gossipSendC, gk, ourAddr)
 					}
-
-					b, err := proto.Marshal(heartbeat)
-					if err != nil {
-						panic(err)
-					}
-
```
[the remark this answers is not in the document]
tbjump: great catch

hendrikhofstadt (approved):
Generally I'd wrap this in a func block and use defer but given this only panics this seems fine here

node/pkg/p2p/p2p_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/assert"
+
+	node_common "github.com/certusone/wormhole/node/pkg/common"
+	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+)
+
+/*
```
bruce-riley, line 18: I don't think you need these comments?

node/pkg/p2p/p2p_test.go
```diff
@@ -0,0 +1,105 @@
...
+			expectSuccess:         true,
+		},
+		// guardian signed a heartbeat for another guardian
+		{
+			timestamp:             time.Now().UnixNano(),
+			gk:                    gk,
+			heartbeatGuardianAddr: gAddr2.String(),
+			expectSuccess:         false,
+		},
+		// old heartbeat
```
bruce-riley, line 59: Add a "too far in the future" test?

node/pkg/p2p/p2p_test.go
tbjump: wanted to remove those, indeed :)

node/pkg/p2p/p2p_test.go
tbjump: added

tbjump:
Addressed reviewer feedback

bruce-riley (approved):

hendrikhofstadt (approved):

PR #2342, opened 2023-02-06
Add top level .nvmrc file

nik-suri:
Node version taken from `node:16-alpine` docker image


evan-gray (approved):
We should definitely consider updating everything to latest LTS (currently 18.14), but matching the dockers is probably a good idea.

hendrikhofstadt (approved):

PR #2349, opened 2023-02-07
Deploy: Make contract upd script support pyth

bruce-riley:
Just a few tweaks to the script that generates contract upgrade governance to make it support pyth. I tried adding the custom RPC as part of the URL but could get it to work, so it is just a note for the user to set that by hand.

Change-Id: I9c40498ef9849979f89e8d48d60333932f8a696f

### DIFF
--- a/scripts/contract-upgrade-governance.sh
+++ b/scripts/contract-upgrade-governance.sh
@@ -104,6 +104,12 @@ case "$chain_name" in
   solana)
     chain=1
     explorer="https://explorer.solana.com/address/"
+    extra=""
+    ;;
+  pythnet)
+    chain=26
+    explorer="https://explorer.solana.com/address/"
+    extra="Be sure to set the custom RPC to https://pythnet.rpcpool.com"
     ;;
   ethereum)
     chain=2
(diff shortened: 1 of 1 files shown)


scripts/contract-upgrade-governance.sh
```diff
@@ -104,6 +104,12 @@ case "$chain_name" in
   solana)
     chain=1
     explorer="https://explorer.solana.com/address/"
+    extra=""
+    ;;
+  pythnet)
+    chain=26
+    explorer="https://explorer.solana.com/address/"
+    extra="Be sure to set the custom RPC to https://pythnet.rpcpool.com"
```
kcsongor, line 112: custom RPC where? should we include more info here, or is it obvious to guardians?

kcsongor (approved):

scripts/contract-upgrade-governance.sh
bruce-riley: I added a few more words.

evan-gray (approved):

panoel (approved):

PR #2350, opened 2023-02-07
Node/Watcher: make pyth ws restart on all errors

bruce-riley:
The pythnet websocket listener encountered an error and was spinning in errors because it was not trying to reconnect. This change causes it to restart on all errors, including deadline exceeded. This is acceptable because with the high rate of pyth messages, the deadline should never be exceeded.

Change-Id: I9956876d4fa8f9fc4614d1e9e294a252cd3b7610

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


evan-gray (approved):
Agree erroring out is likely the right thing to do in any case and should restart the watcher. If we find this behavior to be improper for some reason and are seeing lots of runnable died errors we can reevaluate.

panoel (approved):

PR #2365, opened 2023-02-13
node/grpc: add logging of x-forwarded-for to gRPC middleware

tbjump:
**Issue**: JSON-Web requests are being logged with `"peer.address": "@"` because they go through the `grpc-gateway`. 

**Solution**: Also log the `x-forwarded-for` header, which is set by `grpc-gateway`. This may also help in similar cases where Guardians are running reverse proxies and want to log IP addresses of the original caller. 

**Testing**: I tested this change locally by running `guardiand` outside of Tilt. If you run it inside of Tilt, the kube proxy will unfortunately hide the IP address of the caller.

### DIFF
--- a/node/pkg/common/grpc.go
+++ b/node/pkg/common/grpc.go
@@ -1,23 +1,54 @@
 package common
 
 import (
+	"context"
+
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
 	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


node/pkg/common/grpc.go
```diff
@@ -1,23 +1,54 @@
...
+// This can be identified by `"peer.address": "@"` in the logs and `grpc-gateway` correctly sets the `x-forwarded-for` metadata.
+func xForwardedForStreamServerInterceptor(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	ctx := stream.Context()
+
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		grpc_ctxtags.Extract(ctx).Set("x-forwarded-for", md.Get("x-forwarded-for"))
+	}
+
+	err := handler(srv, stream)
+	return err
```
hendrikhofstadt, line 27: return handler(srv, stream)

hendrikhofstadt (approved):

bruce-riley (approved):

PR #2381, opened 2023-02-16
node/tools: update dlv to v1.9.1

tbjump:


node/tools/go.mod
```diff
@@ -1,28 +1,5 @@
 module github.com/certusone/wormhole/node/tools
 
-go 1.19
+go 1.16
```
bruce-riley, line 3: Why are we rolling back the version of go? If this is intentional, please add a comment.

hendrikhofstadt: This should not be necessary and fail with >1.19. Also why are all the indirects dropped?

hendrikhofstadt (changes requested):
See inline comments

node/tools/go.mod
tbjump: Idk. With go 1.19 I get a build error. 

How do I generate the indirects?

hendrikhofstadt: `go mod tidy`

hendrikhofstadt (approved):

bruce-riley (approved):

PR #2390, opened 2023-02-17
node: add request logging with --publicRpcLogDetail and --logPublicRpcToTelemetry

### LINKED ISSUE #2370: node/grpc: request logging is not configurable
## Expected Behavior

Guardians should be able to configure if/how they want to log gRPC requests: 
* no logging of gRPC requests at all
* only log gRPC methods and payload
* additionally log HTTP headers

Additionally, Guardians should be able to configure if this gRPC request logging is being captured by `telemetry`. 

## Current Behavior
* Since https://github.com/wormhole-foundation/wormhole/pull/2247, request headers are being logged and forwarded to `telemetry`. 
* There was a request to add logging of request bodies (https://github.com/wormhole-foundation/wormhole/pull/2248) which was rejected due to log volume concern and offered to instead do request logging a layer up. I think that the log volume should actually not be a concern because all request bodies are less than 100 bytes unless they're malicious (and bigger request could just be dropped) and logging on the layer up is inconvenient because the gRPC payload parsing would need to be implemented there as well. I therefore propose to make request payload logging optional. 

## Possible Solution
**Proposed `guardiand` options:** 
* `--logPublicRPC`
  * `false` -- no logging of gRPC requests at all
  * `minimal` (default) -- only log gRPC methods and payload. (This default value would match the current behavior of the guardian)
  * `full` -- additionally log HTTP headers and gRPC request payload
* `--logPublicRPCTelemetry`
  * `false`
  * `true` (default) -- (This default would resemble the current behavior of the guardian where request headers, method, path, and remote_addr are logged to telemetry in https://github.com/wormhole-foundation/wormhole/blob/5c41d343ca1d9c6c2f2b436b920fced607dc85f9/node/cmd/guardiand/publicweb.go#L79)

**Request logging implementation**
* Discard any requests to the publicRPC whose body is > 100 bytes to protect against malicious requests that could cause excessive logging or compute. 
* Log HTTP headers where they're currently being logged already (https://github.com/wormhole-foundation/wormhole/blob/5c41d343ca1d9c6c2f2b436b920fced607dc85f9/node/cmd/guardiand/publicweb.go#L79) and add a `traceId`. Add this traceId to the gRPC metadata. 
* Change the gRPC logging middleware (https://github.com/wormhole-foundation/wormhole/blob/4575fe05d91499980140e87c38471a8e628969a0/node/pkg/common/grpc.go#L22) from `UnaryServerInterceptor` to `PayloadUnaryServerInterceptor`, such that the gRPC payload is being logged in JSON format. 
* Add an interceptor to log the traceId from the metadata, such that the gRPC log can be connected to the HTTP log. 

**Telemetry option**
* Not sure why exactly the gRPC middleware is not logging to the telemetry logger. 

**Which HTTP headers to log?** 
Recommend to only log (see below for rationale)
* HTTP method
* path (truncated to max 100 bytes)
* User agent (truncated to max 200 bytes)

## Alternatives considered
* **Do request logging in the layer up (e.g. nginx reverse proxy)**: The layer up wouldn't be able to parse the gRPC payloads and the logs would therefore be less useful. 

### HTTP Headers
Option: Logging all headers
* Size depends on the client. On my machine with the latest Chrome, they were ca. 650 bytes. 
* HTTP spec doesn't limit the size, but most servers limit it to 8kb or something like that (see https://stackoverflow.com/questions/686217/maximum-on-http-header-values). 
* A malicious client could make them arbitrarily large. 

Option: Only log method, path, and user-agent
* UA can be truncated after 200 bytes (common practice)
* Path can be truncated after 100 bytes (the max is in the JSON-Web API where for /v1/signed_vaa/ the request payload is in the URL. The max there is 100 bytes, see discussion about request payload. For other requests it'd just be the message name)
* I think this encompasses the most useful information and the rest doesn't need to be logged. 

### Request Payload
The publicrpc service is defined in https://github.com/wormhole-foundation/wormhole/blob/3413f09b8f63116e22bb8c365d850bf5cfce33b4/proto/publicrpc/v1/publicrpc.proto

Here is the size of all requests: 
* GetLastHeartbeatsRequest: empty
* GetSignedVAARequest: MessageID (enum, string, uint64)
    Hex-encoded (without leading 0x) emitter address. emitter address is 32 bytes, so this should be max 64 characters
* GetSignedBatchVAARequest: BatchID (enum, bytes, uint32)
    Not sure, how long the BatchID will be, since this is not implemented yet, but I assume it won't be more than a messageId
* GetLastHeartbeatsRequest: empty
* GetCurrentGuardianSetRequest: empty
* GovernorGetAvailableNotionalByChainRequest: empty
* GovernorGetEnqueuedVAAsRequest: empty
* GovernorIsVAAEnqueuedRequest: MessageID (see above)
* GovernorGetTokenListRequest: empty

Therefore, request bodies should not be longer than ~80 bytes and enforcing a hard limit of 100 shouldn't break anything.

tbjump:
fixes #2370 

# Release Notes
## New `guardiand` cli options:
* `--publicRpcLogDetail [string]`
  * `none` -- no logging of gRPC requests at all
  * `minimal` -- only log gRPC methods
  * `full` (default) -- additionally log user-agent and gRPC request payload
* `--logPublicRpcToTelemetry [bool]`
  * `false` do not send publicrpc logs to Google Cloud Logging
  * `true` (default) -- (current behavior)

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -178,10 +178,9 @@ var (
 	baseRPC      *string
 	baseContract *string
 
-	logLevel              *string
-	logLevel              *string
-	logPublicRPC          *string
-	logPublicRPCTelemetry *bool
+	logLevel                *string
+	publicRpcLogDetailStr   *string
+	publicRpcLogToTelemetry *bool
 
 	unsafeDevMode   *bool
 	testnetMode     *bool
@@ -332,8 +331,8 @@ func init() {
 	baseContract = NodeCmd.Flags().String("baseContract", "", "Base contract address")
 
 	logLevel = NodeCmd.Flags().String("logLevel", "info", "Logging level (debug, info, warn, error, dpanic, panic, fatal)")
-	logPublicRPC = NodeCmd.Flags().String("logPublicRPC", "full", "Logging of public RPC requests (false=no logging, minimal=only log gRPC methods, full=log gRPC method, payload (up to 200 bytes) and user agent (up to 200 bytes))")
-	logPublicRPCTelemetry = NodeCmd.Flags().Bool("logPublicRPCTelemetry", true, "whether or not to include publicRpc request logs in telemetry")
+	publicRpcLogDetailStr = NodeCmd.Flags().String("publicRpcLogDetail", "full", "The detail with which public RPC requests shall be logged (none=no logging, minimal=only log gRPC methods, full=log gRPC method, payload (up to 200 bytes) and user agent (up to 200 bytes))")
+	publicRpcLogToTelemetry = NodeCmd.Flags().Bool("logPublicRpcToTelemetry", true, "whether or not to include publicRpc request logs in telemetry")
 
 	unsafeDevMode = NodeCmd.Flags().Bool("unsafeDevMode", false, "Launch node in unsafe, deterministic devnet mode")
 	testnetMode = NodeCmd.Flags().Bool("testnetMode", false, "Launch node in testnet mode (enables testnet-only features)")
@@ -672,16 +671,16 @@ func runNode(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	var logPublicRpcDetail common.GrpcLogDetail
-	switch *logPublicRPC {
-	case "false":
-		logPublicRpcDetail = common.GrpcLogDetailNone
+	var publicRpcLogDetail common.GrpcLogDetail
+	switch *publicRpcLogDetailStr {
+	case "none":
+		publicRpcLogDetail = common.GrpcLogDetailNone
 	case "minimal":
-		logPublicRpcDetail = common.GrpcLogDetailMinimal
+		publicRpcLogDetail = common.GrpcLogDetailMinimal
 	case "full":
-		logPublicRpcDetail = common.GrpcLogDetailFull
+		publicRpcLogDetail = common.GrpcLogDetailFull
 	default:
-		logger.Fatal("--logPublicRPC should be one of (false, minimal, true)")
+		logger.Fatal("--publicRpcLogDetail should be one of (none, minimal, full)")
 	}
 
 	if *nodeName == "" {
@@ -965,7 +964,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Fatal("Failed to get peer ID from private key", zap.Error(err))
 		}
 
-		tm, err := telemetry.New(context.Background(), telemetryProject, creds, *logPublicRPCTelemetry, map[string]string{
+		tm, err := telemetry.New(context.Background(), telemetryProject, creds, *publicRpcLogToTelemetry, map[string]string{
 			"node_name":     *nodeName,
 			"node_key":      peerID.Pretty(),
 			"guardian_addr": guardianAddr,
@@ -1460,7 +1459,7 @@ func runNode(cmd *cobra.Command, args []string) {
 		if shouldStart(publicGRPCSocketPath) {
 
 			// local public grpc service socket
-			publicrpcUnixService, publicrpcServer, err := publicrpcUnixServiceRunnable(logger, *publicGRPCSocketPath, logPublicRpcDetail, db, gst, gov)
+			publicrpcUnixService, publicrpcServer, err := publicrpcUnixServiceRunnable(logger, *publicGRPCSocketPath, publicRpcLogDetail, db, gst, gov)
 			if err != nil {
 				logger.Fatal("failed to create publicrpc service socket", zap.Error(err))
 			}
@@ -1470,7 +1469,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			}
 
 			if shouldStart(publicRPC) {
-				publicrpcService, err := publicrpcTcpServiceRunnable(logger, *publicRPC, logPublicRpcDetail, db, gst, gov)
+				publicrpcService, err := publicrpcTcpServiceRunnable(logger, *publicRPC, publicRpcLogDetail, db, gst, gov)
 				if err != nil {
 					log.Fatal("failed to create publicrpc tcp service", zap.Error(err))
 				}
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -332,8 +332,8 @@ func init() {
 	baseContract = NodeCmd.Flags().String("baseContract", "", "Base contract address")
 
 	logLevel = NodeCmd.Flags().String("logLevel", "info", "Logging level (debug, info, warn, error, dpanic, panic, fatal)")
-	logPublicRPC = NodeCmd.Flags().String("logPublicRPC", "minimal", "Logging of public RPC requests (false=no logging, minimal=only log gRPC methods and payload (up to 200 bytes), full=additionally log HTTP method, path (up to 100 bytes), and user agent (up to 200 bytes))")
-	logPublicRPCTelemetry = NodeCmd.Flags().Bool("logPublicRPCTelemetry", true, "false=do not include publicRpc request logs in telemetry")
+	logPublicRPC = NodeCmd.Flags().String("logPublicRPC", "full", "Logging of public RPC requests (false=no logging, minimal=only log gRPC methods, full=log gRPC method, payload (up to 200 bytes) and user agent (up to 200 bytes))")
+	logPublicRPCTelemetry = NodeCmd.Flags().Bool("logPublicRPCTelemetry", true, "whether or not to include publicRpc request logs in telemetry")
 
 	unsafeDevMode = NodeCmd.Flags().Bool("unsafeDevMode", false, "Launch node in unsafe, deterministic devnet mode")
 	testnetMode = NodeCmd.Flags().Bool("testnetMode", false, "Launch node in testnet mode (enables testnet-only features)")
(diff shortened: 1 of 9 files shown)


bruce-riley (approved):

node/cmd/guardiand/publicrpc.go
```diff
@@ -15,7 +15,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-func publicrpcTcpServiceRunnable(logger *zap.Logger, listenAddr string, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) (supervisor.Runnable, error) {
+func publicrpcTcpServiceRunnable(logger *zap.Logger, listenAddr string, logPublicRPC common.GrpcLogDetail, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) (supervisor.Runnable, error) {
```
hendrikhofstadt, line 18: make clear this is the detail level in the variable name

node/cmd/guardiand/publicrpc.go
```diff
@@ -25,13 +25,14 @@ func publicrpcTcpServiceRunnable(logger *zap.Logger, listenAddr string, db *db.D
...
-	grpcServer := common.NewInstrumentedGRPCServer(logger)
+	grpcServer := common.NewInstrumentedGRPCServer(logger, logPublicRPC)
+
 	publicrpcv1.RegisterPublicRPCServiceServer(grpcServer, rpcServer)
 
 	return supervisor.GRPCServer(grpcServer, l, false), nil
 }
 
-func publicrpcUnixServiceRunnable(logger *zap.Logger, socketPath string, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) (supervisor.Runnable, *grpc.Server, error) {
+func publicrpcUnixServiceRunnable(logger *zap.Logger, socketPath string, logPublicRPC common.GrpcLogDetail, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) (supervisor.Runnable, *grpc.Server, error) {
```
hendrikhofstadt, line 35: same as above

node/pkg/telemetry/telemetry.go
```diff
@@ -38,7 +40,23 @@ var logLevelSeverity = map[zapcore.Level]logging.Severity{
...
-func (enc *encoder) EncodeEntry(entry zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
+func (enc *guardianTelemetryEncoder) EncodeEntry(entry zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
+
+	// if skipPrivateLogs==true, then private logs don't go to telemetry
+	if enc.skipPrivateLogs {
+		for _, f := range fields {
+			if f.Type == zapcore.BoolType {
+				if f.Key == "_privateLogEntry" {
+					if f.Integer == 1 {
+						return enc.Encoder.EncodeEntry(entry, fields)
```
hendrikhofstadt, line 51: doesn't this do the opposite i.e. encode logs when _privateLogEntry == true?

node/pkg/telemetry/telemetry.go
```diff
@@ -62,7 +80,16 @@ func (enc *encoder) EncodeEntry(entry zapcore.Entry, fields []zapcore.Field) (*b
 	return buf, nil
 }
 
-func New(ctx context.Context, project string, serviceAccountJSON []byte, labels map[string]string) (*Telemetry, error) {
+func (enc *guardianTelemetryEncoder) Clone() zapcore.Encoder {
```
hendrikhofstadt, line 83: Where is this used?

hendrikhofstadt (changes requested):
See inline comments

node/cmd/guardiand/node.go
```diff
@@ -647,6 +651,19 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Fatal("Please do not specify --neonContract")
 		}
 	}
+
+	var logPublicRpcDetail common.GrpcLogDetail
```
panoel, line 655: It almost seems like this var should be called "publicRpcLogLevel".

node/cmd/guardiand/node.go
```diff
@@ -175,7 +175,9 @@ var (
 	optimismCtcRpc             *string
 	optimismCtcContractAddress *string
 
-	logLevel *string
+	logLevel              *string
+	logPublicRPC          *string
```
panoel, line 179: Maybe to be consistent the name should be "publicRPCLogLevel"?

node/pkg/common/grpc.go
```diff
@@ -2,60 +2,137 @@ package common
...
-// This can be identified by `"peer.address": "@"` in the logs and `grpc-gateway` correctly sets the `x-forwarded-for` metadata.
-func xForwardedForStreamServerInterceptor(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
-	ctx := stream.Context()
+type GrpcLogDetail string
+
+const (
+	GrpcLogDetailNone    GrpcLogDetail = "false"
+	GrpcLogDetailMinimal GrpcLogDetail = "minimal"
+	GrpcLogDetailFull    GrpcLogDetail = "full"
+)
```
panoel, line 25: Sorry, I keep coming back to this.  The semantics keep throwing me off.  But, maybe that's just me.  The following would be clearer to me:
```
type GrpcLogLevel string
const (
	GrpcLoggingNone    GrpcLogLevel = "none"
	GrpcLoggingMinimal GrpcLogLevel = "minimal"
	GrpcLoggingFull    GrpcLogLevel = "full"
)

**Review by panoel — APPROVED** [review]

File: node/cmd/guardiand/node.go [hunk 8]
```diff
@@ -323,6 +325,8 @@ func init() {
 	optimismCtcContractAddress = NodeCmd.Flags().String("optimismCtcContractAddress", "", "Optimism CTC contract address")
 
 	logLevel = NodeCmd.Flags().String("logLevel", "info", "Logging level (debug, info, warn, error, dpanic, panic, fatal)")
+	logPublicRPC = NodeCmd.Flags().String("logPublicRPC", "full", "Logging of public RPC requests (false=no logging, minimal=only log gRPC methods, full=log gRPC method, payload (up to 200 bytes) and user agent (up to 200 bytes))")
```
panoel, line 328: Who is enforcing the 200 byte limit?

node/cmd/guardiand/node.go
tbjump: Hmm. I think that `publicRpcLogLevel` is not a good name because this variable controls the verbosity of the log entries, not the log level of the log entries.

node/pkg/common/grpc.go
tbjump: good suggestion

node/cmd/guardiand/node.go
tbjump: for the body, it's here: https://github.com/wormhole-foundation/wormhole/blob/c957120b5570db63b1c441a26d60c8750a7dad41/node/pkg/common/grpc.go#L71
for the user-agent, it's indeed missing after I refactored some code. Will add it back in, good catch!

node/pkg/telemetry/telemetry.go
tbjump: by returning early here, it short-circuits the publishing to telemetry by going to the underlying encoder directly. I added a corresponding code comment for clarity.

node/cmd/guardiand/publicrpc.go
tbjump: changed to `publicRpcLogDetail`

node/cmd/guardiand/publicrpc.go
tbjump: renamed to `publicRpcLogDetail`

tbjump:
Addressed reviewer feedback and resolve merge conflicts. 
Tried to make the config and variable names more clear.

hendrikhofstadt (dismissed):

node/pkg/common/grpc.go
```diff
@@ -2,60 +2,144 @@ package common
...
-// xForwardedForStreamServerInterceptor adds the `x-forwarded-for` metadata as a tag to cause it to be logged by grpc_zap.StreamServerInterceptor().
-// Note that `x-forwarded-for` can only be trusted if the latest hop in the proxy chain is trusted.
-// For JSON-Web requests, the latest hop is the guardian itself (`grpc-gateway`), which is listening on TCP and forwarding to the gRPC publicrpc UNIX socket.
-// This can be identified by `"peer.address": "@"` in the logs and `grpc-gateway` correctly sets the `x-forwarded-for` metadata.
-func xForwardedForStreamServerInterceptor(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
-	ctx := stream.Context()
+type GrpcLogDetail string
+
+const (
+	GrpcLogDetailNone    GrpcLogDetail = "false"
```
panoel, line 22: Should this be "none" or "false"?

panoel (approved):

bruce-riley (approved):

node/pkg/common/grpc.go
tbjump: good catch, thanks!

tbjump:
Fixed two typos. Hope tilt tests pass now. Will squash on merge to clean up commit history.

bruce-riley (approved):

panoel (approved):

hendrikhofstadt (approved):

PR #2399, opened 2023-02-17
sui: test token bridge `dynamic/registered_emitters`

optke3:
## Updates
- add custom, more informative error types, so we don't throw uninformative standard library errors
- assert preconditions in `registered_emitter.move` `new`, `add`, `external_address` functions
- write tests exercising intended function behavior as well as individual assertions (negative test cases)


a5-pickle (approved):
Will need to clean up comments. Would also like to fix module imports to use namespaces (as opposed to importing methods from a given module).

PR #2400, opened 2023-02-17
wormchain: migrate the jail window using a migration ante handler

conorpp:


conorpp:
shouldn't do migrations in ante handler

PR #2407, opened 2023-02-21
Refactor p2p layer

hendrikhofstadt:
This change modifies the way components of the node interact with the p2p network and the general p2p network segmenting.

It: 
- Moves packet processing and routing code out of the p2p core
- Separates processors into own routines (such as the heartbeat packet)
- Uses separate pubsub topics for different components (now subscribers like the spy can only subscribe to signed VAAs and don' receive other p2p traffic). This is a prerequisite for allowing multiple types of consensus reactors.
- Allows multiple consumers for the same p2p packets (vs the prior single consumer blocking channel concepts)
- Reduces blocking between different components by using the subscription model
- Introduces metered channel pairs that expose metrics on processing (more similar to a primitive work queue)

This change is currently backwards compatible. It runs in a legacy mode where all topics receive / publish from the legacy topic as well.


node/pkg/p2p/gossip.go
```diff
@@ -0,0 +1,146 @@
...
+
+type GossipSender interface {
+	Send(ctx context.Context, msg *gossipv1.GossipMessage) error
+}
+
+type FilteredEnvelope[K any] struct {
+	Message K
+	From    peer.ID
+}
+
```
bruce-riley, line 36: Please add prologues for these two functions.

sdk/vaa/structs.go
```diff
@@ -742,40 +755,49 @@ func (v *VAA) VerifySignatures(addresses []common.Address) bool {
...
+func (b *BatchVAA) serializeBody() []byte {
 	buf := new(bytes.Buffer)
 
-	hashes := v.ObsvHashArray()
+	hashes := b.ObsvHashArray()
 
-	MustWrite(buf, binary.BigEndian, uint8(len(hashes)))
+	numHashes := len(hashes)
+	if numHashes > MaxBatchObservations {
+		panic(fmt.Sprintf("encountered BatchVAA Hashes length outside of expected range: %d", numHashes))
```
bruce-riley, line 789: Is there any chance that an external source publishing too large of a batch could panic the guardian?

evan-gray:
Closing as this is pretty stale now, the base branch is closed (was superseded by https://github.com/wormhole-foundation/wormhole/pull/3140), and I get the impression that future endeavors may separate p2p from the guardian itself.

PR #2410, opened 2023-02-22
accountant: remove double keccak for submitting observations

conorpp:
This updates some of the functions in the sdk to calculate & verify digests in a less error prone way.  Calculating or verifying a digest now requires a VAA type, or a prefix to be used for a custom message.  The prefix must be at least 32 bytes.

The accountant and guardian were using an undocumented way of signing messages.  I.e. they were hashing message twice with keccak which makes the message indistinguishable from a P2P payload.  To better protect from message confusion, we removed the double hash and used the less error prone digest methods in the sdk.

Right now there are still old functions still in place (E.g. `VerifyQuorum` on Wormchain) to ensure we have compatibility throughout the Wormchain upgrade and accountant upgrade.  After the upgrade, we can fully remove the deprecated functions.


evan-gray (changes requested):

evan-gray:
Also, how does a test **not** fail with this change? Shouldn't https://github.com/wormhole-foundation/wormhole/blob/main/wormchain/x/wormhole/keeper/vaa_test.go fail?

wormchain/x/wormhole/keeper/query_plugins.go
```diff
@@ -57,8 +58,8 @@ func WormholeQuerier(keeper Keeper) func(ctx sdk.Context, data json.RawMessage)
 		}
 		if wormholeQuery.VerifySignature != nil {
 			// handle the verify signature query
-			digest := vaa.SigningMsg(wormholeQuery.VerifySignature.Data).Bytes()
-			err := keeper.VerifySignature(ctx, digest, wormholeQuery.VerifySignature.GuardianSetIndex, wormholeQuery.VerifySignature.Signature)
+			digest := ethCrypto.Keccak256Hash(wormholeQuery.VerifySignature.Data)
```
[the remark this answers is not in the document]
conorpp: I think this was written to avoid doing the hash in cosmwasm, but I don't think we're close to bottlenecking in performance there, so I can apply this change.

sdk/vaa/structs.go
```diff
@@ -645,21 +645,42 @@ func (v *BatchVAA) signingBody() []byte {
...
 }
 
-// SigningMsg returns the hash of the signing body. This is used for signature generation and verification
-func (v *BatchVAA) SigningMsg() common.Hash {
-	return SigningMsg(v.signingBody())
+// MessageSigningDigest returns the hash of the data prepended with it's signing prefix.
+// This is intending to be used for signing messages of different types from VAA's.
+// The message prefix help protect from message collisions.
+func MessageSigningDigest(prefix []byte, data []byte) (common.Hash, error) {
+	if len(prefix) < 32 {
```
bruce-riley, line 666: I believe @tbjump said the check should be that the `len(prefix) + len(data) >= 34`.

sdk/vaa/structs.go
```diff
@@ -645,21 +645,42 @@ func (v *BatchVAA) signingBody() []byte {
...
+func DeprecatedSigningDigest(bz []byte) common.Hash {
+	return doubleKeccak(bz)
 }
 
-// SigningMsg returns the hash of the signing body. This is used for signature generation and verification
-func (v *BatchVAA) SigningMsg() common.Hash {
-	return SigningMsg(v.signingBody())
+// MessageSigningDigest returns the hash of the data prepended with it's signing prefix.
+// This is intending to be used for signing messages of different types from VAA's.
+// The message prefix help protect from message collisions.
```
bruce-riley, line 664: Nit: "helps".

sdk/vaa/structs.go
```diff
@@ -645,21 +645,42 @@ func (v *BatchVAA) signingBody() []byte {
...
+func MessageSigningDigest(prefix []byte, data []byte) (common.Hash, error) {
+	if len(prefix) < 32 {
+		// Prefixes must be at least 32 bytes
+		// https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0009_guardian_key.md
+		return common.Hash([32]byte{}), errors.New("prefix must be atleast 32 bytes")
```
bruce-riley, line 669: Nit: "at least".

sdk/vaa/structs.go
```diff
@@ -727,16 +750,34 @@ func VerifySignatures(digest []byte, signatures []*Signature, addresses []common
...
 
+// Operating on bytes directly is error prone.  We should use `vaa.VerifyingSignatures()` whenever possible.
+// This function will be removed in a subsequent release.
+func DeprecatedVerifySignatures(vaaBody []byte, signatures []*Signature, addresses []common.Address) bool {
+	vaaDigest := doubleKeccak(vaaBody)
+	return verifySignatures(vaaDigest[:], signatures, addresses)
+}
+
+func VerifyMessageSignature(prefix []byte, messageBody []byte, signatures *Signature, addresses common.Address) bool {
+	if len(prefix) < 32 {
```
bruce-riley, line 761: Same comment regarding length check.

bruce-riley (approved):

bruce-riley:
I only reviewed the go code.

sdk/vaa/structs.go
tbjump: doesn't matter too much for now, we can come back to this if/once the p2p messages are ported over to use this function.

sdk/vaa/structs.go
```diff
@@ -645,21 +645,42 @@ func (v *BatchVAA) signingBody() []byte {
 	return buf.Bytes()
 }
 
-// SigningMsg returns the hash of the signing body.
```
tbjump, line 648: since this is part of the SDK, could this interface change break users of the sdk and would require a major version increase? Maybe just better to keep the deprecated method for now and just mark it as deprecated instead of removing it

jynnantonix (approved):

bruce-riley (approved):

tbjump (approved):

hendrikhofstadt (approved):

hendrikhofstadt:
My only remaining question was if we didn't want to require exactly 32bytes of prefix for more consistency?

tbjump:
> My only remaining question was if we didn't want to require exactly 32bytes of prefix for more consistency?

I think that'd be nice for consistency, but the prefixes across the codebase are of varying length already and I don't think it's worth the effort to unify them to 32 bytes, given that those are mostly breaking changes.

jynnantonix (approved):

PR #2413, opened 2023-02-22
sui: test `wrapped_asset.move`

optke3:
## Summary
Write a test for creation and destruction of `WrappedAssets`. Since the mint and burn functionalities are already tested in `registered_tokens.move`, we do not add test coverage for them here.


a5-pickle (approved):
Sick

PR #2417, opened 2023-02-23
sui: update coin metadata 

optke3:


sui/token_bridge/sources/create_wrapped.move
```diff
@@ -1,23 +1,26 @@
...
     use std::option::{Self};
+    use sui::coin::{Self, CoinMetadata};
     use sui::transfer::{Self};
     use sui::tx_context::{TxContext};
     use sui::url::{Url};
     use wormhole::state::{Self as wormhole_state, State as WormholeState};
     use wormhole::myvaa as core_vaa;
 
-    use token_bridge::asset_meta::{Self};
+    use token_bridge::asset_meta::{Self, symbol_to_string, name_to_string};
```
kcsongor, line 14: I think we should just import the module and use `asset_meta::symbol_to_string` and `asset_meta::name_to_string`. Little easier to read the code with the module qualifiers IMO (and that's how these two functions are already used in this module)

sui/token_bridge/sources/datatypes/string32.move
```diff
@@ -96,6 +97,18 @@ module token_bridge::string32 {
         string::utf8(bytes)
     }
 
+    public fun to_ascii_string(s: &String32): ascii::String {
```
kcsongor, line 100: we should document what this function does (and what it doesn't do). currently it fails on strings that contain non-ascii characters, which the symbol might well do, since the token bridge spec allows utf8 strings here. Currently it means that tokens with unicode characters in their symbol name are not supported. we can either document this, or fix somehow (e.g. by removing non-ascii characters)

sui/token_bridge/sources/create_wrapped.move
```diff
@@ -118,4 +121,38 @@ module token_bridge::create_wrapped {
             decimals,
         );
     }
+
+    public entry fun update_registered_wrapped_coin_metadata<CoinType>(
```
kcsongor, line 125: currently the `CoinType` is not verified which means that the coin that the VAA refers to might be different from `CoinType`. this allows an attacker to grab an attestation VAA and run it against any wrapped coin type. we should query `state::token_info<CoinType>` and ensure that the token chain and token address match what's in the VAA

kcsongor: that check should probably exist in `state::update_registered_wrapped_coin_metadata<CoinType>`

sui/token_bridge/sources/create_wrapped.move
```diff
@@ -118,4 +121,38 @@ module token_bridge::create_wrapped {
...
+        let meta = asset_meta::deserialize(payload);
+        let origin_chain = asset_meta::token_chain(&meta);
+
+        assert!(
+            origin_chain != wormhole_state::chain_id(),
+            E_UPDATING_NATIVE_COIN_META
+        );
+        assert!(
+            state::is_registered_asset<CoinType>(token_bridge_state),
+            E_WRAPPED_COIN_NOT_INITIALIZED
```
kcsongor, line 148: is this check necessary? the existence of `metadata` implies it's been set up, doesn't it?

sui/token_bridge/sources/objects/registered_tokens.move
```diff
@@ -75,6 +75,15 @@ module token_bridge::registered_tokens {
         !is_wrapped<C>(self)
     }
 
+    public(friend) fun treasury_cap<C>(
+        self: &RegisteredTokens
+    ): &TreasuryCap<C> {
+        assert!(is_wrapped<C>(self), 0);
```
kcsongor, line 81: better error code

sui/token_bridge/sources/objects/registered_tokens.move
```diff
@@ -75,6 +75,15 @@ module token_bridge::registered_tokens {
         !is_wrapped<C>(self)
     }
 
+    public(friend) fun treasury_cap<C>(
```
kcsongor, line 78: should `C` be `CoinType`?

kcsongor (changes requested):

sui/token_bridge/sources/create_wrapped.move
optke3: just because the metadata exists doesn't mean it is registered with the bridge as a wrapped asset native better error message would be `E_COIN_NOT_REGISTERED_AS_WRAPPED`

sui/token_bridge/sources/objects/registered_tokens.move
optke3: we've been using C for brevity in some places

sui/token_bridge/sources/datatypes/string32.move
```diff
@@ -12,7 +12,7 @@ module token_bridge::string32 {
 
     const E_STRING_TOO_LONG: u64 = 0;
 
-    const QUESTION_MARK: u8 = 0x63;
+    const QUESTION_MARK: u8 = 63;
```
a5-pickle, line 15: Nice catch

PR #2422, opened 2023-02-23
Deploy base to testnet

bruce-riley:
This PR deploys the contracts to Base testnet. This will be followed by separate PRs to add guardian and client support.


ethereum/contracts/Implementation.sol
```diff
@@ -53,6 +53,7 @@ contract Implementation is Governance {
             } else if (chain == 23) { evmChainId = 42161;      // arbitrum
             } else if (chain == 24) { evmChainId = 10;         // optimism
             } else if (chain == 25) { evmChainId = 100;        // gnosis
+            } else if (chain == 30) { evmChainId = 84531;      // base
```
kev1n-peters, line 56: We don't need this migration (and in `NftBridgeImplementation.sol`) since the `evmChainId` will already be set during the initial deployment

bruce-riley: True, but since it was already there, I didn't want it to be incomplete. I could delete everything if people prefer.

bruce-riley: Sounds like we should leave this as is then.

scripts/contract-upgrade-governance.sh
```diff
@@ -190,6 +190,13 @@ case "$chain_name" in
     explorer="https://optimistic.etherscan.io/address/"
     evm=true
     ;;
+  base)
+    echo "Need to specify the base explorer URL!"
+    exit 1  
+    chain=30
+    explorer="??/address/"
```
evan-gray, line 197: https://goerli.basescan.org/

ethereum/contracts/Implementation.sol
evan-gray: Isn't this the testnet chain id though? should this be committed with all these mainnet ids? I'd lean towards omitting this for now.

evan-gray (changes requested):

ethereum/contracts/Implementation.sol
bruce-riley: Good catch. Removing base from both core and nft implementation (didn't add it to token bridge).

scripts/contract-upgrade-governance.sh
bruce-riley: Don't think we want testnet here. This is a reminder to add the mainnet one when we know it.

kev1n-peters (approved):

scripts/contract-upgrade-governance.sh
evan-gray: ohhh. 👍

evan-gray (approved):

a5-pickle (approved):
Reviewed changes in the _ethereum_ directory.

PR #2432, opened 2023-02-24
Node: Rename Arbitrum connector

bruce-riley:
The `ArbitrumConnector` in the EVM watcher is now also being used for Optimism Bedrock and Base, so this PR renames it to `ConnectorWithGetTimeOfBlock` since that is what it really provides.


panoel (dismissed):

hendrikhofstadt (commented):
LGTM, however I would recommend adding the docstring to the struct instead of the file beginning. Also I think the naming is misleading because the normal ethereum connector also has the method.

panoel (approved):

hendrikhofstadt (approved):

PR #2465, opened 2023-03-03
Node/Governor: Regenerate token list and floor prices

bruce-riley:
This PR regenerates the token list used by the governor, based on querying a new cloud function that returns all tokens with CoinkGecko IDs, rather than just the top 100.

This new list required removing a couple of tests because the new token list has a variety of decimals, and some tokens do not have symbols.

Also, some tokens were removed from the manual list because they are now auto generated.


tbjump:
Thanks for taking the lead on this. 
1. I was wondering if it was possible to somehow generate a better diff of the [generated_mainnet_tokens](https://github.com/wormhole-foundation/wormhole/pull/2465/files#diff-648f1166c6f409bf3c67dadb1478fc6521ccb032289d4fbf87d9cf77871b8a37) that'll show which tokens are removed? 
2. Do you know how it is possible that there were tokens on the list before that are not in the new list, e.g. `raydium`?

node/pkg/governor/mainnet_tokens_test.go
```diff
@@ -64,25 +64,6 @@ func TestTokenListTokenAddressDuplicates(t *testing.T) {
...
-}
-
-func TestTokenListEmptySymbols(t *testing.T) {
-	tokenConfigEntries := tokenList()
-
-	/* Assume that all governed token entry symbol strings will be greater than zero */
-	for _, tokenConfigEntry := range tokenConfigEntries {
-		assert.Greater(t, len(tokenConfigEntry.symbol), 0)
-	}
-}
```
claudijd, line 84: reading above, the exception case is Solana tokens, could we instead in the loop just continue for Solana and still get said protection for the other chains where we expect to symbols to be non-zero or were there other unexpected cases on other chains?

claudijd (dismissed):
Approving, but had two questions for Bruce to consider on whether the unit-tests that help us prevent errors could be salvaged or not.  I suspect this is very context specific, so we may or may not be able to easily keep them, leaving this to Bruce's discretion.

node/pkg/governor/mainnet_tokens_test.go
```diff
@@ -64,25 +64,6 @@ func TestTokenListTokenAddressDuplicates(t *testing.T) {
...
 }
 
-func TestTokenListDecimalRange(t *testing.T) {
-	tokenConfigEntries := tokenList()
-
-	/* Assume that all governed token entries will have decimals of 6 or 8 */
-	for _, tokenConfigEntry := range tokenConfigEntries {
-		d := tokenConfigEntry.decimals
-		assert.Condition(t, func() bool { return d == 6 || d == 8 })
-	}
```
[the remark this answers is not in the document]
bruce-riley: Yes, there are at least a dozen different number of decimals being used, so it doesn't seem worth trying to test.

node/pkg/governor/mainnet_tokens_test.go
bruce-riley: Put the test back and just have it skip Solana.

claudijd (approved):

evan-gray (approved):
reviewed node/hack

PR #2472, opened 2023-03-06
sui: fix deploy script

heyitaki:
This PR aims to modify the existing deployment script to work within Tilt since changes to the Sui CLI and the Wormhole Sui contracts previously cause this deployment script to break. In addition, manual steps such as having to import a key and modifying `Move.toml` files have been completely automated.

Note that this script is fairly fragile because it relies on the output of Sui CLI commands which are subject to change. Implementing this within the `worm` CLI will improve the reliability of this functionality.


sui/Dockerfile
```diff
@@ -10,13 +10,14 @@ WORKDIR /tmp
 
 RUN sui genesis -f
 COPY scripts/start_node.sh  .
-COPY scripts/funder.sh  .
```
optke3, line 13: If we no longer use the faucet, do we manually fund the deployer account before deploying the contracts?

optke3 (changes requested):
I think we want to make sure `make test` and `sui move test` still pass. Currently, they fail since the Move.toml doesn't have the `dev-addresses` specified. 

Do we want to have separate Move.toml files for the deploy sequence and tests? Were you still going to automate toggling between the toml files?

If this PR is meant to unblock watcher dev, maybe we should hold off on merging into `sui/integration_v2`, since we want the contract tests to pass on there.

heyitaki:
Closing in favor of #2486

PR #2478, opened 2023-03-07
github-actions: update go version in build.yml

SEJeff:
This is a W.I.P. PR to also test moving the default workflow perms to read only vs read-write.


hendrikhofstadt (dismissed):
If we're updating, could also just go to go1.19.7

hendrikhofstadt (approved):

PR #2480, opened 2023-03-08
relayer: update docs on running a spy

SEJeff:
Having two bootstrap nodes allows an outage of the certus one guardian to not cause too much hassle with existing spy instances or new guardians restarting.

Thanks @fchantada and the xLabs team for adding the dns and helping set this up.


evan-gray (approved):
Nice!

PR #2482, opened 2023-03-08
node/telemetry: fix nullpointer, _privateLogEntry, tests

tbjump:
1. Fix a nullpointer dereference bug caused by not cloning everything in `(enc *guardianTelemetryEncoder) Clone()`
2. Fix a bug that caused logs marked with `_privateLogEntry` to still be sent to telemetry because when a child logger is created (`logger.With(zap.Bool("_privateLogEntry", true))`) the extra fields will go into the encoder's buffer and will not be passed to `EncodeEntry(entry zapcore.Entry, fields []zapcore.Field)`. This bug had no production impact because the telemetry logger was broken (see (1))
3. make the telemetry module testable and add unit tests

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -947,7 +947,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Fatal("Failed to get peer ID from private key", zap.Error(err))
 		}
 
-		tm, err := telemetry.NewGoogleCloudTelemetry(context.Background(), telemetryProject, creds, *publicRpcLogToTelemetry, map[string]string{
+		tm, err := telemetry.New(context.Background(), telemetryProject, creds, *publicRpcLogToTelemetry, map[string]string{
 			"node_name":     *nodeName,
 			"node_key":      peerID.Pretty(),
 			"guardian_addr": guardianAddr,
@@ -961,15 +961,6 @@ func runNode(cmd *cobra.Command, args []string) {
 		logger = tm.WrapLogger(logger)
 	} else {
 		logger.Info("Telemetry disabled")
-
-		externalLogger := &telemetry.ExternalLoggerNil{}
-		tm, err := telemetry.New(*publicRpcLogToTelemetry, externalLogger)
-		if err != nil {
-			logger.Fatal("Failed to initialize telemetry", zap.Error(err))
-		}
-		defer tm.Close()
-		logger = tm.WrapLogger(logger)
-
 	}
 
 	// Redirect ipfs logs to plain zap
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -947,7 +947,7 @@ func runNode(cmd *cobra.Command, args []string) {
 			logger.Fatal("Failed to get peer ID from private key", zap.Error(err))
 		}
 
-		tm, err := telemetry.New(context.Background(), telemetryProject, creds, *publicRpcLogToTelemetry, map[string]string{
+		tm, err := telemetry.NewGoogleCloudTelemetry(context.Background(), telemetryProject, creds, *publicRpcLogToTelemetry, map[string]string{
 			"node_name":     *nodeName,
 			"node_key":      peerID.Pretty(),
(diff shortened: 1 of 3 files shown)


node/pkg/telemetry/telemetry.go
```diff
@@ -67,16 +90,16 @@ func (enc *guardianTelemetryEncoder) EncodeEntry(entry zapcore.Entry, fields []z
 	bufCopy := make([]byte, len(buf.Bytes()))
 	copy(bufCopy, buf.Bytes())
 
-	// Convert the zapcore.Level to a logging.Severity
-	severity := logLevelSeverity[entry.Level]
+	// if skipPrivateLogs==true, then private logs don't go to telemetry
+	if enc.skipPrivateLogs {
+		if bytes.Contains(bufCopy, []byte("\"_privateLogEntry\":true")) {
```
hendrikhofstadt, line 95: Couldn't this be tricked by adding this to some user-controlled input (either now or in the future)

hendrikhofstadt (changes requested):
The new design looks good to me and I like that you introduced tests. Main concern is with checking for the tag in the buffer which may contain untrusted input.

node/pkg/telemetry/telemetry.go
```diff
@@ -14,16 +17,53 @@ import (
...
+		Severity:  logLevelSeverity[level],
+		Labels:    logger.labels,
+	})
+}
+
+func (logger *ExternalLoggerGoogleCloud) flush() error {
+	return logger.Flush()
+}
+
+// Mock external logger for integration tests
```
bruce-riley, line 49: Not entirely comfortable with calling this "Mock", since it will actually be used in production if telemetry is disabled. At least update the comment to indicate that it is not only used for tests. Should there be a separate "null logger" used in prod that doesn't include the event counter?

bruce-riley (approved):

panoel (approved):

node/pkg/telemetry/telemetry.go
tbjump: renamed it to `ExternalLoggerNil`

node/cmd/guardiand/node.go
```diff
@@ -961,6 +961,15 @@ func runNode(cmd *cobra.Command, args []string) {
 		logger = tm.WrapLogger(logger)
 	} else {
 		logger.Info("Telemetry disabled")
+
+		externalLogger := &telemetry.ExternalLoggerNil{}
```
hendrikhofstadt, line 965: given you're passing in a pointer anyway I think just accepting nil as input would make sense.

hendrikhofstadt: For tests, a "testExternalLogger" could be used that sits in the test file.

hendrikhofstadt (changes requested):

node/cmd/guardiand/node.go
tbjump: I wanted to include this `ExternalLoggerNil` stub logger in devmode such that the code paths involving the cloning of the logger etc. are covered in integration tests. That's why it's not just local to the test file. 
Wdyt?

hendrikhofstadt: I'd prefer accepting nil and just writing a unit test that exercises the clone.

tbjump: Ok. Reverted changes to `node.go` and made `externalLoggerMock` local to `telemetry_test.go`

hendrikhofstadt (dismissed):

node/pkg/telemetry/telemetry.go
```diff
@@ -14,16 +16,39 @@ import (
 	"google.golang.org/api/option"
 )
 
+const TELEMETRY_LOG_LEVEL = zap.InfoLevel
```
hendrikhofstadt, line 19: small nit, Constants in go are not capital snake case but camel case

tbjump: changed

hendrikhofstadt (approved):

bruce-riley (approved):

PR #2486, opened 2023-03-08
sui: fix deploy script

### LINKED ISSUE #2449: sui: deploy package to Tilt
Recently, the `Move.toml` files had a `[dev-dependencies]` section added for testing. While this is useful for testing, it adds complexity to the build process because the addresses specified under dev dependencies now conflict with the addresses under dependencies (all set to `0x0`). If we follow the recommended convention and turn the addresses under dependencies into `_`, `sui move test` will pass, but `sui move build` will now fail due to unspecified module addresses.

This require us to create some sort of dynamic build process that hotswaps the underscores with the correct addresses before building, and back to the underscores post build.

heyitaki:
This PR aims to modify the existing deployment script to work within Tilt since changes to the Sui CLI and the Wormhole Sui contracts previously cause this deployment script to break. In addition, manual steps such as having to import a key and modifying Move.toml files have been completely automated.

This is meant to unblock any development dependent on a working Tilt environment (eg Sui guardian changes), but note that this script is fairly fragile because it relies on the output of Sui CLI commands which are subject to change. Implementing this functionality within the worm CLI and using those commands here a la the aptos deploy script will improve reliability (#2047).

Resolves #2449


a5-pickle (dismissed):

optke3 (dismissed):
lgtm, except for docker-test

optke3 (approved):
lgtm!

PR #2487, opened 2023-03-08
sui: worm cli parse address

optke3:
## Updates
Fill in the Worm CLI `parseAddress` function for Sui. 

This makes it possible to generate a chain registration VAA for Sui, for example:

```
worm generate registration --chain "sui" -a 0x00000000000000000000000000000000000000000000000000000000deadbeef --guardian-secret cfb12303a19cde580bb4dd771639b0d26bc68353645571a8cff516ab2ee113a0 -m "TokenBridge"
```


clients/js/cmds/generate.ts
```diff
@@ -290,7 +290,9 @@ function parseAddress(chain: ChainName, address: string): string {
   } else if (chain === "osmosis") {
     throw Error("OSMOSIS is not supported yet");
   } else if (chain === "sui") {
-    throw Error("SUI is not supported yet");
+    if (/^(0x)?[0-9a-fA-F]+$/.test(address)) {
```
heyitaki, line 293: What is the material difference between this entire method and this one used in the sdk? If they are the same, we should standardize usage to one of them; I don't have a strong opinion on which https://github.com/wormhole-foundation/wormhole/blob/6def9b306ccbf001185436e985ca58b551ea6e60/sdk/js/src/utils/array.ts#L214

heyitaki (changes requested):

clients/js/cmds/generate.ts
optke3: it does look like there is duplicate functionality

bruce-riley (approved):

evan-gray:
@heyitaki maybe you can look at re-adding this from a less stale branch 🙂

PR #2491, opened 2023-03-09
Node/Acct: Handle large batch status queries

bruce-riley:
When a guardian restarts, it may attempt to query for many pending transfers, which can cause the query to run out of gas and fail. This PR breaks those large queries up into a series of smaller ones. Based on testing by @evan-gray, we believe we could do up to 750 queries per batch. This change uses a more conservative value of 500 queries per batch.

Please do not be put off by the number of lines in this PR. The vast majority is populating data used for testing (as well as the actual tests).

### DIFF
--- a/node/pkg/accountant/query_test.go
+++ b/node/pkg/accountant/query_test.go
@@ -156,7 +156,9 @@ type BatchTransferStatusQueryConnMock struct {
 }
 
 func (qc *BatchTransferStatusQueryConnMock) SubmitQuery(ctx context.Context, contractAddress string, query []byte) ([]byte, error) {
-	if len(query) > 150*maxPendingsPerQuery {
+	// Force a failure if the query is much bigger than what we are allowing. This does not have to be exact, since the chunking tests will be using a lot more than that.
+	// A json encoded transfer key is about 150 characters.
+	if len(query) > 150*maxPendingsPerQuery+1000 {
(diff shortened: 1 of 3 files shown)


node/pkg/accountant/query_test.go
```diff
@@ -145,3 +149,84 @@ func TestParseBatchTransferStatusPendingResponse(t *testing.T) {
...
 	assert.True(t, reflect.DeepEqual(expectedResult, response.Details[0]))
 }
+
+// BatchTransferStatusQueryConnMock allows us to mock batch_transfer_status by implementing SubmitQuery.
+type BatchTransferStatusQueryConnMock struct {
+	resp []byte
+}
+
+func (qc *BatchTransferStatusQueryConnMock) SubmitQuery(ctx context.Context, contractAddress string, query []byte) ([]byte, error) {
+	if len(query) > 150*maxPendingsPerQuery {
```
hendrikhofstadt, line 159: would be best to leave a comment here to explain the 150

hendrikhofstadt (approved):

node/pkg/accountant/query_test.go
bruce-riley: Added a comment.

hendrikhofstadt (approved):

panoel (approved):

PR #2492, opened 2023-03-09
sui: test upgrade contract

optke3:
## Update
- We write a test confirming that the following contract upgrade call sequence successfully updates the version number in the upgrade capability.
```
(upgrade vaa) -> upgrade_contract -> (UpgradeTicket) -> test_upgrade -> (UpgradeReceipt) -> commit_upgrade
```
- Write a negative test for calling commit_upgrade at old call-site.


node/pkg/watchers/sui/watcher.go
```diff
@@ -399,11 +399,14 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			}
 
+			// curl -s -X POST -d '{"jsonrpc":"2.0", "id": 1, "method": "sui_getLatestCheckpointSequenceNumber", "params": []}' -H 'Content-Type: application/json' http://127.0.0.1:9000
 			// Epoch is currently not ticking in 0.16.0.  They also
 			// might release another API that gives a
 			// proper block height as we traditionally
 			// understand it...
 			currentSuiHeight.Set(float64(res.Result.Epoch))
+			logger.Error(fmt.Sprintf("sui_getCommitteeInfo: %d", res.Result.Epoch))
```
bruce-riley, line 408: Don't need to do a sprintf here. That being said, @panoel is reworking this whole file under a separate PR, so it will be rereviewed then.

bruce-riley (approved):
Only reviewed the single golang file, which will be rereviewed when @panoel finishes the watcher PR.

evan-gray:
OBE

PR #2502, opened 2023-03-10
Node/acct minor fixes

bruce-riley:
This PR addresses several small issues that were noticed while getting the accountant working in log-only mode. Some of them are not really related to accountant, but hopefully no one objects to them being included here.

Here's what's included:

- The accountant should not block reobservations off pending transfers when in log only mode.
- Increase the accountant audit interval to be greater than the reobservation window, since reobservations more often than that don't go through anyway.
- We are seeing reobservation requests that the accountant makes on polygon failing, and it would be helpful if the error message included the txHash.
- An unnecessary log message got committed in node.go. (This is the one thing that really is not related to the accountant!)

### DIFF
--- a/node/pkg/accountant/accountant.go
+++ b/node/pkg/accountant/accountant.go
@@ -240,11 +240,12 @@ func (acct *Accountant) SubmitObservation(msg *common.MessagePublication) (bool,
 				zap.String("msgID", msgId),
 				zap.String("oldDigest", oldEntry.digest),
 				zap.String("newDigest", digest),
+				zap.Bool("enforcing", acct.enforceFlag),
 			)
 		} else {
-			acct.logger.Info("acct: blocking transfer because it is already outstanding", zap.String("msgID", msgId))
+			acct.logger.Info("acct: blocking transfer because it is already outstanding", zap.String("msgID", msgId), zap.Bool("enforcing", acct.enforceFlag))
 		}
-		return false, nil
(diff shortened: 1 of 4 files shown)


node/pkg/accountant/accountant.go
```diff
@@ -240,11 +240,12 @@ func (acct *Accountant) SubmitObservation(msg *common.MessagePublication) (bool,
...
 				zap.String("oldDigest", oldEntry.digest),
 				zap.String("newDigest", digest),
+				zap.Bool("enforcing", acct.enforceFlag),
 			)
 		} else {
-			acct.logger.Info("acct: blocking transfer because it is already outstanding", zap.String("msgID", msgId))
+			acct.logger.Info("acct: blocking transfer because it is already outstanding", zap.String("msgID", msgId), zap.Bool("enforcing", acct.enforceFlag))
 		}
-		return false, nil
+		return !acct.enforceFlag, nil
```
panoel, line 248: Does the comment on line 235 need to change, since this is not a blanket false and depends on the enforceFlag?

panoel (approved):

jynnantonix (approved):

hendrikhofstadt (dismissed):

hendrikhofstadt (approved):

panoel (approved):

evan-gray (approved):

PR #2494, opened 2023-03-09
sui/wormhole: refactor and add upgradability

a5-pickle:
## Objective

The purpose of this pull request is to clean up the Wormhole core contract, which includes:

- Re-organizing large modules into smaller modules
- Converting data structures to dynamic fields where appropriate
- Adding `wormhole_scenario` module for integrators to use to initialize a test scenario with Wormhole
- Adding version control for upcoming [Sui upgradability]
  - This includes a temporary module `dummy_sui_package`, which seeks to replicate `sui::package`, so we could build out upgrade logic
- Adding `public entry` methods to the `wormhole` module (with the exception of `init_and_share_state`, which lives in the `setup` module)

Details of the methodology behind upgrading Wormhole is written about in Issue #2495 (this does not close the issue, as this PR may not fully capture everything required to satisfy it).

Current Sui build: **0.25.0**

## How to Review

1. Navigate to _sui/wormhole_ and run `make test` to make sure all Move tests succeed.
3. Review `publish_message_tests` and `publish_test` modules in _publish_message.move_
4. Review `vaa_tests` and `vaa` modules in _vaa.move_.
5. Review `state` module (first to review `new_emitter` method and modules associated with it, then to review `State` organization with its getters/setters).
6. Review governance tests and logic, which live in the _governance_ subdirectory.
7. Review other modules and associated tests.

[sui upgradability]: https://github.com/MystenLabs/sui/issues/2045

PR #2512, opened 2023-03-13
Sui/wormhole: add zero address utility method to external_address

gator-boi:
The purpose of this PR is to add a utility method to the `external_address` module that allows integrators to check if an `ExternalAddress` type contains the zero address.

### DIFF
--- a/sui/wormhole/sources/datatypes/external_address.move
+++ b/sui/wormhole/sources/datatypes/external_address.move
@@ -72,6 +72,18 @@ module wormhole::external_address {
         address::from_bytes(vec)
     }
 
+    public fun is_nonzero(e: &ExternalAddress): bool {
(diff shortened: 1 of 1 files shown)


kcsongor:
what would the use case of such a function be? on EVM, the non-zero checks are required to ensure that the data is not missing (since 0 values are indistinguishable by missing values), but Move doesn't have this problem.

optke3 (approved):
lgtm!

gator-boi:
@kcsongor Dapps that compose on the Token Bridge might allow a user to specify a target wallet address. This could be used to confirm that the caller doesn't pass a vector of zeros.

gator-boi:
A cleaner version of this check has been added to #2494.

PR #2513, opened 2023-03-13
accountant: separate modify_balance into separate module and check er…

conorpp:
* Updates tests in accountant to check for specific errors instead of just that the error occurred.  There were a couple of cases the we weren't getting the error we thought we were, but there no actual bugs.

* Add `accountant` governance message type to rust sdk.

* Changes `ModifyBalance` to be checked only when the "global account" module is used in the governance message.  Otherwise, ModifyBalance can be confused with other defined token bridge governance messages.

* Change guardian to use "global accountant" module for `ModifyBalance`.

* Update wormchain tests since they include accountant.


cosmwasm/contracts/global-accountant/src/contract.rs
```diff
@@ -334,9 +334,21 @@ fn handle_vaa(
...
-        handle_governance_vaa(deps.branch(), info, body.with_payload(govpacket))?
+        let mut module = [0u8; 32];
+        module.clone_from_slice(&body.payload.get()[..32]);
+
+        if let Ok(govpacket) = serde_wormhole::from_slice::<token::GovernancePacket>(body.payload) {
+            handle_token_governance_vaa(deps.branch(), body.with_payload(govpacket))?
+        } else if let Ok(govpacket) =
+            serde_wormhole::from_slice::<accountant_module::GovernancePacket>(body.payload)
+        {
+            handle_accountant_moduleernance_vaa(deps.branch(), info, body.with_payload(govpacket))?
```
evan-gray, line 345: Typo?

evan-gray (dismissed):

sdk/rust/core/src/accountant.rs
```diff
@@ -0,0 +1,424 @@
...
+                                    token_chain,
+                                    token_address,
+                                    kind,
+                                    amount,
+                                    reason,
+                                }
+                            }
+                            v => {
+                                return Err(Error::custom(format_args!(
+                                    "invalid action: {v}, expected one of: 1, 2"
```
hendrikhofstadt, line 317: only expecting 1

sdk/rust/core/src/accountant.rs
```diff
@@ -0,0 +1,424 @@
...
+                        token_chain,
+                        token_address,
+                        kind,
+                        amount,
+                        reason,
+                    }
+                }
+                v => {
+                    return Err(Error::custom(format_args!(
+                        "invalid value {v}, expected one of 1, 2"
```
hendrikhofstadt, line 236: of 1

sdk/rust/core/src/accountant.rs
```diff
@@ -0,0 +1,424 @@
...
+            payload: GovernancePacket {
+                chain: Chain::Solana,
+                action: Action::ModifyBalance {
+                    sequence: 1,
+                    chain_id: 2,
+                    token_chain: 3,
+                    token_address: Address([0x32u8; 32]),
+                    kind: ModificationKind::Add,
+                    amount: Amount([0x31u8; 32]),
+                    reason: "".into(),
```
hendrikhofstadt, line 413: maybe add a reason to the test

cosmwasm/contracts/global-accountant/tests/submit_vaas.rs
```diff
@@ -181,14 +199,19 @@ fn bad_signature() {
...
+    assert_eq!(
+        "generic error: querier contract error: failed to recover verifying key",
+        err.root_cause().to_string().to_lowercase()
+    );
 }
 
 #[test]
 fn non_transfer_message() {
     let (wh, mut contract) = proper_instantiate();
+    register_emitters(&wh, &mut contract, 4);
```
hendrikhofstadt, line 214: why did this work without the registration / why did you add this?

cosmwasm/contracts/global-accountant/tests/chain_registration.rs
```diff
@@ -88,22 +88,30 @@ fn wrong_target() {
...
     body.emitter_chain = Chain::Fantom;
 
     let (_, data) = sign_vaa_body(&wh, body);
-    contract
+    let err = contract
         .submit_vaas(vec![data])
         .expect_err("successfully executed chain registration with non-governance chain");
+
+    // This tries to parse as a normal token bridge vaa and fails.  Not sure how best to differentiate from a governance message.
+    _ = err;
```
hendrikhofstadt, line 114: Why are you explicitly dropping the error and not asserting?

cosmwasm/contracts/global-accountant/tests/chain_registration.rs
```diff
@@ -114,9 +122,12 @@ fn non_governance_emitter() {
     body.emitter_address = Address([0x88; 32]);
 
     let (_, data) = sign_vaa_body(&wh, body);
-    contract
+    let err = contract
         .submit_vaas(vec![data])
         .expect_err("successfully executed chain registration with non-governance emitter");
+
+    // This tries to parse as a normal token bridge vaa and fails.  Not sure how best to differentiate from a governance message.
+    _ = err;
```
hendrikhofstadt, line 130: same here?

cosmwasm/contracts/global-accountant/tests/chain_registration.rs
```diff
@@ -88,22 +88,30 @@ fn wrong_target() {
...
+        err.root_cause().to_string().to_lowercase()
+    );
 }
 
 #[test]
 fn non_governance_chain() {
     let (wh, mut contract) = proper_instantiate();
 
     let mut body = create_vaa_body();
+    body.payload.chain = Chain::Solana;
```
hendrikhofstadt, line 105: why is this required? This tests the emitter chain check.

hendrikhofstadt (changes requested):
See inline comments

bruce-riley (dismissed):
I only reviewed the guardian changes.

node/cmd/guardiand/adminserver.go
```diff
@@ -148,7 +148,7 @@ func tokenBridgeRegisterChain(req *nodev1.BridgeRegisterChain, timestamp time.Ti
 
 // tokenBridgeModifyBalance converts a nodev1.TokenBridgeModifyBalance message to its canonical VAA representation.
 // Returns an error if the data is invalid.
-func tokenBridgeModifyBalance(req *nodev1.BridgeModifyBalance, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+func accountantModifyBalance(req *nodev1.AccountantModifyBalance, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
```
panoel, line 151: Please change comment on line 149 to reflect code changes.

cosmwasm/contracts/global-accountant/src/contract.rs
```diff
@@ -334,9 +334,21 @@ fn handle_vaa(
     let mut evt = if body.emitter_chain == Chain::Solana
         && body.emitter_address == wormhole::GOVERNANCE_EMITTER
     {
-        let govpacket = serde_wormhole::from_slice(body.payload)
-            .context("failed to parse governance packet")?;
-        handle_governance_vaa(deps.branch(), info, body.with_payload(govpacket))?
+        let mut module = [0u8; 32];
+        module.clone_from_slice(&body.payload.get()[..32]);
+
+        if let Ok(govpacket) = serde_wormhole::from_slice::<token::GovernancePacket>(body.payload) {
```
a5-pickle, line 340: This might be better written as `match` where the default (`_`) match results in `bail!`.

`if let` is a nice syntax when there aren't other conditions (either match something specific, otherwise do something else).

cosmwasm/contracts/global-accountant/src/contract.rs
```diff
@@ -378,7 +389,22 @@ fn handle_governance_vaa(
...
+}
+
+fn handle_accountant_moduleernance_vaa(
+    deps: DepsMut<WormholeQuery>,
+    info: &MessageInfo,
+    body: Body<accountant_module::GovernancePacket>,
+) -> anyhow::Result<Event> {
+    ensure!(
+        body.payload.chain == Chain::Any || body.payload.chain == Chain::Wormchain,
+        "this governance VAA is for another chain"
```
a5-pickle, line 403: It would be nice if this error message were distinct from `handle_token_governance_vaa` when it checks its governance target chain.

sdk/rust/core/src/token.rs
```diff
@@ -469,28 +369,6 @@ mod governance_packet_impl {
...
-                        token_chain,
-                        token_address,
-                        kind,
-                        amount,
-                        reason,
-                    }
-                }
                 v => {
                     return Err(Error::custom(format_args!(
                         "invalid value: {v}, expected one of 1, 2, 3"
```
a5-pickle, line 374: This needs to change to either "of 1, 2" or the error message should just be `"invalid value: {v}"`.

If the latter is chosen, we can change the same error in _accountant.rs_.

a5-pickle (changes requested):
Reviewed cosmwasm related changes and left comments.

cosmwasm/contracts/global-accountant/src/contract.rs
conorpp: match won't work here since we need to actually call `serde_wormhole::from_slice:` twice.  Once for `token::GovernancePacket` and once for `accountant::GovernancePacket`.

cosmwasm/contracts/global-accountant/tests/chain_registration.rs
conorpp: I think I added this debugging, just removed.

a5-pickle (dismissed):
Out of scope for this PR, but we may want to move more deserialized output checking to the Rust SDK as opposed to the contract.

Because the Rust SDK already checks for the module name in its deserialization of governance messages, we may want to enforce target chain checking in this deserialization, too. Otherwise we should have all the deserialization output checked in the contract.

cosmwasm/contracts/global-accountant/src/contract.rs
```diff
@@ -334,9 +334,15 @@ fn handle_vaa(
     let mut evt = if body.emitter_chain == Chain::Solana
         && body.emitter_address == wormhole::GOVERNANCE_EMITTER
     {
```
jynnantonix, line 336: If we want to still handle the errors from parsing, we could do something like:

```rust
    if body.payload.len() < 32 {
        bail!("governance module missing");
    }
    let module = &body.payload[..32];

    if module == TOKENBRIDGE_MODULE {
        let govpacket = serde_wormhole::from_slice(body.payload)
            .context("failed to parse tokenbridge governance packet")?;
        handle_token_governance_vaa(deps.branch(), body.with_payload(govpacket))
    } else if module == ACCOUNTANT_MODULE {
        let govpacket = serde_wormhole::from_slice(body.payload)
            .context("failed to parse accountant governance packet")?;
        handle_accountant_governance_vaa(deps.branch(), body.with_payload(govpacket))
    } else {
        bail!("unknown governance module")
    }
```

We could export the constants for the tokenbridge and accountant governance modules from the rust SDK.

sdk/rust/core/src/accountant.rs
```diff
@@ -0,0 +1,424 @@
...
+        where
+            D: Deserializer<'de>,
+        {
+            let arr = <[u8; 32]>::deserialize(deserializer)?;
+
+            if arr == MODULE {
+                Ok(Module)
+            } else {
+                Err(Error::custom(
+                    "invalid governance module, expected \"NFTBridge\"",
```
jynnantonix, line 126: nit: "NFTBridge" -> "GlobalAccountant"

jynnantonix (approved):

jynnantonix:
> Out of scope for this PR, but we may want to move more deserialized output checking to the Rust SDK as opposed to the contract.
> 
> Because the Rust SDK already checks for the module name in its deserialization of governance messages, we may want to enforce target chain checking in this deserialization, too. Otherwise we should have all the deserialization output checked in the contract.

I think there's a difference in the 2 types of checks.  The module name of governance messages is checked during deserialization for type-safety reasons.  Tokenbridge governance messages **always** have the tokenbridge governance module prefixed in the serialized payload and anything that has a different prefix is a completely different type and so should not be parsed as a tokenbridge governance action.  This is always true regardless of the context in which the message is deserialized.

On the other hand, most of the other checks depend on the context: if it's a contract on a specific chain we want to check that the target chain matches or if it's off-chain then we may not care about the exact value of the target chain at all (only that it is one of the supported chains).  These are application-safety checks, which is why they are not done during deserialization but afterwards.

Or to put it another way, the wormhole wire format is generally not self-describing except in the case of governance messages, which is why we take advantage of that to provide additional type safety that's not normally possible for other kinds of messages.

So while I'm not opposed to moving more checks into the SDK we should be careful to not mix up application-safety checks with type-safety checks.

hendrikhofstadt (approved):

bruce-riley (dismissed):

cosmwasm/contracts/global-accountant/src/contract.rs
conorpp: extra safe, will add it now.

hendrikhofstadt (approved):

jynnantonix (approved):

bruce-riley (approved):

PR #2517, opened 2023-03-13
Node/Watcher/EVM: Fix TimeOfBlockByHash

bruce-riley:
Discovered in the geth documentation that it is possible to use `HeaderByHash()` rather than `BlockByHash()` for `TimeOfBlockByHash()` since we don't need all the details. This only requires one RPC call instead of two. Also, `HeaderByHash()` works for all the chains that currently require a special version of `TimeOfBlockByHash()` because `BlockByHash()` fails.


panoel (approved):

evan-gray (changes requested):
If `NewConnectorWithGetTimeOfBlockOverride` is no longer used with this, can it be removed?

hendrikhofstadt (approved):
Nice simplification! LGTM + evan's comment

bruce-riley:
Deleted the obsolete file.

panoel (approved):

hendrikhofstadt (approved):

evan-gray (approved):

PR #2540, opened 2023-03-21
Node: Generate readiness labels programmatically

bruce-riley:
Right now the readiness syncing labels for all the chains are manually defined as strings. Every time a chain is added, the label needs to also be defined. Also, there is no easy way to get the readiness label associated with a chain, which will be needed for the IBC watcher. This PR changes the code to generate the label from the chain ID.

All of the labels exactly match the chain ID except for ethereum, where the label uses "eth" rather than "ethereum". I added a special case for that because I don't know if changing the readiness label will break existing monitoring.

Also, I created a test to make sure that all of the generated labels match the manually created ones. Once this change is merged, we can probably delete that test so we don't have to keep updating it as new chains are added.


node/cmd/guardiand/node.go
```diff
@@ -1331,34 +1349,37 @@ func runNode(cmd *cobra.Command, args []string) {
 		var solanaFinalizedWatcher *solana.SolanaWatcher
 		if shouldStart(solanaRPC) {
 			logger.Info("Starting Solana watcher")
-			readiness.RegisterComponent(common.ReadinessSolanaSyncing)
+			readinessSync := common.ChainIdToReadinessSyncing(vaa.ChainIDSolana)
+			readiness.RegisterComponent(readinessSync)
```
hendrikhofstadt, line 1353: It's safe to do `RegisterComponent` in ` ChainIdToReadinessSyncing` and remove the necessary registration everywhere. Register is currently *not* safe to be called multiple times but since we own the package that can be changed.

hendrikhofstadt (changes requested):
See inline comments

node/cmd/guardiand/node.go
```diff
@@ -1137,81 +1140,91 @@ func runNode(cmd *cobra.Command, args []string) {
...
 			if err := supervisor.Run(ctx, "avalanchewatch",
-				common.WrapWithScissors(evm.NewEthWatcher(*avalancheRPC, avalancheContractAddr, "avalanche", common.ReadinessAvalancheSyncing, vaa.ChainIDAvalanche, chainMsgC[vaa.ChainIDAvalanche], nil, chainObsvReqC[vaa.ChainIDAvalanche], *unsafeDevMode).Run, "avalanchewatch")); err != nil {
+				common.WrapWithScissors(evm.NewEthWatcher(*avalancheRPC, avalancheContractAddr, "avalanche", readinessSync, vaa.ChainIDAvalanche, chainMsgC[vaa.ChainIDAvalanche], nil, chainObsvReqC[vaa.ChainIDAvalanche], *unsafeDevMode).Run, "avalanchewatch")); err != nil {
 				return err
 			}
 		}
 		if shouldStart(oasisRPC) {
 			logger.Info("Starting Oasis watcher")
+			readinessSync := common.ChainIdToReadinessSyncing(vaa.ChainIDOasis)
+			readiness.RegisterComponent(readinessSync)
```
panoel, line 1154: Did oasis not register a component previously?  Or is it somewhere else?

node/pkg/common/readiness_test.go
```diff
@@ -0,0 +1,103 @@
...
+	ReadinessInjectiveSyncing readiness.Component = "injectiveSyncing"
+	ReadinessXplaSyncing      readiness.Component = "xplaSyncing"
+	ReadinessPythNetSyncing   readiness.Component = "pythnetSyncing"
+	ReadinessArbitrumSyncing  readiness.Component = "arbitrumSyncing"
+	ReadinessOptimismSyncing  readiness.Component = "optimismSyncing"
+	ReadinessBaseSyncing      readiness.Component = "baseSyncing"
+	ReadinessWormchainSyncing readiness.Component = "wormchainSyncing"
+)
+
+// This test is just to make sure that nothing go broken when we switched from manually specifying the readiness syncing labels.
```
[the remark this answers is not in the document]
bruce-riley: Good eye! Fixed.

node/cmd/guardiand/node.go
bruce-riley: It was not there before. I assumed it was an oversight. @evan-gray do you know if it was intentional that Oasis did not register for readiness?

node/cmd/guardiand/node.go
```diff
@@ -1251,37 +1266,40 @@ func runNode(cmd *cobra.Command, args []string) {
...
-				common.WrapWithScissors(cosmwasm.NewWatcher(*xplaWS, *xplaLCD, *xplaContract, chainMsgC[vaa.ChainIDXpla], chainObsvReqC[vaa.ChainIDXpla], common.ReadinessXplaSyncing, vaa.ChainIDXpla).Run, "xplawatch")); err != nil {
+				common.WrapWithScissors(cosmwasm.NewWatcher(*xplaWS, *xplaLCD, *xplaContract, chainMsgC[vaa.ChainIDXpla], chainObsvReqC[vaa.ChainIDXpla], readinessSync, vaa.ChainIDXpla).Run, "xplawatch")); err != nil {
 				return err
 			}
 		}
 
 		if shouldStart(algorandIndexerRPC) {
 			logger.Info("Starting Algorand watcher")
-			readiness.RegisterComponent(common.ReadinessAlgorandSyncing)
+			readiness.RegisterComponent(common.ChainIdToReadinessSyncing(vaa.ChainIDAlgorand))
```
panoel, line 1302: Should someone create an issue so that all watchers either get a readiness component passed in via the NewWatcher or creates a watcher component inside the watcher object?  Literally, half do it one way and the other half do it the other.

panoel (dismissed):

node/cmd/guardiand/node.go
bruce-riley: I'm changing all the watchers to get the readiness tag internally rather than having it passed in.

node/cmd/guardiand/node.go
bruce-riley: Added a separate method to generate the readiness component and register it.

hendrikhofstadt (dismissed):
Generally looks good to me. However is there a reason we're explicitly registering components? Couldn't we just auto-register it the first time someone calls ChainIDToReadinessSyncing?

node/pkg/common/readiness.go
```diff
@@ -1,33 +1,41 @@
...
+
+// RegisterReadinessSyncing registers the specified chain for readiness syncing. It will panic if the chain ID is invalid
+// so it should only be used during initialization.
+func RegisterReadinessSyncing(chainID vaa.ChainID) {
+	readiness.RegisterComponent(ChainIdToReadinessSyncing(chainID))
+}
+
+// ChainIdToReadinessSyncing maps a chain ID to a readiness syncing value. It will panic if the chain ID is invalid
+// so it should only be used during initialization. Otherwise use ChainIdToReadinessSyncingWithError.
+func ChainIdToReadinessSyncing(chainID vaa.ChainID) readiness.Component {
```
hendrikhofstadt, line 22: The generall naming convention is "Must<FunctionName>" and the error returning method is the default

node/pkg/common/readiness.go
```diff
@@ -1,33 +1,41 @@
...
-	ReadinessArbitrumSyncing  readiness.Component = "arbitrumSyncing"
-	ReadinessOptimismSyncing  readiness.Component = "optimismSyncing"
-	ReadinessBaseSyncing      readiness.Component = "baseSyncing"
-	ReadinessWormchainSyncing readiness.Component = "wormchainSyncing"
+	ReadinessEthSyncing readiness.Component = "ethSyncing"
 )
+
+// RegisterReadinessSyncing registers the specified chain for readiness syncing. It will panic if the chain ID is invalid
+// so it should only be used during initialization.
+func RegisterReadinessSyncing(chainID vaa.ChainID) {
```
hendrikhofstadt, line 16: See comment on review, we could do this automatically to make usage of the API more intuitive.

node/pkg/common/readiness.go
bruce-riley: Renamed the functions.

node/pkg/common/readiness.go
bruce-riley: I'm not crazy about auto registering (especially since `readiness.RegisterComponent` currently panics in that case.

I made `MustRegisterReadinessSyncing` both register and return the tag, and am calling it from watcher start up, so there is no redundancy.

hendrikhofstadt (approved):

panoel (approved):

evan-gray (approved):
Gave it a once over, concept looks good. 🙂

PR #2542, opened 2023-03-22
Enhancing the SDK repairVAA functionality for multipleChain

wasif28:
Enhancing the SDK repairVAA functionality for multipleChain out of the box to support all chains for MAINNET, TESTNET, DEVNET


sdk/js/src/utils/repairVaa.ts
```diff
@@ -17,14 +15,19 @@ interface GuardianSetData {
 }
 
 export async function getCurrentGuardianSet(
-  provider: ethers.providers.JsonRpcProvider
+  provider: ethers.providers.JsonRpcProvider,
+  chainId: ChainId,
```
kev1n-peters, line 19: I think we should add support for non-EVM chains or at least set the type here to `EVMChainId`.

wasif28:
@kev1n-peters  have done the suggested changes kindly give it a glance

evan-gray (changes requested):
@wasif28 can you provide some background as to why this function needs to provide support for other chains or environments? Wormhole currently uses Ethereum as it's source of truth for guardian sets, so my understanding was this function was meant to match that (as it's what the guardian uses).

Furthermore, there is only one guardian set in devnet and testnet, so what is the rationale for adding those to this function?

evan-gray:
Closing per my comment above. Feel free to open a discussion regarding your use case.

PR #2561, opened 2023-03-30
node: add benchmarking capability

tbjump:
Goals
* refactor the Guardian Node into its own package such that it can be tested through golang unit-tests.

Non-goals
* Any kind of refactoring beyond the stated goal is not in-scope. The interfaces of the various components could probably be further cleaned up and the handling of the configuration flags could probably be improved (e.g. with `viper`). 

High-level overview
* The `node` package is added with `type GuardianNode`. This allows creating multiple `GuardianNode` in the unit tests. 
* Configuration of `GuardianNode` is implemented through currying. This helps to separate the *specification* of the configuration from the *application* of the configuration. 
    * The configuration of the Guardian previously took 710 lines (https://github.com/wormhole-foundation/wormhole/blob/main/node/cmd/guardiand/node.go#L783-L1493). It is now reduced to 501 lines which are much more clear and less error-prone. 
    * Watchers are now configured through the `WatcherConfig` interface, which I implemented for each watcher. This can probably be cleaned up further, but is not in scope of this PR. 

Functional changes
* *All* watchers will be run with `common.WrapWithScissors`. Before, the following ran without: Wormchain, Aptos, Sui.
* A few log messages might've been changed
* Other than that, there should be no functional changes in this PR!

### DIFF
--- a/node/cmd/guardiand/adminverify.go
+++ b/node/cmd/guardiand/adminverify.go
@@ -1,20 +1,15 @@
 package guardiand
 
 import (
-	"encoding/hex"
-	"fmt"
 	"log"
 	"os"
-	"time"
 
-	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+	"github.com/certusone/wormhole/node/pkg/adminrpc"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/spf13/cobra"
 	"google.golang.org/protobuf/encoding/prototext"
 
 	nodev1 "github.com/certusone/wormhole/node/pkg/proto/node/v1"
-	"github.com/status-im/keycard-go/hexutils"
 )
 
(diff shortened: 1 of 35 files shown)


evan-gray (changes requested):
Just seeing this. I feel this should be broken up into smaller, more reviewable PRs. How about starting with just one that changes the env from int to common.Environment and building a stack from there. This also has conflicts and has sat out for a while. Maybe writing this plan up as an issue, getting feedback, and iterating would be a more productive approach.

tbjump:
Hey, 
thanks for the feedback, but this draft PR was mostly to share an idea with another contributor and not intended to be merged, definitely not in its current state.

tbjump:
Closing in favor of https://github.com/wormhole-foundation/wormhole/pull/3024 which is broken up into graphite stacks.

PR #2573, opened 2023-04-03
Node/Gov: Split up CoinGecko queries

bruce-riley:
As the number of tokens monitored by the governor increases (to something over 500), the CoinGecko query will eventually fail with a "rate limit exceeded" error. This PR breaks the queries up into chunks of 200 tokens, with a one second sleep between each query.

### DIFF
--- a/node/pkg/governor/governor_prices.go
+++ b/node/pkg/governor/governor_prices.go
@@ -26,12 +26,25 @@ import (
 // The CoinGecko API is documented here: https://www.coingecko.com/en/api/documentation
 // An example of the query to be generated: https://api.coingecko.com/api/v3/simple/price?ids=gemma-extending-tech,bitcoin,weth&vs_currencies=usd
 
+// coinGeckoQueryIntervalInMins specifies how often we query CoinGecko for prices.
 const coinGeckoQueryIntervalInMins = 15
 
+// tokensPerCoinGeckoQuery specifies how many tokens will be in each CoinGecko query. The token list will be broken up into chunks of this size.
+const tokensPerCoinGeckoQuery = 200
+
+// initCoinGecko builds the set of CoinGecko queries that will be used to update prices. It also starts a go routine to periodically do the queries.
 func (gov *ChainGovernor) initCoinGecko(ctx context.Context, run bool) error {
+	queryIdx := 0
+	tokenIdx := 0
 	ids := ""
 	first := true
 	for coinGeckoId := range gov.tokensByCoinGeckoId {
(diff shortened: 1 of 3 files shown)


node/pkg/governor/governor_prices.go
```diff
@@ -81,50 +104,25 @@ func (gov *ChainGovernor) PriceQuery(ctx context.Context) error {
...
-	defer func() {
-		err = response.Body.Close()
+	result := make(map[string]interface{})
+	for queryIdx, query := range gov.coinGeckoQueries {
+		thisResult, err := gov.queryCoinGeckoChunk(query)
 		if err != nil {
-			gov.logger.Error("cgov: failed to close coin gecko query")
-			// We can't safely call revertAllPrices() here because we don't know if we hold the lock or not.
-			// Also, we don't need to because the prices have already been updated / reverted by this point.
+			gov.logger.Error("cgov: coin gecko query failed", zap.Int("queryIdx", queryIdx), zap.String("query", query), zap.Error(err))
```
panoel, line 116: Do you want to keep "cgov:" in the log messages?

panoel (approved):

node/pkg/governor/governor_prices.go
bruce-riley: Fixed "coin gecko".

The change from "cgov: " to using a logger component will be done with a separate PR.

panoel (approved):

node/pkg/governor/governor_prices.go
```diff
@@ -26,42 +26,82 @@ import (
...
+
+	// Create the set of queries, breaking the IDs into the appropriate size chunks.
+	gov.coinGeckoQueries = createCoinGeckoQueries(ids, tokensPerCoinGeckoQuery)
+	for queryIdx, query := range gov.coinGeckoQueries {
+		gov.logger.Info("cgov: coingecko query: ", zap.Int("queryIdx", queryIdx), zap.String("query", query))
+	}
+
+	if len(gov.coinGeckoQueries) == 0 {
+		gov.logger.Info("cgov: did not find any tokens, nothing to do!")
+		return nil
```
claudijd, line 51: Should this hard error when we cannot get prices or is this ok with this because there is a floor price to fall back on?

claudijd (approved):

node/pkg/governor/governor_prices.go
bruce-riley: This is the way it has been since the beginning. I think we were okay with it because of the floor prices. If CG ever becomes unavailable, we may want to be able to continue. But I can make it exit if you prefer. Let me know.

PR #2574, opened 2023-04-03
Node/Acct: Logging changes

bruce-riley:
This PR makes the following changes to make it easier to sift through the accountant logs:

- Changes a couple of errors that are not really errors to warnings.
- Changes the "acct:", "acctaudit:" and "acctwatcher:" log prefixs to add a "g" on the front so we can search for "gacct" rather than "acct" (which gives lots of messages not related to accountant.

### DIFF
--- a/node/pkg/accountant/accountant.go
+++ b/node/pkg/accountant/accountant.go
@@ -127,7 +127,7 @@ func NewAccountant(
 
 // Run initializes the accountant and starts the watcher runnable.
 func (acct *Accountant) Start(ctx context.Context) error {
-	acct.logger.Debug("acct: entering Start", zap.Bool("enforceFlag", acct.enforceFlag))
+	acct.logger.Debug("gacct: entering Start", zap.Bool("enforceFlag", acct.enforceFlag))
 	acct.pendingTransfersLock.Lock()
 	defer acct.pendingTransfersLock.Unlock()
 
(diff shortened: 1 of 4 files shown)


evan-gray (changes requested):
should we just change all the logs with `acct:`? There's some in `node/cmd/guardiand` and `node/hack/accountant` as well

panoel (approved):

hendrikhofstadt (changes requested):
Have we considered using a logger field for module? I think we're kind of adding an implicit formatting standard here that is not enforced. Would probably make sense to do logger.With() and store a properly tagged logger in accountant.

panoel (dismissed):

evan-gray (changes requested):
Had questions about two files not captured:
https://github.com/wormhole-foundation/wormhole/blob/58f86e17a5529fd14c8b7ffda8fa38c230c1fe83/node/pkg/db/accountant.go#L68
I think this is now passed in the logger, so it doesn't need the `acct:`?

https://github.com/wormhole-foundation/wormhole/blob/58f86e17a5529fd14c8b7ffda8fa38c230c1fe83/node/pkg/processor/processor.go#L199

This one is errors, not logs so I'm not sure if it's relevant 🤷

panoel (dismissed):

hendrikhofstadt (dismissed):

bruce-riley:
@evan-gray pointed out I don't need the `.With` in `AcctGetData` since it gets passed the accountant logger. @hendrikhofstadt , @panoel , that is the only change since your last approval (node/pkg/db/accountant.go).

evan-gray (approved):

hendrikhofstadt (approved):

kev1n-peters (approved):

panoel (approved):

PR #2592, opened 2023-04-06
sui: fix upgrade cap check

a5-pickle:
The existing `UpgradeCap` check would fail when we try to set up the contract after publishing. The reason is `0x0` is the published package address (which originates from _Move.toml_ when a contract is built prior to publishing).

Added a generic method to assert that an `UpgradeCap` belongs to `T` object's package.


heyitaki (approved):
Amazing :D

PR #2591, opened 2023-04-06
Ibc cosmwasm contracts

nik-suri:
#2198 #2199 

This PR adds the two new cosmwasm contract for IBC generic message emission: `wormchain-ibc-receiver` and `wormhole-ibc`. These two contracts implement a custom IBC protocol. The `wormhole-ibc` contract should be deployed to IBC-enabled chains, and the various `wormhole-ibc` contract should connect via IBC to the `wormchain-ibc-receiver` contract, which will live on Wormchain.

For existing Cosmos deployments, the `wormhole` cosmwasm contract should be migrated to the `wormhole-ibc` cosmwasm contract after it is uploaded to the chain.

PR #2593, opened 2023-04-06
Tilt devnet deployment for ibc generic messaging

nik-suri:


cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -48,18 +47,6 @@ if (missing_artifacts.length) {
   process.exit(1);
 }
 
-const unexpected_artifacts = actual_artifacts.filter(
```
kcsongor, line 51: why delete this code?

cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -15,14 +15,13 @@ import { zeroPad } from "ethers/lib/utils.js";
   deterministic.
 */
 const artifacts = [
-  "wormhole.wasm",
+  "wormhole_ibc.wasm",
```
kcsongor, line 18: did the core contract get renamed from wormhole to wormhole_ibc? if so, why? if not, don't we need to deploy the actual core contract?

cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -251,7 +240,19 @@ for (const [contract, registrations] of Object.entries(
         memo: "",
       })
       .then((tx) => terra.tx.broadcast(tx))
-      .then((rs) => console.log(rs));
+      .then((rs) => console.log(rs))
+      .catch((error) => {
```
kcsongor, line 244: we're catching the error and logging, but not rethrowing - this means the exit code will be success even when deployment fails, right? shouldn't we exit with an error in that case?

cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -251,7 +240,19 @@ for (const [contract, registrations] of Object.entries(
...
+        if (error.response) {
+          // Request made and server responded
+          console.log(error.response.data, error.response.status, error.response.headers);
```
kcsongor, line 247: console.log prints to stdout, since the messages are errors, we should log to stdere with console.error

kcsongor (changes requested):

cosmwasm/deployment/terra2/tools/deploy.js
nik-suri: No, see the IBC cosmwasm contracts PR: https://github.com/wormhole-foundation/wormhole/pull/2591

`wormole_ibc` is a new contract that we will be deploying to IBC enabled chains such as terra2. It is meant to replace the core bridge contract. For the devnet deployment, we can just replace the contract entirely. For the production migration on existing Cosmos deployments, we will first upload the `wormhole-ibc` code, and then migrate the `wormole` contract to use the `wormhole-ibc` code.

cosmwasm/deployment/terra2/tools/deploy.js
nik-suri: I removed `global_accountant` from the artifacts list - there's no need to deploy it on terra2 devnet. Unless I remove this code, this script will always throw an error since it'll find the `global_account` in the artifacts folder but not in the artifacts list.

cosmwasm/deployment/terra2/tools/deploy.js
nik-suri: sure, makes sense to exit with an error. Since the token bridge and the contract registrations are effectively needed to run on devnet, there's no reason to consider it the deployment a success when this call fails.

cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -116,6 +103,8 @@ const govChain = 1;
 const govAddress =
   "0000000000000000000000000000000000000000000000000000000000000004";
 
+const wormchainIbcReceiverAddress = "0000000000000000000000000000000000000000000000000000000000000000";
```
hendrikhofstadt, line 106: unused?

hendrikhofstadt (approved):

cosmwasm/deployment/terra2/tools/deploy.js
evan-gray: Could the IBC version be deployed _in addition to_ the existing core wormhole contract. Then the IBC specific testing could be done alongside the existing tests. Making this strictly additional and not a replacement.

evan-gray (commented):
I had anticipated testing the IBC route in addition to the existing core code. Perhaps it is worth considering setting the second, IBC route as a different chain id in devnet. Open to discussion, just concerned about losing a test path.

cosmwasm/deployment/terra2/tools/deploy.js
```diff
@@ -154,8 +141,13 @@ if (!init_guardians || init_guardians.length === 0) {
   throw "failed to get initial guardians from .env file.";
 }
 
+<<<<<<< HEAD
```
bruce-riley, line 144: Merge conflicts.

bruce-riley (changes requested):

cosmwasm/deployment/terra2/tools/deploy.js
evan-gray: Fixed this up

bruce-riley (dismissed):

evan-gray (approved):
Lots of churn to get tilt to pass but you did it! 🙌

conorpp (approved):

kcsongor (approved):

PR #2602, opened 2023-04-08
clients/js: refactor worm start-validator command

AlberErre:
This PR extracts `worm start-validator <chain>` command into chain-specific commands.

As a result, the following commands are created ✅:

```
worm aptos start-validator

worm evm start-validator
```

And the `worm start-validator <chain>` has been deleted ❌

This change was discussed on #2598, following a `worm <chain> <sub-commands>` pattern.

Documentation has been updated accordingly 📝

PR #2603, opened 2023-04-09
clients/js: add worm info namespace

### LINKED ISSUE #2598: clients/js: refactor worm so that chain name is top-level
Currently, commands in `worm` CLI are split by either chain or by functionality (eg `worm sui deploy` and `worm start-validator aptos`). This is confusing because functionality for any given chain may live in different top-level commands. We may want to consider refactoring the CLI such that all commands follow the format `worm <chain> <sub-commands>`

AlberErre:
🚨 Note: This PR is a follow up from #2602 

Resolves #2598 

This PR adds a new namespace `worm info`, which includes commands showing static data of `chain-id`, `contract addresses` and `rpc endpoints`.

Therefore the following commands are available ✅:

```
worm info contract <network> <chain> <module>
worm info rpc <network> <chain>
worm info chain-id <chain>
```

As a result of this new namespace, the old root commands has been deleted ❌:
```
worm contract <network> <chain> <module>
worm rpc <network> <chain>
worm chain-id <chain>
```

This change also originates from https://github.com/wormhole-foundation/wormhole/issues/2598, where was discussed as a follow up from the `worm <chain> <sub-commands>` pattern.

Documentation has been updated to cover this new `worm info` namespace 📝, as well as missing commands.


heyitaki (dismissed):

AlberErre:
Thank you @kcsongor and @heyitaki 🏄‍♂️!

By debugging where the CI pipelines failed, I notice `worm contract`, `worm rpc` and `worm chain-id` are used extensively 🕵️‍♂️

I just updated those **scripts** files to use the new `worm info` namespace ✅ to avoid breaking changes 🙏🏼 

This way we would avoid maintaining two different versions of the same commands

AlberErre:
Also, in case there are more external dependencies using these commands. I will just rollback these changes. 

Let me know as you have more context on this 🙏🏻 @kcsongor @heyitaki

AlberErre:
Totally, I'm going to take a look to aliases. 

In case that becomes a dead-end, we could extract this functionality into functions and use them in both commands. 

That way we could avoid duplication 👍

AlberErre:
Hurray! 🎉 there is a way to export/import command logic as [Command Modules](https://github.com/yargs/yargs/blob/main/docs/advanced.md#providing-a-command-module).

I just re-implement the `worm info` namespace by referencing the root commands directly ✅

There is an alternative way of doing this by using [Command Aliases](https://github.com/yargs/yargs/blob/main/docs/advanced.md#command-aliases), but unfortunately it does not cover this use case, as these aliases only works on the same command hierarchy level (it doesn't work with child/sub-commands)

@kcsongor @heyitaki could you take a final review on this? 🙏🏼

kcsongor (approved):
awesome!

heyitaki (approved):
very nice~ 

as far as I can see, this branch was checked out from the one in #2602, so closing that PR and squashing this one to merge it in seems cleanest to me

AlberErre:
thanks! 🎉  @kcsongor @heyitaki 

I don't have permissions to merge it, but looks like ready to go 🚀

Agree about closing the other one in favour of this one ✅

PR #2610, opened 2023-04-10
Batch deployment configuration for `RelayProvider`

scnale:
This PR adds interfaces for batch updates of several state attributes in the `RelayProvider` and a function that allows setting the entire state (minus irrelevant internal structures like the `initialized` mapping for upgrades) in one call.

This reduces the typical deployment of a `RelayProvider` instance in a given chain to just four transactions: three for the deployment of contracts (setup, implementation and proxy) and one for the configuration of the `RelayProvider` itself.

It also removes a few redundant configuration transactions for the `CoreRelayer` itself.


derpy-duck (approved):
Looks good to me, as long as it is tested somehow (which I assume it is tested in Tilt)

Quick thing: readContractStatus.sh and tiltMessageTest.sh show up in this PR as empty files. Don't know if this is a github bug or not

scnale:
>Quick thing: readContractStatus.sh and tiltMessageTest.sh show up in this PR as empty files. Don't know if this is a github bug or not

Ah no, only their diffs should be empty. The scripts should be intact otherwise. I set the execute flags on all the scripts in `ethereum/ts-scripts/relayer/shell/`

PR #2616, opened 2023-04-11
node/pkg/watchers: algorand fix

panoel:
We have found that the 
```
defer close(errC)
```
line has been causing panics.
Also, cleanup reading the errC channel so that only one go function does the read.

### DIFF
--- a/node/pkg/watchers/algorand/watcher.go
+++ b/node/pkg/watchers/algorand/watcher.go
@@ -186,7 +186,6 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 	// Create an error channel
 	errC := make(chan error)
-	defer close(errC)
 
 	// Signal that basic initialization is complete
 	readiness.SetReady(e.readinessSync)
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):

node/pkg/watchers/algorand/watcher.go
```diff
@@ -186,7 +186,6 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 	// Create an error channel
 	errC := make(chan error)
-	defer close(errC)
```
kcsongor, line 189: 🤔 I feel like not closing the channel is introducing a code smell. Perhaps deleting this line just masks the bug instead of fixing it?

kcsongor: what was the panic?

panoel:
The original commit was reverted.

kcsongor:
@panoel what was the original commit?

PR #2618, opened 2023-04-11
ethereum/bridge: enforce vm.emitterAddress != 0

tbjump:
Wormhole should never produce messages with vm.emitterAddress == 0, but this is an additional sanity check. No need to do an on-chain contract upgrade for now, but could be picked up in the next release.


tbjump:
@gator-boi if this change looks good to you, could you please also apply it to https://github.com/wormhole-foundation/wormhole-scaffolding/blob/main/evm/src/01_hello_world/HelloWorld.sol#L151?

PR #2629, opened 2023-04-12
cosmwasm: clean up unused dependencies

nik-suri:
Seems like a lot of dependencies have been copy/pasted — would be good to stop this practice in the future so we avoid any unnecessary compatibility issues, which makes it harder for others to import these libs and build on top of wormhole.

### DIFF
--- a/cosmwasm/Cargo.lock
+++ b/cosmwasm/Cargo.lock
@@ -93,16 +93,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b645a089122eccb6111b4f81cbc1a49f5900ac4666bb93ac027feaecf15607bf"
 
 [[package]]
-name = "bigint"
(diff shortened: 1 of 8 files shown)


jynnantonix (approved):
Thanks for doing this!  Unfortunately, there's no built-in linter to warn about unused deps so we have to periodically clean things up ourselves.

kcsongor (approved):

nik-suri:
@jynnantonix @kcsongor could we integrate `cargo udeps` into the ci pipeline?

jynnantonix:
> @jynnantonix @kcsongor could we integrate `cargo udeps` into the ci pipeline?

It looks like `cargo udeps` requires a nightly toolchain to run.  Adding a whole second toolchain to the CI pipeline just for this seems like a bit much but I don't feel that strongly about it either way.

PR #2630, opened 2023-04-12
cosmwasm: bump k256 dependency to 0.11

nik-suri:
See https://github.com/RustCrypto/elliptic-curves/blob/master/k256/CHANGELOG.md for changes to the `k256` library from `0.9.4` to `0.11`.

Most notable are:
1. The function renaming of `recover_verify_key_from_digest_bytes => recover_verifying_key_from_digest_bytes`
2. Using the `sec1` crate, which changes the `EncodedPoint` type. The `EncodedPoint` type no longer has a `decompress` method, so we use the affine point conversion: https://github.com/RustCrypto/elliptic-curves/pull/435/files

### DIFF
--- a/cosmwasm/Cargo.lock
+++ b/cosmwasm/Cargo.lock
@@ -194,25 +194,19 @@ checksum = "591ff76ca0691bd91c1b0b5b987e5cf93b21ec810ad96665c5a569c60846dd93"
 
 [[package]]
 name = "const-oid"
-version = "0.6.2"
+version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


jynnantonix (approved):
Code change looks good to me but may be good to have someone more familiar with the crypto algorithm to verify the implementation change.

kcsongor (approved):
the code changes check out, I like that the decoding is now more explicit about the affine representation - the failing test is safe to update to the new output (it's an expected failure, which now fails with a slightly more precise error message)

a5-pickle (approved):

kcsongor (approved):

PR #2628, opened 2023-04-12
sdk/js: add attestFromSui

heyitaki:
This PR adds attest functionality for the WH Sui SDK

PR #2632, opened 2023-04-13
sdk/js: add createWrappedOnSui

heyitaki:
This PR adds the ability to create a wrapped asset via the WH Sui SDK. This functionality is necessarily broken into two steps, since we must call `prepare_registration` and `complete_registration` and the latter takes in an object that is transferred to us as a result of a successful execution of the former.

EDIT: SDK now simply uses pre-built bytecode and uses the following steps to test that the bytecode is accurate to the latest version of the `Coin` contract.

1. Create a temporary directory out of which the wrapped asset will be deployed
2. Recursively copy the `wormhole` and `token_bridge` dependencies to a sibling directory
3. Set up the `Move.toml` files of these dependencies and substitute on-chain addresses
4. Publish the coin
5. Search the signer's owned objects for the WrappedAssetSetup object
6. Call `complete_registration` with the object

I'm sure this can be made more resilient/elegant but it's a good start.


heyitaki:
* sui/integration_v2
  * **PR #2628** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2628" target="_blank">Graphite</a> 
    * **PR #2632** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2632" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2632?utm_source=stack-comment).

sdk/js/src/sui/publish.ts
```diff
@@ -0,0 +1,257 @@
...
+  const defaultTomlPath = getDefaultTomlPath(packagePath);
+  const tempTomlPath = getTempTomlPath(packagePath);
+  if (fs.existsSync(tempTomlPath)) {
+    fs.renameSync(tempTomlPath, defaultTomlPath);
+  }
+};
+
+const cloneDependencies = async (): Promise<void> => {
+  const { tree, sha: latestSha } = await fetchWormholeTree(
+    "sui/integration_v2"
```
a5-pickle, line 61: Can we make this a const at the top of the file so it's clear that we'll need to change this?

It might be best if `fetchWormholeTree` took an optional argument for the ref, which will default to `main`. And for the time being, we are explicitly setting it to a const value (e.g. `TEMPORARY_SUI_BRANCH`) using that const value here.

Either that or a TODO comment would be helpful.

sdk/js/src/sui/template/coin.move
```diff
@@ -0,0 +1,28 @@
...
+            create_wrapped::prepare_registration(
+                witness,
+                VAA,
+                ctx
+            ),
+            tx_context::sender(ctx)
+        );
+    }
+
+    public fun token_meta(): AssetMeta {
```
a5-pickle, line 23: We may not want this here unless it'll be used to test something (which I didn't find in this PR).

sdk/js/src/token_bridge/__tests__/sui-integration.ts
```diff
@@ -69,38 +67,75 @@ afterAll(async () => {
...
-    assertIsNotNullOrUndefined(coin8TreasuryCapObjectId);
+    // Complete create wrapped on Sui
+    const publishEvents = getPublishedObjectChanges(
+      suiPrepareRegistrationTxRes
+    );
+    expect(publishEvents.length).toBe(1);
 
-    const suiAttestTxPayload = await attestFromSui(
+    const coinPackageId = publishEvents[0].packageId;
+    const coinType = `${coinPackageId}::coin::COIN`;
```
a5-pickle, line 124: Should there be a utility method to create a wrapped coin type? For example, someone calls `getWrappedCoinType(coinPackageId)` so an integrator doesn't have to think about this too much.

If we feel that generating this wrapped coin type is only relevant internally for the SDK, maybe we just pass the `coinPackageId` to `createWrappedOnSui` instead of the coin type.

We may want other utility methods to generate types, too. Not sure where these would go.

sdk/js/src/token_bridge/createWrapped.ts
```diff
@@ -164,3 +166,59 @@ export function createWrappedOnAptos(
...
+  );
+}
+
+export async function createWrappedOnSui(
+  provider: JsonRpcProvider,
+  tokenBridgeAddress: string,
+  coreBridgeStateObjectId: string,
+  tokenBridgeStateObjectId: string,
+  signerAddress: string,
+  coinType: string
```
a5-pickle, line 192: Made a comment in _sui-integration.ts_ about this, but we may want to pass in the coin package ID and generate the coin type within this method.

a5-pickle (changes requested):
Looks good so far. I left a few comments.

sdk/js/src/sui/publish.ts
heyitaki: Hoisted the const and added TODO comment, didn't add the `main` default param since we will require a code change to remove the usage of that const anyway.

sdk/js/src/sui/template/Move.toml
```diff
@@ -0,0 +1,19 @@
+[package]
```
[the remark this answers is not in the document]
heyitaki: Ack

sdk/js/src/sui/template/coin.move
heyitaki: Removed

sdk/js/src/token_bridge/__tests__/sui-integration.ts
heyitaki: good thinking, I think the documentation for such a method would be especially helpful for integrators

sdk/js/src/token_bridge/createWrapped.ts
heyitaki: Hmm I think type would be just as easy to pass in because they could get it from the `WrappedAssetSetup` type vs getting the bare `packageId` from the `Coin` publish event, but I don't have strong opinions, so I'll make the change

a5-pickle (approved):

evan-gray (changes requested):
Until this code can be run from a browser, I suggest this not live in the sdk. Perhaps the CLI is more appropriate. I haven't reviewed the exact implementation but there's likely more to consider.

bruce-riley (approved):
Only reviewed go code.

sdk/js/src/bridge/parseSequenceFromLog.ts
```diff
@@ -180,3 +181,13 @@ export function parseSequenceFromLogAptos(
 
   return null;
 }
+
+export function parseSequenceFromLogSui(
+  coreBridgePackageId: string,
```
evan-gray, line 186: just want to throw out that this will need to be the _original_ package id.

evan-gray (dismissed):

evan-gray:
a better step forward

sui/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/wormhole-foundation/sui:0.32.0@sha256:a07bc5fccb81c56f27d38aa4422a16aafe71949f0dd8ed791908b68f22badab4 as sui
+FROM ghcr.io/wormhole-foundation/sui:0.31.0_2@sha256:64c47edd2ab6e6146c07e8835bb9e6d56bb64b4023903e9ff00205cafbeb0fc8 as sui
```
evan-gray, line 1: you sure?

evan-gray (approved):

PR #2649, opened 2023-04-14
Node: RunWithScissors should not hang on error if no listener

bruce-riley:
The existing RunWithScissors will hang if it tries to write to the error channel if there is no listener. This PR changes that. It also adds tests and metrics to allow us to verify the behavior.

### DIFF
--- a/node/pkg/common/scissors_test.go
+++ b/node/pkg/common/scissors_test.go
@@ -5,9 +5,22 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
 )
+
+func getCounterValue(metric *prometheus.CounterVec, runnableName string) float64 {
+	var m = &dto.Metric{}
+	if err := metric.WithLabelValues(runnableName).Write(m); err != nil {
+		return 0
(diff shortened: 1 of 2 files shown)


panoel (approved):

node/pkg/common/scissors_test.go
```diff
@@ -78,3 +91,101 @@ func TestSupervisor(t *testing.T) {
...
+	errC := make(chan error)
+
+	itRan := false
+	RunWithScissors(ctx, errC, "TestRunWithScissorsCleanExit", func(ctx context.Context) error {
+		itRan = true
+		return nil
+	})
+
+	time.Sleep(100 * time.Millisecond)
+	require.Equal(t, true, itRan)
```
hendrikhofstadt, line 106: waiting on a channel would be cleaner than sleeping a fixed amount of time

hendrikhofstadt (approved):

node/pkg/common/scissors_test.go
bruce-riley: I actually thought about waiting on a channel rather than timing. The problem is that the channel could wake up as soon as the go routine posts to it, which could be before the scissor code completes, and in this case it is important to let the scissors code complete to confirm it doesn't hang.

hendrikhofstadt: That makes sense. We should be good then.

panoel (approved):

hendrikhofstadt (approved):

PR #2652, opened 2023-04-14
Node: Add tests for invalid messages

bruce-riley:
Add tests for the code that checks for invalid chain / zero address in messages. Move the code into a separate function to facilitate testing.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -931,16 +931,7 @@ func runNode(cmd *cobra.Command, args []string) {
 						)
 					}
 					if shouldPublish {
-						select {
-						case msgWriteC <- msg:
-						default:
-							logger.Error("Failed to post message to processor, dropping it",
-								zap.Stringer("tx", msg.TxHash),
-								zap.Stringer("emitter_chain", msg.EmitterChain),
-								zap.Stringer("emitter_address", msg.EmitterAddress),
-								zap.Uint64("sequence", msg.Sequence),
-							)
-						}
+						msgWriteC <- msg
 					}
 				}
 			}
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1620,14 +1620,13 @@ func makeChannelPair[T any](cap int) (<-chan T, chan<- T) {
 	return out, out
 }
 
-var zeroAddress vaa.Address
-
 func validateMessage(logger *zap.Logger, msg *common.MessagePublication, chainId vaa.ChainID) (bool, error) {
 	if msg.EmitterChain != chainId {
 		// SECURITY: This should never happen. If it does, a watcher has been compromised.
 		return false, fmt.Errorf("received observation from a chain that was not marked as originating from that chain")
 	}
 
+	var zeroAddress vaa.Address
 	if msg.EmitterAddress == zeroAddress {
 		// SECURITY: This could happen based on external input. We want to block it but not restart the guardian.
 		logger.Error("SECURITY ERROR: received observation with EmitterAddress == 0x00",
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -921,7 +921,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				case msg := <-c:
 					shouldPublish, err := validateMessage(logger, msg, chainId)
 					if err != nil {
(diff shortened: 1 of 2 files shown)


bruce-riley:
@hendrikhofstadt, I added code so that posting a message to the process won't block. But now I'm not so sure that's such a good idea. Do you think the watcher should hang until the processor can accept the observation, or should the message just get dropped (like I have it in this PR)?

node/cmd/guardiand/validate_msg_test.go
```diff
@@ -0,0 +1,81 @@
...
+		Sequence:         789101112131415,
+		EmitterChain:     vaa.ChainIDEthereum,
+		EmitterAddress:   emitterAddress,
+		Payload:          []byte("Hello"),
+		ConsistencyLevel: 32,
+	}
+
+	shouldPublish, err := validateMessage(logger, msg, vaa.ChainIDEthereum)
+	require.NoError(t, err)
+	assert.Equal(t, true, shouldPublish)
```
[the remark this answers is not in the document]
hendrikhofstadt: Also, there is a shortcut require.True

node/cmd/guardiand/validate_msg_test.go
```diff
@@ -0,0 +1,81 @@
...
+		Sequence:         789101112131415,
+		EmitterChain:     vaa.ChainIDEthereum,
+		EmitterAddress:   emitterAddress,
+		Payload:          []byte("Hello"),
+		ConsistencyLevel: 32,
+	}
+
+	shouldPublish, err := validateMessage(logger, msg, vaa.ChainIDBSC)
+
+	assert.Error(t, err)
```
hendrikhofstadt, line 59: require, see above

node/cmd/guardiand/node.go
```diff
@@ -1573,3 +1565,26 @@ func makeChannelPair[T any](cap int) (<-chan T, chan<- T) {
 	out := make(chan T, cap)
 	return out, out
 }
+
+var zeroAddress vaa.Address
```
panoel, line 1569: Why not put this inside the function validateMessage()?

node/cmd/guardiand/validate_msg_test.go
```diff
@@ -0,0 +1,81 @@
...
+		ConsistencyLevel: 32,
+	}
+
+	shouldPublish, err := validateMessage(logger, msg, vaa.ChainIDEthereum)
+	assert.NoError(t, err)
+	assert.True(t, shouldPublish)
+}
+
+func TestValidateMessageWrongChainShouldReturnError(t *testing.T) {
+	logger, _ := zap.NewDevelopment()
```
panoel, line 41: Why is this one different than the other 2 tests?

panoel (approved):

node/cmd/guardiand/node.go
bruce-riley: It's at least some overhead to set it to zero. Don't think it hurts to have it global.

hendrikhofstadt (dismissed):

panoel (approved):

node/cmd/guardiand/node.go
```diff
@@ -914,31 +914,23 @@ func runNode(cmd *cobra.Command, args []string) {
...
-			zeroAddress := vaa.Address{}
 			for {
 				select {
 				case <-rootCtx.Done():
 					return
 				case msg := <-c:
-					if msg.EmitterChain != chainId {
-						// SECURITY: This should never happen. If it does, a watcher has been compromised.
-						logger.Fatal("SECURITY CRITICAL: Received observation from a chain that was not marked as originating from that chain",
+					shouldPublish, err := validateMessage(logger, msg, chainId)
```
tbjump, line 922: I think the code becomes less readable this way, because now there are three different paths in which errors are handled 
1. by setting shouldPublish to false
2. by returning an error
3. by logging an error within validateMessage itself

tbjump (changes requested):
I appreciate the spirit of adding unit tests for security checks like this one. But I think it'd be better to instead do the test on a higher level, i.e. inject a wrong message into a watcher channel and assert that it won't get published. That way we'd check not only that validateMessage is working correctly, but also that it is being used in the right place. 

I am working on a PR to refactor the node.go that will allow you to create a guardian within a unit test and then simulate a watcher (https://github.com/wormhole-foundation/wormhole/pull/2561). I'd recommend we wait until that PR is merged and then add these tests on top there.

PR #2653, opened 2023-04-14
cosmwasm: token bridge: add new CompleteTransferResponse struct as da…

nik-suri:
…ta to complete_transfer executions


cosmwasm/contracts/token-bridge/src/msg.rs
```diff
@@ -117,3 +117,16 @@ pub struct IsVaaRedeemedResponse {
...
+pub struct CompleteTransferResponse {
+    // contract address if this minted or unlocked a cw20, otherwise none
+    pub contract: Option<String>,
+    // denom if this unlocked a native token, otherwise none
+    pub denom: Option<String>,
+    pub recipient: String,
+    pub amount: Uint128,
+    pub relayer: String,
+    pub fee: Uint128,
+}
```
jynnantonix, line 132: Since the point of this change is to improve the developer experience, I recommend keeping the actual types in the struct rather than using strings and leave the encoding / decoding to the `to_binary` and `from_binary` methods.

Otherwise the users will have to convert the strings back into the native types before doing anything with them.

If we're going to keep these as strings then we need to clarify in the comments how things like addresses are encoded.  Are they hex, base64, bech32, something else?

nik-suri: The actual types for `recipient`, `relayer`, `contract`, and `denom` are all `String`.

The `relayer_address` is a `&String` and the `recipient` is a `Addr(String)`. Do you mean to leave these as `&String` and `Addr`?

kcsongor (approved):

cosmwasm/contracts/token-bridge/src/msg.rs
jynnantonix: No, that's fine.  If the native types are strings as well then I have no issue.

jynnantonix (approved):

PR #2658, opened 2023-04-15
sdk/js: bump injective sdk

evan-gray:
I encountered issues installing the sdk on a fresh Ubuntu image. This patch release of the injective sdk fixes a `postinstall` script issue.

https://github.com/InjectiveLabs/injective-ts/commit/adb7c764ad00a0cfa38223cecf9b221873cd31b8

Furthermore, in the interim versions, Injective now provides esm builds. This along with marking of `"sideEffects":false` should allow for drastically reduced builds for consumers of the Wormhole sdk. See [Webpack Tree Shaking](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) for more details.

To increase tree shaking potential, moving the rest of the chain-specific functions to their own files (instead of grouping them by function) could be considered.

### DIFF
--- a/sdk/js/CHANGELOG.md
+++ b/sdk/js/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.9.13
+
+### Added
+
+"sideEffects": false
+
+### Changed
+
+injective dependencies updated 
+
 ## 0.9.12
 
 ### Added
--- a/sdk/js/package-lock.json
+++ b/sdk/js/package-lock.json
@@ -1,20 +1,20 @@
 {
   "name": "@certusone/wormhole-sdk",
-  "version": "0.9.12",
+  "version": "0.9.13",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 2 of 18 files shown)


panoel (approved):

panoel:
Are we going towards a pattern of putting all sdk functions for a chain in its own file instead of spreading them among function oriented files?

evan-gray:
> Are we going towards a pattern of putting all sdk functions for a chain in its own file instead of spreading them among function oriented files?

I think we should consider this for the potential tree-shaking gains (grouping them by dependency and limiting the dependencies required). It's certainly more difficult to analyze the dependency usage when all the files import the same dependencies because they span all the chains.

Additionally, it might be nice to make types / interfaces for the exports to implement that way we can ensure each implementation maintains a standard naming convention, but this might not be so possible with the difference in arguments and such 🤔 would at least likely be a breaking change. Probably better saved for a larger re-write.

kev1n-peters (approved):

PR #2660, opened 2023-04-17
Sei testnet support

bruce-riley:
This PR adds testnet support for Sei using the IBC watcher.

### DIFF
--- a/node/pkg/watchers/ibc/watcher.go
+++ b/node/pkg/watchers/ibc/watcher.go
@@ -48,7 +48,7 @@ type (
 
 var (
 	// Chains defines the list of chains to be monitored by IBC. Add new chains here as necessary.
-	Chains = []vaa.ChainID{}
+	Chains = []vaa.ChainID{vaa.ChainIDSei}
 
 	// Features is the feature string to be published in the gossip heartbeat messages. It will include all chains that are actually enabled on IBC.
 	Features = ""


node/cmd/guardiand/adminserver.go
```diff
@@ -879,6 +879,7 @@ func (s *nodePrivilegedService) DumpRPCs(ctx context.Context, req *nodev1.DumpRP
 	rpcMap["polygonRPC"] = *polygonRPC
 	rpcMap["pythnetRPC"] = *pythnetRPC
 	rpcMap["pythnetWS"] = *pythnetWS
+	rpcMap["sei"] = "IBC"
```
panoel, line 882: Will this be the case for all IBC chains?

panoel (approved):

node/cmd/guardiand/adminserver.go
evan-gray: Should this be the wormchain WS?

bruce-riley: For other chains that are configured for full nodes, I guess it would have that RPC. Sei is only supported over IBC, so that's why it says "IBC". The IBC WS is listed above. I thought it would be nice for it to stand out as something special (IBC) rather than just another RPC.

But we can make it whatever people want.

evan-gray: Ah, I see now that I looked harder 😆

evan-gray (approved):

hendrikhofstadt (approved):

PR #2734, opened 2023-04-18
sui: add coin type mapping using token info as key

a5-pickle:
## Objective

Add a convenient mapping in the `TokenRegistry` that maps chain ID and address to a specific coin type as string. This mapping is helpful for someone interacting with the contract via RPC where a coin type needs to be specified when interacting with the Token Bridge. It is especially helpful when completing a transfer, where the coin type is not encoded the token transfer VAA (only the canonical token info).

## How to Review

1. Make sure Move tests still run via `make test`.
2. Review the new member `coin_type` and how it is used when adding native and wrapped assets.
3. Review tests that validate `coin_type` entries.


gator-boi (approved):

gator-boi:
Looks great!

kev1n-peters (approved):

PR #2739, opened 2023-04-19
cosmwasm: remove terraswap dependency

nik-suri:
I think it makes sense to remove the `terraswap` dependency from the `cw_token_bridge` for a couple reasons:

1. It’s an extra dependency that cosmwasm integrators need to pull in to build token bridge `ExecuteMsg` types.
2. The type is very simple and the dependency has functionality specific to `terra`, so it shouldn't be included in the general cosmwasm contract.

These `Asset` and `AssetInfo` types are generally quite useful in cosmwasm, and should ideally be pulled out of terraswap and committed upstream to one of the cosmwasm std libs. However, that is a longer process, so defining them in the token bridge is a good solution for the interim.

### DIFF
--- a/cosmwasm/Cargo.lock
+++ b/cosmwasm/Cargo.lock
@@ -1232,15 +1232,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "protobuf"
-version = "2.28.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


jynnantonix (dismissed):
Code change seems fine but we should have a test to ensure that the serialization / deserialization works the same way before and after the change.

jynnantonix (approved):

kcsongor (approved):

PR #2740, opened 2023-04-19
Update docker push workflow

JoeHowarth:
Allow pushing the relayer engine docker image to ghcr so that relay providers can pull it automatically


derpy-duck (approved):
LGTM

PR #2747, opened 2023-04-19
sui: lower rpc processing latency

### LINKED ISSUE #2670: sui: lower `signAndExecuteTransactionBlock` RPC processing latency
## Description and context
While interacting with the public testnet RPC, we have encountered 504 (timeout) errors, but only using the Typescript SDK. Somehow when we use the `sui` CLI tool (specifically the `client` command) to execute, we get a successful transaction. **NOTE: This behavior does not occur when we interact with a local validator.**

We want to explore using [WaitForEffectsCert] when executing `signAndExecuteTransactionBlock`, where we would have to poll using `getTransaction` until we receive a valid transaction. This may help alleviate the timeout issue.

## Definition of done
At least in `worm` CLI, modify all Sui transaction execution using `signAndExecuteTransactionBlock` to use the above option and add polling. Test this change by trying to execute the `init-wormhole` command successfully.

[waitforeffectscert]: https://github.com/MystenLabs/sui/blob/ee8a72527084f8e8196631654711fae739ee1307/sdk/typescript/src/signers/signer-with-provider.ts#L127

AlberErre:
Resolves #2670 

This PR solves 504 errors (Gateway Time-out) while using `filter` params on API method `suix_getOwnedObjects`

A new method `findOwnedObjectByType` has  been added to directly download all `OwnedObjects` and navigate them until finding our target. 

This implementation uses `cursor` to search in every response page. You can check the API docs [here](https://docs.sui.io/sui-jsonrpc#suix_getOwnedObjects).

Further context is available [here]( https://github.com/wormhole-foundation/wormhole/issues/2670#issuecomment-1518415668) 👀

### DIFF
--- a/clients/js/sui/publish.ts
+++ b/clients/js/sui/publish.ts
@@ -10,7 +10,6 @@ import fs from "fs";
 import { resolve } from "path";
 import { Network } from "../utils";
 import { MoveToml } from "./MoveToml";
-import { pollTransactionForEffectsCert } from "./utils";
 
 export const publishPackage = async (
   signer: RawSigner,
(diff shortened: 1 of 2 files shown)


heyitaki (approved):
Approving, this PR looks good to me and @gator-boi has tested it successfully locally. Thanks a lot for this contribution @AlberErre, it is much appreciated 🙏

PR #2753, opened 2023-04-20
sui: fix vaa redemption and other things

a5-pickle:
## Objective

- Fix VAA redemption by removing encoded VAA buffers as arguments in favor of passing `VAA` by value
- Fix `complete_transfer_with_payload` to adopt the same technique used for `VAA` redemption by separating redeeming this transfer into two separate steps
- Fix vulnerability in `TokenRegistry`
- Add `UpgradeCap` warehousing for creating wrapped assets (which are derived from untrusted coin types)
- Tweak upgrade logic to remove gotchas
- Remove `entry` methods (since `public` methods can be used in transaction blocks)

## How to Review PR
1. Make sure Move tests run via `make test`.
2. Review each commit sequentially.

**Note: When this PR is approved and ready to merge, please do not squash commits.**


sui/token_bridge/sources/state.move
```diff
@@ -134,14 +134,6 @@ module token_bridge::state {
         self: &mut State,
         implementation_digest: Bytes32
     ): UpgradeTicket {
```
kcsongor, line 136: since it's not obvious that the version check is performed by the runtime, we should include a comment here that states that explicitly, to save future readers a minor heart attack

sui/wormhole/sources/governance_message.move
```diff
@@ -281,12 +246,13 @@ module wormhole::governance_message_tests {
         let worm_state = take_state(scenario);
         let the_clock = take_clock(scenario);
 
-        let msg =
-            governance_message::parse_and_verify_vaa_test_only(
+        let parsed =
```
kcsongor, line 249: nit: the name `parsed` is misleading, because these `VAA`s are not only parsed, but also (sig)verified

sui/token_bridge/sources/complete_transfer_with_payload.move
```diff
@@ -1,7 +1,31 @@
...
-/// This module implements the method `complete_transfer_with_payload` which
-/// allows a contract to redeem a Token Bridge transfer with arbitrary payload.
+/// This module implements two methods: `authorize_transfer` and `redeem_coin`,
+/// which are to be executed in a transaction block in this order.
+///
+/// `authorize_transfer` allows a contract to complete a Token Bridge transfer
+/// with arbitrary payload. This deserialized `TransferWithPayload` with the
+/// bridged balance and source chain ID are packaged in a `RedeemerTicket`.
+///
+/// `redeem_coin` unpacks the `RedeemerTicket` and checks whether the speciifed
```
kcsongor, line 10:
```suggestion
/// `redeem_coin` unpacks the `RedeemerTicket` and checks whether the specified
```

sui/token_bridge/sources/complete_transfer_with_payload.move
```diff
@@ -1,7 +1,31 @@
 // SPDX-License-Identifier: Apache 2
 
-/// This module implements the method `complete_transfer_with_payload` which
-/// allows a contract to redeem a Token Bridge transfer with arbitrary payload.
+/// This module implements two methods: `authorize_transfer` and `redeem_coin`,
```
kcsongor, line 3: smart!

sui/token_bridge/sources/state.move
```diff
@@ -150,34 +150,42 @@ module token_bridge::state {
...
         //
         // A separate method is required because `state` is a dependency of
         // `migrate`. This method warehouses state modifications required
         // for the new implementation plus enabling any methods required to be
         // gated by the current implementation version. In most cases `migrate`
-        // is a no-op. But it still must be called in order to reset the
-        // migration control to `false`.
+        // is a no-op.
+        //
+        // The only case where this would fail is if `migrate` were not called
```
kcsongor, line 186: I would be careful here. `migrate` must be called in a separate transaction, which means the `upgrade` + `migrate` flow is not atomic, which opens up the possibility that `upgrade` succeeds, but `migrate` fails for some reason. An example of a failure that would not come up in local testing is if there's a loop that iterates over some state that with mainnet data exceeds the maximum transaction budget. If the `migrate` call fails, then further upgrades become impossible. We need to be able to recover from such scenarios

kcsongor (changes requested):

sui/token_bridge/sources/state.move
a5-pickle: Added a note just above `authorize_upgrade` and `commit_upgrade` in both Wormhole and Token Bridge. How does it look?

sui/token_bridge/sources/state.move
a5-pickle: It wasn't clear that I changed `migrate` to be optional, so I amended the commit where I added the `MigrateTicket`: https://github.com/wormhole-foundation/wormhole/pull/2753/commits/0997424c11afef6a5e53b74f4118e5cb865e37c7.

sui/token_bridge/sources/vaa.move
```diff
@@ -11,17 +11,26 @@
...
 
     use token_bridge::state::{Self, State};
     use token_bridge::version_control::{Vaa as VaaControl};
 
-    // All friends need `parse_and_verify`.
     friend token_bridge::create_wrapped;
     friend token_bridge::complete_transfer;
     friend token_bridge::complete_transfer_with_payload;
 
+    struct TokenBridgeMessage {
```
kcsongor, line 26: could you document the assumptions about this type here? so they're easy to review + ensure that future changes don't violate them

kcsongor (dismissed):
Nice! 👍

sui/token_bridge/sources/create_wrapped.move
```diff
@@ -114,9 +115,9 @@ module token_bridge::create_wrapped {
     /// `WrappedAsset`).
```
gator-boi, line 115: UpgradeCap was added, I think this comment is stale.

sui/token_bridge/sources/vaa.move
```diff
@@ -34,62 +42,96 @@ module token_bridge::vaa {
     /// NOTE: This method has `friend` visibility so it is only callable by this
```
gator-boi, line 42: Stale comment? This method is no longer a `friend` method.

gator-boi (approved):

evan-gray (approved):
Approving to merge into integration only

PR #2756, opened 2023-04-20
gRelayer: Redelivery test & comment fixes

chase-45:


derpy-duck (dismissed):
Looks good, the only note is I assume we'll probably want three different error codes for clarity rather than combining them all into InvalidOverride

derpy-duck:
Actually quick note: We don't want to revert in the case of an invalid override, we actually want to emit an invalidredelivery event right? Because this is something that could happen due to user error?

ethereum/contracts/relayer/coreRelayer/CoreRelayerMessages.sol
```diff
@@ -344,9 +344,6 @@ contract CoreRelayerMessages is CoreRelayerGetters {
         bool roundUp,
         IRelayProvider provider
     ) internal view returns (uint256 targetAmount) {
-        if (!provider.isChainSupported(targetChain)) {
```
JoeHowarth, line 347: Are we making the isChainSupported check implicit now? Basically if the src and dst native currency price !== 0 implies it is supported?

JoeHowarth (approved):
approved w/ comment ^

ethereum/contracts/relayer/coreRelayer/CoreRelayerMessages.sol
derpy-duck: Oh no, I just moved it to the beginning of 'send/forward/resend'

chase-45:
> Actually quick note: We don't want to revert in the case of an invalid override, we actually want to emit an invalidredelivery event right? Because this is something that could happen due to user error?

I think a revert is more appropriate here, as the InvalidRedelivery can be calculated off-chain just as easily as on chain. Even if the InvalidRedelivery event were posted on-chain, it would not be immediately clear if the Relayer or Integrator was at fault, and we would also want to log an invalid delivery where the VAAs were not properly submitted. Overall I think it's more pragmatic to have invalid deliveries simply not be submitted, as all the error cases can't be detected on-chain.

PR #2764, opened 2023-04-21
Client/js: Add new commands

bruce-riley:
This PR adds the following commands to the worm client. Please see the changes to README.md for details.


clients/js/README.md
```diff
@@ -31,6 +36,7 @@ Commands:
   worm near                                       NEAR utilities
   worm submit <vaa>                               Execute a VAA
   worm update                                     Update this tool by rebuilding it
+  worm verify-vaa --vaa <vaa> --network <network> Verifies a VAA by querying the core contract on Ethereum
```
panoel, line 39: How hard is it to write a generic version for all supported chains?

bruce-riley: This command is meant to verify the VAA rather than the chain. It just does the verification on Ethereum because that is (probably) the easiest. If the VAA is valid on Ethereum, it should be valid throughout the entire wormhole network. I don't think there is any need or advantage to doing this on multiple chains.

bruce-riley:
This PR is being broken up into smaller ones.

PR #2780, opened 2023-04-24
Node: Don't lock memory in testnet

bruce-riley:


node/cmd/guardiand/node.go
```diff
@@ -415,7 +415,12 @@ func runNode(cmd *cobra.Command, args []string) {
 		fmt.Print(devwarning)
 	}
 
-	common.LockMemory()
+	if *testnetMode {
+		fmt.Println("This is testnet mode, not locking in memory.")
+	} else {
+		common.LockMemory()
+	}
+
```
panoel, line 423: Should it be locked or unlocked in devnet?

panoel (approved):

node/cmd/guardiand/node.go
scnale: Just want to chime in a bit here: I've personally been commenting this out when using devnet so I would be in favor of that. In fact, here's what I was using for this particular fragment of code:

```diff
-   common.LockMemory()
+   if !*unsafeDevMode {
+       common.LockMemory()
+   }
```

I'm not sure if this makes sense for the use case you were trying to address here :thinking:. They're probably separate.

bruce-riley: Changed it for devnet as well.

panoel (approved):

evan-gray (approved):

PR #2786, opened 2023-04-25
sui: harden migrating to new build

a5-pickle:
## Objective

Here we are again. As we have been reviewing the consequences of upgrading contracts, we have been valuating the execution of version control. The existing upgradability design allowed for too much room for error when configuring hard-coded versions via `u64` (to be in line with the `UpgradeCap`).

This change aims to dissociate version control from how `UpgradeCap` tracks its package versions by establishing specific types that denote build versions. When a package is upgraded, we enforce the implementor to add a new version struct in _version_control.move_ (with release notes justifying the new struct).

Calling `migrate` is now a requirement to establish the upgraded build as the only functioning build for both Wormhole and Token Bridge.

Given the new paradigm of passing package types to limit the blast radius of upgrading contracts, we can now enforce that the package called on via transaction block is the latest package (as opposed to tracking minimum required versions).

### How to Review PR

_Carefully..._

- Make sure Move tests run via `make test`.
- Review _version_control.move_ in both Wormhole and Token Bridge **EXTREMELY CAREFULLY**.
- Review _version_control.move_ tests.
- Review `migrate` modules (plus tests).
- Review tests that enforce breaking changes when version control updates its version.

### DIFF
--- a/sui/wormhole/sources/governance_message.move
+++ b/sui/wormhole/sources/governance_message.move
@@ -7,7 +7,9 @@
 module wormhole::governance_message {
     use wormhole::bytes::{Self};
     use wormhole::bytes32::{Self, Bytes32};
+    use wormhole::consumed_vaas::{Self, ConsumedVAAs};
     use wormhole::cursor::{Self};
+    use wormhole::external_address::{ExternalAddress};
     use wormhole::state::{Self, State, chain_id};
     use wormhole::vaa::{Self, VAA};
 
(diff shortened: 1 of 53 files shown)


sui/wormhole/sources/governance_message.move
```diff
@@ -182,6 +175,9 @@ module wormhole::governance_message {
     /// Aborts if the VAA is not governance (i.e. sent from the governance
     /// emitter on the governance chain)
     fun assert_governance_emitter(wormhole_state: &State, verified_vaa: &VAA) {
+        // This state capability ensures that the current build version is used.
+        state::assert_current(wormhole_state);
```
kcsongor, line 179: why doesn't this function just take a `StateCap`? Currently, when I read `verify_vaa` it's not immediately obvious to me that it's version protected, since that check is performed in `assert_governance_emitter` internally.

Proposal: let's remove `state::assert_current` altogether, and just propagate `StateCap`s. The `StateCap` should in turn be created at the "top-level" (i.e. entry point). That way, it will be easier to see that functions like `verify_vaa` are version protected.

sui/wormhole/sources/utils/package_utils.move
```diff
@@ -38,4 +53,45 @@ module wormhole::package_utils {
...
+        let _: Old = field::remove(id, CurrentVersion {});
+
+        let new_type = type_name::get<New>();
+        // Make sure the new type does not equal the old type, which means there
+        // is no protection against either build.
+        assert!(new_type != type_name::get<Old>(), E_SAME_VERSION);
+
+        // Also make sure `New` originates from this module.
+        let module_name = into_bytes(type_name::get_module(&new_type));
+        assert!(module_name == b"version_control", E_TYPE_NOT_ALLOWED);
```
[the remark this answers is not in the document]
a5-pickle: The purpose of this is to make sure the types are defined in a module called "version_control" so they're always defined in the same location.

We can add more checks to this function, though. But I'd like to keep this here so it forces code modifiers to keep everything in one spot.

sui/wormhole/sources/governance_message.move
```diff
@@ -31,203 +30,167 @@ module wormhole::governance_message {
...
-
-        // Finally deserialize Wormhole payload as governance message.
-        let cur = cursor::new(vaa::take_payload(verified_vaa));
-        let module_name = bytes32::take_bytes(&mut cur);
-        let action = bytes::take_u8(&mut cur);
-        let chain = bytes::take_u16_be(&mut cur);
-        let payload = cursor::take_rest(cur);
-
-        GovernanceMessage { module_name, action, chain, payload, vaa_hash }
+    public fun authorize_verify_local(
```
kcsongor, line 61: this function needs access control, currently it allows bypassing the governance check entirely I think

kcsongor (changes requested):

sui/wormhole/sources/governance_message.move
kcsongor: pushed a fix, @a5-pickle please mark this as resolved if you agree with it

gator-boi (dismissed):

gator-boi (approved):

PR #2788, opened 2023-04-25
G relayer governance changes

derpy-duck:


proto/node/v1/node.proto
```diff
@@ -93,6 +93,9 @@ message GovernanceMessage {
     WormchainStoreCode wormchain_store_code = 14;
     WormchainInstantiateContract wormchain_instantiate_contract = 15;
     WormchainMigrateContract wormchain_migrate_contract = 16;
+
+    // Wormhole Relayer module
+    WormholeRelayerSetDefaultRelayProvider wormhole_relayer_set_default_relay_provider = 18;
```
bruce-riley, line 98: Going to have problems when you merge because 18-20 are taken by Circle Integration.

sdk/vaa/payloads.go
```diff
@@ -31,6 +39,9 @@ var (
 	ActionRegisterChain      GovernanceAction = 1
 	ActionUpgradeTokenBridge GovernanceAction = 2
 	ActionModifyBalance      GovernanceAction = 3
+
+	// Wormhole relayer governance actions
+	WormholeRelayerSetDefaultRelayProvider GovernanceAction = 3
```
bruce-riley, line 44: Is there a reason you are using 3 rather than starting at 1?

sdk/vaa/payloads.go
```diff
@@ -170,6 +187,33 @@ func (r BodyWormchainMigrateContract) Serialize() []byte {
 	return serializeBridgeGovernanceVaa(WasmdModuleStr, ActionMigrateContract, ChainIDWormchain, r.MigrationParamsHash[:])
 }
 
+<<<<<<< HEAD
```
bruce-riley, line 190: Merge problem.

proto/node/v1/node.proto
derpy-duck: Got it, will use 21

sdk/vaa/payloads.go
derpy-duck: Yeah; the actions for the Generic Relayer will be registering chain, upgrading contract, and setting the default relay provider

The idea is to reuse the 'ActionRegisterChain' and 'ActionUpgradeTokenBridge' governance actions (just as NFTbridge does), but then also add an additional one; to distinguish it from those two it has to be '3'.

bruce-riley (approved):
Did not review solidity code.

PR #2798, opened 2023-04-28
relayer: forge fmt

JoeHowarth:


derpy-duck (approved):
Looks good, could you do the test files as well if its easy? (No worries if you think that should be seperate)

derpy-duck (approved):
Looks great, thanks!

PR #2804, opened 2023-04-29
js: initialize workspace

evan-gray:
This PR moves the `sdk/js` and `clients/js` packages to the workspace. This should enable faster iteration of co-dependent sdk + client changes. Reviewers, please spot check the versions in the package-lock.json match those from the removed package-lock.jsons. The workspace and lock were generated with npm 9.6.4 using npm init -w <folder> for each of the folders.

This also removes any local filesystem interaction with tilt. The experience is now entirely containerized.

> Note: After rebasing on the clients/js restructure, the clients build is larger. This could be due to an inadvertent dependency change or something more nefarious. It's possible there's another, more efficient approach to this change.


evan-gray:
Sui now fails in tilt maybe because it doesn't have a compatible node version compared to what the client was built with? :shrug:

evan-gray:
Closing for now, as at least we got the config containerization in https://github.com/wormhole-foundation/wormhole/pull/2886

PR #2807, opened 2023-04-29
sdk/js: added sui testnet addrs

kev1n-peters:


heyitaki (changes requested):
Do you mind updating clients/js as well? https://github.com/wormhole-foundation/wormhole/blob/sui/sdk-testnet-addrs/clients/js/consts.ts

heyitaki (approved):

PR #2812, opened 2023-04-30
sui - add VAA sequence number to `DecreeReceipt`

optke3:
## Summary
See https://github.com/wormhole-foundation/wormhole/pull/2811.


a5-pickle:
This will break what we are using in testnet. #2811 will be what we will review.

PR #2813, opened 2023-04-30
sdk/js: fix getDynamicFieldObject response handling

kev1n-peters:


evan-gray (approved):
pro changes. love that we're getting some distance out of `unnormalizeSuiAddress` 😬

PR #2818, opened 2023-05-01
Relayer/resend improvements

chase-45:


derpy-duck (approved):
Looks good to me. Out of curiosity, do you have a formatter on for typescript? lmk; I probably should use the same one

chase-45:
> Looks good to me. Out of curiosity, do you have a formatter on for typescript? lmk; I probably should use the same one

I believe I'm using the standard prettier config

derpy-duck (approved):

PR #2821, opened 2023-05-01
node/pkg/watcher: sui refactor plus tests

panoel:
This refactors runWithScissors() to call functions instead of running the function inline.
Also, adds a test file that can be run with **go run**
The diff in vscode is much nicer than the one here.

### DIFF
--- a/node/pkg/watchers/sui/watcher.go
+++ b/node/pkg/watchers/sui/watcher.go
@@ -223,12 +223,12 @@ func (w *Watcher) inspectBody(body SuiResult) error {
 
 	seq, err := strconv.ParseUint(*fields.Sequence, 10, 64)
 	if err != nil {
-		w.logger.Info("Sequence decode error", zap.String("Sequence", *fields.Sequence))
+		w.logger.Error("Sequence decode error", zap.String("Sequence", *fields.Sequence))
 		return err
 	}
 	ts, err := strconv.ParseInt(*fields.Timestamp, 10, 64)
 	if err != nil {
-		w.logger.Info("Timestamp decode error", zap.String("Timestamp", *fields.Timestamp))
+		w.logger.Error("Timestamp decode error", zap.String("Timestamp", *fields.Timestamp))
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


node/pkg/watchers/sui/watcher_test.go
```diff
@@ -0,0 +1,123 @@
...
+	fmt.Println("inspectBody had an error", err)
+	assert.Contains(tst, err.Error(), "Transaction hash is not 32 bytes")
+
+	hashBase := "This needs to be a 32 byte hash."
+	txd = base58.Encode([]byte(hashBase))
+	err = watcher.inspectBody(res)
+	require.NoError(tst, err)
+	require.Equal(tst, 1, len(msgChan))
+	obsv := <-msgChan
+	fmt.Println("Received msgChan msg...")
```
bruce-riley, line 108: Don't really need all these print statements since assert will tell you the line number of an error.

bruce-riley (dismissed):

node/pkg/watchers/sui/watcher_test.go
panoel: Fixed

node/pkg/watchers/sui/watcher.go
```diff
@@ -222,12 +223,12 @@ func (e *Watcher) inspectBody(logger *zap.Logger, body SuiResult) error {
 
 	seq, err := strconv.ParseUint(*fields.Sequence, 10, 64)
 	if err != nil {
-		logger.Info("Sequence decode error", zap.String("Sequence", *fields.Sequence))
+		w.logger.Info("Sequence decode error", zap.String("Sequence", *fields.Sequence))
```
bruce-riley, line 226: Should "decode errors" be info logs, or error logs?

bruce-riley (approved):

node/pkg/watchers/sui/watcher.go
panoel: Changed to Error

panoel:
The SUI watcher ha been running without incident for quite some time.  This refactor is not necessary at this time.  It will be reevaluated should we need to address issues in the Sui watcher in the future.

PR #2824, opened 2023-05-02
sdk/js: sui get latest package from CurrentPackage

kev1n-peters:


evan-gray:
pulled into #2830 2830

PR #2834, opened 2023-05-02
relayer: faster tilt builds - move 'make' to docker build

JoeHowarth:


derpy-duck (approved):
Yess thank you I would love faster tilt iteration

PR #2838, opened 2023-05-02
Client/js: Add cmds to edit and verify VAAs

bruce-riley:
This PR adds two new commands to the worm client:

- `worm edit-vaa` - Allows user to edit / create / sign a VAA.
- `worm verify-vaa` - Verifies a VAA by calling `parseAndVerifyVM` on the core contract on Ethereum.

### DIFF
--- a/clients/js/README.md
+++ b/clients/js/README.md
@@ -32,7 +32,7 @@ Commands:
   worm near                                       NEAR utilities
   worm submit <vaa>                               Execute a VAA
   worm update                                     Update this tool by rebuilding it
-  worm verify-vaa --vaa <vaa> --network <network> Verifies a VAA by querying the core contract on Ethereum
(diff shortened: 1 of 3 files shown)


clients/js/README.md
```diff
@@ -17,6 +17,7 @@ private keys, based on `.env.sample` in this folder.
 worm [command]
 
 Commands:
+  worm edit-vaa                                   Allows editing / creating / signing VAAs
```
panoel, line 20: Nit: since you call out that verify-vaa only works on Ethereum, should you call it out here, too?

bruce-riley: It's not that these commands "only work on Ethereum". They are chain independent. They just access the core contract on Ethereum to get the guardian set and verify signatures. 

I will remove the reference to Ethereum from verify-vaa.

panoel (approved):

PR #2841, opened 2023-05-03
scripts/scope checkers: reduce false negatives

tbjump:
This change reduces false negatives in the scope checking scripts by
* supporting spaces and other special characters in file paths
* using jq instead of a simple grep to parse the json
* being more restrictive in the regexes on what to allow to pass through

This addresses all review comments in the previously abandoned PR https://github.com/wormhole-foundation/wormhole/pull/1774


SEJeff (approved):
Using null termination is the way to do this correctly. You're on 🔥 recently @tbjump. Awesome!

kcsongor (approved):

PR #2854, opened 2023-05-04
sdk/js: Added optional Sui package ID parameters to functions

kev1n-peters:
The package IDs can be passed in to avoid fetching them via RPC

### DIFF
--- a/sdk/js/package-lock.json
+++ b/sdk/js/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@certusone/wormhole-sdk",
-  "version": "0.9.14",
+  "version": "0.9.16",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@certusone/wormhole-sdk",
-      "version": "0.9.14",
+      "version": "0.9.16",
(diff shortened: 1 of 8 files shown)


heyitaki:
Why don't we just cache this in getPackageId the first time we call it for a certain value?

kev1n-peters:
> 

IMO the "low-level" SDK should be stateless and the integrator should cache the values on their end

heyitaki:
While I agree that that is generally a good abstraction, I think in cases such as this, dev UX should be prioritized. Given that the UX isn't actually changing too much here (since the added parameters are optional), I'm fine pushing this change through.

Separately, I'm not sure I understand the addition of the `hasInstallScript` flag. From the npm docs: 
```
hasInstallScript: A flag to indicate that the package has a preinstall, install, or postinstall script.
```
but I don't see and pre- or post-install commands in `package.json`. What do you think?

evan-gray (approved):

heyitaki (approved):
May make sense to abstract this functionality into a util since its used so often

evan-gray (approved):

PR #2856, opened 2023-05-04
sdk: Go Fuzz example experiment

claudijd:
This is a PR to demonstrate a bit about how Go Fuzz could be used to test various parts of the guardiand code.

I can't say these example cases are particularly compelling, but shows an example of how they could be built out and how there is incremental value in finding corner cases.

I didn't do it in this example, but these could be things we could run with say N second thresholds in builds to act as complementary test coverage for regression tests.  It's really hard to see the future of some of the guardiand interfaces, but the hope would be that if we had the fuzz tests running lightly in CI to catch low hanging fruit, we could additionally have more heavy fuzz runs (unbounded by time) run independently by security contributors and the security research community.

I'm opening this as a PR, but really trying to frame this more of a discussion on whether folks are interested in seeing more examples like this being produced to further harden the primatives by with guardiand depends on.

To try this out on the bench, you can run a localized fuzz test of one of the examples like so:

`cd sdk`
`go test -v -fuzz FuzzStringToAddress -run FuzzStringToAddress ./vaa/`

It will produce an unbounded fuzz execution of the fuzzer and premably, if you throw more compute at a thing, and the fuzz test itself is continuing to better explore new execution space, you might increase your chances over time to find a bug.  I think the examples I chose are somewhat benign, but I hope this conveys the concept.

On invocation, you will see something like this...

```
$ go test -v -fuzz FuzzStringToAddress -run FuzzStringToAddress ./vaa/
=== RUN   FuzzStringToAddress
fuzz: elapsed: 0s, gathering baseline coverage: 0/53 completed
fuzz: elapsed: 0s, gathering baseline coverage: 53/53 completed, now fuzzing with 32 workers
fuzz: elapsed: 3s, execs: 335309 (111612/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 6s, execs: 535132 (66538/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 9s, execs: 642623 (35804/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 12s, execs: 697864 (18414/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 15s, execs: 722709 (8284/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 18s, execs: 738365 (5216/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 21s, execs: 748284 (3310/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 24s, execs: 748284 (0/sec), new interesting: 1 (total: 54)
fuzz: elapsed: 27s, execs: 748284 (0/sec), new interesting: 1 (total: 54)
...
```


kcsongor:
We could write more interesting tests by incorporating something like https://github.com/trailofbits/go-fuzz-utils. They provide a way to fuzz structs and fixed size arrays. For example, we could write a test that ensure that for any random address, turning it into a string then turning it back to an address results in the original address.

That seems to me like a better approach than repeatedly calling `t.Skip()` when certain conditions are not met, because skipping will waste a lot of test cycles.

bruce-riley (approved):

PR #2879, opened 2023-05-05
relayer: import sdk locally for gr relayer engine app

JoeHowarth:


relayer/generic_relayer/relayer-engine-v2/package.json
```diff
@@ -34,7 +36,7 @@
     "@types/yargs": "^17.0.23",
     "nodemon": "^2.0.20",
     "prettier": "^2.8.7",
-    "ts-node": "^10.9.1",
+    "tsx": "^3.12.7",
```
JoeHowarth, line 39: made this change bc the local sdk takes very very long to load with ts-node

sdk/js/package.json
```diff
@@ -87,5 +84,10 @@
     "elliptic": "^6.5.4",
     "js-base64": "^3.6.1",
     "near-api-js": "^1.0.0"
+  },
+  "optionalDependencies": {
```
JoeHowarth, line 88: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies
>build failures do not cause installation to fail.

Since injective doesn't build on local imports not in a sibling directory, ignore the dep if it fails. The injective sdk is only used in a couple places that interact with injective, so this should not affect evm only generic relaying any time soon

chase-45 (approved):

PR #2885, opened 2023-05-06
clients/js: set strict = true in tsconfig

heyitaki:
This PR adds `strict: true` to `tsconfig.json` and takes care of any resulting errors. In general, most of this code doesn't touch logic, and instead focuses on null checks and legibility.

Things to review:
* NEAR CLI. Whatever code was previously in there was definitely broken, referencing multiple non-existent args, so this is probably strictly better even without testing. Still required some more involved changes, so would be good to look at.
* Type used in `yargs` handler (`Awaited<ReturnType<typeof builder>["argv"]>`). This type isn't perfect but should be good enough because it will always contain the types of the arguments (+ some extra fields). It doesn't seem possible to get the actual type that would appear if `handler` was passed into `yargs.command` along with `builder` since `yargs` throws the type away.

### DIFF
--- a/clients/js/package-lock.json
+++ b/clients/js/package-lock.json
@@ -45,6 +45,7 @@
         "@truffle/hdwallet-provider": "^2.0.15",
         "@types/bn.js": "^5.1.0",
         "@types/bs58": "^4.0.1",
+        "@types/node-fetch": "^2.6.3",
         "@types/yargs": "^17.0.2",
         "copy-dir": "^1.3.0",
         "typescript": "^4.6"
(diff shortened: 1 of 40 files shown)


heyitaki:
* main
  * **PR #2885** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2885" target="_blank">Graphite</a>  👈
    * **PR #2886** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2886" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/2885?utm_source=stack-comment).

evan-gray (approved):
If it passes tsc, it passes me. (for this)

kev1n-peters (approved):

PR #2886, opened 2023-05-06
clients/js: refactor worm CLI

heyitaki:
This PR:
1. Adds `worm info` as top-level cmd
2. Moves building `clients/js` into a shared image
3. Adds `worm info wrapped` and `worm info origin` (see below). It's possible to pass in custom network and/or RPC.
4. Refactors various parts of `worm`: organize chain-specific and generic code, enforce consistent typing & styling

Example output:
```sh
$ worm info wrapped ethereum 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 sui
0x5d4b302506645c37ff133b98c4b50a5ae14841659738d6d733d59d0d217a93bf::coin::COIN

$ worm info origin sui 0x5d4b302506645c37ff133b98c4b50a5ae14841659738d6d733d59d0d217a93bf::coin::COIN
{
  isWrapped: true,
  chainId: 2,
  assetAddress: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
}
```

PR #2887, opened 2023-05-07
worm: reduce wormhole sdk bundle size

kcsongor:
By only importing the esm dist, all the cjs modules are dropped from the bundle, eliminating about 1mb.


heyitaki:
Possible to merge this in after #2885 and #2886?

heyitaki (approved):

heyitaki:
Merging this in now so that we have the `sideEffects` list for #2908. Will rebase #2885 stack on top of this later today.

tonyjin (approved):

PR #2895, opened 2023-05-09
Relayer: Large CoreRelayer Refactor

nonergodic:


ethereum/contracts/interfaces/relayer/IWormholeReceiver.sol
```diff
@@ -1,48 +1,67 @@
 // SPDX-License-Identifier: Apache 2
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.19;
```
derpy-duck, line 3: Quick question: does that mean that users who want to use this interface file would need to have solidity 0.8.19?

ethereum/contracts/interfaces/relayer/IWormholeReceiver.sol
```diff
@@ -1,48 +1,67 @@
...
-     */
-    struct DeliveryData {
-        bytes32 sourceAddress;
-        uint16 sourceChain;
-        uint256 maximumRefund;
-        bytes32 deliveryHash;
-        bytes payload;
-    }
+interface IWormholeReceiver {
+  /**
```
derpy-duck, line 29: I think most of the diff in this file is actually not intended, as I think you started this PR before commit d560a95e08c08c774fbbd2e69cc9c1e811851313 (which is reflected in the old code but not here). Probably the main diff you intended here is removing the struct out of the interface?

ethereum/truffle-config.js
```diff
@@ -370,12 +370,13 @@ module.exports = {
 
   compilers: {
     solc: {
-      version: "0.8.4",
+      version: "0.8.19",
```
scnale, line 373: Ah, now I see that you had to modify the truffle config :P
I think if we could somehow tweak truffle to not get in our way, we would be fine. I might take a look at this.

Note that modifying the truffle config implies modifying the wormhole core + token bridge build which we probably won't want to do.

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -1,376 +1,579 @@
...
+error NotAnEvmAddress(bytes32);
+error OutOfBounds(uint256 offset, uint256 length);
+
+enum VaaKeyType {
+  EMITTER_SEQUENCE,
+  VAAHASH
 }
+
+/**
+ * @notice This 'VaaKey' struct identifies a wormhole message from the current transaction
```
derpy-duck, line 66: on second glance (this was incorrect earlier as well!) we should remove the 'from the current transaction' line! I can do this (as well as address any other of the small nits I point out) as well; just want to point it out so we don't forget

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -1,376 +1,579 @@
...
+   * - the relay provider passed in at least [(one wormhole message fee) + instruction.maximumRefundTarget + instruction.receiverValueTarget] of this chain's currency as msg.value
+   * - the instruction's target chain is this chain
+   * - the relayed signed VAAs match the descriptions in container.messages (the VAA hashes match, or the emitter address, sequence number pair matches, depending on the description given)
+   *
+   * @param targetParams struct containing the signed wormhole messages and encoded delivery instruction container (and other information)
+   */
+  function deliver(TargetDeliveryParameters memory targetParams) external payable;
+}
+
+interface IWormholeRelayer is IWormholeRelayerSend, IWormholeRelayerDelivery {}
```
derpy-duck, line 579: Users will only need IWormholeRelayerSend. Do you think it is preferable to hand them this full interface rather than simply have IWormholeRelayerSend in a seperate file that users can be given? I don't have a strong opinion here but curious about the decision to put both in one file

ethereum/forge-test/relayer/MockGenericRelayer.sol
```diff
@@ -2,13 +2,12 @@
 
 pragma solidity ^0.8.0;
 
-import {IWormholeRelayer} from "../../contracts/interfaces/relayer/IWormholeRelayer.sol";
-import {IDelivery} from "../../contracts/interfaces/relayer/IDelivery.sol";
-import {IWormholeRelayerInstructionParser} from
-    "../../contracts/interfaces/relayer/IWormholeRelayerInstructionParser.sol";
+import "../../contracts/interfaces/relayer/IWormholeRelayer.sol";
+import "../../contracts/interfaces/relayer/IWormholeRelayer.sol";
```
derpy-duck, line 6: this can be done once

ethereum/truffle-config.js
scnale: Update: if we use [truffle v5.5.14](https://github.com/trufflesuite/truffle/releases/tag/v5.5.14) or higher I think we could avoid compiling generic relayer contracts while building the core with v0.8.4.

ethereum/foundry.toml
```diff
@@ -1,8 +1,8 @@
 [profile.default]
-solc_version = "0.8.17"
+solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = false
+via_ir = true
```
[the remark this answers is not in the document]
derpy-duck: Also the primary reason I wanted to have it able to be compiled in the legacy pipeline is it helped me tremendously with quick iteration when I was debugging the forge tests. Now that hopefully the contracts are more stable / less forge test development is needed, maybe it is not as necessary to have quick iteration.

ethereum/truffle-config.js
derpy-duck: True, I wonder if this might be okay (to modify the wormhole core + token bridge) because they might be currently built with Truffle anyways? I'm not sure. Probably we should flag this in the description of the big PR (generic-relayer-merge -> main) we end up making.

ethereum/contracts/relayer/coreRelayer/CoreRelayer.sol
```diff
@@ -1,68 +1,41 @@
...
+
+  //the only normal storage variable - everything else uses slot pattern
+  //no point doing it for this one since it is entirely one-off and of no interest to the rest
+  //  of the contract and it also can't accidentally be moved because we are at the bottom of
+  //  the inheritance hierarchy here
+  bool private initialized;
+
+  constructor(
+    address wormhole,
+    address _initialDefaultRelayProvider
```
scnale, line 30: I'm a bit out of date on decisions related to CREATE2 deployment, but this means that we introduce a dependency on the address of the relay provider to get a deterministic address for the `CoreRelayer` proxy, right? Oh, and the wormhole core address is a problem too now that I think about it. I think we should move these to the `initialize` parameters.

ethereum/contracts/interfaces/relayer/IWormholeReceiver.sol
nonergodic: Just another huge diff due to code formatting... (I recall that some comments had 2 spaces between the asterisk and the text and other comments only had one and I fixed that - not sure if this was here).

I originally branched before commit [d560a95](https://github.com/wormhole-foundation/wormhole/commit/d560a95e08c08c774fbbd2e69cc9c1e811851313) but then did a caveman rebase (couldn't get my actual branch to properly rebase on the latest upstream and so I just cloned the entire thing again and submitted all changes as a single commit there...).

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
nonergodic: No, this should definitely be separated again. I didn't find the time to give the interfaces the amount of love that they need. I only restructured things and threw them together because it was easy at the time and then I didn't find the time to neatly separate them out again.

I also mentioned this on the call with @JoeHowarth just now: The errors should be grouped too so you have an idea which ones come from Send, which from Delivery and which from both and what are the circumstances when they are thrown.

ethereum/truffle-config.js
derpy-duck: Oh just realized this is the truffle-config. Yeah haha this for sure we'd need to get checked since it modifies the Core/Token bridge

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
```diff
@@ -0,0 +1,264 @@
...
+
+    uint256 amount;
+    unchecked {amount = maxTransactionFee + receiverValue;}
+    //TODO AMO: what if pay fails? (i.e. returns false)
+    pay(relayProvider.getRewardAddress(), amount);
+  }
+
+  // ----------------------- delivery transaction temorary storage functions -----------------------
+
+  function startDelivery(address targetAddress) internal {
```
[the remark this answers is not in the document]
nonergodic: As mentioned on the call: Because they form a coherent whole that belongs together (Send and Delivery coordinate around DeliveryTmpState. Really personal preference of whether to keep them as a logical unit together in this spot or split them to the subclasses where the individual functions are actually used.

ethereum/contracts/relayer/coreRelayer/CoreRelayer.sol
nonergodic: That's true, but do we care about the address of the logic contracts? The proxies should all be deployed to the same, but logic contract addresses shouldn't really matter, no?

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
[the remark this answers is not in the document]
derpy-duck: I assume because this file assumes the role of 'setters' and modifies state directly

ethereum/contracts/mock/MockRelayerIntegration.sol
```diff
@@ -86,20 +88,20 @@ contract MockRelayerIntegration is IWormholeReceiver {
...
-            IWormholeRelayer.VaaKeyType.EMITTER_SEQUENCE,
+        vaaKeys = new VaaKey[](2);
+        vaaKeys[0] = VaaKey(
+            VaaKeyType.EMITTER_SEQUENCE,
             wormhole.chainId(),
-            relayer.toWormholeFormat(address(this)),
+            toWormholeFormat(address(this)),
             sequence1,
             bytes32(0x0)
         );
```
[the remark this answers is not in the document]
scnale: Hm, yeah, in this case I'd use the object notation because `bytes32(0x0)` doesn't tell me anything :sweat_smile:
When several parameters have the same type, I think it might be worth it. Even with good variable naming the compiler won't catch some argument swaps. Only tests would.

ethereum/contracts/relayer/coreRelayer/CoreRelayer.sol
scnale: Yeah, the proxies addresses should be fine I think. I think at some point this would have posed a problem. I have to get up to speed on the current deployment logic. If this still poses a problem, we should be able to solve it by adjusting the way the proxy is deployed, I think.

PR #2897, opened 2023-05-09
Relayer: CoreRelayer large refactor 

JoeHowarth:
Moving from https://github.com/wormhole-foundation/wormhole/pull/2895 for easier editing


ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
-    error MsgValueTooHigh(); // msg.value is higher than the amount you specified (i.e., msg.value is greater than (wormhole message fee) + (maxTransactionFee) + (receiverValue))
-    // Specifically, (msg.value) + (any leftover funds if this is a forward) is less than (maxTransactionFee + receiverValue), summed over all of your requests if this is a multichainSend/multichainForward
-    error MaxTransactionFeeNotEnough(); // maxTransactionFee is less than the minimum needed by your chosen relay provider
-    error NoDeliveryInProgress(); // Forwards can only be requested within execution of 'receiveWormholeMessages', or when a delivery is in progress
-    error ForwardRequestFromWrongAddress(); // A forward was requested from an address that is not the 'targetAddress' of the original delivery
-    error RelayProviderDoesNotSupportTargetChain(); // Your relay provider does not support the target chain you specified
+//TODO AMO: Decide on use regarding '' vs `` quotation of variables/function names in comments
+//            and then make it uniform.
```
derpy-duck, line 6: Comment on 'TODO AMO': Yes this would be nice. Probably not very high priority though

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
+
+//TODO AMO: 132 because presumably 4 (function selector) + 4*32 (4 32-byte words)
```
derpy-duck, line 8: Yup. Addressed this in https://github.com/wormhole-foundation/wormhole/pull/2899/files

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
+ * To use the default relay provider, set this field to be getDefaultRelayProvider()
+ * @custom:member vaaKeys Array of VaaKey structs identifying each message to be relayed. Each VaaKey struct specifies a wormhole message, either by the VAA hash, or by the (chain id, emitter address, sequence number) triple.
+ * The relay provider will set the second parameter in the call to receiveWormholeMessages to be an array of signed VAAs specified by this vaaKeys array.
+ * Specifically, the 'signedVaas' array will have the same length as 'vaaKeys', and additionally for each 0 <= i < vaaKeys.length, signedVaas[i] will match the description in vaaKeys[i]
+ * @custom:member consistencyLevel The level of finality to reach before emitting the Wormhole VAA corresponding to this 'send' request. See https://book.wormhole.com/wormhole/3_coreLayerContracts.html#consistency-levels
+ * @custom:member payload an optional payload which will be delivered to the receiving contract.
+ * @custom:member relayParameters This should be 'getDefaultRelayParameters()'
+ */
+  //TODO AMO: Why does this struct exist as an externally visible struct?
```
derpy-duck, line 151: Agreed, thanks for calling out! Joe is working on both of these TODOs I believe

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
+   * @title IWormholeRelayer
+   * @notice Users may use this interface to have wormhole messages (VAAs) in their transaction
+   *   relayed to destination contract(s) of their choice.
+   */
+
+  /**
+   * @notice This 'send' function emits a wormhole message (VAA) that alerts the default wormhole
+   * relay provider to call the receiveWormholeMessage(DeliveryData memory deliveryData, bytes[]
+   * memory signedVaas) endpoint of the contract on chain 'targetChainId' and address
+   //TODO AMO: another duplication of DeliveryData documentation
```
derpy-duck, line 185: addressed in https://github.com/wormhole-foundation/wormhole/pull/2899/files

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
+    uint256 receiverValue,
+    bytes memory payload,
+    VaaKey[] memory vaaKeys,
+    uint8 consistencyLevel
+  ) external payable returns (uint64 sequence);
+
+  /**
+   *  @notice This 'send' function emits a wormhole message (VAA) that alerts the relay provider specified by sendParams.relayProviderAddress to
+   *  call the receiveWormholeMessage(DeliveryData memory deliveryData, bytes[] memory signedVaas) endpoint of the contract on chain 'sendParams.targetChainId' and address 'sendParams.targetAddress'
+   //TODO AMO: Jesus H Christ, how often is this duplicated?!?
```
derpy-duck, line 301: addressed in https://github.com/wormhole-foundation/wormhole/pull/2899/files

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
 }
+
+/**
+ * @notice TargetDeliveryParameters is the struct that the relay provider passes into 'deliver'
+ * containing an array of the signed wormhole messages that are to be relayed
+ *
+ * @custom:member encodedVMs An array of signed wormhole messages (all from the same source chain transaction)
+ * @custom:member encodedDeliveryVAA signed wormhole message from the source chain's CoreRelayer contract with payload being the encoded delivery instruction container
+ * @custom:member relayerRefundAddress The address to which any refunds to the relay provider should be sent
+ * @custom:member overrides. Optional overrides field which must parse to executionParameters. //TODO AMO: this seems wrong
```
derpy-duck, line 501: addressed in https://github.com/wormhole-foundation/wormhole/pull/2899/files

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -2,375 +2,580 @@
...
+ */
+//TODO AMO: Why does this struct exist in the first place?
```
derpy-duck, line 503: will be addressed by Joe

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
```diff
@@ -0,0 +1,264 @@
...
+} from "../../interfaces/relayer/IWormholeRelayer.sol";
+import {
+  ForwardInstruction,
+  DeliveryTmpState,
+  getDeliveryTmpState,
+  getRegisteredCoreRelayersState
+} from "./CoreRelayerStorage.sol";
+
+abstract contract CoreRelayerBase is IWormholeRelayerBase {
+  //TODO AMO: see https://book.wormhole.com/wormhole/3_coreLayerContracts.html#consistency-levels
```
derpy-duck, line 27: agreed that these should be provided by IWormhole. In lieu of that, is 15 the best choice?

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
```diff
@@ -0,0 +1,264 @@
...
+    IRelayProvider relayProvider
+  ) internal returns (uint64 sequence) { 
+    sequence =
+      getWormhole().publishMessage{value: wormholeMessageFee}(0, message, consistencyLevel);
+
+    emit SendEvent(sequence, maxTransactionFee, receiverValue);
+
+    uint256 amount;
+    unchecked {amount = maxTransactionFee + receiverValue;}
+    //TODO AMO: what if pay fails? (i.e. returns false)
```
derpy-duck, line 71: Good point that if this fails, the value is stored in the contract. Is there an alternative you'd suggest?

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
+import {BytesParsing} from "./BytesParsing.sol";
+import {CoreRelayerSerde} from "./CoreRelayerSerde.sol";
+import {ForwardInstruction} from "./CoreRelayerStorage.sol";
+import {CoreRelayerBase} from "./CoreRelayerBase.sol";
+
+abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelivery {
+  using CoreRelayerSerde for *; //somewhat yucky but unclear what's a better alternative
+  using BytesParsing for bytes;
+
+  //TODO AMO: Why does TargetDeliveryParameters exist in the first place?
```
derpy-duck, line 41: will be addressed by Joe

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
-                Utils.pay(payable(fromWormholeFormat(instruction.refundAddress)), refundAmount);
-            refundStatus =
-                refundPaidToRefundAddress ? RefundStatus.REFUND_SENT : RefundStatus.REFUND_FAIL;
-        } else {
-            address providerAddress = fromWormholeFormat(relayerAddress);
-            (refundPaidToRefundAddress, refundStatus) =
-                payRefundRemote(instruction, refundAmount, providerAddress);
-        }
-    }
+    //TODO AMO: At this point forwards (which are payable!) might have increased contract balance
```
derpy-duck, line 220: We store the msg value that the forwarder puts in, in state, and then later pay it out as part of the forward relayer's reward, so I don’t think we need to account for that here? Those funds are been payed out to the relay provider for the forward. Let me know if I am understanding this comment wrong.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
-            deliveryInstruction.receiverValueTarget = overrides.receiverValue;
-            deliveryInstruction.maximumRefundTarget = overrides.maximumRefund;
-            return (deliveryInstruction, overrides.redeliveryHash);
-        }
+    //Refund the relayer (their extra funds) + (the amount that the relayer spent on gas)
+    //  + (the users refund if that refund didn't succeed)
+    uint256 relayerRefundAmount = (
+      msg.value - deliveryInstruction.receiverValueTarget - deliveryInstruction.maximumRefundTarget
+    ) + (deliveryInstruction.maximumRefundTarget - transactionFeeRefundAmount)
+    //TODO AMO: Isn't this a bug? We add the same amount regardless of whether we hit the max or not
```
derpy-duck, line 436: - Don’t believe this is a bug, but definitely an interesting quirk: for cross chain refunds, if the refund amount ends up being more than the maximum budget, the relay provider gets paid the full amount and only has to deliver the ‘maximum budget’ amount on the refund chain. This was an intentional decision and I think it is fine? because, 1) I assume the maximum budget will be quite large 2) we can make the guarantee: you will receive refunds up to the maximum budget if you are doing a cross-chain refund; I think that is reasonable, and 3) we can consider the price graph not linear but piecewise linear and then straight vertical, in which case this is correct. Lmk if I am missing something about your comment

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
+      + ((refundStatus == RefundStatus.REFUND_SENT ||
+          refundStatus == RefundStatus.CROSS_CHAIN_REFUND_SENT ||
+          refundStatus == RefundStatus.CROSS_CHAIN_REFUND_SENT_MAXIMUM_BUDGET
+         ) ? 0 : refundToRefundAddress);
+
+    //TODO AMO: what if pay fails? (i.e. returns false)
```
derpy-duck, line 442: If it fails, I guess it just gets stored in the contract. What would you suggest we do in that case?

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
+    unchecked{refundSubMessageFee = refundAmount - wormholeMessageFee;}
+
+    uint256 receiverValueTarget = calculateTargetDeliveryMaximumRefund(
+      refundChainId, refundSubMessageFee, relayProvider
+    );
+    
+    //If refundAmount is not enough to pay for one wei of receiver value, then do not perform the
+    //  cross-chain refund (i.e. if (delivery overhead) + (wormhole message fee) + (cost of one wei
+    //  of receiver value) is larger than the remaining refund)
+    //TODO AMO: but what happens to the value then?
```
derpy-duck, line 478: it gets paid to the relayer refund address as part of line 437-440

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
+    if (receiverValueTarget == 0)
+      return RefundStatus.CROSS_CHAIN_REFUND_FAIL_NOT_ENOUGH;
+
+    uint256 maxBudget = relayProvider.quoteMaximumBudget(refundChainId);
+    bool exceedsMaxBudget = false;
+    if (receiverValueTarget > maxBudget) {
+      //TODO AMO: What happens with the difference if the maximum budget is exceeded?
```
derpy-duck, line 485: It gets paid to the relay provider as part of the cost (so the price doesn't change, the amount given just gets upper bounded by the maximum budget)

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
-            return (vaaKey.vaaHash == parsedVaa.hash);
-        } else {
-            return false;
-        }
-    }
+    publishAndPay(
+      wormholeMessageFee,
+      0,
+      refundSubMessageFee,
+      //TODO AMO: Why does this not use convertSendToDeliveryInstruction()?
```
derpy-duck, line 494: Now it does with https://github.com/wormhole-foundation/wormhole/pull/2898 ! good callout :)

ethereum/contracts/relayer/coreRelayer/CoreRelayerGovernance.sol
```diff
@@ -0,0 +1,208 @@
...
+//
+//   //This could potentially be combined with ContractUpgrade
+//   struct UpdateDefaultProvider {
+//     bytes32 module;
+//     uint8 action;
+//     uint16 chainId;
+//     bytes32 newProvider; //TODO AMO: why is this not an address?
+//   }
+
+  //TODO AMO: What's this module business about? (and also the action business for that matter)
```
derpy-duck, line 57: Module is 'CoreRelayer' keccak256-ed. (perhaps this should be written as a comment, hmm)

ethereum/contracts/relayer/coreRelayer/CoreRelayerSend.sol
```diff
@@ -1,283 +1,287 @@
...
+      instruction.encode(),
+      sendParams.consistencyLevel,
+      relayProvider
+    );
+  }
+
+  function forward(Send memory sendParams) public payable {
+    checkMsgSenderInDelivery();
+
+    //TODO AMO: Introduce basic sanity checks on sendParams (e.g. all valus below 2^128?)
```
derpy-duck, line 132: Joe working on this

ethereum/contracts/relayer/coreRelayer/CoreRelayerSend.sol
```diff
@@ -1,283 +1,287 @@
...
+    receiverValue = assetConversionHelper(
+      targetChainId, targetAmount, getChainId(), numerator, denominator, true, provider
+    );
+  }
+
+  function getDefaultRelayProvider() public view returns (address relayProvider) {
+    relayProvider = getDefaultRelayProviderState().defaultRelayProvider;
+  }
+
+  //TODO AMO: solc suggested changing view to pure - is this fine given that IWormholeRelayerSend
```
derpy-duck, line 250: done in https://github.com/wormhole-foundation/wormhole/pull/2899/files

ethereum/contracts/relayer/coreRelayer/CoreRelayerSerde.sol
```diff
@@ -0,0 +1,272 @@
...
+library CoreRelayerSerde {
+  using BytesParsing for bytes;
+
+  // ---------------------- "public" (i.e implicitly internal) encode/decode -----------------------
+
+  uint8 private constant PAYLOAD_ID_VAAKEY = 1;
+  uint8 private constant PAYLOAD_ID_EXECUTION_PARAMETERS = 1;
+  uint8 private constant PAYLOAD_ID_DELIVERY_INSTRUCTION = 1;
+  uint8 private constant PAYLOAD_ID_DELIVERY_OVERRIDE = 1;
+  //TODO AMO: Why are we at version 2 here?
```
derpy-duck, line 28: When reading wormhole messages emitted from the contract (in the relayer engine), we parse the first byte to determine whether it is a DeliveryInstruction or a RedeliveryInstruction.

ethereum/contracts/relayer/coreRelayer/CoreRelayerSerde.sol
```diff
@@ -0,0 +1,272 @@
...
+  uint8 private constant PAYLOAD_ID_REDELIVERY_INSTRUCTION = 2;
+
+  //TODO AMO: All the recursive abi.encodePacked calls in here are _insanely_ gas inefficient unless
```
derpy-duck, line 31: Ah the only reason for this was stack-too-deep related reasons, which are alleviated now that we have via-ir.

derpy-duck: I took a stab at removing some in https://github.com/wormhole-foundation/wormhole/pull/2899, but all the ones related to encoding arrays seem tough to remove

ethereum/contracts/relayer/coreRelayer/CoreRelayerGovernance.sol
scnale: Btw, I'd be in favor of just setting the `module` constant to `bytes32(CoreRelayer)` :P

scnale: AFAIK, module adds domain information for the governance VAA and action, the same but as a way to choose a particular governance action for that module. It's an entirely custom dispatch for contract interaction. Another option could be to use the Ethereum ABI, but we would be imposing that on other VMs and that may end up being a bit impractical. I don't know which parts of this are flexible and which are not in regards to the guardian codebase.

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
nonergodic: Yes, @chase-45 signed off on it

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
nonergodic: I'm happy to supply a backup address that is guaranteed to be payable :zany_face:

More seriously, I'm not sure what a good solution for this stupid edge case looks like.
We could have a mapping that keeps track of unredeemed funds per relay provider and provide a fallback mechanism to claim those funds, but the juice hardly seems worth the squeeze.

Probably the better solution is just to revert and have the relay provider fix their shit (since it's in their interest in the first place)

nonergodic: Just trapping the funds seems like the worst of all worlds though. So I'd say: revert with a `RewardAddressNotPayable(address)` error

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,560 @@
...
-        clearForwardInstructions();
-
-        // unlock the contract
-        setContractLock(false);
-        setLockedTargetAddress(address(0));
-
-        // set delivery status and additionalStatusInfo if there was a failure
-        stack.transactionFeeRefundAmountPostForward = stack.transactionFeeRefundAmount;
-        DeliveryStatus status;
+    //TODO AMO: removed the outer gasUsed calculation since it is returned by executeInstruction
```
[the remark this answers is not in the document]
nonergodic: I think that fixed overhead for Core Relayer + variable for whatever the user consumes is totally fine. I don't think measuring it with higher precision (while possible) makes much sense.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
nonergodic: No, this is perfectly fine. I wrote this earlier on when I hadn't seen/digested the code in emitForwards yet, so this really is just outdated.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
nonergodic: No, sounds good! Definitely deserves a comment in the code though, I think. If you truncate the funds, it's a very natural to ask "but what about the rest?" and so there should be a comment there to answer that. (Or perhaps just point to the part of the bigger doc that explains it)

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
nonergodic: As I said for the other pay too: We probably want to enforce that `pay` is successful because (unless I'm missing something) this is fully under the control of the relayProvider. Seems kinda bad if we don't "warn" them (via a failing transaction simulation) that they won't get rewarded for their relay because their setup is somehow screwed up.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -1,623 +1,565 @@
...
+        }),
+        vaaInfo.encodedVMs
+      )
+    returns (bool targetCallSucceeded, uint32 _gasUsed, bytes memory targetRevertDataTruncated) {
+      gasUsed = _gasUsed;
+      transactionFeeRefundAmount =
+        calculateTransactionFeeRefundAmount(vaaInfo.deliveryInstruction, gasUsed);
+      if (targetCallSucceeded) {
+        //Retrieve the forward instruction created during execution of 'receiveWormholeMessages'
+        ForwardInstruction[] storage forwardInstructions = getForwardInstructions();
```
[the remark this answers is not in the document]
nonergodic: The clearForwardInstructions happens in `finishDelivery` which is invoked at the end of `delivery`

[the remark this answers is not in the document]
derpy-duck: Ah. missed that! my bad :)

ethereum/contracts/relayer/coreRelayer/CoreRelayerBase.sol
scnale: Yeah, agreed, I had seen this before and was not sure about how to handle it. Reverting seemed excessive to me but at the same time, the reward address should be maintained extremely well, especially if it contains any code.

ethereum/contracts/relayer/coreRelayer/CoreRelayerSerde.sol
scnale: I think we can't encode arrays efficiently with the current expressive power of solidity. We always end with a quadratic cost in terms of memory. At some point I wanted to write some assembly to optimize this part and concatenate elements in a linear algorithm but I think it can be done after the first release.

ethereum/contracts/relayer/coreRelayer/CoreRelayerGovernance.sol
nonergodic: It's actually not the same as `bytes32("CoreRelayer")` because the string version is left-aligned while the hex number version is right-aligned.

scnale: I know. I just don't see any purpose in choosing a particular alignment, it just obfuscates what it is :pensive:

derpy-duck (approved):

ethereum/contracts/relayer/coreRelayer/CoreRelayerGovernance.sol
derpy-duck: True. the problem is the guardian code for upgrade/register chain has the module left-padded so this would need to be the case in order to reuse that :(

derpy-duck (approved):

PR #2899, opened 2023-05-09
gRelayer: address some TODOs

derpy-duck:


ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -488,7 +459,7 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
    * @notice Returns default relay parameters
    * @return relayParams default relay parameters
    */
-  function getDefaultRelayParams() external view returns (bytes memory relayParams);
+  function getDefaultRelayParams() external pure returns (bytes memory relayParams);
```
nonergodic, line 462: I'm not sure to what extent this matters, but I would keep this `view` instead of `pure` because in the future whatever the default params are, they might be read/computed from state and then suddenly the function would have to go back to being view.

PR #2903, opened 2023-05-10
Relayer: remove send struct from api, enforce input size limits

JoeHowarth:
Input size limits rationale:

**receiverValue and maxTransactionFee u128**
- These both need to pass through assetConversion source -> target, which will add (log(targetNativePrice / sourceNativePrice)) decimals, which could be as high as 64

**nativePrices u64**
- This is enough for a single unit of currency to be 2^64/10^6 =>  $18,446,744,073,709 -> 10^13 dollars

**gasPrice u64**
- 1 gas is ~10^10 wei (10 gwei) -> 2^32
- This leaves room for gas prices to be 10 billion times higher than on eth
- can probably lower this safely

**quoteGasPrice u88**
- allow gas prices to be different by up to 2^24 (i.e. target is 2^64 and source is 2^40) 

**quoteDeliveryOverhead u128**
- same scale as maxTransactionFee 

**maximumBudget u192**
- allow u128 * u64 => u192

Up next is removing the DeliveryInfo struct


ethereum/forge-test/relayer/WormholeRelayer.t.sol
```diff
@@ -101,9 +103,23 @@ contract WormholeRelayerTests is Test {
...
+            (
+                uint256(gasParams.targetGasPrice) * feeParams.targetNativePrice
+                    + feeParams.sourceNativePrice - 1
+            ) / uint256(feeParams.sourceNativePrice) < type(uint88).max
+        );
+
         vm.assume(
             feeParams.targetNativePrice
-                < (uint256(2) ** 238)
+                < (uint256(2) ** 126)
```
derpy-duck, line 122: why 126 and not 128? if we are trying to test maxTransactionFees up to 2^128

ethereum/forge-test/relayer/WormholeRelayer.t.sol
```diff
@@ -526,6 +548,9 @@ contract WormholeRelayerTests is Test {
 
         assertTrue(keccak256(setup.target.integration.getMessage()) == keccak256(message));
 
+        console.log("refund addr diff", (setup.target.refundAddress.balance - refundAddressBalance));
```
derpy-duck, line 551: are these intentionally kept in?

ethereum/forge-test/relayer/WormholeRelayer.t.sol
```diff
@@ -534,17 +559,21 @@ contract WormholeRelayerTests is Test {
             * (setup.source.rewardAddress.balance - rewardAddressBalance)
             - feeParams.targetNativePrice * (relayerBalance - setup.target.relayer.balance);
 
+        console.log("relayerProfit", relayerProfit);
         uint256 howMuchGasRelayerCouldHavePaidForAndStillProfited =
             relayerProfit / gasParams.targetGasPrice / feeParams.targetNativePrice;
-        assertTrue(howMuchGasRelayerCouldHavePaidForAndStillProfited >= 30000); // takes around this much gas (seems to go from 36k-200k?!?)
+        console.log(
```
derpy-duck, line 565: maybe remove these console logs?

derpy-duck (approved):

PR #2904, opened 2023-05-10
cosmwasm: add comment emphasizing payload3 security checks

nik-suri:


a5-pickle (approved):

cosmwasm/contracts/token-bridge/src/contract.rs
```diff
@@ -834,6 +834,9 @@ fn handle_complete_transfer_token(
     let target_address = (&transfer_info.recipient.as_slice()).get_address(0);
     let recipient = deps.api.addr_humanize(&target_address)?;
 
+    // SECURITY CRITICAL CHECK FOR PAYLOAD3 VAAs.
+    // This check ensures that a malicious sender cannot
```
kcsongor, line 838: not sure if "malicious sender" is the right phrase here. The payload3 spec says that only the intended recipient of the tx can redeem the transaction, because only the recipient knows how to process the additional payload (since it's application specific). This check enforces that

evan-gray (changes requested):
Requesting @kcsongor 's comment change 🙂

kcsongor:
I'll close this for now, as the check is required by the specification (and not for security reasons). Feel free to reopen if you think a version of this comment should still be included @nik-suri

PR #2908, opened 2023-05-10
clients/js: move worm parse to print valid json

SEJeff:
This was really useful when looking at a list of all governance vaas to see which ones were guardian set upgrades. I did it as a local change and @tbjump expressed interest in the functionality.


heyitaki (approved):

SEJeff:
Example usage:
```
$ worm parse --json 01000000010d02aa3ffb28f2c401f90b8b5fb7af57bf9249204970053b1ba06bca147c2dab31b21d50c2a7695e8ebed00b07f0b1cb9c1658f41ee59a56b9fb4b289d9c15ce838f01030976bd5450631765db2c4017f33dd8fbd61dae364d1dfd43cdb866f72f8d4a4b47960026e64560a8823c422dce7c1e2a850aa5226c6b7092606c5224b69450e20104597391bda61d86dcb900d6b0062cf72b52b1ca250d625b3f62f89981041c16ec4b457038388058cdec9c5486baf6e84c1481e124c9f809ced7d73c235fcd400d0005c3356b43ed505d26fd55841925f8dcd4ae176990c4e9654742a20fe3438e0bb7367bd35019a546afe71af895f31e6e70de33ac800f75ee48a9dedac8f9817af5000609f9b2bd1c3f06bf23ed10eca5d13762b28872e1b44dbdb5d1f079c40c396f2a2ad2ac07278bd4cb140f61b86305f89a339dd0d39862f3a1f2cd8a9b7fd3eafe000779956931cd35e5e54934cadb37d9ad530164d58d9cbf2e39dbd73826a879705350ecd93ac7e48b21f64232b24b423d2d15b897e442dc4df9be7fdb92ebbaa2c100099af9249cfc3300edb5da61fd425371244a9a30fbfc93df455a3ac59cf3e56a4900da0566b38f8b56059f1795593b92c30c833dc6106657f4554a86f501b1f6b2010badf2b90028e98eef4aea4094c4cd0d3b411837f0469e13468b5016cc955510e808b4aafaa87aea8bc07c7506f3782d0428468df405b21fc36ed2ccfb32877642010da3f78c15d55d8efd5646a31a81c9316de35d61abd2c8a2df419b888a62a5c80771cd045db60a49d4f87c42f18de743faf6883313dba558c7da4ad209254b9b83010e5f398f609913e6d39420d44eaeff12e3431778ede50b119f0aa7ddd926de1aa34d855013984f2ca785e610e0efae0ffdb424d9caedac7f2e89ada0dee9ac6acb0110e19223db4b8837625732237d3c0f6041c28794c58178c0466ca6d646b9fb6c0941db7854a2f48a2383ea46d047e76a81f06185378c0598ead8535c0b080aa55b0111d8647c197db25ada7d55f559c65a7196fbd7361e92c68203c1842544b61bfd837117a8ab1ff10937737768c7a5fa93f65a39de2bb15370d714f1e0b4e24b6ea801127ec8835b907102f92a14a259054a1914a86afdf424108e83ae6b64b1dca0b69a6e55051cdce368fac7f3b4a50c8d9affae89d4253730eb4f12f38e056f1657ef01000000002c4929070001000000000000000000000000000000000000000000000000000000000000000411c7893f86b34cf020000000000000000000000000000000000000000000546f6b656e427269646765020004000000000000000000000000ee91c335eab126df5fdb3797ea9d6ad93aec9722 | jq '. | "type=\(.payload.type) chain=\(.payload.chain) emitterChain=\(.emitterChain) digest=\(.digest)"'
"type=ContractUpgrade chain=4 emitterChain=1 digest=0x353bb7417bf9d0873091e29fc07a7b776adb780981aa217b51a6a167941c211a"
```

kcsongor:
Great! Instead of introducing a new flag, why don't we just simply change the way the default printing works? The intention for the default format is to be valid JSON, so the current big number serialisation can be considered a bug.

clients/js/src/cmds/parse.ts
```diff
@@ -7,8 +7,22 @@ export const builder = (y: typeof yargs) => {
...
     type: "string",
+  })
+  .option("json", {
+    alias: "j",
+    describe: "Print results in json format",
+    type: "boolean",
+    default: false,
   });
 };
+// eslint-disable-next-line @typescript-eslint/no-redeclare
```
kcsongor, line 18: instead of redeclaring `BigInt`'s `toJSON` implementation, maybe we could use a replacer in `JSON.stringify`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

SEJeff: I figured the mozilla folks knew better than me and [used the first approach they suggested](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json). I can make it use a replacer.

SEJeff:
@kcsongor I'll change the default, but you lose the pretty colorization when doing that FYI.

heyitaki:
Feels to me that patching BigInt is more elegant than patching JSON.stringify so we don't add code to calls that require BigInt serialization. RE: replacers, I agree with @kcsongor that we should patch the default behavior since this is more akin to a bug, so I'd rather use the BigInt prototype extension than have to add a replacer for each call that need to serialize BigInts. Thoughts? @kcsongor @SEJeff

SEJeff:
I tend to prefer patching BigInt, however I don't do js / ts that often, so I'm not sure what is more idiomatic or preferred. That said, I don't much care.

What do I need to do to get @heyitaki and @kcsongor to approve this? :)

heyitaki (approved):

evan-gray (changes requested):
Unfortunately this is the way https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json

I’ll pour one out for the loss of colorization, but we should really keep side effects to dedicated files. That is, do this once in a dedicated file that is imported at the very beginning of the program.  (Or rename quiet.ts and put it in there?)

heyitaki (approved):

PR #2622, opened 2023-04-11
node: update golang to 1.20 from 1.19 (and upgrade libp2p to 0.27 from 0.22)

tbjump:
# Release notes
Since libp2p and many indirect dependencies were upgraded, I'd recommend thorough testing in testnet and a staged roll-out in mainnet. Guardians can see other Guardian's version and boot time in the [Wormhole Dashboard](https://wormhole-foundation.github.io/wormhole-dashboard/). 
# Rationale
Go’s [release policy](https://go.dev/doc/devel/release) states, “Each major Go release is supported until there are two newer major releases.”
Therefore it is important to stay current with the latest golang version. 

# Changes
Golang 1.20 introduced some breaking changes: 
1. `libp2p` and related modules needed to be updated to the latest version 
2. `ecdsa.GenerateKey` is no longer deterministic and hence we needed to switch to hardcoded devnet keys rather than generating them with a hardcoded seed. 

The go.mod was generated the following way: 
```
go1.20.3 mod edit -go=1.20
go1.20.3 mod why -m github.com/lucas-clemente/quic-go
go1.20.3 get github.com/libp2p/go-libp2p@v0.27.0
go1.20.3 get github.com/libp2p/go-libp2p-pubsub@v0.9.3
go1.20.3 get github.com/libp2p/go-libp2p-kad-dht@v0.23.0
go1.20.3 mod tidy
```

PR #2919, opened 2023-05-11
node: update golang to 1.20 from 1.19

tbjump:
same as https://github.com/wormhole-foundation/wormhole/pull/2622 but merging to main


DEVELOP.md
```diff
@@ -4,7 +4,7 @@
 
 The following dependencies are required for local development:
 
-- [Go](https://golang.org/dl/) >= 1.19.3
+- [Go](https://golang.org/dl/) >= 1.20.3
```
SEJeff, line 7:
```suggestion
- [Go](https://golang.org/dl/) >= 1.20.4
```

SEJeff (approved):

hendrikhofstadt (approved):

DEVELOP.md
tbjump: since golang 1.19 is still *supported*, I won't change this for now. The *recommended* version is 1.20 though, which will be in the release notes.

evan-gray (approved):

DEVELOP.md
SEJeff: That’s totally fine but you should make them consistent. You use go 1.20.4 everywhere except this specific one. You shouldn’t do that.

make them consistent was my review feedback.

PR #2946, opened 2023-05-17
Update forge-std to the latest

SEJeff:
Newer versions of forge-std [like v1.0.0+](https://twitter.com/msolomon44/status/1587171153918840833?lang=en) include much improved fuzzing and invariant testing. To take advantage of that functionality, this PR upgrades both forge and forge-std.

To do that, it required changing the inheritance of the forge tests. I changed a bit of the logs in the test for truncateAddress with some help from @scnale since the previous version led to somewhat undefined behavior. I verified the changed tests run ✅ with the old and new version of forge-std.

It is highly recommended in your local checkout you run:

    rm -rf ethereum/lib/forge-std
    make forge_dependencies

### DIFF
--- a/ethereum/Makefile
+++ b/ethereum/Makefile
@@ -35,7 +35,7 @@ node_modules: package-lock.json
 forge_dependencies: lib/forge-std
 
 lib/forge-std:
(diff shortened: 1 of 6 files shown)


kcsongor (approved):

scnale (approved):
Looks good to me.

PR #2953, opened 2023-05-18
sdk/js: add sui/aptos to uint8ArrayToNative

heyitaki:
Implements Sui/Aptos for `uint8ArrayToNative`.

### DIFF
--- a/sdk/js/CHANGELOG.md
+++ b/sdk/js/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.9.18
+
+### Changed
+
+Update uint8ArrayToNative
+
 ## 0.9.17
 
 ### Changed
--- a/sdk/js/package.json
+++ b/sdk/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@certusone/wormhole-sdk",
-  "version": "0.9.17",
+  "version": "0.9.18",
   "description": "SDK for interacting with Wormhole",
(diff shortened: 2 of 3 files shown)


evan-gray (commented):
Just to check, what are the expected input and output types for Sui and Aptos here, certainly if I pass in the Aptos emitter it will just give me the emitter in hex (and not something more meaningful, which may not be possible anyway). For Sui, when passing in a unit8Array of... a token? it wouldn't return the actual coin type. Just want to be clear on the use case. Would be great to have some unit tests to demonstrate all of these (but I wouldn't block on it right now).

heyitaki:
> Just to check, what are the expected input and output types for Sui and Aptos here, certainly if I pass in the Aptos emitter it will just give me the emitter in hex (and not something more meaningful, which may not be possible anyway). For Sui, when passing in a unit8Array of... a token? it wouldn't return the actual coin type. Just want to be clear on the use case. Would be great to have some unit tests to demonstrate all of these (but I wouldn't block on it right now).

As far as I can see, `uint8ArrayToNative` is mostly used to convert the output of methods such as `getOriginalAsset` (which returns the asset address as a `UInt8Array`) to something that is human readable. For Sui, this array corresponds to the object ID of the CoinMetadata object of the given asset.

evan-gray:
Closing as this has been outstanding for a while and likely better served by [wormhole-sdk-ts](https://github.com/wormhole-foundation/wormhole-sdk-ts)

PR #2954, opened 2023-05-18
root: add go.work file

panoel:


evan-gray (approved):
Finally won't have to open multiple vscode windows / folders! 🙌 
This matches what I get when I run
```
go work init ./node
go work use ./sdk
```

evan-gray:
Closing for now since @bruce-riley pointed out that this could change the build outputs and dependency resolution 😿 Guess we can take a closer look in the future to synchronize the `go.sum`s into the `go.work.sum` and get some additional advisement.

PR #2957, opened 2023-05-19
Update README.md

SamayXD:
Rewrite the crypto warning.


evan-gray:
Not sure there is value to re-writing the existing, already approved disclaimer.

PR #2960, opened 2023-05-20
Fix: strict set to true in clients/js/tsconfig.json

### LINKED ISSUE #2869: clients/js: set `"strict": true` in `tsconfig.json`
Currently, `strict` is set to `false` in `clients/js`, but should absolutely be set to `true` to prevent bugs involving undefined/null values.

haseeb-xd:
issue it resolves #2869


heyitaki:
As I mentioned in #2869, #2885 makes this change and also resolves all resulting compilation errors, so I'm closing this PR in favor of that one.

PR #2964, opened 2023-05-22
Node: Add wormchain deploy admin commands

bruce-riley:
This PR adds guardiand commands to generate prototxt for the Wormchain deploy commands:
```shell
$ build/bin/guardiand template wormchain-store-code --wasm-hash c9262d14fd2f5fe7ca2c272be609be2b25a22f10afe79b4726941da51eb27489
current_set_index:  3
messages:  {
  sequence:  21314180802357929
  nonce:  456824973
  wormchain_store_code:  {
    wasm_hash:  "c9262d14fd2f5fe7ca2c272be609be2b25a22f10afe79b4726941da51eb27489"
  }
}
$ build/bin/guardiand template wormchain-instantiate-contract --code-id 5 --label ibcReceiver --instantiation-msg "{}"
current_set_index:  3
messages:  {
  sequence:  13300675307875321402
  nonce:  3949257006
  wormchain_instantiate_contract:  {
    code_id:  5
    label:  "ibcReceiver"
    instantiation_msg:  "{}"
  }
}
$ build/bin/guardiand template wormchain-migrate-contract --code-id 5 --contract-address wormhole1nc5tatafv6eyq7llkr2gv50ff9e22mnf70qgjlv737ktmt4eswrq0kdhcj --instantiation
-msg "{}"
current_set_index:  3
messages:  {
  sequence:  3161097595250295933
  nonce:  1174708629
  wormchain_migrate_contract:  {
    code_id:  5
    contract:  "wormhole1nc5tatafv6eyq7llkr2gv50ff9e22mnf70qgjlv737ktmt4eswrq0kdhcj"
    instantiation_msg:  "{}"
  }
}
$

### DIFF
--- a/node/cmd/guardiand/admintemplate.go
+++ b/node/cmd/guardiand/admintemplate.go
@@ -532,7 +532,7 @@ func runWormchainStoreCodeTemplate(cmd *cobra.Command, args []string) {
 
 	// Validate the string is the correct length.
 	if len(buf) != 32 {
-		log.Fatal("wasm-hash (expected 32 bytes but received %d bytes)", len(buf))
+		log.Fatalf("wasm-hash (expected 32 bytes but received %d bytes)", len(buf))
 	}
 
 	m := &nodev1.InjectGovernanceVAARequest{
--- a/node/cmd/guardiand/admintemplate.go
+++ b/node/cmd/guardiand/admintemplate.go
@@ -523,9 +523,16 @@ func runWormchainStoreCodeTemplate(cmd *cobra.Command, args []string) {
 	if *wormchainStoreCodeWasmHash == "" {
 		log.Fatal("--wasm-hash must be specified.")
 	}
-	wasmHash, err := parseAddress(*wormchainStoreCodeWasmHash)
+
+	// Validate the string is valid hex.
+	buf, err := hex.DecodeString(*wormchainStoreCodeWasmHash)
 	if err != nil {
-		log.Fatal("failed to parse new implementation address: ", err)
+		log.Fatal("invalid wasm-hash (expected hex): %w", err)
+	}
+
+	// Validate the string is the correct length.
+	if len(buf) != 32 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by panoel — DISMISSED** [review]

File: node/cmd/guardiand/admintemplate.go [hunk 1]
```diff
@@ -472,6 +519,115 @@ func runCircleIntegrationUpgradeContractImplementationTemplate(cmd *cobra.Comman
...
 }
 
+func runWormchainStoreCodeTemplate(cmd *cobra.Command, args []string) {
+	if *wormchainStoreCodeWasmHash == "" {
+		log.Fatal("--wasm-hash must be specified.")
+	}
+	wasmHash, err := parseAddress(*wormchainStoreCodeWasmHash)
+	if err != nil {
+		log.Fatal("failed to parse new implementation address: ", err)
+	}
```
nik-suri, line 529: Why do we call `parseAddress` here? The hash should already be a 32-byte string, no need for padding. Calling `parseAddress` would technically allow a template to be generated with a `base58` or `bech32` encoded string. Not much danger since the VAA then needs to be signed, but can be a little misleading.

Can we use the same validation here that the `adminserver` does? Similar to
```
	// validate the length of the hex passed in
	b, err := hex.DecodeString(req.WasmHash)
	if err != nil {
		return nil, fmt.Errorf("invalid cosmwasm bytecode hash (expected hex): %w", err)
	}

	if len(b) != 32 {
		return nil, fmt.Errorf("invalid cosmwasm bytecode hash (expected 32 bytes but received %d bytes)", len(b))
	}

	wasmHash := [32]byte{}
	copy(wasmHash[:], b)
```

nik-suri (changes requested):

node/cmd/guardiand/admintemplate.go
bruce-riley: Fixed.

nik-suri (approved):

panoel (approved):

PR #2967, opened 2023-05-23
Relayer: Cleanup interfaces and adds untyped ifaces

JoeHowarth:
Added build script to remove UDVTs (aka Wei, GasPrice etc.) from IWormholeRelayer.sol and IRelayProvider.sol 

- Moves util libs to libraries folder
- Removes all structs from public api (besides VaaKey)
- Simplifies VaaKey (only Emitter Sequence since hash doesn't really exist yet)


derpy-duck (approved):
Looks great! 
Can you make sure that the contracts build though before you merge this
(delete contracts/mock, delete forge-test locally, and then run forge build)

PR #2971, opened 2023-05-23
node/near: Fix unit tests

tbjump:
The NEAR watcher unit tests have been intermittently failing, see e.g. 
https://github.com/wormhole-foundation/wormhole/actions/runs/4884237090/jobs/8716699257?pr=2850
https://github.com/wormhole-foundation/wormhole/actions/runs/4884164541/jobs/8716516861
https://github.com/wormhole-foundation/wormhole/actions/runs/4865612488/jobs/8676147944
https://github.com/wormhole-foundation/wormhole/actions/runs/4886760889/jobs/8722535032?pr=2858
https://github.com/wormhole-foundation/wormhole/actions/runs/5061553539/jobs/9085941996?pr=2968

The main issue seems to be in https://github.com/wormhole-foundation/wormhole/pull/2971/files#diff-541054d2cf5abedef3e9ef612b46f581e8d58ffab2479254b60986f57c7c12c2R57, where errors were previously not properly propagated. 

If the error is not properly propagated, the transaction processing job is not re-queued. Errors can occur if the block of the transaction receipt outcome is not yet in the finalized block cache. 

This unit test `TestWatcherDelayedFinal` is supposed to test this cache miss behavior, but due to a race condition it can be that more blocks are polled before the transaction is being processed, hence there not being a cache miss. 

I modified the unit test to cause an increased time between transaction processing and finalization of the outcomeReceipt block, thereby reducing the likelihood of encountering the race condition. Unfortunately, eliminating this race condition in the test would require extensive re-writes. 

This PR also adds some logging and fixes a few minor issues.

### DIFF
--- a/node/pkg/watchers/near/poll.go
+++ b/node/pkg/watchers/near/poll.go
@@ -34,7 +34,6 @@ func (e *Watcher) recursivelyReadFinalizedBlocks(logger *zap.Logger, ctx context
 
 	// SECURITY: Sanity checks for the block header
 	if startBlock.Header.Hash == "" || startBlock.Header.Height == 0 || startBlock.Header.PrevBlockHash == "" {
-		logger.Debug("recursivelyReadFinalizedBlocks: json parse error")
 		return errors.New("json parse error")
 	}
 
@@ -51,9 +50,8 @@ func (e *Watcher) recursivelyReadFinalizedBlocks(logger *zap.Logger, ctx context
 	// we want to avoid going too far back because that would increase the likelihood of error somewhere in the recursion stack.
 	// If we go back too far, we just report the error and terminate early.
 	if recursionDepth > maxFallBehindBlocks {
-		logger.Debug("recursivelyReadFinalizedBlocks: maxFallBehindBlocks")
 		e.eventChan <- EVENT_NEAR_WATCHER_TOO_FAR_BEHIND
-		return nil
+		return errors.New("recursivelyReadFinalizedBlocks: maxFallBehindBlocks")
 	}
 
 	// recursion + stop condition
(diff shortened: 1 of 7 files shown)


tbjump:
After running the tests 5 times in CI they failed, so I have new logs to look at: https://github.com/wormhole-foundation/wormhole/actions/runs/5062973862/jobs/9089682811
Converting to draft.

tbjump:
I replicated the bug by modifying a testcase, identified it, and fixed it. It was actually a logic bug that only occurred due to a race condition in the test.

node/pkg/watchers/near/poll.go
```diff
@@ -34,15 +34,24 @@ func (e *Watcher) recursivelyReadFinalizedBlocks(logger *zap.Logger, ctx context
 
 	// SECURITY: Sanity checks for the block header
 	if startBlock.Header.Hash == "" || startBlock.Header.Height == 0 || startBlock.Header.PrevBlockHash == "" {
+		logger.Debug("recursivelyReadFinalizedBlocks: json parse error")
```
SEJeff, line 37: Idiomatic go says to log errors you don't return. Is it possible to log one level up where you return the error to?

SEJeff (approved):
Nice job!

node/pkg/watchers/near/watcher.go
```diff
@@ -174,7 +174,10 @@ func (e *Watcher) runChunkFetcher(ctx context.Context) error {
 				continue
 			}
 			for _, job := range newJobs {
-				e.schedule(ctx, job, job.delay)
+				err := e.schedule(ctx, job, job.delay)
+				if err != nil {
+					logger.Info("error scheduling transaction processing job", zap.Error(err))
```
bruce-riley, line 179: logger.Error? Same comment at lines 209 and 239.

tbjump: changed it to logger.Error in 211 but Debug in the other two cases and added a reason in the comment: `// Debug-level logging here because it could be very noisy (one log entry for *any* transaction on the NEAR blockchain)`

bruce-riley (approved):

SEJeff (approved):
Tobi 4 Prez

evan-gray (approved):
Approving top level timeout change.

PR #2988, opened 2023-05-25
aptos: pin tag instead of deleted branch

heyitaki:
This is the commit hash of tag `aptos-node-mainnet_20221115.0`


kcsongor (approved):
nice one, thanks 🎉

PR #2995, opened 2023-05-26
CCQ: Add request retries

bruce-riley:
This PR changes it so that query responses flow through the query handler, rather than directly from the watcher to p2p. This allows the handler to keep a cache of pending queries, and retry them. It retries a query every ten seconds for up to a minute, before giving up.

### DIFF
--- a/node/cmd/guardiand/query.go
+++ b/node/cmd/guardiand/query.go
@@ -2,6 +2,7 @@ package guardiand
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"strings"
 
@@ -21,6 +22,8 @@ func handleQueryRequests(
 	signedQueryReqC <-chan *gossipv1.SignedQueryRequest,
 	chainQueryReqC map[vaa.ChainID]chan *gossipv1.SignedQueryRequest,
 	allowedRequestors map[ethCommon.Address]struct{},
+	queryResponseReadC <-chan *common.QueryResponse,
+	queryResponseWriteC chan<- *common.QueryResponsePublication,
 	env common.Environment,
 ) {
 	qLogger := logger.With(zap.String("component", "ccqhandler"))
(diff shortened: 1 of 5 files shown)


node/pkg/common/queryResponse.go
```diff
@@ -29,6 +42,13 @@ type QueryResponsePublication struct {
 	Response EthCallQueryResponse
 }
 
+func (resp *QueryResponsePublication) RequestID() string {
+	if resp == nil || resp.Request == nil {
+		return "nil"
+	}
+	return hex.EncodeToString(resp.Request.Signature)
```
evan-gray, line 49: It's possible we should make this the hash of the request (body + sig for offchain), but this should work for now. Once we add on-chain request support, we will likely need to change the request to a struct anyway. :thinking: Actually wondering if it should be the hash of that first part of the serialized request (chainId + sig or vaa hash)

node/cmd/guardiand/query.go
```diff
@@ -100,3 +202,25 @@ func ccqParseAllowedRequesters(ccqAllowedRequesters string) (map[ethCommon.Addre
...
+		qLogger.Debug("forwarded query request to watcher", zap.String("requestID", pq.reqId), zap.Stringer("chainID", pq.chainId))
+		pq.lastUpdateTime = pq.receiveTime
+	default:
+		// By leaving lastUpdateTime unset, we will retry next interval.
+		qLogger.Warn("failed to send query request to watcher, will retry next interval", zap.String("requestID", pq.reqId), zap.Uint16("chain_id", uint16(pq.chainId)))
+	}
+}
+
+// requestID returns the request signature as a hex string.
+func requestID(req *gossipv1.SignedQueryRequest) string {
```
evan-gray, line 221: I think we can share this with the response somehow.

node/cmd/guardiand/query.go
```diff
@@ -2,34 +2,69 @@ package guardiand
...
 	"google.golang.org/protobuf/proto"
 )
 
-// Multiplex observation requests to the appropriate chain
+const (
+	// requestTimeout indicates how long before a request is considered to have timed out.
+	requestTimeout = 1 * time.Minute
+
+	// retryInterval specifies how long we will wait between retry intervals. This is the interval of our ticker.
+	retryInterval = 10 * time.Second
```
evan-gray, line 26: We should consider moving the ctx timeout in the watcher for the batch call here so that we can easily spot check that it is definitively less than the retryInterval, otherwise requests could pile up.

evan-gray (approved):
I think this all makes sense. Couple comments but don't necessarily need to be addressed now.

node/pkg/common/queryResponse.go
bruce-riley: Made a note to rethink the request ID.

PR #2996, opened 2023-05-26
sdk: add CalculateQuorum documentation and fuzz testing

claudijd:
This PR adds comments to better document CalculateQuorum and additionally adds some fuzz testing and a shortened run to the GitHub action flow to ensure it's always run briefly during each CI run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -270,9 +270,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: "1.20.4"
-      - run: cd sdk/vaa
-      - run: go test
-      - run: go test -v -fuzz FuzzCalculateQuorum -run FuzzCalculateQuorum -fuzztime 15s
+      - run: cd sdk/vaa && go test && go test -v -fuzz FuzzCalculateQuorum -run FuzzCalculateQuorum -fuzztime 15s
   
   # Run Go linters
   node-lint:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -262,7 +262,7 @@ jobs:
           git diff --name-only --exit-code && echo "✅ Generated proto matches committed proto" || (echo "❌ Generated proto differs from committed proto, run \`make proto -B\` and commit the result" >&2 && exit 1)
           make test
 
-  # Verify go sdk unit tests
+  # Verify go sdk unit/fuzz tests
   sdk_vaa:
     runs-on: ubuntu-20.04
     steps:
@@ -270,8 +270,10 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: "1.20.4"
-      - run: cd sdk/vaa && go test
-
+      - run: cd sdk/vaa
+      - run: go test
+      - run: go test -v -fuzz FuzzCalculateQuorum -run FuzzCalculateQuorum -fuzztime 15s
+  
   # Run Go linters
   node-lint:
     # The linter is slow enough that we want to run it on the self-hosted runner
--- a/sdk/vaa/quorum.go
+++ b/sdk/vaa/quorum.go
@@ -5,8 +5,18 @@ package vaa
 // The canonical source is the calculation in the contracts (solana/bridge/src/processor.rs and
 // ethereum/contracts/Wormhole.sol), and this needs to match the implementation in the contracts.
 func CalculateQuorum(numGuardians int) int {
+	// A safety check to avoid caller from ever supplying a negative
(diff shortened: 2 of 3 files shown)


sdk/vaa/quorum_test.go
```diff
@@ -55,3 +55,48 @@ func TestCalculateQuorum(t *testing.T) {
...
+			t.Skip()
+		}
+
+		// Let's determine how many guardians are needed for quorum
+		num := CalculateQuorum(numGuardians)
+
+		// Let's always be sure that there are enough guardians to maintain quorum
+		if num > numGuardians {
+			t.Errorf("%v", "fuzz violation: quorum cannot be acheived because we require more guardians than we have")
+		}
```
SEJeff: Consider using [assert.Less()](https://pkg.go.dev/github.com/stretchr/testify/assert#Less)
```suggestion
		assert.Less(t, num, numGuardians, "fuzz violation: quorum cannot be achieved because we require more guardians than we have")
```

sdk/vaa/quorum_test.go
```diff
@@ -55,3 +55,48 @@ func TestCalculateQuorum(t *testing.T) {
...
+
+		// Let's always be sure that num is never zero
+		if num == 0 {
+			t.Errorf("%v", "fuzz violation: no guardians are required to acheive quorum")
+		}
```
SEJeff: Using [assert.NotZero()](https://pkg.go.dev/github.com/stretchr/testify/assert#NotZero)

```suggestion
		assert.NotZero(t, num, "fuzz violation: no guardians are required to achieve quorum")
```

sdk/vaa/quorum_test.go
```diff
@@ -55,3 +55,48 @@ func TestCalculateQuorum(t *testing.T) {
...
+		// Let's always make sure that the int division does not violate the floor of our float division
+		var floorFloat float64 = 0.66666666666666666
+		numGuardiansFloat := float64(numGuardians)
+		numFloat := float64(num)
+
+		actualFloat := numFloat / numGuardiansFloat
+
+		if actualFloat < floorFloat {
+			t.Errorf("%v", "fuzz violation: quorum has dropped below 2/3rds threshold")
+		}
```
SEJeff: Using [assert.Greater()](https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.Greater)

```suggestion
		assert.Greater(t, actualFloat, floorFloat, "fuzz violation: quorum has dropped below 2/3rds threshold")
```

.github/workflows/build.yml
```diff
@@ -262,16 +262,16 @@ jobs:
...
   sdk_vaa:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
           go-version: "1.20.4"
-      - run: cd sdk/vaa && go test
-
+      - run: cd sdk/vaa && go test && go test -v -fuzz FuzzCalculateQuorum -run FuzzCalculateQuorum -fuzztime 15s
```
claudijd, line 273: I'm currently adding fuzz tests explicitly based on their need to be added, but maybe in if we get a enough of these can add a script harness and then pick them all up.  I think the main consideration is how much time penalty these may or may not add to the run, which is why this is timebound to 15s.

claudijd: The supplied harness however, allows anyone with more interest, compute and willingness to guide the fuzzer to quickly dive in and make some progress there.

sdk/vaa/quorum.go
```diff
@@ -5,8 +5,18 @@ package vaa
...
 		panic("Invalid numGuardians is less than zero")
 	}
+
+	// The goal here is to acheive a 2/3 quorum, but since we're
+	// dividing on int, we need to +1 to avoid the rounding down
+	// effect of integer division
+	//
+	// For example sake, 5 / 2 == 2, but really that's not an
+	//   effective 2/3 quorum, so we add 1 for safety to get to 3
+	//
```
claudijd, line 20: I know this is obvious to most wormhole core devs, but just spelling this out further for the benefit of white-hats and auditors.

SEJeff (commented):
These are suggestions (you'll have to import testify/assert), but I think make the code more straightforward.

I inverted the logic because your logic fired if the bad case happens, mine is an assertion so it only fires if it is false.

SEJeff (approved):

bruce-riley (approved):

evan-gray (approved):
Approving the workflow change

PR #2998, opened 2023-05-26
relayer: Catches executeInstruction reverts & Deals with bad reward addresses

derpy-duck:
- Only decodes the revert message from executeInstruction if it matches one of three known error types
- Revert when ‘pay()’ fails

### DIFF
--- a/ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
+++ b/ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
@@ -358,8 +358,6 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
     Wei transactionFeeRefundAmount,
     ForwardInstruction[] storage forwardInstructions
   ) private {
-     //despite being external, we only allow ourselves to call this function (via CALL opcode)
-    //  used as a means to retroactively revert the call if the calls to the relay provider here revert
     if (msg.sender != address(this))
       revert RequesterNotCoreRelayer();
     Wei wormholeMessageFee = getWormholeMessageFee();
@@ -508,7 +506,7 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
     uint offset = 0;
     bytes4 selector;
 
-    if(revertData.length < 4) {
+    if(revertData.length < 36) {
      return (0, false);
     }
     (selector, offset) = revertData.asBytes4Unchecked(offset);
--- a/ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
+++ b/ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
@@ -418,18 +418,12 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
       receiverValueRefundAmount + (status == DeliveryStatus.FORWARD_REQUEST_SUCCESS ? Wei.wrap(0) : transactionFeeRefundAmount);
     
     //Refund the user
-    try this.payRefundToRefundAddress(
+    refundStatus = payRefundToRefundAddress(
       deliveryInstruction.refundChainId,
       deliveryInstruction.refundAddress,
       refundToRefundAddress,
       deliveryInstruction.refundRelayProvider
-    )
-    returns (RefundStatus _refundStatus) {
-      refundStatus = _refundStatus;
-    } 
-    catch (bytes memory) {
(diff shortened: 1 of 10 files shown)


ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -488,17 +502,17 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
...
   }
 
-  function decodeCancelled(
+  function decodeExecuteInstructionError(
     bytes memory revertData
-  ) private pure returns (uint32 gasUsed, uint256 available, uint256 required) {
+  ) private pure returns (uint32 gasUsed, bool knownError) {
     uint offset = 0;
     bytes4 selector;
     (selector, offset) = revertData.asBytes4Unchecked(offset);
```
scnale, line 510: Do we know that the buffer has at least 4 bytes at this point? We might want to check for the entire exact size or a lower bound for it.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -488,17 +502,17 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
     }
   }
 
-  function decodeCancelled(
+  function decodeExecuteInstructionError(
```
scnale, line 505: Perhaps we could name this `tryDecodeExecuteInstructionError`.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
derpy-duck: Ah good point. Fixed!

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
derpy-duck: done

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
derpy-duck: Ah yes oops, corrected!

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -356,6 +358,10 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
     Wei transactionFeeRefundAmount,
     ForwardInstruction[] storage forwardInstructions
   ) private {
+     //despite being external, we only allow ourselves to call this function (via CALL opcode)
+    //  used as a means to retroactively revert the call if the calls to the relay provider here revert
+    if (msg.sender != address(this))
+      revert RequesterNotCoreRelayer();
```
scnale, line 364: Should this be placed here? This is a private function.

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
```diff
@@ -488,17 +502,21 @@ abstract contract CoreRelayerDelivery is CoreRelayerBase, IWormholeRelayerDelive
...
     (selector, offset) = revertData.asBytes4Unchecked(offset);
-    offset += 28;
-    (gasUsed, offset) = revertData.asUint32Unchecked(offset);
-    (available, offset) = revertData.asUint256Unchecked(offset);
-    (required, offset) = revertData.asUint256Unchecked(offset);
-    assert(offset == revertData.length && selector == Cancelled.selector);
+    if((selector == Cancelled.selector) || (selector == RelayProviderReverted.selector) || (selector == RelayProviderPaymentFailed.selector)) {
+      knownError = true;
+      offset += 28;
+      (gasUsed, offset) = revertData.asUint32Unchecked(offset);
```
scnale, line 518: I think we should have a bounds check for this too. I think I'd just unify the bounds check with the above. We would check that the `revertData` has at least 36 bytes; enough to hold the selector and the ABI encoded `uint32` argument.

derpy-duck: agreed!

ethereum/contracts/relayer/coreRelayer/CoreRelayerDelivery.sol
derpy-duck: ah good catch, fixed

scnale (approved):

PR #3002, opened 2023-05-28
CCQ design

evan-gray:


tbjump:
Thanks for kicking off the work on this, I can see how this could be a really cool and important feature. 

My biggest concerns are: 
1. DoS: The design doc states "DoS mitigation" as a design goal, but does not address it and it is instead mentioned in the "future considerations" section. I think that fees or rate-limiting need to be considered from the start. 
2. Requestor identity: For fees or rate-limiting to be implemented, requests need to be associated to a requestor identity. The design doc vaguely talks about public keys being used, but this needs to be further detailed. 
3. Off-chain request initialization: It is proposed to initiate requests through the gossip network. The gossip network is currently read-only for non-guardians and I think it'd be better to keep it that way. Adding a large amount of data to it could overwhelm it and bring down core consensus. The gossip network is also a scalability bottleneck for wormhole and I could see it being replaced by something more efficient in the future. Another issue is that clients would need to import libp2p, which places unnecessary burden on them and complicates upgradability of the networking layer. I think a better approach would be to have HTTP or gRPC API gateways into the wormhole network. 
4. For a major feature like this there should be a full specification of the API. The design doc provides a sketch but lacks details.

evan-gray:
> Thanks for kicking off the work on this, I can see how this could be a really cool and important feature.
> 
> My biggest concerns are:
> 
> 1. DoS: The design doc states "DoS mitigation" as a design goal, but does not address it and it is instead mentioned in the "future considerations" section. I think that fees or rate-limiting need to be considered from the start.
> 2. Requestor identity: For fees or rate-limiting to be implemented, requests need to be associated to a requestor identity. The design doc vaguely talks about public keys being used, but this needs to be further detailed.
> 3. Off-chain request initialization: It is proposed to initiate requests through the gossip network. The gossip network is currently read-only for non-guardians and I think it'd be better to keep it that way. Adding a large amount of data to it could overwhelm it and bring down core consensus. The gossip network is also a scalability bottleneck for wormhole and I could see it being replaced by something more efficient in the future. Another issue is that clients would need to import libp2p, which places unnecessary burden on them and complicates upgradability of the networking layer. I think a better approach would be to have HTTP or gRPC API gateways into the wormhole network.
> 4. For a major feature like this there should be a full specification of the API. The design doc provides a sketch but lacks details.

1. Currently, in the work-in-progress implementation, the requestors are allow-listed. Certainly before recommending a deploy to production fees and/or rate-limiting would be considered.
2. More details coming soon, but an off-chain request contains the signature of the requestor - that can be recovered to the public key (which identifies them) - in the WIP this needs to match the allow-list. For an on-chain request, that's still TBD but, for example, the `msg.sender` would be identified.
3. That is part of what @kev1n-peters is working on - a REST endpoint which handles submitting the requests to gossip. The guardian could then allow-list the p2p requests from those services if some known parties host them. We can also move the traffic to a separate channel if it becomes a concern. Personally, I don't feel this needs to be solved all at once - we can pick a limited way to start and go from there. 
4. Definitely needs some updates (I thought I left this in draft). 🙂

evan-gray:
Closing while the contributors iterate.

PR #3006, opened 2023-05-30
Client/js: Add submit for all chains

bruce-riley:
This PR extends the `worm submit` command to add the `--all-chains` option, which will submit the VAA to all chains defined on the network (mainnet/testnet), with the following exceptions:
- Only works for `RegisterChain` and `AttestMeta` VAAs.
- Skips the emitter / origin chain.
- The RPC must be defined in `network.ts`
- The target contract must be defined in `consts.ts`.

### DIFF
--- a/clients/js/src/cmds/submit.ts
+++ b/clients/js/src/cmds/submit.ts
@@ -1,8 +1,11 @@
 import {
   assertChain,
+  ChainId,
   ChainName,
   CHAINS,
   coalesceChainName,
+  Contracts,
+  CONTRACTS,
   isEVMChain,
   isTerraChain,
   toChainName,
@@ -19,8 +22,10 @@ import { execute_near } from "../near";
 import { execute_solana } from "../solana";
 import { execute_terra } from "../terra";
 import { assertNetwork } from "../utils";
-import { assertKnownPayload, impossible, parse } from "../vaa";
+import { assertKnownPayload, impossible, parse, Payload, VAA } from "../vaa";
 import { execute_xpla } from "../xpla";
+import { NETWORKS } from "../networks";
+import { Network } from "../utils";
 
 export const command = "submit <vaa>";
 export const desc = "Execute a VAA";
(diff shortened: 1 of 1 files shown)


clients/js/src/cmds/submit.ts
```diff
@@ -67,93 +79,150 @@ export const handler = async (
...
-    throw Error("Wormchain is not supported yet");
-  } else if (chain === "btc") {
-    throw Error("btc is not supported yet");
-  } else if (chain === "sei") {
-    throw Error("sei is not supported yet");
-  } else {
-    // If you get a type error here, hover over `chain`'s type and it tells you
-    // which cases are not handled
-    impossible(chain);
+  for (let chain of chains) {
```
kcsongor, line 144: Two notes about this:
1. The loop body is large, and there's some conditional behaviour based on whether the all chains flag is set to true, it would be nicer if the `execute_` logic was factored out into a separate function so it's easier to read.
2. The `for ... { await doStuff(..) }` pattern sequentialises the async operations, forcing each one to wait for the previous one. A better way is to collect all the actions into an array of promises, and call `Promise.all` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all on them. In pseudo-ish code
```typescript
await Promise.all(chains.map((chain) => execute(chain, vaa)))
```
instead of the loop

bruce-riley: I refactored the code to make the for loop smaller.

Regarding using Promise.all, I considered that, but decided against it. Since the promises would be acknowledged in random order (based on the RPC responses), the output from the various chains would be comingled.

I think the most common use case for the submit command will still be for a single chain, so Promise.all wouldn't help. The most likely case when we will use `--all-chains` is when we are doing registrations for a new deployment, so the extra delay is not a big deal. In that case, I think it's more important for the output to be clear. I also don't think it's worth the additional code complexity.

So I would like to leave the code sequential. If you disagree, maybe we can discuss more off line?

heyitaki (approved):
Generally we should use camelcase instead of snakecase with Typescript var/method names. 
Ex `submit_to_all` => `submitToAll`

Also when accessing object properties, it's better to use dot notation instead of bracket notation (unless it's a dynamic property or the property name has non-alphanumeric letters in it).
Ex: `argv["rpc"] => argv.rpc` but argv["all-chains"] is correct

Don't want to block on these things if you want to push this through and fix them later

clients/js/src/cmds/submit.ts
```diff
@@ -131,24 +177,86 @@ export const handler = async (
...
+        vaa_hex,
+        parsed_vaa,
+        buf,
+        network,
+        chain,
+        undefined,
+        undefined
+      );
+    } catch (e) {
+      console.log(`Failed to submit to ${chain}: `, e);
```
kcsongor, line 259: `console.error` or `console.warn` would be more appropriate here

clients/js/src/cmds/submit.ts
kcsongor: That makes sense, alternatively we could capture the output of each process and return in an array at the end to get the ordering, but it won't make too much of a difference given how infrequently this code needs to be run.

kcsongor (dismissed):

kcsongor (approved):

heyitaki (approved):

PR #3008, opened 2023-05-30
CCQ: Improve retry logic

bruce-riley:


node/pkg/watchers/evm/watcher.go
```diff
@@ -683,22 +688,10 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 							zap.String("block", block),
 							zap.String("component", "ccqevm"),
 						)
+						w.ccqSendQueryResponse(logger, common.QueryRetryNeeded, signedQueryRequest, nil)
```
evan-gray, line 691: We may be able to do some extensive testing to see if this could be a fatal error. We would have to be sure that this happens deterministically and that the above cases passing is informative - e.g. the block exists, but the state has been pruned means retries are not expected to succeed. This may be in contrast to when the block does not yet exist, if the block call fails, a retry may work.

evan-gray (approved):

PR #3009, opened 2023-05-30
Client/js: Add info registrations command

bruce-riley:


clients/js/src/aptos.ts
```diff
@@ -325,3 +328,71 @@ function hex(x: string): Buffer {
...
+      state_object_id = contracts.token_bridge;
+      if (state_object_id === undefined) {
+        throw Error(`Unknown token bridge contract on ${network} for Aptos`);
+      }
+      break;
+    default:
+      throw new Error(`Invalid module: ${module}`);
+  }
+
+  console.log(
```
kcsongor, line 352: I know this is a draft still, but I think it would be nice for this command to output valid json (without any additional logging)

bruce-riley: I addressed this comment in #3035 which replaces this PR.

PR #3017, opened 2023-05-31
CCQ: Validation and marshalling changes

bruce-riley:
This PR eliminates the extra unmarshalling that was happening with query requests. It also factors out the request validation code so it can be more easily reused and tested.


evan-gray (approved):
Validation and cached unmarshal seem reasonable. I'll yield on what the best struct use is.

evan-gray (approved):
Validation and cached unmarshal seem reasonable. I'll yield on what the best struct use is.

PR #3018, opened 2023-05-31
Relayer: Natspec documentation in IWormholeRelayer

derpy-duck:


ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -114,6 +299,42 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
...
+     * The refund from the delivery currently in progress will not be sent to the user; it will instead
+     * be paid to the delivery provider to perform the instruction specified here
+     * 
+     * Publishes an instruction for the delivery provider at 'deliveryProviderAddress' 
+     * to relay a payload and VAAs specified by 'vaaKeys' to the address 'targetAddress' on chain 'targetChain' 
+     * with gas limit 'gasLimit' and with msg.value equal to 
+     * receiverValue + (arbitrary amount that is paid for by paymentForExtraReceiverValue of this chain's wei) in targetChain wei.
+     * 
+     * Any refunds will be sent to 'refundAddress' on chain 'refundChain'
+     * targetAddress must implement the IWormholeReceiver interface
```
nonergodic, line 315: This should probably say that we are now talking about refunds on the new target chain (because current refunds are slapped on to the first forward)

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -105,6 +266,30 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
         uint256 gasLimit
     ) external payable;
 
+    /**
+     * @notice Performs a send, and uses the current refund to fund the 'send'
```
nonergodic, line 270: I think we want to make the main difference between forward an refund clear.

E.g. something along the lines of:
The same as send, except that
1. it only checks whether there are sufficient funds to execute all forwards taking both leftover gas and passed value into account
2. it's only executed after receiveWormholeMessages has completed and hence does not return a sequence number

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -142,6 +399,21 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
         uint8 consistencyLevel
     ) external payable;
 
+    /**
+     * @notice Requests a previously published delivery instruction to be redelivered
```
nonergodic, line 403: I think we want to spell out the primary use case: To switch delivery providers in case of a non-delivery or in case of delivery being too slow. Otherwise the natural question would be "why on god's green earth would I ever want to redeliver anything?"

ethereum/contracts/interfaces/relayer/IWormholeRelayer.sol
```diff
@@ -158,26 +445,71 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
         address newDeliveryProviderAddress
     ) external payable returns (uint64 sequence);
 
+    /**
+     * @notice Returns the price to request a relay to chain 'targetChain', using the default delivery provider
+     * 
+     * @param targetChain destination chain, in Wormhole Chain ID format
```
nonergodic, line 451: I don't think we want to introduce "destination" as additional verbiage - let's just stick to target - the parameter itself is sufficiently clear, imho

PR #3020, opened 2023-06-01
draft: make the docs work

barnjamin:
Locally I had to muck around with versions to get the docs to build. 

truffle was complaining about getting an object instead of a numeric "code" on exit, bumping the version in `ethereum` package.json made it compile.

Also had to bump typedoc version to make the markdown plugin work but I'm not positive we'll stick with it.

With the typedoc change, some type errors in solana/algorand js packages caused doc to fail, so added a directive to tsconfig to ignore libs

draft for now so someone can tell me I was doing it wrong or suggest something else


sdk/js/package.json
```diff
@@ -56,12 +56,13 @@
     "ethers": "^5.6.8",
     "jest": "^27.3.1",
     "prettier": "^2.3.2",
-    "truffle": "^5.7.2",
+    "truffle": "5.7.2",
```
barnjamin, line 59: caret probably shouldn't have disappeared

PR #3028, opened 2023-06-02
relayer: improve delivery process debugging, add utility functions

chase-45:


relayer/generic_relayer/pricing/src/utils/currentPricing.ts
```diff
@@ -0,0 +1,229 @@
...
+        whAddress: await deliveryProvider.getTargetChainAddress(chain),
+      });
+
+      deliveryOverheads.push({
+        chainId: chain,
+        deliveryOverhead: await deliveryProvider.quoteDeliveryOverhead(chain),
+      });
+      maximumBudgets.push({
+        chainId: chain,
+        maximumBudget: await deliveryProvider.maximumBudget(chain),
```
derpy-duck, line 117: Is this a xLabs specific pricing process? or a generic one
if the latter, I don't think we can expect that the delivery provider exposes (maximumBudget, nativeCurrencyPrice, assetConversionBuffer, etc) right

Tiltfile
```diff
@@ -485,22 +485,40 @@ if spy_relayer or redis or generic_relayer or ci_tests:
 
 if generic_relayer:
     k8s_resource(
-        "relayer-engine",
+        "gen-relay-delivery",
```
JoeHowarth, line 488: nit: wormhole relayer instead of generic relayer

relayer/generic_relayer/pricing/src/app.ts
```diff
@@ -0,0 +1,97 @@
...
+        metrics: { registry: app.metricsRegistry },
+      })
+    );
+  }
+  if (opts.fetchSourceTxhash) {
+    app.use(sourceTx());
+  }
+
+  // Set up middleware
+  app.use(async (ctx: PricingContext, next: Next) => {
```
JoeHowarth, line 53: How is this suppossed to be triggered? It appears to be listening for deliveryVaas and running the pricing based off that? What's the rationale here vs. a background job?

JoeHowarth (commented):
Can you write a readme describing how the pricing piece should be used?

Can we also get a review from @solanoepalacio? My understanding was someone on his team was also working on the pricing oracle updater

relayer/generic_relayer/pricing/src/utils/currentPricing.ts
chase-45: This is meant for anyone who intends to use the reference DeliveryProvider contracts

relayer/generic_relayer/pricing/src/app.ts
chase-45: It would be good in the future to trigger based off a cron timer or something similar, but currently the trigger isn't meaningful, I agree

derpy-duck:
Probably review shouldn't come from me (maybe Joe or Solano?) but just quick note that it would be nice if after you merge this to generic-relayer-merge, to cherry-pick that commit into #3042

JoeHowarth (dismissed):

derpy-duck (approved):

PR #3033, opened 2023-06-04
Update Governance.sol

deep0072:
added custom error a gas-efficient way to explain to users why an operation failed through the use of custom errors.


kcsongor:
Thanks for opening this PR @deep0072! Ideally functional changes should not include formatting changes, it makes reviewing the code harder.

I'm going to close this PR for two reasons:
1. Changing the error format is a breaking change, some off-chain integrators might depend on the specific error returned.
2. Such changes are better done consistently, i.e. applied across all of the errors uniformly, as opposed to just picking one.

In general, non-trivial changes like this should first be discussed via either a github discussion or an issue, so that contributors can evaluate them before the implementation is done. (and even then, some quantification of the gas improvement would also be helpful)

deep0072:
Yeah sure. Thanks for explaining. I accidentally put that code in vs code editor. And formatted. At that time I did not notice. Thanks again

PR #3035, opened 2023-06-05
Client/js: Add Info Registrations

bruce-riley:
This PR adds a new worm cli command that can be used to dump the chain registrations for most of the chains:
`worm info registrations testnet aptos TokenBridge`

You can also use the `--verify` option to audit the returned registrations against the values in the `consts.ts` file.

This PR supersedes #3009 which is hopelessly out of date due to extensive refactoring in client/js.

### DIFF
--- a/clients/js/package-lock.json
+++ b/clients/js/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@celo-tools/celo-ethers-wrapper": "^0.1.0",
-        "@certusone/wormhole-sdk": "^0.9.16",
+        "@certusone/wormhole-sdk": "^0.9.18",
         "@cosmjs/encoding": "^0.26.2",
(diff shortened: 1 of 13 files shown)


heyitaki (approved):

kev1n-peters:
This doesn't seem to be working for terra and the cosmwasm chains (terra2, injective, xpla), I'm getting this error:
```
'rpc error: code = Internal desc = Error parsing into type token_bridge_terra::msg::QueryMsg: unknown variant `chain_registration`, expected `wrapped_registry` or `transfer_info`: contract query failed: unknown request'
```
The terra contract doesn't define the `chain_registration` `QueryMsg` variant. The cosmwasm contract does define it, but I don't think it's been deployed yet

bruce-riley:
It is correct that these commands don't work for the cosmwasm chains in mainnet yet. They will when the contracts are eventually updated.

kev1n-peters (approved):

PR #3036, opened 2023-06-05
SDK: tweak typedefs to make doc generation easier

barnjamin:
Typescript doesn't allow us to easily go from string literal types to array so lots of manual effort needs to happen in multiple places to do checks like isXXXChain. 

This modifies the chain name definitions to be arrays first and infer the string literal types from the arrays. 

1) should we export the XXXChainNames const?
2) should we add other chain types even for those that have only 1 instance? e.g. Algorand/Bitcoin
3) Left in a TODO for the terra chains since they're in both the CosmWasm and Terra definitions, do they need to be?

### DIFF
--- a/sdk/js/src/utils/consts.ts
+++ b/sdk/js/src/utils/consts.ts
@@ -41,7 +41,7 @@ export type ChainId = (typeof CHAINS)[ChainName];
  *
  * All the EVM-based chain names that Wormhole supports
  */
-export const EVMChainNames = [
+export const EVMChainNames: ReadonlyArray<ChainName> = [
   "ethereum",
   "bsc",
   "polygon",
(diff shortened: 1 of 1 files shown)


sdk/js/src/utils/consts.ts
```diff
@@ -790,51 +811,33 @@ export function coalesceChainName(chain: ChainId | ChainName): ChainName {
...
-    chainId === CHAIN_ID_BASE ||
-    chainId === CHAIN_ID_SEPOLIA
-  ) {
-    return isEVM(chainId);
+  const chainName = coalesceChainName(chain);
+  if (EVMChainNames.includes(chainName)){
+    return isEVM(chainName as EVMChainName);
   } else {
-    return notEVM(chainId);
+    return notEVM(chainName as NonEVMChainName);
```
kcsongor, line 818: the `notEVM` was used to ensure that the membership check above was performed correctly (i.e. if you missed an EVM chain ID in the conditional, it would complain), but with the array version, the check should just be correct by construction. Thus we should be able to just remove the if-then-else block entirely and return `EVMChainNames.includes(chainName)`.
That way, we can get rid of `notEVM` and `NotEVMChainName` entirely

barnjamin: took the if/else out and type definition I added. Should we also remove the `isEVM`/`notEVM` and `evm_chain_subset`? they're not exported and don't appear used

kcsongor (approved):

kev1n-peters (approved):

PR #3039, opened 2023-06-05
Relayer: Setup for Merging into Main

derpy-duck:
Start of a sequence of PRs to get the branch 'generic-relayer-merge' merged into main! 

- Adds codeowners
- Adds forge to build workflow
- Installs forge in sdk-ci-tests container

Note: We use 'slim' instead of 'alpine' to make forge and several other processes easier to use


devnet/tests.yaml
```diff
@@ -13,7 +13,7 @@ spec:
           command:
             - /bin/sh
             - -c
-            - "sh /app/testing/sdk.sh && touch /app/testing/success"
+            - "bash /app/testing/sdk.sh && touch /app/testing/success"
```
heyitaki, line 16: This command invokes `/bin/sh`, should that be `/bin/bash` instead? What's the motivation behind this change?

testing/Dockerfile.sdk.test
```diff
@@ -1,6 +1,13 @@
-FROM node:18-alpine@sha256:44aaf1ccc80eaed6572a0f2ef7d6b5a2982d54481e4255480041ac92221e2f11
+FROM node:19.6.1-slim@sha256:a1ba21bf0c92931d02a8416f0a54daad66cb36a85d2b73af9d73b044f5f57cfc
```
heyitaki, line 1: Should we use a node 18 slim image?

heyitaki (changes requested):

devnet/tests.yaml
derpy-duck: without this change, here is what happens: https://github.com/wormhole-foundation/wormhole/actions/runs/5189716428/jobs/9355110068

(The SDK-ci-tests container starts up way too early)

testing/Dockerfile.sdk.test
derpy-duck: @chase-45

chase-45: I don't think anything about the CI process is sensitive to 19 vs 18, so either should be acceptable

devnet/tests.yaml
derpy-duck: Oh also currently, this emulates how the readiness check for 'accountant-ci-tests' is done right below in this file

a5-pickle (approved):

heyitaki (approved):

evan-gray (approved):

PR #3042, opened 2023-06-05
Relayer/merge into wormhole relayer engine

derpy-duck:
Note: The time limit for CI needs to be increased because now the relayer engine needs to build the SDK from source (the ethereum folder) in it's docker image. 

Once the SDK is published, I can have the relayer engine docker image simply import the SDK, which should save a lot of time and allow us to reduce the CI time limit again.


Tiltfile
```diff
@@ -452,8 +457,9 @@ docker_build(
     dockerfile = "./ethereum/Dockerfile",
 
     # ignore local node_modules (in case they're present)
-    ignore = ["./ethereum/node_modules"],
+    ignore = ["./node_modules"],
```
JoeHowarth, line 460: should this also ignore ./ethereum/node_modules ?

JoeHowarth (approved):

Tiltfile
derpy-duck: I believe this is a bug that existed earlier. the context is './ethereum' so here it is ignoring './ethereum/node_modules'.

Earlier it was ignoring './ethereum/ethereum/node_modules'

JoeHowarth (approved):

relayer/generic_relayer/relayer-engine-v2/src/app.ts
```diff
@@ -0,0 +1,125 @@
...
+    privateKeys,
+    name,
+    spyEndpoint,
+    redis,
+    redisCluster,
+    redisClusterEndpoints,
+    wormholeRpcs,
+  } = opts;
+  app.spy(spyEndpoint);
+    const store = new RedisStorage({
```
JoeHowarth, line 43: nit: can you run a formatter over this

relayer/generic_relayer/relayer-engine-v2/src/env.ts
```diff
@@ -0,0 +1,278 @@
...
+      ? JSON.parse(process.env.WORMHOLE_RPCS)
+      : base.wormholeRpcs,
+    providers: process.env.BLOCKCHAIN_PROVIDERS
+      ? JSON.parse(process.env.BLOCKCHAIN_PROVIDERS)
+      : base.providers,
+    fetchSourceTxhash: process.env.FETCH_SOURCE_TX_HASH
+      ? JSON.parse(process.env.FETCH_SOURCE_TX_HASH)
+      : base.fetchSourceTxhash,
+
+    // concurrency: Number(process.env.RELAY_CONCURRENCY) || 5,
```
JoeHowarth, line 184: nit: remove comment

JoeHowarth (changes requested):
This is missing @chase-45 's recent pr actually
We should either merge that into this, or do a follow-up pr to main with those changes

JoeHowarth (approved):
Chatted with @derpy-duck and we will merge https://github.com/wormhole-foundation/wormhole/pull/3028 after this pr

Will also address nits in a flup

Tiltfile
```diff
@@ -464,25 +470,61 @@ docker_build(
...
     docker_build(
         ref = "redis",
         context = ".",
         only = ["./third_party"],
         dockerfile = "third_party/redis/Dockerfile",
     )
 
-    k8s_yaml_with_ns("devnet/redis.yaml")
-
+if spy_relayer or redis or ci_tests:
```
evan-gray, line 481: or this

testing/sdk.sh
```diff
@@ -3,4 +3,5 @@ set -e
 while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' guardian:6060/readyz)" != "200" ]]; do sleep 5; done
 while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' spy:6060/metrics)" != "200" ]]; do sleep 5; done
 while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' ibc-relayer:7597/debug/pprof/)" != "200" ]]; do sleep 5; done
+while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' relayer-engine:3000/metrics)" != "200" ]]; do sleep 5; done
```
evan-gray, line 6: does the sdk test rely on relayer-engine now?

.github/workflows/generic-relayer-docker.yml
```diff
@@ -0,0 +1,44 @@
+name: Publish generic relayer container image
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main", "generic-relayer-merge", "gr/re-app-docker-push"]
```
evan-gray, line 6: this should probably just be `main`?

.github/workflows/build.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           kubectl config use-context ci
 
       - run: tilt ci -- --ci --namespace=$DEPLOY_NS --num=2
-        timeout-minutes: 20
+        timeout-minutes: 40
```
evan-gray, line 38: don't love this, but as long as we can bring this back down with optimizations we can live with it being higher temporarily (I hope)

evan-gray (changes requested):
a few quick things - only looked at non-relayer changes

testing/sdk.sh
derpy-duck: Yes, the generic relayer integration tests run as part of SDK-ci-tests, so the relayer engine is needed there

.github/workflows/build.yml
derpy-duck: CI will definitely run pretty quick if the ethereum folder / relayer engine code aren't touched (because the docker images are cached)

If the ethereum folder is touched, now (after all the generic relayer changes) CI will be 6 minutes slower than it would have been before, since we compile contracts in CI using via-ir. 

If the relayer engine is touched, in the current moment there is a significant CI slowdown (at least 6 if not more minutes) because the docker image has to be re-built meaning the ethereum folder (as a dependency) will have to be rebuilt. However, once the change of importing the SDK is made, this slowdown will be much less significant because the relayer engine will no longer have /ethereum as a dependency.

Tiltfile
```diff
@@ -464,25 +470,61 @@ docker_build(
...
+        port_forwards = [
+            port_forward(3003, container_port=3000, name = "Bullmq UI [:3003]", host = webHost),
+        ],
+        labels = ["relayer-engine"],
+        trigger_mode = trigger_mode,
+    )
+    docker_build(
+        ref = "relayer-engine",
+        context = ".",
+        only = ["./ethereum", "./relayer/generic_relayer", "./sdk", "./solana"],
```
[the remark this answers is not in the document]
derpy-duck: The ethereum folder, sdk folder, and solana folders all can be removed from here after the change of importing the SDK is made. It is currently needed because the sdk depends on /ethereum and /solana. But, once the sdk is published, this can be imported easily without all these additional dependencies.

.github/workflows/generic-relayer-docker.yml
derpy-duck: agreed, fixed

Tiltfile
```diff
@@ -464,25 +470,61 @@ docker_build(
     ],
 )
 
-if spy_relayer:
+if spy_relayer or redis or generic_relayer or ci_tests:
```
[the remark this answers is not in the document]
derpy-duck: agreed, fixed

Tiltfile
```diff
@@ -464,25 +470,61 @@ docker_build(
...
             port_forward(6379, name = "Redis Default [:6379]", host = webHost),
         ],
-        labels = ["spy-relayer"],
+        labels = ["redis"],
+        trigger_mode = trigger_mode,
+    )
+
+    k8s_yaml_with_ns("devnet/redis.yaml")
+
+if generic_relayer or ci_tests:
```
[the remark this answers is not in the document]
derpy-duck: agreed, fixed

Tiltfile
derpy-duck: agreed, fixed

JoeHowarth (approved):

evan-gray (approved):
thanks for addressing! reviewed everything outside relayer

PR #3045, opened 2023-06-06
docs: checking in js sdk docs so they can be served from github pages directly

barnjamin:
Preview here: https://barnjamin.github.io/wormhole/sdk-docs/

We need to enable the pages setting to serve from `/docs` which is the only option other than '/' (unless you use gh actions?)

Changed the npm script for docs to move them to `/docs/sdk-docs` so we can have multiple docs sections there. 

It would make sense to write a small `index.html` page to link off to any other docs we stuff in there.

Can we get the URL on the repo front page/description changed to the docs URL once its setup?


evan-gray:
Let's try making a mirror repo quickly to see if that could work 🙂

PR #3046, opened 2023-06-06
Relayer: Price according to whether they want delivery VAA verified

derpy-duck:
Allows a codepath where you can request your delivery to be executed without verifying the delivery vaa

In which case, it will call the IWormholeReceiverUnsafe.receiveWormholeMessagesUnsafe endpoint

---------------------------------
**Note 1: The SLA changes** - it becomes: 'we guarantee that in all scenarios, an event will be emitted for your delivery instruction **where the 'deliveryVaa' input is a correctly signed VAA' and all 'additionalVaa's are correctly signed VAAs**

This means that we'd want the SDK to check that the delivery VAA (and all additional VAAs) are indeed signed correctly to truly validate the SLA. 

As such, I updated the SDK to make this check (so it ignores any relays that have invalid delivery VAAs or invalid additional VAAs)

(I think we should make this part of the SLA whether or not we do not implement this feature, because relayers can pass in invalid 'additionalVaas' even currently and get the event to show up. It should be the case that if you did a relay with invalid 'additionalVaas', the SLA is not satisfied, and we can easily check this when we check for status in the SDK)

Note 2: We also make the optimization of not having the delivery VAA be verified for cross-chain refunds! No reason for this to be verified as long as it is clear that the SLA involves it being signed correctly. 

Note 3: It is important to think about how users could accidently request their delivery to be unverified if they intended for it to be verified. The only way a user could possibly do this is if:

- they use the most generic 'send' and specifically request their delivery VAA to not be verified, by encoding it into executionparams
- they use the most generic 'forward' and specifically request their delivery VAA to not be verified, by encoding it into executionparams

(note: only power users are ever going to touch these most generic functions probably)

- they use the most generic 'forward' and specifically request their delivery VAA to not be verified. If this is a big concern, we can have a seperate forwardVaasToEvmWithoutVerification() endpoint just like we do for send

- they use the function sendVaasToEvmWithoutVerification() endpoint **which doesn't have a payload field**

But in all four of these cases, the user would also have to explicitly implement a IWormholeReceiverUnsafe endpoint that explicitly only has a 'vaas' field and no other fields!
`function receiveWormholeMessagesUnsafe(
        bytes[] memory vaas
    ) external payable;`

### DIFF
--- a/ethereum/contracts/interfaces/relayer/IWormholeRelayerTyped.sol
+++ b/ethereum/contracts/interfaces/relayer/IWormholeRelayerTyped.sol
@@ -200,6 +200,25 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
         VaaKey[] memory vaaKeys,
         uint8 consistencyLevel
     ) external payable returns (uint64 sequence);
+
+    /**
+     * @notice Same as 'sendToEvm' above, except the delivery VAA will not be verified on the target chain, 
+     * 
+     * This function must be called with `msg.value` equal to 
+     * quoteEVMDeliveryPrice(targetChain, receiverValue, gasLimit, deliveryProviderAddress, false) + paymentForExtraReceiverValue
+     */
+    function sendVaasToEvmWithoutVerification(
+        uint16 targetChain,
+        address targetAddress,
+        TargetNative receiverValue,
+        LocalNative paymentForExtraReceiverValue,
+        Gas gasLimit,
+        uint16 refundChain,
+        address refundAddress,
+        address deliveryProviderAddress,
+        VaaKey[] memory vaaKeys,
+        uint8 consistencyLevel
+    ) external payable returns (uint64 sequence);
     
     /**
      * @notice Publishes an instruction for the delivery provider at `deliveryProviderAddress` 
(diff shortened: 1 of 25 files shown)


ethereum/contracts/interfaces/relayer/IWormholeRelayerTyped.sol
```diff
@@ -424,22 +452,32 @@ interface IWormholeRelayerSend is IWormholeRelayerBase {
...
+     *         - newReceiverValue >= receiver value of the old instruction
+     *         - newDeliveryProvider's `targetChainRefundPerGasUnused` >= old relay provider's `targetChainRefundPerGasUnused`
+     *         - (verifyDeliveryVaa, oldVerifyDeliveryVaa) is not (false, true)
      */
     function resendToEvm(
         VaaKey memory deliveryVaaKey,
         uint16 targetChain,
         TargetNative newReceiverValue,
         Gas newGasLimit,
+        bool verifyDeliveryVaa,
```
JoeHowarth, line 480: What's the usecase for allowing this to be changed by a resend? Is it safer to require the original send to specify whether it's verified or not? 

Is there a reason this has to be an option here?

PR #3058, opened 2023-06-07
Docs: CLI README update

barnjamin:
Adding a script to autogenerate the help text for the subcommands.

This should be modified with any new commands and run as arguments are changed to get the README to reflect the options available.

I intend to steal the README from this file for use in the docs site so I'd like all the subcommands to be fully documented.


kcsongor:
Very cool! would it be worth adding a CI action to ensure the README is always up to date?

kcsongor (approved):
nice!

kev1n-peters (dismissed):

.github/workflows/build.yml
```diff
@@ -246,7 +246,10 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "16"
-      - run: cd clients/js && make test
+      - run: |
+          make docs
```
evan-gray, line 250: Is this running `make` in the correct folder? I think this might always pass because it's not `cd`'ed into the clients/js folder, but maybe I'm missing something.

evan-gray (changes requested):

.github/workflows/build.yml
barnjamin: good catch!

barnjamin:
Task Failed Successfully: https://github.com/wormhole-foundation/wormhole/actions/runs/5242695294/jobs/9466426054?pr=3058

I added some space to some of the autogen'd docs to test the gh action

evan-gray (approved):
dope

kcsongor (approved):

PR #3059, opened 2023-06-07
ethereum: fix view error with newer version of forge

claudijd:
In newer versions of forge, an assignment inside a view function generates an error, just tweaking this to it's one less thing to fix on our next forge update.


SEJeff (approved):
The irony is that I have a local change with this exact same update

tbjump (approved):

a5-pickle (approved):

kcsongor:
Can you @claudijd or @SEJeff expand on why this change is needed? I was under the impression that forge doesn't perform any checks on the source code, just invokes solc. And looking at the code, I don't see anything that would make this a non-view function.
Now we get a warning each time the tests run 🤔

PR #3071, opened 2023-06-09
[sdk/js] Add sei native denom check to isNativeCosmWasmDenom

M-Picco:
The condition for the native Sei denomination (`usei`) in `isNativeCosmWasmDenom` is missing, which leads to the `buildTokenId` function returning an id with the first byte set to `00` instead of `01`, which is the intended marker value for native denominations.


kev1n-peters (approved):

evan-gray (approved):
good catch, thanks Martin!

PR #3072, opened 2023-06-09
Node/Watcher/Algorand: Handle Algorand inner transactions

barnjamin:
Currently we're just picking up the first level of inner transactions. Since Algorand can have inner transactions up to 8 layers deep, we need to recurse through to find any calls to the core contract and pass them as observations into the channel.

This pr includes a fix to find all relevant transactions and a (bad) test for the case I was trying to debug.


jumpsiegel (approved):

node/pkg/watchers/algorand/watcher_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/algorand/go-algorand-sdk/client/v2/algod"
+	"github.com/certusone/wormhole/node/pkg/common"
+	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+	"go.uber.org/zap"
+)
+
+// TODO: grab these from const file somewhere?
+const (
+	IDX_URL   = "https://testnet-idx.algonode.cloud"
```
bruce-riley, line 16: Should a go test be accessing external endpoints? Would it be better to either hard code the data here are directly invoke gatherObservations, or move this test into a ci test?

evan-gray: @barnjamin other tests use a hardcoded response as seen in this PR https://github.com/wormhole-foundation/wormhole/pull/3080. could you do something like that?

node/pkg/watchers/algorand/watcher_test.go
```diff
@@ -0,0 +1,62 @@
...
+	err = json.Unmarshal(b, &txn)
+	if err != nil {
+		t.Fatalf("failed to unmarshal block: %s", err)
+	}
+
+	// Because we are using a json blob and the type of logs array is []string
+	// and because go json package will refuse to properly encode/decode
+	// invalid utf8 characters, the json blob has the relevant log encoded as base64
+	// and we base64 decode it and convert it to a string _manually_ so we can
+	// make sure we got the right sequence number
```
barnjamin, line 42: 🤮

panoel (approved):

node/pkg/watchers/algorand/watcher_test.go
barnjamin: added a flat file with a json payload containing the transaction with nested inner transactions.

had to do some gross stuff for the log message, detailed in the comments

bruce-riley (approved):
Thanks for changing the test!

PR #3073, opened 2023-06-12
relayer/Address review. 

derpy-duck:
Only non-comment changes to the contracts being audited can be found in WormholeRelayerGovernance.sol:

We add two new errors: 
```
error ChainAlreadyRegistered(uint16 chainId, bytes32 registeredWormholeRelayerContract);
error InvalidDefaultDeliveryProvider(bytes32 defaultDeliveryProvider);
```

The first one is emitted if the current registered address is not the 0 address, and a registration is requested. This is so we can disallow re-registrations
The second is to disallow setting the default delivery provider to the 0 address (a simple sanity check)

and an event which we emit whenever the contract is upgraded
`event ContractUpgraded(address indexed oldContract, address indexed newContract);`

Additionally, we set the 'allowUnset' flag on parseAndCheckPayloadHeader in parseAndCheckRegisterDefaultDeliveryProviderVm to false. 

This is to disable the feature of having 'governance VAAs' where a default delivery provider is set, and no chainId is specified, meaning this VAA can register the default delivery provider onto any chain.


ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
```diff
@@ -28,16 +28,19 @@ contract DeliveryProvider is DeliveryProviderGovernance, IDeliveryProvider {
...
-        Wei transactionFee =
-            quoteDeliveryOverhead(targetChain) + gasLimit.toWei(quoteGasPrice(targetChain));
-        Wei receiverValueCost = quoteAssetCost(targetChain, receiverValue);
-        nativePriceQuote = (
-            transactionFee.max(costOfProvidingFullGasLimit) + receiverValueCost
-        ).asLocalNative();
+        (uint16 buffer, uint16 denominator) = assetConversionBuffer(targetChain);
+        targetChainRefundPerUnitGasUnused = GasPrice.wrap(gasPrice(targetChain).unwrap() * (denominator) / (uint256(denominator) + buffer));
+        
+        TargetNative targetCostIfNoGasLimitUsed = gasLimit.toWei(targetChainRefundPerUnitGasUnused).asTargetNative();
```
JoeHowarth, line 34: This calc does not appear to do what the var name suggests. As written it is "how much target chain currency would be refunded if no gas was used", not the cost to the user if no gas was used. 
The name implies the fixed overhead is included, but that is added later in the fn. 

Perhaps, max refund is more descriptive?

ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
```diff
@@ -80,12 +83,10 @@ contract DeliveryProvider is DeliveryProviderGovernance, IDeliveryProvider {
         return sourceChainAmount.asNative().convertAsset(
             nativeCurrencyPrice(sourceChain),
             nativeCurrencyPrice(targetChain),
-            (buffer),
+            (bufferDenominator),
```
JoeHowarth, line 86: Just confirming, the denom and num are meant to be swapped here correct?

ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
```diff
@@ -28,16 +28,19 @@ contract DeliveryProvider is DeliveryProviderGovernance, IDeliveryProvider {
...
-        Wei receiverValueCost = quoteAssetCost(targetChain, receiverValue);
-        nativePriceQuote = (
-            transactionFee.max(costOfProvidingFullGasLimit) + receiverValueCost
-        ).asLocalNative();
+        (uint16 buffer, uint16 denominator) = assetConversionBuffer(targetChain);
+        targetChainRefundPerUnitGasUnused = GasPrice.wrap(gasPrice(targetChain).unwrap() * (denominator) / (uint256(denominator) + buffer));
+        
+        TargetNative targetCostIfNoGasLimitUsed = gasLimit.toWei(targetChainRefundPerUnitGasUnused).asTargetNative();
+        TargetNative targetCostIfFullGasLimitUsed = gasLimit.toWei(gasPrice(targetChain)).asTargetNative();
+        LocalNative costIfNoGasLimitUsed = quoteAssetCost(targetChain, targetCostIfNoGasLimitUsed);
```
JoeHowarth, line 36: can this this use quoteGasCost?

ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
```diff
@@ -113,28 +114,37 @@ contract DeliveryProvider is DeliveryProviderGovernance, IDeliveryProvider {
...
-        Wei gasPriceInSourceChainCurrency =
-            assetConversion(targetChain, gasPrice(targetChain).priceAsWei(), chainId());
-        require(gasPriceInSourceChainCurrency.unwrap() <= type(uint88).max, "Overflow");
-        return GasPrice.wrap(uint88(gasPriceInSourceChainCurrency.unwrap()));
+    function quoteGasPrice(uint16 targetChain) public view returns (GasPrice price) {
+        price = GasPrice.wrap(quoteGasCost(targetChain, Gas.wrap(1)).unwrap());
+        require(price.unwrap() <= type(uint88).max, "Overflow");
     }
 
+    error PriceIsZero(uint16 chain);
```
JoeHowarth, line 134: Nit: move this to interface, or top of file

ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerGovernance.sol
```diff
@@ -93,7 +95,11 @@ abstract contract WormholeRelayerGovernance is WormholeRelayerBase, ERC1967Upgra
             foreignAddress;
     }
 
+    event ContractUpgraded(address indexed oldContract, address indexed newContract);
```
JoeHowarth, line 98: I believe ERC1967Upgrade already emits a similar event

JoeHowarth (changes requested):

ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerGovernance.sol
derpy-duck: yup - this is the same event from TokenBridge, NFTBridge, CoreBridge, etc for consistency

ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
derpy-duck: done

ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
derpy-duck: fixed

JoeHowarth (approved):

PR #3077, opened 2023-06-12
node: remove pre-bedrock optimism code

panoel:
Now that Optimism Bedrock is in mainnet, there is no need to support pre-bedrock Optimism.


bruce-riley (approved):

evan-gray (approved):
🙌

PR #3080, opened 2023-06-13
Node: Make IBC watcher publish wormchain version

bruce-riley:
This PR adds the wormchain version string to the IBC feature flag. The flag will look something like this:
`ibc:sei|v0.0.1` where `v0.0.1` is the wormchain version string from devnet.

### DIFF
--- a/node/pkg/watchers/ibc/watcher_test.go
+++ b/node/pkg/watchers/ibc/watcher_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


node/pkg/watchers/ibc/watcher_test.go
```diff
@@ -213,3 +214,39 @@ func TestParseIbcAllChannelChainsQueryResults(t *testing.T) {
...
+	assert.Equal(t, "http://wormchain:26657", convertWsUrlToHttpUrl("ws://wormchain:26657"))
+	assert.Equal(t, "http://wormchain:26657", convertWsUrlToHttpUrl("wormchain:26657"))
+}
+
+func TestParseAbciInfoResults(t *testing.T) {
+	// This came from the following query: http://localhost:26659/abci_info
+	respJson := []byte(`
+{
+  "jsonrpc": "2.0",
+  "id": -1,
```
panoel, line 229: It's okay to use a negative number?  That works?
Learn something new everyday...

panoel (approved):

PR #3084, opened 2023-06-13
tilt: deploy core and token bridge to wormchain

panoel:
Deploy core and token bridge contracts in tilt

### DIFF
--- a/wormchain/contracts/tools/deploy_wormchain.ts
+++ b/wormchain/contracts/tools/deploy_wormchain.ts
@@ -42,10 +42,10 @@ const readFileAsync = util.promisify(fs.readFile);
 type ContractName = string;
 const artifacts: ContractName[] = [
   "global_accountant.wasm",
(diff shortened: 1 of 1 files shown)


wormchain/contracts/tools/deploy_wormchain.ts
```diff
@@ -42,6 +42,9 @@ const readFileAsync = util.promisify(fs.readFile);
 type ContractName = string;
 const artifacts: ContractName[] = [
   "global_accountant.wasm",
+  "cw_wormhole.wasm",
+  "cw20_wrapped_2.wasm",
+  "cw_token_bridge.wasm",
```
evan-gray, line 47: Changing the order here may change the addresses (which might cause some inconsistencies). We should probably double check the IBC receiver didn't change.

evan-gray (dismissed):

evan-gray (approved):

PR #3093, opened 2023-06-14
Ccq: Messaging changes

bruce-riley:
This PR switches from using protobuf encoding for query requests to using binary encoding similar to what is used for VAAs. This is so the messages can more efficiently be encoded and decoded on chain.


panoel:
What is that 60 MB binary `send_req` file?

node/cmd/guardiand/query_test.go
```diff
@@ -410,80 +407,86 @@ func TestInvalidQueries(t *testing.T) {
...
-	require.Nil(t, md.waitForResponse())
-
-	// A per-chain query with no call data should fail.
-	md.resetState()
-	perChainQueries = []*gossipv1.PerChainQueryRequest{createPerChainQueryForTesting(vaa.ChainIDPolygon, "0x28d9630", 0)}
-	signedQueryRequest, _ = createSignedQueryRequestForTesting(md.sk, perChainQueries)
-	md.signedQueryReqWriteC <- signedQueryRequest
-	require.Nil(t, md.waitForResponse())
+	/*
+		 TODO: These tests should be done in common/query_test.go. Make sure they are there.
```
panoel, line 436: The directive to make sure the tests are there is for future you and not the code inspectors, correct?

panoel (approved):

node/pkg/common/queryResponse.go
```diff
@@ -206,103 +149,182 @@ func UnmarshalQueryResponsePublication(data []byte) (*QueryResponsePublication,
...
+//
+
+// Marshal marshalls a per chain query response.
+func (pcr *PerChainQueryResponse) Marshal() ([]byte, error) {
+	if err := pcr.Validate(); err != nil {
+		return nil, err
+	}
+
+	buf := new(bytes.Buffer)
+	vaa.MustWrite(buf, binary.BigEndian, pcr.ChainID)
```
evan-gray, line 244: If it's possible that ChainID isn't unique in a request (which seems reasonable), then should we avoid writing it here if this response list should match up 1-1 with the request list?

evan-gray (changes requested):
I think the request is good, but the response marshalling might need a little tweaking. I like the direction it's going with the typing overall though.

node/cmd/guardiand/query_test.go
bruce-riley: Correct. The tests have been updated and the TODO removed.

node/pkg/common/queryRequest.go
```diff
@@ -14,15 +14,63 @@ import (
...
+type ChainSpecificQuery interface {
+	Type() ChainSpecificQueryType
+	Marshal() ([]byte, error)
+	Unmarshal(data []byte) error
+	UnmarshalFromReader(reader *bytes.Reader) error
+	Validate() error
+}
+
+// ChainSpecificQueryType is used to interpret the data in a per chain query request.
+type ChainSpecificQueryType uint8
```
panoel, line 43: What kind of values do you think will be used here?  Is there any way it can explode to someone wanting more than 256 different values?

node/pkg/common/queryRequest.go
```diff
@@ -50,233 +98,350 @@ func PostSignedQueryRequest(signedQueryReqSendC chan<- *gossipv1.SignedQueryRequ
...
+		queryRequest.PerChainQueries = append(queryRequest.PerChainQueries, &perChainQuery)
 	}
+
+	return nil
+}
+
+// Validate does basic validation on a received query request.
+func (queryRequest *QueryRequest) Validate() error {
+	// Nothing to validate on the Nonce.
+	if len(queryRequest.PerChainQueries) == 0 {
```
panoel, line 160: This is a Go question, I guess.  len returns an int instead of a uint.  In general, should we check for <= 0 or == 0 is sufficient?

node/pkg/common/queryRequest.go
```diff
@@ -50,233 +98,350 @@ func PostSignedQueryRequest(signedQueryReqSendC chan<- *gossipv1.SignedQueryRequ
...
-		default:
-			return fmt.Errorf("received invalid message from query module")
+		if len(callData.To) != EvmContractAddressLength {
+			return fmt.Errorf("invalid length for To contract")
+		}
+		if callData.Data == nil || len(callData.Data) == 0 {
+			return fmt.Errorf("no call data data")
+		}
+		if len(callData.Data) > math.MaxUint32 {
+			return fmt.Errorf("request data too long")
```
panoel, line 405: request or call data?

node/pkg/common/queryResponse.go
```diff
@@ -34,72 +33,67 @@ const (
...
-	QUERY_REQUEST_TYPE_ETH_CALL = uint8(1)
-)
 
-func (resp *QueryResponsePublication) RequestID() string {
-	if resp == nil || resp.Request == nil {
-		return "nil"
-	}
-	return hex.EncodeToString(resp.Request.Signature)
+	// Results is the array of responses matching CallData in EthCallQueryRequest
+	Results [][]byte
```
panoel, line 85: Out of curiosity, why is Number on line 80 a big.Int when the (un)marshal treats it like a uint64?

panoel (approved):

node/pkg/common/queryRequest.go
bruce-riley: It's hard to imagine more than 255 query types? (You know, like it was hard to imagine Unix lasting longer than 2^32 seconds since 1970. . .) Seriously though, I think this is fine.

node/pkg/common/queryRequest.go
bruce-riley: Seems like something lint might complain about some day. I changed it.

node/pkg/common/queryResponse.go
bruce-riley: Changed it to use uint64 in the internal struct.

node/pkg/common/queryRequest.go
evan-gray: worst case, 255 can always be an "extended" type with more sub-types.

node/pkg/common/queryResponse.go
evan-gray: We can play with this later 🙂

evan-gray (approved):
The serialization looks reasonable. We can revisit if we want to repeat the chain id and type in the response. 👍

PR #3105, opened 2023-06-19
Docs: fix Aptos worm send message command and URL to view events

barnjamin:
Worm command is missing a couple params.

The current cURL results in a `resource_not_found` message. I think this is just because the wrong address is used so changed it to the core contract address.


kcsongor (approved):
nice!

evan-gray (approved):

PR #3111, opened 2023-06-21
sdk: make explicit types for transferPayloads

0xksure:
This PR exports the `TransferPayload` so it's possible for clients to use them . 

This is my first PR into wormhole so I would appreciate some feedback on whether this is a good idea, if the types will be exposed to clients and if they are defined in the right place.


sdk/js/src/utils/parseVaa.ts
```diff
@@ -27,49 +48,35 @@ export function parseNFTPayload(payload: Buffer) {
...
+export function parseTransferPayload(payload: Buffer): TokenTransferPayload {
   const parsed = parseTokenTransferPayload(payload);
   return {
     amount: parsed.amount,
     originAddress: parsed.tokenAddress.toString("hex"),
     originChain: parsed.tokenChain,
     targetAddress: parsed.to.toString("hex"),
     targetChain: parsed.toChain,
-    fee: parsed.fee === null ? undefined : parsed.fee,
+    fee: parsed.fee === null ? null : parsed.fee,
```
kev1n-peters, line 76: Is there a reason for changing this from `undefined` to `null`? Would prefer to keep it as `undefined` if possible to not break anyone not expecting `null`.

0xksure: The reason was because the `TokenTransfer` uses `fee: bigint | null` and since `TokenTransferPayload` uses fee it would make sense to use the same type for consistency. But yea this might be breaking so can revert to undefined.

evan-gray (approved):
Not sure about the usefulness of the `Pick` here, but approving because I agree with having types

kev1n-peters (approved):

PR #3113, opened 2023-06-21
ccq: query server

kev1n-peters:
The CCQ Query Server is an HTTP server / RESTful service that handles query requests, publishes them to the gossip network, waits for the responses to reach quorum, and then returns the response.

### DIFF
--- a/node/cmd/ccq/http.go
+++ b/node/cmd/ccq/http.go
@@ -1,0 +1,133 @@
+package ccq
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sort"
+	"time"
+
+	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
+	"github.com/certusone/wormhole/node/pkg/query"
+	"github.com/gorilla/mux"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"google.golang.org/protobuf/proto"
+)
+
+type queryRequest struct {
+	Bytes     string `json:"bytes"`
+	Signature string `json:"signature"`
+}
+
+type queryResponse struct {
+	Bytes      string   `json:"bytes"`
+	Signatures []string `json:"signatures"`
+}
+
+type httpServer struct {
+	topic            *pubsub.Topic
+	pendingResponses *PendingResponses
+}
+
+func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
+	var q queryRequest
+	err := json.NewDecoder(r.Body).Decode(&q)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	queryRequestBytes, err := hex.DecodeString(q.Bytes)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


node/cmd/ccq/http.go
```diff
@@ -0,0 +1,143 @@
...
+	err = s.topic.Publish(r.Context(), b)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		s.pendingResponses.Remove(pendingResponse)
+		return
+	}
+
+	// Wait for the response or timeout
+	select {
+	case <-time.After(1 * time.Minute):
```
bruce-riley, line 89: The query handler waits a minute. Maybe this should wait a bit longer to allow for roundtrip over gossip?

PR #3112 makes `query.RequestTime` available. Maybe you could use that plus some number of seconds?

node/cmd/ccq/p2p.go
```diff
@@ -0,0 +1,267 @@
...
+					if len(responses[requestSignature][digest]) >= q {
+						s := &SignedResponse{
+							Response:   &queryResponse,
+							Signatures: responses[requestSignature][digest],
+						}
+						delete(responses, requestSignature)
+						pendingResponse.ch <- s
+					}
+				} else {
+					logger.Warn("received observation by unknown guardian - is our guardian set outdated?",
```
bruce-riley, line 254: I wonder if this should flip a message to the guardian set routine to make it read immediately? Maybe not worth attempting to retry this message, but might reduce the length of an outage due to a guardian set update.

node/cmd/ccq/http.go
kev1n-peters: I think for now we could simply wait an additional 5 seconds for the message to propagate

node/cmd/ccq/p2p.go
kev1n-peters: We decided to just fetch the guardian set once at startup. For now the plan is to have some sort of manual intervention during guardian set upgrades

node/cmd/ccq/p2p.go
```diff
@@ -0,0 +1,237 @@
...
+				}
+				requestSignature := hex.EncodeToString(queryResponse.Request.Signature)
+				// Check that we're handling the request for this response
+				pendingResponse := pendingResponses.Get(requestSignature)
+				if pendingResponse == nil {
+					logger.Debug("skipping query response for unknown request", zap.String("signature", requestSignature))
+					continue
+				}
+				// Make sure that the request bytes match
+				if !bytes.Equal(queryResponse.Request.QueryRequest, pendingResponse.req.QueryRequest) ||
```
panoel, line 179: Is it possible that some request gets stuck in here from a previous error leg and all future responses get passed over because the requests don't match?

panoel (dismissed):

node/cmd/ccq/p2p.go
kev1n-peters: The request signature is removed from the `pendingResponses` container after the http request is handled (either successfully or an error) so future requests with the same signature shouldn't get stuck

bruce-riley (approved):

panoel (approved):

PR #3119, opened 2023-06-23
relayer/deploy generic relayer mainnet

derpy-duck:


ethereum/ts-scripts/relayer/config/mainnet/chains.json
```diff
@@ -0,0 +1,76 @@
+{
+  "guardianSetIndex": 3,
+  "description": "This file contains the chains against which all the scripts should run.",
+  "chains": [
+    {
+      "description": "Ethereum",
+      "evmNetworkId": 1,
+      "chainId": 2,
+      "rpc": "https://rpc.ankr.com/eth",
+      "wormholeAddress": "0x98f3c9e6E3fAce36bAAd05FE09d375Ef1464288B"
```
kcsongor, line 10: why duplicate these from the sdk consts file?

derpy-duck: This is to keep consistency with how the rest of the configs are done. I think ideally these should not be duplicated here though, and should be taken from the SDK - but probably that should be done in a seperate PR along with all of the other files in relayer/config?

ethereum/ts-scripts/relayer/config/mainnet/scriptConfigs/configureDeliveryProvider.json
```diff
@@ -0,0 +1,176 @@
+{
+  "addresses": [
+    {
+      "chainId": 2,
+      "rewardAddress": "0x53207E216540125e322CdA8A693b0b89576DEb46",
+      "approvedSenders": [
+        {
+          "address": "0xdFE31C0aE49bc9ad7F184977064fBefef5d76ee9",
+          "approved": true
+        }
```
scnale, line 10: I just realized these fields are obsolete.

JoeHowarth (approved):

JoeHowarth (approved):

PR #3129, opened 2023-06-26
node: graceful shutdown

tbjump:
This helps with the benchmarking (next PR on the stack) if there are multiple different benchmarks in a row.

### DIFF
--- a/node/pkg/node/options.go
+++ b/node/pkg/node/options.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/benbjohnson/clock"
 	"github.com/certusone/wormhole/node/pkg/accountant"
@@ -16,6 +17,7 @@ import (
 	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
 	"github.com/certusone/wormhole/node/pkg/readiness"
 	"github.com/certusone/wormhole/node/pkg/reporter"
+	"github.com/certusone/wormhole/node/pkg/supervisor"
 	"github.com/certusone/wormhole/node/pkg/watchers"
 	"github.com/certusone/wormhole/node/pkg/watchers/ibc"
 	"github.com/certusone/wormhole/node/pkg/watchers/interfaces"
@@ -151,7 +153,7 @@ func GuardianOptionGovernor(governorEnabled bool) *GuardianOption {
 func GuardianOptionStatusServer(statusAddr string) *GuardianOption {
 	return &GuardianOption{
 		name: "status-server",
-		f: func(ctx context.Context, logger *zap.Logger, g *G) error {
+		f: func(_ context.Context, _ *zap.Logger, g *G) error {
 			if statusAddr != "" {
 				// Use a custom routing instead of using http.DefaultServeMux directly to avoid accidentally exposing packages
 				// that register themselves with it by default (like pprof).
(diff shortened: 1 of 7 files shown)


node/pkg/node/options.go
```diff
@@ -171,11 +173,32 @@ func GuardianOptionStatusServer(statusAddr string) *GuardianOption {
...
+				}
+
+				g.runnables["status-server"] = func(ctx context.Context) error {
+					logger := supervisor.Logger(ctx)
+					go func() {
+						if err := server.ListenAndServe(); err != http.ErrServerClosed {
+							logger.Error("status server crashed", zap.Error(err))
+						}
+					}()
+					time.Sleep(time.Millisecond * 10) // we don't know exactly when the server will be ready to accept connections, but a few ms should be enough time
```
SEJeff, line 192: Famous last words :)

node/pkg/p2p/p2p.go
```diff
@@ -224,9 +224,35 @@ func Run(
...
+			<-ctx.Done()
+			logger.Info("context canceled, shutting down p2p...")
+
+			if sub != nil {
+				sub.Cancel()
+			}
+
+			if th != nil {
+				if err := th.Close(); err != nil {
+					logger.Error("Error closing the topic", zap.Error(err))
```
SEJeff, line 241:
```suggestion
					logger.Error("error closing the topic", zap.Error(err))
```

node/pkg/p2p/p2p.go
```diff
@@ -224,9 +224,35 @@ func Run(
...
+				}
+			}
+
+			if err := h.Close(); err != nil {
+				logger.Error("Error closing the host", zap.Error(err))
```
SEJeff, line 246:
```suggestion
				logger.Error("error closing the host", zap.Error(err))
```

node/pkg/node/options.go
tbjump: should be fine, nothing depends on this.

bruce-riley (dismissed):

node/pkg/p2p/p2p.go
```diff
@@ -247,9 +247,35 @@ func Run(
 			panic(err)
 		}
 
+		var sub *pubsub.Subscription
+		var th *pubsub.Topic
+
+		go func() {
```
evan-gray, line 253: How will setting this up here interact if something fails below and the function returns? Would it be better (and less lines of code) to put this cleanup at the end of the function?

node/pkg/node/options.go
evan-gray: seems hacky but :shrug: what is the importance of waiting here?

evan-gray (dismissed):
Couple questions, but generally on board.

node/pkg/p2p/p2p.go
tbjump: I changed it to have defers instead of waiting on ctx.Done in a go-routine. I also split it up into multiple defers right after the resource creation, which I think is the more common practice. Lmk what you think.

node/pkg/node/options.go
tbjump: I actually just removed this since it's not necessary and probably just adds more confusion.

node/pkg/node/publicrpcRunnable.go
```diff
@@ -15,21 +16,28 @@ import (
 	"google.golang.org/grpc"
 )
 
-func publicrpcTcpServiceRunnable(logger *zap.Logger, listenAddr string, publicRpcLogDetail common.GrpcLogDetail, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) (supervisor.Runnable, error) {
-	l, err := net.Listen("tcp", listenAddr)
+func publicrpcTcpServiceRunnable(logger *zap.Logger, listenAddr string, publicRpcLogDetail common.GrpcLogDetail, db *db.Database, gst *common.GuardianSetState, gov *governor.ChainGovernor) supervisor.Runnable {
+	return func(ctx context.Context) error {
+		l, err := net.Listen("tcp", listenAddr)
```
evan-gray, line 21: What happened to the deferred `l.close` you had added?

evan-gray (approved):

node/pkg/p2p/p2p.go
evan-gray: don't immediately see why the defers wouldn't work and I don't recall if there was a reason I hadn't done that in another script.

bruce-riley (approved):

node/pkg/node/publicrpcRunnable.go
tbjump: that's actually already in `supervisor.GRPCServer(grpcServer, l, false)`.

PR #3130, opened 2023-06-26
node/processor: add mutex on state

tbjump:


SEJeff:
Please add context on what this change is for and why it is necessary. It wasn't apparently necessary before. What does this do and how does the network benefit from this change?

tbjump:
This change seems to be actually not necessary. Closing.

PR #3131, opened 2023-06-26
Ensure optin transaction is the current transaction and address the a…

marcosnc:
…lgo_seed transaction in a relative manner.

### DIFF
--- a/algorand/admin.py
+++ b/algorand/admin.py
@@ -1151,7 +1151,7 @@ class PortalCore:
             off += 2
             ret["Fee"] = vaa[off:(off + 32)].hex()
 
-        if int.from_bytes((vaa[off:off+1]), "big") == 3:
+        if len(vaa[off:]) > 133 and int.from_bytes((vaa[off:off+1]), "big") == 3:
             ret["Meta"] = "TokenBridge Transfer With Payload"
             ret["Type"] = int.from_bytes((vaa[off:off+1]), "big")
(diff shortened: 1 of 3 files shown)


hernandp (approved):
looks good ;)

jumpsiegel (approved):

jumpsiegel (approved):

marcosnc:
> The change seems sensible, it allows the optin transaction to appear anywhere within a group rather than only at the beginning, right? I'm a little confused why `admin.py` needs to change due to this -- in the comment message you mention that it's to fix a test. Can you explain why the test fails after allowing optin to appear anywhere within a group, and how the length check fixes that failure? Thanks!

Indeed, the change ensures the validations are done on the actual optin, the one received, instead of the one in the absolute position 1.
Regarding the `admin.py` change, it seemed the test was failing with something not related to this change and I followed the recommendations from @barnjamin.

barnjamin:
The admin.py change was necessary because the VAA parser was overwriting some fields since the test VAA happened to have a `3` in the offset being checked. The change here is a bit smelly but it works to make the tests pass. 

Fwiw I did rework the parser today here, though its not thoroughly tested.
https://github.com/barnjamin/wh-metrics/blob/main/vaa.py

kcsongor:
thanks @marcosnc!

kcsongor (approved):

PR #3096, opened 2023-06-14
node: Wait for bootstrap guardian in devnet

tbjump:
This PR fixes a race condition in devnet where the p2p network may silently fail to start because at the time guardian-1 tries to connect to guardian-0 (bootstrat node) it may not be online yet. The race condition was more frequenly hit after a refactor, see https://github.com/wormhole-foundation/wormhole/pull/3024#issuecomment-1591803726

PR #3128, opened 2023-06-26
node/node_test: improve logging

tbjump:
This PR drastically decreases the log volume in the node_test.go.

PR #3127, opened 2023-06-26
node/node: Move options to options.go

tbjump:
This PR only moves the Guardian from node.go to options.go and documents them. 

How to review: Diff node.go and options.go and ensure there are no functional changes.

PR #3097, opened 2023-06-15
ci: set Tilt guardiand_loglevel=info

tbjump:
There seem to be multiple race conditions in the Tilt tests and they are hard to reproduce. I therefore propose to set `guardiand_loglevel=info` in CI. 

https://github.com/wormhole-foundation/wormhole/issues/3052

PR #3134, opened 2023-06-27
node/p2p: improve startup

tbjump:
This change reduces the runtime of TestConsensus() in node_test.go from ~16s to ~3s and makes startup more reliable in devnet. 

# Delay first heartbeat by 2 seconds
Before this change, the heartbeat was sent immediately (after 1ns), but by then the network connections are not established yet and the heartbeat gets lost, so the tests needed to wait another 15s for the next heartbeat. 

After this change, we wait 2s to send out the first heartbeat, and by then the network is likely to be established. 

# Enforce connection to bootstrap node on startup
This change solves a problem during Guardian startup in devnet / Tilt: 
guardian-1 may start up before guardian-0 and therefore has no bootstrap node to connect to. 
guardian-1's messages then go into the void. 
This has been previously addressed in https://github.com/wormhole-foundation/wormhole/pull/3096 by essentially having guardian-1 wait for 10s before startup. 
However, this 10s number is arbitrary and we don't want to depend on it. With this change, we instead force a connection to the bootstrap node and kill the guardian if not successful (at which point it would get restarted by Kubernetes/Tilt).

### DIFF
--- a/node/pkg/p2p/p2p.go
+++ b/node/pkg/p2p/p2p.go
@@ -255,6 +255,49 @@ func Run(
 			return fmt.Errorf("failed to subscribe topic: %w", err)
 		}
 
+		// Make sure we connect to at least 1 bootstrap node (this is particularly important in a local devnet and CI
+		// as peer discovery can take a long time).
+
+		// Count number of successful connection attempts. If we fail to connect to any bootstrap peer, kill the service
+		// TODO: Currently, returning from this function will lead to rootCtxCancel() being called in the defer() above.
+		// 		The service will then be restarted by Tilt/kubernetes
+		successes := 0
+		// Are we a bootstrap node? If so, it's okay to not have any peers.
+		bootstrapNode := false
+
+		for _, addr := range strings.Split(bootstrapPeers, ",") {
+			if addr == "" {
+				continue
+			}
+			ma, err := multiaddr.NewMultiaddr(addr)
+			if err != nil {
+				logger.Error("Invalid bootstrap address", zap.String("peer", addr), zap.Error(err))
+				continue
+			}
+			pi, err := peer.AddrInfoFromP2pAddr(ma)
+			if err != nil {
+				logger.Error("Invalid bootstrap address", zap.String("peer", addr), zap.Error(err))
+				continue
+			}
+
+			if pi.ID == h.ID() {
+				logger.Info("We're a bootstrap node")
+				bootstrapNode = true
+				continue
+			}
+
+			if err = h.Connect(ctx, *pi); err != nil {
+				logger.Error("Failed to connect to bootstrap peer", zap.String("peer", addr), zap.Error(err))
+			} else {
+				successes += 1
+			}
+		}
+
+		if successes == 0 && !bootstrapNode {
+			return fmt.Errorf("failed to connect to any bootstrap peer")
+		}
+		logger.Info("Connected to bootstrap peers", zap.Int("num", successes))
+
 		logger.Info("Node has been started", zap.String("peer_id", h.ID().String()),
 			zap.String("addrs", fmt.Sprintf("%v", h.Addrs())))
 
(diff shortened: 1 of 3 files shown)


tbjump:
* main
  * **PR #3097** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3097" target="_blank">Graphite</a> 
    * **PR #3096** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3096" target="_blank">Graphite</a> 
      * **PR #3024** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3024" target="_blank">Graphite</a> 
        * **PR #3067** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3067" target="_blank">Graphite</a> 
          * **PR #3134** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3134" target="_blank">Graphite</a>  👈
          * **PR #3127** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3127" target="_blank">Graphite</a> 
            * **PR #3128** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3128" target="_blank">Graphite</a> 
              * **PR #3129** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3129" target="_blank">Graphite</a> 
                * **PR #3130** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3130" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3134?utm_source=stack-comment).

node/pkg/p2p/p2p.go
```diff
@@ -280,7 +280,9 @@ func Run(
 			ourAddr := ethcrypto.PubkeyToAddress(gk.PublicKey)
 
 			ctr := int64(0)
-			timer := time.NewTimer(time.Nanosecond) // Guardians should send out their first heartbeat immediately to speed up test runs.
+			// Guardians should send out their first heartbeat immediately to speed up test runs.
+			// But we also want to wait a little bit such that network connections can be established by then.
+			timer := time.NewTimer(time.Second * 2)
```
SEJeff, line 285: Is there a way to actually verify the connections are successful? Any way other than a hardcoded sleep?

tbjump: Even if the first heartbeat after 2s is not received by the other guardians, one of the next heartbeats (in 15s intervals) will eventually be received. In other words, we don't depend on the 2 seconds to be sufficient time, it's just a good value because it makes the tests run faster in practice.

SEJeff: 2 seconds is just less time. I meant making no timer and just verifying the connection is successful, but this is an improvement. It's just improving a sloppy approach to a faster still sloppy approach (not your fault)

SEJeff (approved):
Great stuff, but we want to let something like this bake in testnet. If you fancy a slight refactor, consider splitting out that code with the `for _, addr := range strings.Split(bootstrapPeers, ",") {` into a separate function. Bonus that the code you added would be relatively straightforward to test, and putting it in it's own function allows you to more easily test the logic you wrote is correct.

That `Run()` is ~500 LOC, which screams that it needs to be chopped into slightly smaller functions. Also, the deep nesting makes it much harder to contextualize in your head, and as a result, simply maintain.

panoel (approved):

node/pkg/p2p/p2p.go
tbjump: afaik you can force the connection (which we're doing already), but there's no way to verify that gossip is actually connected, so this was the best I could do.

tbjump:
Addressed @SEJeff 's feedback

panoel (approved):

bruce-riley (approved):

panoel (approved):

bruce-riley (approved):

PR #3146, opened 2023-06-28
Add tokenfactory module to wormchain

misko9:
Pulls in x/tokenfactory initially developed by osmosis, iterated upon for cosmos-sdk chains by CosmWasm and now CosmosContracts/juno.


wormchain/app/app.go
```diff
@@ -135,6 +140,15 @@ func GetWasmOpts(app *App, appOpts servertypes.AppOptions) []wasm.Option {
 	// add the custom wormhole query handler
 	wasmOpts = append(wasmOpts, wasmkeeper.WithQueryPlugins(wormholemodulekeeper.NewCustomQueryHandler(app.WormholeKeeper)))
 
+	// Move custom query of token factory to stargate, still use custom msg which is tfOpts[1]
+	var bankBaseKeeper bankkeeper.BaseKeeper
+	bankBaseKeeper, ok := app.BankKeeper.(bankkeeper.BaseKeeper)
```
nik-suri, line 145: nit: don't need the `var` declaration on line 144, can declare directly on line 145.

wormchain/app/app.go
```diff
@@ -180,6 +195,13 @@ var (
 		wormholemoduletypes.ModuleName: nil,
 		// this line is used by starport scaffolding # stargate/app/maccPerms
 		wasm.ModuleName: {authtypes.Burner},
+		tokenfactorytypes.ModuleName: {authtypes.Minter, authtypes.Burner},
+	}
+
+	tokenFactoryCapabilities = []string{
+		tokenfactorytypes.EnableBurnFrom,
+		tokenfactorytypes.EnableForceTransfer,
+		tokenfactorytypes.EnableSetMetadata,
```
nik-suri, line 204: Add comment on why we need this capabilities?

wormchain/app/app.go
```diff
@@ -725,6 +762,10 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
 }
 
+func (app *App) GetWasmKeeper() *wasmkeeper.Keeper {
+	return &app.wasmKeeper
+}
```
nik-suri, line 767: Why is this function needed?

wormchain/app/apptesting/test_suite.go
```diff
@@ -0,0 +1,295 @@
+package apptesting
```
nik-suri, line 1: Thank you for adding these tests! Can we rename the file to `app_test` or `suite_test` so that go detects it as a test file?

wormchain/app/helpers/mock.go
```diff
@@ -0,0 +1,50 @@
+package helpers
+
+import (
```
nik-suri, line 3: I like the way you've structured this, similar comment here though -- can we rename this file to `mock_key.go` or `mock_private_key.go` since we may add similar mock files in the future? 

Might also be a good idea to move the `helpers` folder entirely under `apptesting`, unless you were imagining that these would be used in non-testing code somewhere.

wormchain/app/params/proto.go
```diff
@@ -0,0 +1,22 @@
+package params
+
+import (
```
nik-suri, line 3: I believe that we can get all of the `EncodingConfig` using `cosmoscmd.MakeEncodingConfig`. Does this work? If so, we can probably remove this file as well as `encoding.go`.

wormchain/app/test_helpers.go
```diff
@@ -0,0 +1,260 @@
+package app
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
```
nik-suri, line 7: lets move this file under the `apptesting` folder as well?

nik-suri (changes requested):
Reviewed everything other than `x/tokenfactory`. Will follow up with a review for `x/tokenfactory`

wormchain/app/app.go
misko9: Will remove

wormchain/app/app.go
misko9: I can remove them as the contract does not require those features. Their associated test cases will need to be commented out too for the test suite to pass. This also surfaced an upstream bug in IsCapabilityEnabled() when the list is empty which I'll update and eventually upstream too.

wormchain/app/app.go
misko9: This is required for the x/tokenfactory/bindings/ test cases to access the wasm keeper to perform store/instantiate/execute/query using test contracts to test the wasm bindings. Other chains have wasmkeeper as a public field that can be accessed by the test cases instead of using a method.

wormchain/app/apptesting/test_suite.go
misko9: If that is done, the file is considered a test file and the `apptesting` package won't include it for `x/tokenfactory/keeper` tests to use. I think it is currently safe as `x/tokenfactory/keeper/keeper_test.go` is the only file that imports that package.

wormchain/app/helpers/mock.go
misko9: Yes, renamed and moved to apptesting to consolidate the testing packages.

wormchain/app/params/proto.go
misko9: yes, good catch, cleans things up nicely.

wormchain/app/test_helpers.go
misko9: Yes, moving to apptesting folder was possible and is a better fit than the app package. Thanks!

wormchain/x/tokenfactory/keeper/createdenom.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	k.addDenomFromCreator(ctx, creatorAddr, denom)
+	return nil
+}
+
+func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdenom string) (newTokenDenom string, err error) {
+	// Temporary check until IBC bug is sorted out
```
nik-suri, line 54: What is this IBC bug?

wormchain/x/tokenfactory/types/codec.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
+)
+
+func RegisterCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgCreateDenom{}, "osmosis/tokenfactory/create-denom", nil)
+	cdc.RegisterConcrete(&MsgMint{}, "osmosis/tokenfactory/mint", nil)
+	cdc.RegisterConcrete(&MsgBurn{}, "osmosis/tokenfactory/burn", nil)
+	cdc.RegisterConcrete(&MsgForceTransfer{}, "osmosis/tokenfactory/force-transfer", nil)
+	cdc.RegisterConcrete(&MsgChangeAdmin{}, "osmosis/tokenfactory/change-admin", nil)
+}
```
nik-suri, line 19: Should we rename these to `"wormhole/tokenfactory/..."`?

nik-suri (changes requested):
Reviewed tokenfactory changes using a diff from the juno tokenfactory - https://github.com/CosmosContracts/juno/tree/main/x/tokenfactory

Changes look good. Two thoughts:
1. Let's include the tokenfactory proto files as well, so we can make tokenfactory codegen an easily reproducible process in the wormhole repository.
2. I think your choice to remove the `MsgUpdateParams` makes sense, since it's a governance action that should only be done via a governance VAA on Gateway. We can do this in a follow-up PR, but let's follow the same strategy here as we currently do for the `x/wasm` store, instantiate, and migrate commands - We can disable the message handling in the `msg_server` and re-implement `UpdateParams` in the `x/wormhole` module, which will handle parsing the governance VAA and then calling into the underlying method in `x/tokenfactory`. See here for an example of how this is done - https://github.com/wormhole-foundation/wormhole/blob/main/wormchain/x/wormhole/keeper/msg_server_wasmd.go#L18-L65.

wormchain/proto/osmosis/tokenfactory/v1beta1/authorityMetadata.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+package osmosis.tokenfactory.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/CosmosContracts/juno/x/tokenfactory/types";
```
nik-suri, line 7: I think we'll need to change these references in every proto file to `"github.com/wormhole-foundation/wormchain/x/tokenfactory/types"`

misko9: yes, thanks. I've updated those files, regenerated them and tested so we are now sourcing from our proto files.

wormchain/x/tokenfactory/types/codec.go
misko9: I think we could, but since those types are registered with osmosis, it makes sense to me to keep osmosis unless we change the package name too. And the chance of the current package name staying the same is probably greatest when there is a tokenfactory module we can import in the future. That would keep some consistency.

wormchain/x/tokenfactory/keeper/createdenom.go
misko9: Took some digging, but it was this bug: https://github.com/cosmos/ibc-go/issues/894 which is present in ibc-go v3.0.0, fixed for ibc-go v4 and backported to ibc-go v3.1.0+.
Since we are using a version without the bug, I have removed the check.

nik-suri (approved):

PR #3147, opened 2023-06-28
CCQ: Message tweaks

bruce-riley:
Minor tweaks to the message spec to facilitate on-chain parsing.

### DIFF
--- a/node/pkg/query/request.go
+++ b/node/pkg/query/request.go
@@ -219,6 +219,9 @@ func (perChainQuery *PerChainQueryRequest) Marshal() ([]byte, error) {
 	}
 
 	// Write the length of the query to facilitate on-chain parsing.
+	if len(queryBuf) > math.MaxUint32 {
+		return nil, fmt.Errorf("query too long")
+	}
 	vaa.MustWrite(buf, binary.BigEndian, uint32(len(queryBuf)))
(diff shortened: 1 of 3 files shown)


node/pkg/query/response.go
```diff
@@ -260,6 +281,7 @@ func (perChainResponse *PerChainQueryResponse) Marshal() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	vaa.MustWrite(buf, binary.BigEndian, uint32(len(respBuf)))
```
evan-gray, line 284: should there be a check to make sure this is < max uint32

node/pkg/query/request.go
```diff
@@ -207,6 +217,10 @@ func (perChainQuery *PerChainQueryRequest) Marshal() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Write the length of the query to facilitate on-chain parsing.
+	vaa.MustWrite(buf, binary.BigEndian, uint32(len(queryBuf)))
```
evan-gray, line 222: should there be a check to make sure this is < max uint32

panoel (approved):

evan-gray (approved):

PR #3150, opened 2023-06-28
Node: Drop inbound signed vaas with quorum faster

bruce-riley:
When handling inbound signed VAAs with quorum, the processor looks to see if they are already in the database, and if so, it just returns without doing anything. It currently doesn't do that until after it verifies the signatures, which is wasted compute. This change moves the DB check ahead of that verification.


SEJeff (dismissed):
niiice

tbjump (approved):

evan-gray (approved):

PR #3151, opened 2023-06-28
spy: fixed deadlock on subscriber disconnects

kev1n-peters:
Added the "sendTimeout" flag that specifies the timeout duration for sending messages to subscribers.

### DIFF
--- a/node/cmd/spy/spy.go
+++ b/node/cmd/spy/spy.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/certusone/wormhole/node/pkg/common"
 	"github.com/certusone/wormhole/node/pkg/p2p"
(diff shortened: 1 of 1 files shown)


bruce-riley (dismissed):

panoel (dismissed):

evan-gray (approved):
Agree especially with bumping the channels. Fundamentally, these have to be `> sendTimeoutSeconds * expectedVAAsPerSecond` to avoid dropping VAAs.

bruce-riley (approved):

kev1n-peters:
I did some simple profiling of the Spy with my change that spawns a go-routine before each call to Send() for handling the timeout. Profiling in mainnet, I noticed that messages are being dropped a bit more frequently than previously when looking at this metric: wormhole_p2p_receive_channel_overflow{type="signed_vaa_with_quorum"}. The receive channel drops messages when the Spy server is too slow to Send() them out to subscribers (the signedInC gets full). I ran the Spy with 2 subscribers and saw the signedInC grow to a maximum length of 175. So I bumped the signedInC from a size of 50 to 1024 and haven't seen any dropped messages. We agreed that it's worth bumping that size here.

Issue #3152, opened 2023-06-29
Client/js: Fake Worm CLI commands on README documentation


AlberErre:
## Description and context

Currently, there are many non-existent (fake) Worm CLI commands present on the README documentation.

This issue was detected while reviewing PR #2631, here: https://github.com/wormhole-foundation/wormhole/pull/2631#issuecomment-1612868768

e.g.

Command `worm verify-vaa init-token-bridge` does not exist as a valid command on the current Worm CLI. However, it is present on the `clients/js/README` file.

## Steps to reproduce

1. Build worm CLI tool locally using `make install` on `clients/js` folder.
2. Visit `clients/js/README` file
3. Try some fake commands from README file on `worm` CLI

## Experienced behavior

There are several commands described on README file that does not exist on the current `worm` CLI implementation. 

e.g. 
`worm verify-vaa init-token-bridge`

## Expected behavior

Every commands present on README documentation should exist and work properly.

## Solution recommendation

Debug and fix `clients/js/src/doc.ts`, which is the source file that triggers `make docs` command. This source code is generating the incorrect README documentation.

AlberErre:
@aadam-10 just reported this issue detected while reviewing the Worm CLI tests 👀

kcsongor:
Looks like the issue is that https://github.com/wormhole-foundation/wormhole/blob/aa03d417193814c6a82f28dee83bf2e715941dd4/clients/js/src/doc.ts#L22-L24
mutates the same underlying `yargs` instance, and subcommands are not differentiated in their respective builders, e.g.

https://github.com/wormhole-foundation/wormhole/blob/aa03d417193814c6a82f28dee83bf2e715941dd4/clients/js/src/cmds/verifyVaa.ts#L12-L20.

To fix it, during doc generation we could attempt to create a fresh clone of `yargs` for each invocation.

PR #3153, opened 2023-06-29
CLI: Fix bug in doc generator

barnjamin:
As reported in https://github.com/wormhole-foundation/wormhole/issues/3152 the `docs.ts` script meant to produce documentation for the CLI tool was producing incorrect subcommands. 

This was caused by re-using the `yargs` singleton that had state built up from other commands.

The fix here is to create a fresh copy for every invocation of `builder`

### DIFF
--- a/clients/js/src/doc.ts
+++ b/clients/js/src/doc.ts
@@ -17,10 +17,14 @@ import * as submit from "./cmds/submit";
 import * as sui from "./cmds/sui";
 import * as verifyVaa from "./cmds/verifyVaa";
 
+
 const MD_TAG = "<!--CLI_USAGE-->";
 
-async function getHelpText(name: string, cmd: any): Promise<string> {
-  return await cmd.builder(yargs).scriptName(`worm ${cmd.command}`).getHelp();
+async function getHelpText(cmd: any): Promise<string> {
+  // Note that `yargs` is called as a function to produce a fresh copy.
+  // Otherwise the imported module is effectively a singleton where state from 
+  // other commands is accumulated from repeat calls.
+  return await cmd.builder(yargs()).scriptName(`worm ${cmd.command}`).getHelp();
 }
 
 (async function () {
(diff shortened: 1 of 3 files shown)


clients/js/src/doc.ts
```diff
@@ -39,8 +40,8 @@ async function getHelpText(name: string, cmd: any): Promise<string> {
   ];
 
   const helpOutputs: Buffer[] = [];
-  for (const [name, cmd] of Object.entries(cmds)) {
-    const helpText = await getHelpText(name, cmd);
+  for (const cmd of Object.values(cmds)) {
```
kcsongor, line 43: wouldn't
```suggestion
  for (const cmd of cmds) {
```
achieve the same thing?

barnjamin: yep! just changed locally, thanks!

clients/js/src/doc.ts
```diff
@@ -17,10 +17,11 @@ import * as submit from "./cmds/submit";
 import * as sui from "./cmds/sui";
 import * as verifyVaa from "./cmds/verifyVaa";
 
+
 const MD_TAG = "<!--CLI_USAGE-->";
 
-async function getHelpText(name: string, cmd: any): Promise<string> {
-  return await cmd.builder(yargs).scriptName(`worm ${cmd.command}`).getHelp();
+async function getHelpText(cmd: any): Promise<string> {
+  return await cmd.builder(yargs()).scriptName(`worm ${cmd.command}`).getHelp();
```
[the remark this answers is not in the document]
barnjamin: good point

kcsongor (approved):

kcsongor:
Nice!
As one final thing, could we include a note here reminding future people that when they add a new command, it should be added to the doc generator too?

https://github.com/wormhole-foundation/wormhole/blob/aa03d417193814c6a82f28dee83bf2e715941dd4/clients/js/src/main.ts#L21-L37

AlberErre:
Great! 🎉 I don't approve permissions but this looks good ✅

barnjamin:
> could we include a note here reminding future people that when they add a new command

Good point! just pushed it

kcsongor:
Thanks for the quick fix @barnjamin and for catching it @AlberErre

kcsongor (approved):

evan-gray (approved):

PR #3162, opened 2023-07-03
[WIP] Support Celo native's credit and debit gas fee functions

martinvol:
- [x] Add modifier so that only vm can call
- [x]  Add modifier so that it can only be called on Celo (by checking Celo Registry address has bytecode)
- [x] Add credit and debit functions
- [x] Add Fundry tests
- [ ] Check with wormhole team how to refactor to comply with repo guidelines

### DIFF
--- a/ethereum/contracts/bridge/token/TokenImplementation.sol
+++ b/ethereum/contracts/bridge/token/TokenImplementation.sol
@@ -10,7 +10,9 @@ import "@openzeppelin/contracts/proxy/beacon/BeaconProxy.sol";
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
 // Based on the OpenZepplin ERC20 implementation, licensed under MIT
-contract TokenImplementation is TokenState, Context {
+contract TokenImplementation is TokenState, Context
+    // TODO create interface for credit/debit token
+ {
     event Transfer(address indexed from, address indexed to, uint256 value);
     event Approval(address indexed owner, address indexed spender, uint256 value);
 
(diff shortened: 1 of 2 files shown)


ethereum/contracts/bridge/token/TokenImplementation.sol
```diff
@@ -348,4 +350,81 @@ contract TokenImplementation is TokenState, Context {
...
+    */
+    modifier onlyVm() {
+        require(msg.sender == address(0), "Only VM can call");
+        _;
+    }
+
+    /**
+     * @notice Reverts for non-Celo chains
+    */
+    modifier onlyCeloChains() {
```
evan-gray, line 365: Perhaps we should use the Wormhole chain id here.

evan-gray: Or maybe not because this is in the token 🤔

gator-boi (changes requested):
A few of us have discussed how to include this change in the existing `TokenImplementation` contract. 

How do you feel about creating a second `TokenImplementation` contract with the Celo-specific functionality that inherits from the existing contract? This way the Celo-specific code is not deployed to other blockchains when the `TokenImplementation` contract is upgraded.

martinvol:
Yeah, makes sense.

My original intention with this PR was to pull it in front of your team so you could suggest me the better course of action.

Even if this code gets deployed only on Celo, I'd still include the checks to make sure the chain is actually Celo compatible.

SEJeff:
@martinvol do you still want to work on this?

djb15:
Closing as part of cleanup. The branch can be retrieved [here](https://github.com/wormhole-foundation/wormhole-branch-clone) and the PR reopened from a fork if required.

PR #2187, opened 2023-01-02
node: switch to use proper thread recovery

jumpsiegel:

PR #3166, opened 2023-07-05
Node: Processor multithreading (simple)

bruce-riley:


alokeutpal (approved):

bruce-riley:
The following graphs were generated by listening to mainnet gossip and mainnet pythnet, using 24 cores.

- Multithreading with a factor 2.0 (48 workers)
- The multithreading code but with a factor of 0.0, meaning a single worker (new code but single threaded)
- The old single threaded code

Screenshot from 2023-07-18 13-25-18

bruce-riley:
Here are some charts comparing the original single threaded processor with three versions of multithreading.

image

bruce-riley:
Some further data. I ran the [benchmarks](https://github.com/wormhole-foundation/wormhole/blob/f11299b888892d5b4abaa624737d0e5117d1bbb2/node/pkg/node/node_test.go#L1033) for the various configurations.

- Existing single threaded: 10.989, 10.964, 11.688 seconds
- Simple multithreading: 11.099, 11.509. 11.801 seconds
- Multithreading with dispatcher: 11.361, 11.860, 11.722 seconds
- Direct dispatcher: 11.0557, 9.677, 10.060 seconds

So it seems like multithreading doesn't have much impact on the benchmarks.

bruce-riley:
Although this PR does not have as big of an impact as we had hoped, it is an improvement, and it positions us to make additional improvements going forward. Therefore I am marking it ready for review.

node/pkg/processor/broadcast.go
```diff
@@ -50,20 +49,20 @@ func (p *Processor) broadcastSignature(
...
 	hash := hex.EncodeToString(digest.Bytes())
 
-	if p.state.signatures[hash] == nil {
-		p.state.signatures[hash] = &state{
-			firstObserved: time.Now(),
-			nextRetry:     time.Now().Add(nextRetryDuration(0)),
-			signatures:    map[ethcommon.Address][]byte{},
-			source:        "loopback",
-		}
+	state, created := p.state.getOrCreateState(hash)
```
tbjump, line 52: how about using a different name than `state` here to not confuse it with `p.state`? Maybe `observationState` to signify that it's for one observation only or simply `s` to signify that it has a more local context.

node/pkg/processor/broadcast.go
```diff
@@ -50,20 +49,20 @@ func (p *Processor) broadcastSignature(
...
-	p.state.signatures[hash].ourObservation = o
-	p.state.signatures[hash].ourMsg = msg
-	p.state.signatures[hash].txHash = txhash
-	p.state.signatures[hash].source = o.GetEmitterChain().String()
-	p.state.signatures[hash].gs = p.gs // guaranteed to match ourObservation - there's no concurrent access to p.gs
+	state.ourObservation = o
+	state.ourMsg = msg
+	state.txHash = txhash
+	state.source = o.GetEmitterChain().String()
+	state.gs = p.gs.Load() // guaranteed to match ourObservation - there's no concurrent access to p.gs: TODO: Is this comment a problem??
```
tbjump, line 65: There will be concurrent access to p.gs now, so by the time we get here, p.gs may (in theory) be overwritten by a newer guardian set. So the assertion in the comment no longer holds. But the polling for guardian sets happens asynchronously from the message watching anyways, so I think this comment was never relevant. 
I wonder if `state.gs` is needed at all or if it would be sufficient to use `p.gs` everywhere.

node/pkg/processor/processor.go
```diff
@@ -252,9 +326,12 @@ func (p *Processor) Run(ctx context.Context) error {
...
 			p.handleInboundSignedVAAWithQuorum(ctx, m)
 		case <-cleanup.C:
-			p.handleCleanup(ctx)
+			if workerId == 1 {
+				cleanup.Reset(CleanupInterval)
+				p.handleCleanup(ctx)
+			}
 		case <-govTimer.C:
-			if p.governor != nil {
+			if p.governor != nil && workerId == 1 {
```
tbjump, line 334: Instead of always checking workerId == 1, how about splitting it up into a RunBasic and a RunConcurrent, as in https://github.com/wormhole-foundation/wormhole/blob/2972a68700ec2adfb0fe159c4513cc6fbf3f1206/node/pkg/processor/processor.go. 
I think the code would be cleaner that way, but you'll always have to run at least two workers: One for non-concurrent functions and one for potentially concurrent functions.

node/pkg/processor/processor.go
```diff
@@ -76,10 +83,36 @@ type (
...
+		}
+	}
+
+	return s.signatures[hash], created
+}
+
+// delete removes a state entry from the map. It grabs the lock.
+func (s *aggregationState) delete(hash string) {
+	s.signaturesLock.Lock()
+	defer s.signaturesLock.Unlock()
```
tbjump, line 112: instead of using a defer here, I'd switch it with the line below. It's a tiny performance improvement.

node/cmd/guardiand/node.go
```diff
@@ -367,6 +368,7 @@ func init() {
 	bigTableKeyPath = NodeCmd.Flags().String("bigTableKeyPath", "", "Path to json Service Account key")
 
 	chainGovernorEnabled = NodeCmd.Flags().Bool("chainGovernorEnabled", false, "Run the chain governor")
+	processorWorkerFactor = NodeCmd.Flags().Float64("processorWorkerFactor", 0.0, "Multiplier used to compute number of processor workers, 0.0 means single threaded")
```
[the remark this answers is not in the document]
bruce-riley: Changed.

node/pkg/processor/broadcast.go
bruce-riley: Renamed to `obsState`.

node/pkg/processor/processor.go
bruce-riley: Changed.

node/pkg/processor/broadcast.go
bruce-riley: Had an offline discussion with @tbjump and @evan-gray and we agreed that it is not necessary to save the per-observation guardian set. I've upgraded the code to reflect this.

node/pkg/processor/processor.go
bruce-riley: I split it into a housekeeper and a worker.

node/pkg/processor/processor.go
```diff
@@ -234,31 +234,40 @@ func (p *Processor) Run(ctx context.Context) error {
...
 				}
 				p.logger.Info("processor worker done", zap.Int("workerId", workerId))
 				w.Done()
 			}(ctx, workerId)
 		}
-
-		w.Wait()
 	}
 
+	w.Wait()
```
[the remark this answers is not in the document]
tbjump: Each worker should also use its own logger through `supervisor.Logger(ctx)` and `p.logger` should be removed. Otherwise it's hard to trace logs from multiple concurrent processes.

node/pkg/node/node_test.go
```diff
@@ -193,7 +193,7 @@ func mockGuardianRunnable(t testing.TB, gs []*mockGuardian, mockGuardianIndex ui
 			GuardianOptionPublicWeb(cfg.publicWeb, cfg.publicSocket, "", false, ""),
 			GuardianOptionAdminService(cfg.adminSocket, nil, nil, rpcMap),
 			GuardianOptionStatusServer(fmt.Sprintf("[::]:%d", cfg.statusPort)),
-			GuardianOptionProcessor(),
+			GuardianOptionProcessor(0.0),
```
tbjump, line 196: Should test with concurrency. Maybe set workFactor to 3/numCPU?

node/pkg/processor/processor.go
```diff
@@ -117,7 +148,7 @@ type Processor struct {
 	// Runtime state
 
 	// gs is the currently valid guardian set
-	gs *common.GuardianSet
+	gs atomic.Pointer[common.GuardianSet]
```
tbjump, line 151: could also just use `gst.Get()` everywhere instead of maintaining another variable.

tbjump (changes requested):

node/pkg/processor/cleanup.go
```diff
@@ -124,36 +124,32 @@ func (p *Processor) cleanUpStateEntry(hash string, s *state) bool {
 		s.settled = true
 
 		// Use either the most recent (in case of a observation we haven't seen) or stored gs, if available.
```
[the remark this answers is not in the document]
bruce-riley: Fixed

node/pkg/processor/processor.go
```diff
@@ -293,6 +303,50 @@ func (p *Processor) RunOne(ctx context.Context, workerId int) error {
...
+		}
+	}
+}
+
+// runWorker performs the per-observation tasks that can be distributed to the workers. There will be at least one of these.
+func (p *Processor) runWorker(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			if p.acct != nil {
```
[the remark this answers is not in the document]
bruce-riley: Moved up to after the wait.

node/pkg/processor/processor.go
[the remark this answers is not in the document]
tbjump: `RunWithScissors` actually only catches panics, it doesn't do any re-starting. 
* I think panics in the processor should still lead to a Guardian panic and restarting the entire Guardian, so `RunWithScissors` should *not* be used here, although I'd be open to discussing that. 
* The routines should be restarted if they merely return an error. E.g. right now if the accountant returns an error the worker will just return that error. You can accomplish that by running the workers through the supervisor. The supervisor will restart runnables if they return an error or did not signal `supervisor.SignalDone` before returning.

node/pkg/processor/processor.go
```diff
@@ -241,33 +243,34 @@ func (p *Processor) Run(ctx context.Context) error {
...
-			p.logger.Error("housekeeper failed", zap.Error(err))
-		}
-	}(ctx)
+	common.RunWithScissors(ctx, errC, "processor_housekeeper",
+		func(ctx context.Context) error {
+			err := p.runHousekeeper(ctx)
+			if err != nil {
+				errC <- fmt.Errorf("processor housekeeper failed: %v", err)
+			}
+			return err
```
tbjump, line 252: I think there's no need to wrap this in an extra function. The way it is right now will actually result in two write attempts to errC: One by this function and another one by RunWithScissors. Each channel should only have one writer and one consumer as a general practice. 

Instead `common.RunWithScissors(ctx, errC, "processor_housekeeper", p.runHousekeeper)` should work, no? 

PS: Still not sure if panics should be caught here or just propagated (see my other comment).

node/pkg/processor/processor.go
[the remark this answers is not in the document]
tbjump: Ah I see, because of https://github.com/wormhole-foundation/wormhole/blob/02a33451b80552e47d133e0cb7c92c9d970bd21e/node/pkg/processor/processor.go#L272 the processor would return the error from any worker, which in turn would lead to the supervisor canceling all children and restarting the processor.

tbjump:
### Catching panics or not discussion:
We discussed a bit in the code comments, but starting this higher-level comment so the discussion doesn't get buried in the code comments. 

First of all, *errors* should be caught and handled appropriately, nobody would say anything else. 

When it comes to panics, I think we have to differentiate
1. panics that are expected to occur, e.g. if calling a library that is known to panic on certain inputs: These panics should be recovered as close to the panic as possible. This is the same as handling an error at this point. 
2. panics that are unexpected and therefore don't have any recover code around them. Since the developer did not expect this panic, chances are that returning at this point may leak resources and/or leave the program in an inconsistent state. Maybe a transaction stopped half-way, maybe a lock wasn't released, maybe a context not canceled, a socket not closed, etc. In these cases, a balance needs to be struck between correctness and availability. For correctness, it is safer to just crash the whole program and restart everything cleanly. But that jeopardizes availability since the crash would make concurrent processes fail and the restart takes some time to carry out. 

In the case of watchers, it was agreed (https://github.com/wormhole-foundation/wormhole/pull/2187) that panics should be caught generically and watchers always just restarted. Because watchers are pretty self-contained, restarting them isn't very dangerous and because they're the most panic-prone components, it improves reliability a lot. 

But when it comes to the processor, I don't think that there should be attempts to recover from unexpected panics because the processor has more potential to end up in an inconsistent state.

tbjump:
* main
  * **PR #3166** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3166" target="_blank">Graphite</a>  👈
    * **PR #3267** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3267" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3166?utm_source=stack-comment).

bruce-riley:
It is desirable that the multithreaded processor handle errors / panics in the same way as the current processor, meaning that panics kill the guardian and errors restart the entire processor. It is also desirable to use the `RunWithScissors` pattern since that has been adopted by the team as the way to do error handling. Rather than create a one-off approach for the processor, I created a new `StartRunnable` function in the `RunWithScissors` module that allows you to specify if you want to catch errors or not. I am now using `StartRunnable` in the processor. After this PR is merged, I plan to change all references to `RunWithScissors` to the appropriate `StartRunnable` call, after careful consultation with the various code owners.

node/pkg/processor/processor.go
```diff
@@ -23,7 +26,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	dto "github.com/prometheus/client_model/go"
+	// dto "github.com/prometheus/client_model/go"
```
panoel, line 29: Remove?

node/pkg/processor/processor.go
```diff
@@ -234,31 +234,40 @@ func (p *Processor) Run(ctx context.Context) error {
...
+	w.Add(1)
+	go func(ctx context.Context) {
+		err := p.runHousekeeper(ctx)
+		if err != nil {
+			p.logger.Error("housekeeper failed", zap.Error(err))
+		}
+	}(ctx)
+
 	var ret error
 	if p.workerFactor == 0.0 {
```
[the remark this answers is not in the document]
panoel: Were you going to do it this way?  It doesn't look like the code reflects that?

panoel (dismissed):

node/pkg/processor/processor.go
bruce-riley: No, I want to keep that because it goes with the commented out code to dump the metrics on shutdown, which I want to keep because it's useful when running benchmarks.

tbjump:
Hey, what's the state of this PR? It has conflicts.

PR #3171, opened 2023-07-06
Node: Remove more info logging

bruce-riley:
Change some more high runner info logs into debugs

### DIFF
--- a/node/pkg/processor/vaa.go
+++ b/node/pkg/processor/vaa.go
@@ -1,8 +1,6 @@
 package processor
 
 import (
-	"encoding/hex"
-
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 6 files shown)


node/pkg/processor/vaa.go
```diff
@@ -26,16 +24,10 @@ func (v *VAA) HandleQuorum(sigs []*vaa.Signature, hash string, p *Processor) {
 		Payload:          v.Payload,
 		ConsistencyLevel: v.ConsistencyLevel,
 	}
-	vaaBytes, err := signed.Marshal()
-	if err != nil {
-		panic(err)
-	}
 
 	// Store signed VAA in database.
 	p.logger.Info("signed VAA with quorum",
```
evan-gray, line 29: agree with ditching the whole bytes in logs, as it's hefty and unknown size, but may want to consider still logging the signatures.

evan-gray (approved):
less is more

panoel (approved):

node/pkg/processor/vaa.go
bruce-riley: After an offline discussion, we decided not to log the sigs, because the message grows from 280 bytes to 2800 bytes.

PR #3173, opened 2023-07-06
evm: add TransferRedeemed event to Bridge.sol

gator-boi:
The purpose of this PR is to add the event `TransferRedeemed` to the token bridge. This will allow off-chain processes to track when a transfer has been redeemed.

### DIFF
--- a/ethereum/contracts/bridge/Bridge.sol
+++ b/ethereum/contracts/bridge/Bridge.sol
@@ -20,6 +20,18 @@ import "./token/TokenImplementation.sol";
 contract Bridge is BridgeGovernance, ReentrancyGuard {
     using BytesLib for bytes;
 
+    /**
+     * @notice Emitted when a transfer is completed by the token bridge.
+     * @param emitterChainId Wormhole chain ID of emitter on the source chain.
+     * @param emitterAddress Address (bytes32 zero-left-padded) of emitter on the source chain.
+     * @param sequence Sequence of the Wormhole message.
+     */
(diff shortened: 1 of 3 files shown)


a5-pickle (approved):

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -20,6 +20,18 @@ import "./token/TokenImplementation.sol";
 contract Bridge is BridgeGovernance, ReentrancyGuard {
     using BytesLib for bytes;
 
+    /**
+     * @notice Emitted when a transfer is completed by the token bridge.
+     * @param emitterChainId Wormhole chain ID of emitter on the source chain.
+     * @param emitterAddress Address (bytes32 zero-left-padded) of emitter on the source chain.
+     * @param sequence Sequence of the Wormhole message.
+     */
+    event TransferRedeemed(
```
kcsongor, line 29: naming nit: wouldn't `TransferCompleted` be a more appropriate name, since it's emitted by the `completeTransfer*` family of functions?

kcsongor (approved):
the day has finally come!

ethereum/contracts/bridge/Bridge.sol
gator-boi: The Sui contract emits an event named `TransferCompleted` already, I figured we should be consistent. Although, I agree that's a better name.

tbjump (commented):
lgtm

PR #3174, opened 2023-07-06
Add ibc-translator contract to the cosmwasm workspace and it's governance packet to the rust sdk.

misko9:


cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,197 @@
...
+    // if we haven't created the denom, then create the denom
+    // info.sender contains the cw20 contract address
+    if !CW_DENOMS.has(deps.storage, contract_addr.clone()) {
+        // call into token factory to create the denom
+        let create_denom = SubMsg::reply_on_success(
+            TokenMsg::CreateDenom {
+                subdenom,
+                metadata: None,
+            },
+            CREATE_DENOM_REPLY_ID,
```
nik-suri, line 125: why do we need a reply handler for the `CreateDenom` call?

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,197 @@
...
+        .load(deps.storage, chain_id)
+        .context("chain id does not have an allowed channel")?;
+
+    let payload_decoded = match payload {
+        Some(payload) => String::from_utf8(payload.to_vec())
+            .context(format!("failed to convert {payload} to utf8 string"))?,
+        None => "".to_string(),
+    };
+
+    // Create MsgTransfer protobuf message for Stargate
```
nik-suri, line 152: Let's link to the ICS20 transfer payload spec here so it's clear what protobuf object we are serializing to.

cosmwasm/contracts/ibc-translator/src/execute.rs
```diff
@@ -0,0 +1,295 @@
...
+        .save(deps.storage, vaa.hash.as_slice(), &true)
+        .context("failed to save governance VAA to archive")?;
+
+    // match the governance action and execute the corresponding logic
+    match govpacket.action {
+        Action::UpdateChannelChain {
+            channel_id,
+            chain_id,
+        } => {
+            ensure!(chain_id != Chain::Wormchain, "the wormchain-ibc-receiver contract should not maintain channel mappings to wormchain");
```
nik-suri, line 274: nit - `ibc-translator` instead of `wormchain-ibc-receiver`

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,197 @@
...
+    amount: u128,
+    contract_addr: String,
+    chain_id: u16,
+    payload: Option<Binary>,
+) -> Result<Response<TokenFactoryMsg>, anyhow::Error> {
+    // check the recipient and contract addresses are valid
+    // recipient will have a different bech32 prefix and fail
+    //deps.api
+    //    .addr_validate(&recipient)
+    //    .context(format!("invalid recipient address {}", recipient))?;
```
nik-suri, line 99: good callout - we can remove this code and just keep the comment for validating the cw20 contract address

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,197 @@
...
+            )
+        }
+    }
+}
+pub fn convert_cw20_to_bank_and_send(
+    deps: DepsMut,
+    env: Env,
+    recipient: String,
+    amount: u128,
+    contract_addr: String,
```
nik-suri, line 91: nit - rename to `cw20_contract_addr` for clarity

nik-suri (changes requested):
Great work! Leaving some comments after a first pass review.

cosmwasm/contracts/ibc-translator/src/bindings.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub denom: String,
+    /// exponent represents power of 10 exponent that one must
+    /// raise the base_denom to in order to equal the given DenomUnit's denom
+    /// 1 denom = 1^exponent base_denom
+    /// (e.g. with a base_denom of uatom, one can create a DenomUnit of 'atom' with
+    /// exponent = 6, thus: 1 atom = 10^6 uatom).
+    exponent: u32,
+    /// aliases is a list of string aliases for the given denom
+    aliases: Vec<String>,
+}
```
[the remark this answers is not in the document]
misko9: We'll need to wait for our wasmd to upgrade

cosmwasm/contracts/ibc-translator/src/reply.rs
misko9: We don't, I removed it

cosmwasm/contracts/ibc-translator/src/contract.rs
```diff
@@ -0,0 +1,111 @@
...
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, anyhow::Error> {
+    TOKEN_BRIDGE_CONTRACT
+        .save(deps.storage, &msg.token_bridge_contract)
+        .context("failed to save token bridge contract address to storage")?;
+
+    WORMHOLE_CONTRACT
+        .save(deps.storage, &msg.wormhole_contract)
+        .context("failed to save wormhole contract address to storage")?;
```
nik-suri, line 34: We can remove the wormhole contract storage variable - I don't believe it's used anywhere.

cosmwasm/contracts/ibc-translator/src/execute.rs
```diff
@@ -0,0 +1,298 @@
...
+        .load(deps.storage)
+        .context("could not load token bridge contract address")?;
+
+    let vaa: ParsedVAA = deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: wormhole_contract,
+        msg: to_binary(&WormholeQueryMsg::VerifyVAA {
+            vaa,
+            block_time: env.block.time.seconds(),
+        })?,
+    }))?;
```
nik-suri, line 242: Since we have the `wormhole-bindings` package, which lets us call into the wormhole core module, let's call into that instead of making the contract query to the wormhole contract. See an example here -- 

https://github.com/wormhole-foundation/wormhole/blob/main/cosmwasm/contracts/wormchain-ibc-receiver/src/contract.rs#L58-L73

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,175 @@
+use crate::{
+    msg::GatewayIbcTokenBridgePayload,
+    state::{CHAIN_TO_CHANNEL_MAP, CURRENT_TRANSFER, CW_DENOMS},
+};
+#[cfg(not(feature = "library"))]
+use anybuf::Anybuf;
```
nik-suri, line 6: Why do we need to gate `anybuf` with a feature flag?

cosmwasm/contracts/ibc-translator/src/contract.rs
misko9: yes, using the custom query, we can then remove this

cosmwasm/contracts/ibc-translator/src/reply.rs
misko9: thanks, those shouldn't be there

nik-suri (dismissed):

cosmwasm/contracts/ibc-translator/src/contract.rs
```diff
@@ -0,0 +1,107 @@
...
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, anyhow::Error> {
+    TOKEN_BRIDGE_CONTRACT
+        .save(deps.storage, &msg.token_bridge_contract)
+        .context("failed to save token bridge contract address to storage")?;
+
+    Ok(Response::new()
+        .add_attribute("action", "instantiate")
+        .add_attribute("owner", info.sender))
```
a5-pickle, line 34: What is the purpose of this owner?

cosmwasm/contracts/ibc-translator/src/msg.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[cw_serde]
+pub enum GatewayIbcTokenBridgePayload {
+    Simple {
+        chain: u16,
+        recipient: Binary,
+        fee: u128,
+        nonce: u32,
+    },
+    ContractControlled {
```
a5-pickle, line 80: Is there a naming convention this is following (same goes for `Simple`)? Every Token Bridge contract refers to these transfers as Transfer and TransferWithPayload.

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    // deserialize payload into the type we expect
+    let payload: GatewayIbcTokenBridgePayload = serde_json_wasm::from_slice(&transfer_info.payload)
+        .context("failed to deserialize transfer payload")?;
+
+    match payload {
+        GatewayIbcTokenBridgePayload::Simple {
+            chain,
+            recipient,
+            fee: _,
```
a5-pickle, line 50: How is the relayer fee handled? It looks like in `convert_cw20_to_bank_and_send` that the recipient receives the full amount. Does the spec require that this fee be zero?

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,174 @@
...
+        .append_string(2, channel) // source channel
+        .append_message(
+            3,
+            &Anybuf::new()
+                .append_string(1, tokenfactory_denom)
+                .append_string(2, amount.to_string()),
+        ) // Token
+        .append_string(4, env.contract.address) // sender
+        .append_string(5, recipient) // receiver
+        .append_message(6, &Anybuf::new().append_uint64(1, 0).append_uint64(2, 0)) // TimeoutHeight
```
a5-pickle, line 155: How is the timeout height determined? Is this according to a spec somewhere?

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,174 @@
...
+        .append_uint64(7, env.block.time.plus_days(365).nanos()) // TimeoutTimestamp
```
a5-pickle, line 156: And is a year out typical for a timeout timestamp?

cosmwasm/contracts/ibc-translator/src/execute.rs
```diff
@@ -0,0 +1,310 @@
...
+            msg: token_bridge_query_msg,
+        }))
+        .context("could not parse token bridge payload3 vaa")?;
+
+    // DEFENSE IN-DEPTH CHECK FOR PAYLOAD3 VAAs
+    // ensure that the transfer vaa recipient is this contract.
+    // we should never process any VAAs that are not directed to this contract.
+    let target_address = (&transfer_info.recipient.as_slice()).get_address(0);
+    let recipient = deps.api.addr_humanize(&target_address)?;
+    ensure!(
```
a5-pickle, line 83: Is there a vulnerability in the existing token bridge contract that would allow someone else to redeem this transfer with payload?

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,174 @@
...
+    let tokenfactory_denom = "factory/".to_string()
+        + env.contract.address.to_string().as_ref()
+        + "/"
+        + subdenom.as_ref();
+
+    let mut response: Response<TokenFactoryMsg> = Response::new();
+
+    // check contract storage see if we've created a denom for this cw20 token yet
+    // if we haven't created the denom, then create the denom
+    // info.sender contains the cw20 contract address
```
a5-pickle, line 110: What does this comment mean exactly? I tried to find context for it somewhere but I failed.

cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -0,0 +1,174 @@
...
+    if !CW_DENOMS.has(deps.storage, cw20_contract_addr.clone()) {
+        // call into token factory to create the denom
+        let create_denom = SubMsg::new(TokenMsg::CreateDenom {
+            subdenom,
+            metadata: None,
```
a5-pickle, line 115: Is the metadata retrievable via some sort of explorer? Would it be worth it to put token info here?

a5-pickle (commented):
Left comments for some clarification.

I think it would have helped if I had some context to what the contract is supposed to do. Is there a doc somewhere?

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: For going from external chains to Cosmos chains, the `fee` parameter isn't used today. We primarily added the `fee` parameter for compatibility with the wormhole token bridge when going from Cosmos chains back out to external chains - when we call the token bridge, we need to provide a `fee`, which, if non-zero, is used to incentivize specialized relayers.

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: Good catch, we can remove this comment.

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: We chose to use a year timeout since we decided it's effectively long enough to eliminate the chance that a transfer times out. I think that it should be sufficient for all chains.

@misko9 to confirm - what are your thoughts on configuring timeout as:
1. max uint64
2. different timeouts per chain_id

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: This line of code should be configuring the `TimeoutHeight` to 0. `TimeoutHeight` should be 0 since we are using the `TimeoutTimestamp` to specify timeouts.

@misko9 to confirm

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: Currently, if a user wanted to get metadata on one of these tokens, they'd need to:
1. Convert the base58 subdenom into the bech32 cw20 contract address
2. Query the cw20 contract address to get token metadata

We could set token metadata here, I think it'd just require an extra cross-contract query to get the cw20 metadata from the cw20 contract. I don't know if any cosmos explorers automatically integrate denom metadata. 

@misko9 what are your thoughts?

cosmwasm/contracts/ibc-translator/src/reply.rs
misko9: That is correct, TimeoutHeight is set to 0 to use TimeoutTimestamp.

cosmwasm/contracts/ibc-translator/src/reply.rs
misko9: 1. we should re-visit those timeouts. Looking at the light client code, it may not be possible to submit an ibc transfer after the trusting period (typically 2/3 of bonding period) has elapsed on the counter-party chain. If so, I don't see any reason it should be greater than that and it then gives options for a different solution instead of the funds being stuck.
2. seems reasonable

cosmwasm/contracts/ibc-translator/src/execute.rs
misko9: I don't think so, this check here should restrict that:
https://github.com/wormhole-foundation/wormhole/blob/main/cosmwasm/contracts/token-bridge/src/contract.rs#L838

nik-suri: We had added this to the sei-token-translator contract (which this contract was adapted from) as part of the internal security review for a defense in depth check.
https://github.com/everstake/sei-token-translator/tree/main

cosmwasm/contracts/ibc-translator/src/reply.rs
misko9: I haven't seen any explorers use tokenfactory's metadata yet either, but I think it's a good idea to add it. @nik-suri, can we sync on the correct mapping of CW20 metadata to tokenfactory metadata since it is not all 1:1?

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: @a5-pickle synced with @misko9 offline - instead of a year, we'll set the timeout to be 2 weeks. The unbonding period for Gateway is 21 days and all the connections will be established via a trusted setup process that will set the trusting_period to 14 days.  After the trusting_period has elapsed, one cannot submit an IBC packet on the destination chain anyways.

If this ever happens, the funds will be unlocked and stuck on Gateway. We could introduce a recovery mechanism for those funds in the future.

misko9:
> Also the payloads "simple" and "contract controlled" should be renamed to "transfer" and "transfer with payload" to be consistent with the token bridge terminology.

@a5-pickle - We can make that change. But an external chain -> cosmos chain vaa payload would be a token bridge "transfer with payload" with a Gateway/IbcTranslator payload which would then also be named "transfer" or "transfer with payload". Could this be confusing to an integrator?

cosmwasm/contracts/ibc-translator/src/reply.rs
nik-suri: @misko9 and I discussed offline. It's useful to populate the tokenfactory metadata to prepare for a future where clients such as wallets and explorers can show the correct decimals for the ibc tokens. 

Since cosmos clients interpret decimals using a `base_denom` and `denom` structure (where `base_denom` is used as reference to compute decimals), we'll populate the tokenfactory metadata with 2 denom units - 
1. `base_denom` where the name is `base_<symbol>`
2. `denom` where the name is `<symbol>` and the exponent equals the number of decimals in for the cw20 token.

a5-pickle:
> > Also the payloads "simple" and "contract controlled" should be renamed to "transfer" and "transfer with payload" to be consistent with the token bridge terminology.
> 
> @a5-pickle - We can make that change. But an external chain -> cosmos chain vaa payload would be a token bridge "transfer with payload" with a Gateway/IbcTranslator payload which would then also be named "transfer" or "transfer with payload". Could this be confusing to an integrator?
> 

That's a good point. Can we name them something consistent with the token bridge messages names? Maybe TranslatedTransfer? ChannelTransfer? GatewayTransfer?

nik-suri (approved):

a5-pickle (approved):

PR #3177, opened 2023-07-06
Node: Fix reobservation limits

bruce-riley:
The code in the processor cleanup routine was incorrectly retrying reobservations for **50 days** rather than five days. This PR changes it to retry for one day for observations we have seen and one hour for those we have not (which was previously 50 minutes, although the original comments said five minutes).

### DIFF
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -54,8 +54,10 @@ var (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


panoel (approved):

evan-gray (commented):
I like this, seems cleaner.

evan-gray (approved):

PR #3180, opened 2023-07-07
ethereum/ Makefile and readme update

derpy-duck:


ethereum/README.md
```diff
@@ -31,9 +31,9 @@ assets.
 
 For each test run:
 
-Run `npx ganache-cli --wallet.deterministic --chain.time "1970-01-01T00:00:00+00:00"` to start a chain.
+Run `npx ganache-cli --chain.vmErrorsOnRPCResponse --chain.chainId 1 --wallet.defaultBalance 10000 --wallet.deterministic --chain.time="1970-01-01T00:00:00+00:00" --chain.asyncRequestProcessing=false` to start a chain.
 
-Run the all ethereum tests using `make test`
+Run the all ethereum tests using `make test-dev`
```
kcsongor, line 36: We could achieve the same thing without having to modify the makefile at all by just invoking `DEV=True make test`, then we would not need to introduce (and maintain) duplicated make targets

kcsongor (changes requested):

ethereum/README.md
derpy-duck: agreed - that is better! What about now

kcsongor (approved):

ethereum/README.md
kcsongor: nice!

evan-gray (approved):

PR #3182, opened 2023-07-09
Clients/js/refactor cmds & CLI docs generator

AlberErre:
ℹ️ This PR exposes all CLI commands modules into a constant array called `CLI_COMMAND_MODULES`.

Looks like all commands can be imported into a singe `.command()` call 👀, if we export these commands as a JS module 🎉

More info about this approach can be found on the [official documentation](https://github.com/yargs/yargs/blob/main/docs/advanced.md#example-command-hierarchy-using-indexmjs).

This refactor is needed to provide a single source of truth for all CLI commands. By this approach, we would be able to implement testing on CLI commands against auto-generated documentation, as pointed out here -> https://github.com/wormhole-foundation/wormhole/pull/2631#issuecomment-1620832442

Also, this PR refactors `clients/js/src/doc.ts` documentation generator:
- A default `english` locale  has been added (without this, documentation is generated on the language of the OS machine, it was generated in spanish in my case).
- `CLI_COMMAND_MODULES` variable has been used to derive commands. This avoids maintaining multiple commands lists over the CLI, now we have a single source of truth.


AlberErre:
@aadam-10 @kcsongor @evan-gray @a5-pickle @heyitaki

just created this MR to extract the refactor done on https://github.com/wormhole-foundation/wormhole/pull/2631 regarding cmds modules & auto-generated documentation 👀

This way is easier to review 🎉🏄‍♂️

clients/js/src/cmds/info/index.ts
```diff
@@ -1,22 +1,12 @@
...
-    .command(chainId)
-    .command(contract)
-    .command(emitter)
-    .command(origin)
-    .command(registrations)
-    .command(rpc)
-    .command(wrapped);
+  // Commands can be imported as an array of commands.
+  // Documentation about command hierarchy can be found here: https://github.com/yargs/yargs/blob/main/docs/advanced.md#example-command-hierarchy-using-indexmjs
+  y.command(INFO_COMMANDS as unknown as YargsCommandModule[]);
```
barnjamin, line 11: do we need the `as unkown` now? I see some others changed to remove the unknown intermediate

clients/js/src/doc.ts
```diff
@@ -1,74 +1,51 @@
...
-    info,
-    near,
-    parse,
-    recover,
-    submit,
-    sui,
-    verifyVaa,
-    status
-  ];
+  const cmds = CLI_COMMAND_MODULES;
```
barnjamin, line 22: nice

clients/js/src/doc.ts
```diff
@@ -1,74 +1,51 @@
...
 async function getHelpText(cmd: any): Promise<string> {
   // Note that `yargs` is called as a function to produce a fresh copy.
-  // Otherwise the imported module is effectively a singleton where state from 
+  // Otherwise the imported module is effectively a singleton where state from
   // other commands is accumulated from repeat calls.
-  return await cmd.builder(yargs()).scriptName(`worm ${cmd.command}`).getHelp();
+  return await cmd
+    .builder(yargs())
+    .scriptName(`worm ${cmd.command}`)
+    .locale("en") //NOTE: 'locale' needed to override auto-detected locale from the user’s operating system
```
barnjamin, line 17: why do we want to override the users setting?

clients/js/src/cmds/info/index.ts
AlberErre: Yes, I tried to get rid of `unknown` here, but typescript complains a lot about incompatible types.

On the root command this `unknown` is not necessary anymore, but looks like we still need to cast it here ( probably because this is a subcommand) 👀

clients/js/src/doc.ts
AlberErre: This is needed because we want the `README.md` file to always be written in english.

I noticed this while running `make docs` locally. In my particular case, my machine locale is set to spanish by default. As a result, the commands generated on the `README.md` file after running `make docs` was written in spanish.

I think you could reproduce it locally if you change your system's language preferences 👀

barnjamin: I see, fair enough!

barnjamin (dismissed):

AlberErre:
@barnjamin just synced with `main`, and needed to solve some conflicts 👀

could you approve it again? ✅ 

thanks!

barnjamin (approved):

kcsongor (approved):

PR #3188, opened 2023-07-12
clients: allow custom block range

derpy-duck:


kev1n-peters (approved):

kev1n-peters (approved):

clients/js/src/cmds/status.ts
```diff
@@ -32,6 +32,14 @@ export const builder = (y: typeof yargs) =>
       describe: "Source transaction hash",
       type: "string",
       demandOption: true,
+    } as const)
+    .positional("blockStart", {
```
kcsongor, line 36: for arg names we use `block-start` format elsewhere

clients/js/src/cmds/status.ts
```diff
@@ -32,6 +32,14 @@ export const builder = (y: typeof yargs) =>
...
+      describe: "Starting Block Range, i.e. -2048",
+      type: "string"
+    } as const)
+    .positional("blockEnd", {
```
kcsongor, line 40: same as above

kcsongor (changes requested):
overall makes sense, it would make sense to suggest the flag in the error string when the delivery is not found

kcsongor (approved):

kev1n-peters (approved):

PR #3208, opened 2023-07-14
scripts: adding ubuntu to dev setup

panoel:


evan-gray (approved):

scripts/dev-setup.sh
```diff
@@ -49,14 +51,14 @@ esac
...
-    Debian-*)             sudo apt-get update && sudo apt-get upgrade -y ;;
+    Debian-*|Ubuntu-*)    sudo apt-get update && sudo apt-get upgrade -y ;;
     RedHatEnterprise-8.*) sudo dnf upgrade -y ;;
     *) echo "Internal error: $DISTRO not matched in case block." && exit 1 ;;
 esac
 
 # Install dependencies
 case "$DISTRO" in
-    Debian-*)             sudo apt-get -y install bash-completion git git-review vim ;;
+    Debian-*|Ubuntu-*)             sudo apt-get -y install bash-completion git git-review vim ;;
```
kcsongor, line 61:
```suggestion
    Debian-*|Ubuntu-*)             sudo apt-get -y install curl bash-completion git git-review vim ;;
```

scripts/dev-setup.sh
```diff
@@ -4,15 +4,17 @@ set -euo pipefail
 # This script provisions a working Wormhole dev environment on a minimally provisioned VM.
 # It expects to run as a user without root permissions.
 #
+# It is expected that curl is already installed.
```
kcsongor, line 7: the purpose of this script is to set up the environment, so we might as well just install curl instead of expecting it to be available. see below

kcsongor (changes requested):
I ran this script in `ubuntu:latest`, it works with the proposed modification

scripts/dev-setup.sh
panoel: Fixed

scripts/dev-setup.sh
panoel: Fixed.  Thanks.

kcsongor (approved):

evan-gray (approved):
Thanks @kcsongor for the call outs

PR #3228, opened 2023-07-21
node/node_test: Add RSA to crypto benchmark

tbjump:
I was wondering if it was worthwhile to consider moving from ED25519 in the p2p network to RSA because RSA allows for faster signature verification, albeit larger signature size, and libp2p is verification-heavy. 
The benchmarks show that the libp2p RSA seems to be unoptimized and is actually slower than ED25519.

### DIFF
--- a/node/pkg/node/node_test.go
+++ b/node/pkg/node/node_test.go
@@ -1061,6 +1061,44 @@ func BenchmarkCrypto(b *testing.B) {
 		})
 	})
 
+	/*
+		RSA is an option for libp2p.
+		In an optimized RSA implementation, signature verification in RSA can be faster than with elliptic curves, while signature generation is always slower.
+		Since libp2p is verification-heavy, this might overall still be a faster option.
+		This benchmarks show that the libp2p RSA sigverify seems to be unoptimized and is actually slower than ED25519, as of go-libp2p v0.29.2:
+			libp2p_(Ed25519)/sign-64		36178 ns/op
+			libp2p_(Ed25519)/verify-64		85326 ns/op
+			libp2p_(RSA)/sign-64		  2226550 ns/op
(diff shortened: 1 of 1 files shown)


tbjump:
* main
  * **PR #3228** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3228" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3228?utm_source=stack-comment).

bruce-riley (dismissed):

node/pkg/node/node_test.go
```diff
@@ -1004,6 +1004,34 @@ func BenchmarkCrypto(b *testing.B) {
 		})
 	})
 
+	b.Run("libp2p (RSA)", func(b *testing.B) {
```
kcsongor, line 1007: This is a nice to have, but please include a comment above giving context on why it's here, otherwise it will get confusing in the future to see benchmarks for cryptography algorithms that are not used anywhere in the codebase.

evan-gray (changes requested):
I agree with @kcsongor's sentiment here. My understanding is, as long as it's a benchmark it won't clutter test runs, but having code we agree we'll never use in the code base could be less than ideal. Might it be worth just storing these results in an issue or discussion? I'm on the fence. At least agree we should have a comment stating why this is here.

tbjump:
I added a comment explaining why the code is there. 
As you said, Benchmarks don't run in tests. I think it's alright to keep the code there since in future versions of libp2p the RSA implementation may get optimized and then it's easy to benchmark here.

evan-gray (approved):

bruce-riley (approved):

PR #3234, opened 2023-07-24
Add wormhole middleware to wormchain

misko9:
This PR is a stacked PR and includes #3232 and #3233. Comments regarding x/ibc-hooks should be in #3232. Comments regarding PFM should be in #3233.

Comments regarding wormhole-mw, app.go integration, and wormhole module/VAA changes belong here.


sdk/vaa/payloads.go
```diff
@@ -51,12 +51,13 @@ var (
 	ActionCoreTransferFees   GovernanceAction = 4
 	ActionCoreRecoverChainId GovernanceAction = 5
 
-	// Wormchain cosmwasm governance actions
+	// Wormchain cosmwasm/middlware governance actions
```
nik-suri, line 54: nit: `middleware`

wormchain/x/wormhole/types/message_middleware_contract.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (msg *MsgSetWormholeMiddlewareContract) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Signer)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid signer address (%s)", err)
+	}
+
+	_, err = sdk.AccAddressFromBech32(msg.Address)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid allowlist address (%s)", err)
```
nik-suri, line 39: super nit: "invalid contract address"

wormchain/app/app.go
```diff
@@ -420,9 +440,13 @@ func New(
 
 	// this line is used by starport scaffolding # stargate/app/keeperDefinition
 
+	app.ContractKeeper = wasmkeeper.NewDefaultPermissionKeeper(app.wasmKeeper)
+	app.Ics20WasmHooks.ContractKeeper = app.ContractKeeper
+	app.WormholeMwKeeper.SetWasmKeeper(&app.wasmKeeper)
```
nik-suri, line 445: Why can't we initialize all the middlewares after we initialize the wasmkeeper?
It looks like it might be because we edit the transferKeeper in `app.WireICS20PreWasmKeeper`? But if the transferKeeper is a pointer, we should be able to remove this?

nik-suri (commented):
Looks great! Couple of nits and one question. One more thing - could we rename `wormhole-mw` to something like `ibc-composability-mw`? There may be more custom middlewares we implement on Gateway, so I'd like to keep the name specific.

wormchain/app/app.go
misko9: The transferKeeper, which needs the middlewares setup, needs to be created before the wasmKeeper is created and uses it.

PR #3235, opened 2023-07-25
Node/Deploy: Rootstock readonly

bruce-riley:
This PR adds read only support for Rootstock. Read only meaning to deploy the Wormhole core contract only, without guardian support.


kev1n-peters (dismissed):

sdk/js/src/utils/consts.ts
```diff
@@ -263,6 +265,11 @@ const MAINNET = {
       "sei1smzlm9t79kur392nu9egl8p8je9j92q4gzguewj56a05kyxxra0qy0nuf3",
     nft_bridge: undefined,
   },
+  rootstock: {
+    core: "0xBeBDB6C8dDc678ffa9f8748f85C815c556dD8AC6",
```
scnale, line 269: Note that this uses a different checksum than [EIP-55](https://eips.ethereum.org/EIPS/eip-55) and could cause trouble with ethers as is.

bruce-riley: This is the address you gave me. If it is the output from the truffle migrate, then it should be okay. That's what we always put here.

scnale: I actually copied that from one of the RSK block explorers, that's why it uses a different checksum :sweat_smile:
It's better to use this one just in case: `0xbebdb6C8ddC678FfA9f8748f85C815C556Dd8ac6`

evan-gray (approved):

a5-pickle (approved):
Reviewed ethereum directory

PR #3239, opened 2023-07-26
sdk/js: Replace deprecated function with signSendAndConfirmTransaction

hmzakhalid:
## Summary

This pull request addresses the deprecation of the `sendAndConfirmTransactionsWithRetry` function in the `postVaaWithRetry` method. This function was replaced with the new `signSendAndConfirmTransaction` function as per the recent updates in the library. The retries are done with the `ConfirmOption` and handled internally in the Solana library.

## Changes

The main changes involve replacing the deprecated `sendAndConfirmTransactionsWithRetry` function with the `signSendAndConfirmTransaction` function in the `postVaaWithRetry` method. The `signSendAndConfirmTransaction` function handles each transaction individually, so the code has been adjusted to process each transaction in a loop.

This change was required because `sendAndConfirmTransactionsWithRetry` was marked as deprecated in the library, and the library maintainers recommended using `signSendAndConfirmTransaction` instead.

### DIFF
--- a/sdk/js/src/solana/sendAndConfirmPostVaa.ts
+++ b/sdk/js/src/solana/sendAndConfirmPostVaa.ts
@@ -59,13 +59,13 @@ export async function postVaaWithRetry(
 
   // While the signature_set is used to create the final instruction, it doesn't need to sign it.
   responses.push(
-    (await signSendAndConfirmTransaction(
+    await signSendAndConfirmTransaction(
       connection,
       payer,
       signTransaction,
       postVaaTransaction,
       options
-    ))
+    )
   );
   return responses;
 }
--- a/sdk/js/src/solana/sendAndConfirmPostVaa.ts
+++ b/sdk/js/src/solana/sendAndConfirmPostVaa.ts
@@ -20,6 +20,10 @@ import {
 } from "./wormhole";
 import { isBytes, ParsedVaa, parseVaa, SignedVaa } from "../vaa/wormhole";
 
+/**
+ * @deprecated Please use {@link postVaa} instead, which allows
+ * retries and commitment to be configured in {@link ConfirmOptions}.
+ */
 export async function postVaaWithRetry(
   connection: Connection,
   signTransaction: SignTransaction,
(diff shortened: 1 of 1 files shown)


sdk/js/src/solana/sendAndConfirmPostVaa.ts
```diff
@@ -40,22 +40,32 @@ export async function postVaaWithRetry(
...
-  );
-  //While the signature_set is used to create the final instruction, it doesn't need to sign it.
+  const options: ConfirmOptions = {
+    commitment,
+    maxRetries,
+  };
+
+  const responses: TransactionSignatureAndResponse[] = [];
+  for (const transaction of unsignedTransactions) {
+    const response = await signSendAndConfirmTransaction(
```
barnjamin, line 50: should we be stacking these promises instead of waiting for each?

barnjamin (approved):

barnjamin:
@kev1n-peters or @nonergodic can you plz take a peek at this one?

nonergodic:
Given that the max retries can now be specified via options, isn't the right call here to deprecate `postVaaWithRetry` entirely? (And let's face it, it should have never existed in the first place.)

hmzakhalid:
Hey @nonergodic, I'm in complete favor of deprecating the `postVaaWithRetry` function, as the functionality is identical to `postVaa`, except for the additional `asyncVerifySignatures` parameter. 

Would this deprecation message suffice?
```
/**
 * @deprecated Please use {@link postVaa} instead, which allows
 * retries and commitment to be configured in {@link ConfirmOptions}.
*/
 ```

nonergodic:
Looks good to me :+1:

barnjamin (approved):

hmzakhalid:
@nonergodic Perfect! I've committed the suggested changes.

sdk/js/src/solana/sendAndConfirmPostVaa.ts
hmzakhalid: It's deprecated now so fine either way. The `postVaa` function has an Async parameter for stacking the promises

evan-gray (approved):

PR #3247, opened 2023-07-31
clients/js: fix low gas budget in Sui VAA submit.

scnale:
The default gas budget for Sui transaction blocks is `1_000_000` which is too low to submit VAAs.

### DIFF
--- a/clients/js/src/chains/sui/utils.ts
+++ b/clients/js/src/chains/sui/utils.ts
@@ -38,6 +38,7 @@ export const executeTransactionBlock = async (
   const consoleWarnTemp = console.warn;
   console.warn = () => {};
 
+  transactionBlock.setGasBudget(100_000_000);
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):

clients/js/src/chains/sui/utils.ts
```diff
@@ -38,6 +38,7 @@ export const executeTransactionBlock = async (
   const consoleWarnTemp = console.warn;
   console.warn = () => {};
 
+  transactionBlock.setGasBudget(100_000_000);
```
kcsongor, line 41: a comment pointing to an explorer link for a successful transaction would be nice to give some context to this number

scnale: Yeah, when I made this change I was actually copying what the sdk already did at that point but I think we should have proper gas estimation instead.

PR #3248, opened 2023-07-31
clients/js: fixes hardcoded Near deployer implicit account id.

scnale:
The submit VAA logic for Near in `worm` assumes that the deployer account id is a specific one. This makes it difficult to switch to another key when submitting VAAs to Near.

This adds an environment variable to set the implicit account id to register with the private key but I'm not sure if this is the way we want to go. There might be better alternatives; I haven't looked at the Near SDK that much.

### DIFF
--- a/clients/js/src/consts/networks.ts
+++ b/clients/js/src/consts/networks.ts
@@ -92,7 +92,6 @@ const MAINNET = {
     rpc: "https://rpc.mainnet.near.org",
     key: getEnvVar("NEAR_KEY"),
     networkId: "mainnet",
-    deployerAccount: getEnvVar("NEAR_IMPLICIT_ACCOUNT_ID"),
   },
   injective: {
     rpc: "http://sentry0.injective.network:26657",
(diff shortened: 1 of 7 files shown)


clients/js/src/consts/networks.ts
```diff
@@ -92,8 +92,7 @@ const MAINNET = {
     rpc: "https://rpc.mainnet.near.org",
     key: getEnvVar("NEAR_KEY"),
     networkId: "mainnet",
-    deployerAccount:
-      "85957f38de1768d6db9eab29bee9dd2a01462aff9c8d83daefb9bcd2506c32d2",
+    deployerAccount: getEnvVar("NEAR_IMPLICIT_ACCOUNT_ID"),
```
kev1n-peters, line 95: It looks like there are some typescript compiler errors with this change: https://github.com/wormhole-foundation/wormhole/actions/runs/5720831163/job/15502443893#step:4:87. Maybe we could throw if `deployerAccount` is `undefined` here: https://github.com/wormhole-foundation/wormhole/blob/f11299b888892d5b4abaa624737d0e5117d1bbb2/clients/js/src/near.ts#L162

evan-gray: fixed

evan-gray:
picked this up and implemented @kcsongor 's suggestion

evan-gray (approved):

bruce-riley (approved):

PR #3250, opened 2023-08-01
SDK: fix chain id/name types

barnjamin:
Digging into an [issue ](https://github.com/wormhole-foundation/wormhole/actions/runs/5647923071/job/15299026706?pr=3206)while trying to upgrade the cli tool to the latest SDK I found that the `ChainProvider` was producing the incorrect type given a chain like `solana`

I traced this back to the `utils.ts` file and blame provided the jerk that committed [this change](https://github.com/wormhole-foundation/wormhole/commit/7de46f68b424723522e15d994c3851231748ba97)

Before:
image

After (+npm link the local version):
image

tbh I'm not crazy about this solution but it illustrates the issue. Happy to field other suggestions

### DIFF
--- a/sdk/js/src/utils/consts.ts
+++ b/sdk/js/src/utils/consts.ts
@@ -41,7 +41,7 @@ export type ChainId = typeof CHAINS[ChainName];
  *
  * All the EVM-based chain names that Wormhole supports
  */
-export const EVMChainNames: ReadonlyArray<ChainName> = [
+export const EVMChainNames = [
   "ethereum",
   "bsc",
   "polygon",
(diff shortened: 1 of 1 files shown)


derpy-duck (approved):

sdk/js/src/utils/consts.ts
```diff
@@ -815,28 +809,28 @@ export function isEVMChain(
   chain: ChainId | ChainName
 ): chain is EVMChainId | EVMChainName {
   const chainName = coalesceChainName(chain);
-  return EVMChainNames.includes(chainName);
+  return EVMChainNames.includes(chainName as unknown as EVMChainName);
```
kev1n-peters, line 812: Is it necessary to cast to `unknown` here (and below)? Seems to build for me when we don't do this

barnjamin: It was to make my editor happy, otherwise it cries about `EVMChainNames` having the subset and `chainName` being the full set. 

When it was ReadOnlyArray it was fine because it was (accidentally) evaluating to the full set of chain names despite the values being a subset.

kev1n-peters (approved):

PR #3255, opened 2023-08-02
Node/Gov: Update floor prices

bruce-riley:
This PR updates to governor floor prices. Note that a few symbols were added / removed from the manual token list because the were removed / added from the TVL  list.

I verified that the CoinGecko query still works.

The script to generate the token list (index.ts) shows a ton of changes due to prettier. The main change there is to query Sui to look up the token addresses.


tbjump (changes requested):
It would be nice if there was some way to specify a list of important tokens (e.g. all native tokens, all stablecoins) that should be included no matter what the TVL list produces. 
Bonus if there could be a hardcoded floor price and a test that enforces that the new generated floor price is greater or equal to the hardcoded floor price. 
That would provide overall more assurance that the generation of generated_mainnet_tokens.go is correct. 

If you don't want to solve it in this PR, I'd also be fine with having that as a ticket to follow-up on.

evan-gray (approved):
looks like the appropriate generated output to me

panoel (approved):

tbjump (approved):

PR #3256, opened 2023-08-02
wormchain and others: upgrade to golang 1.20.7

tbjump:
First attempt to upgrade wormchain golang version. 

There may be dependencies that need upgrading as well. 

Not sure what's needed for validation other than Tilt test passing. 
Not sure what's needed for upgrading Guardians in prod.


SEJeff:
Superceded by #3461

PR #3271, opened 2023-08-08
Add PFM to wormchain

misko9:
This is a stacked PR containing PFM and its integration.


evan-gray (commented):
Why do the `node/go.*` files need to change here? 

edit: ...indirect dependency, I see now

evan-gray (changes requested):
I think this PR needs to be rebased. It includes the changes from the previous PR in its commits.

bruce-riley (approved):

evan-gray (approved):
approving node changes

nik-suri (approved):

PR #3272, opened 2023-08-08
AutoRelayer: replay protection

JoeHowarth:
Previously provided as an sdk modifier, we now believe it makes more sense to always have this check done in the base contract.

- Stores block number for successful and failed deliveries
- Implements replay protection if a successful delivery has already occurred
- Also useful for off-chain indexing
- No special treatment of redeliveries


derpy-duck (changes requested):
Not only success, but Forward request success should trigger replay protection

chase-45:
If there is no payload content, then the replay protection doesn't prevent users from seeing an identical relay only differing by transaction hash. In this case the replay protection adds to the cost of the transaction without providing any additional safety it would seem.

chase-45:
> Not only success, but Forward request success should trigger replay protection Also, we should have the mapping be public so that people can query if deliverys happened ALSO (!!!!!!) it would be massive if we could store the block number rather than just a boolean - so that we don't even need wormscan anymore (!!) the SDK itself can provide all the functionalities needed
> 
> also - it shouldn’t revert, it should just consider it a ‘receiver failure’ and deliver the refund right (to avoid the case of a different relayer sniping the delivery

I think this is only true in the happy path case. It doesn't correctly diagnose forwarding or receiver failures.

derpy-duck:
N

> If there is no payload content, then the replay protection doesn't prevent users from seeing an identical relay only differing by transaction hash. In this case the replay protection adds to the cost of the transaction without providing any additional safety it would seem.

this is true but I think the main purpose of replay protection is to prevent the same delivery vaa from being executed twice - if a different delivery vaa is generated that is probably expected protocol behavior

derpy-duck (approved):

derpy-duck:
@JoeHowarth things to consider:
1) should the failure mapping be cleared out when the success one is set
2) should we expose a getter 'isAttempted' that checks both the success and failure mapping

derpy-duck (approved):

chase-45 (approved):

PR #3275, opened 2023-08-08
Gateway/Node: Watcher support

bruce-riley:
To enable the gateway watcher in tilt, set the following guardian parameters:

```
--gatewayWS ws://wormchain:26657/websocket
--gatewayLCD http://wormchain:1317
--gatewayContract wormhole17p9rzwnnfxcjp32un9ug7yhhzgtkhvl9jfksztgw5uh69wac2pgshdnj3k
```


node/pkg/governor/mainnet_chains.go
```diff
@@ -35,5 +35,6 @@ func chainList() []chainConfigEntry {
 		chainConfigEntry{emitterChainID: vaa.ChainIDXpla, dailyLimit: 200_000, bigTransactionSize: 20_000},
 		chainConfigEntry{emitterChainID: vaa.ChainIDSui, dailyLimit: 5_000_000, bigTransactionSize: 500_000},
 		chainConfigEntry{emitterChainID: vaa.ChainIDSei, dailyLimit: 200_000, bigTransactionSize: 20_000},
+		chainConfigEntry{emitterChainID: vaa.ChainIDWormchain, dailyLimit: 200_000, bigTransactionSize: 20_000},
```
panoel, line 38: Is this for gateway?  Or do you need to add one for gateway?

panoel (approved):

node/pkg/governor/mainnet_chains.go
claudijd: This is a bit confusing is this is "Gateway" what happens when/if there's another aggregation of an ecosystem that uses Wormchain as it's mechanism?

evan-gray: @panoel `vaa.ChainIDWormchain` is Gateway, yes.

@claudijd depends on where the ecosystem is aggregated. In the sense of Gateway (the IBC Translator in particular), it deals entirely in wrapped assets custodied on the Gateway chain (aka Wormchain). From the Governor and Global Accountant perspective, when these assets are inbound on the Token Bridge route, they are destined to Wormchain, and when they are outbound (destined to a non-Gateway chain) they are emitted on the Token Bridge from Wormchain.

claudijd: My feedback was mostly on the label fwiw.

claudijd (approved):

PR #3277, opened 2023-08-09
node: update various dependencies

tbjump:
Updating dependencies where I think it makes sense to stay current with latest version and where I think there is good test coverage


tbjump:
* main
  * **PR #3277** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3277" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3277?utm_source=stack-comment).

SEJeff (dismissed):
Numbers go up!

I'd love to see some sort of an actual check for the go version for building that fails the node at startup if it is using an older version of go than is allowed.

tbjump:
I fear that `runtime.Version()` only returns a string that doesn't have hard guarantees around the format. There's no semver check function and doing parsing of the string could be error prone. Ideas? 
Maybe we can explore this in a separate PR?

SEJeff (approved):

PR #3284, opened 2023-08-11
Node/Gov: Don't enqueue already signed VAAs

bruce-riley:
There is currently a problem where old Oasis transactions are being reobserved and they are flooding the governor. This is not really necessary because the transactions have most likely previously reached quorum. 

This PR changes the governor to check to see if a transfer has been previously been approved, and if so, allows it to be published immediately, without impacting the notional value of the governor.

The following checks are done to determine if the transfer can be approved immediately:

- The VAA is already in the signed VAA database.
- The current digest matches what is in the database.
- The VAA in the database was signed by this guardian.

If the above conditions are met, the VAA can be approved immediately. Additionally, this check is made when pending transfers are reloaded on start up, which should immediately clean up the backlog when a guardian upgrades.

### DIFF
--- a/node/pkg/db/governor.go
+++ b/node/pkg/db/governor.go
@@ -22,29 +22,81 @@ type GovernorDB interface {
 	DeleteTransfer(t *Transfer) error
 	DeletePendingMsg(k *PendingTransfer) error
 	GetChainGovernorData(logger *zap.Logger) (transfers []*Transfer, pending []*PendingTransfer, err error)
+	StoreSignedVAA(v *vaa.VAA) error
+	GetSignedVAABytes(id VAAID) (b []byte, err error)
 }
 
 type MockGovernorDB struct {
+	storedVAAs       map[VAAID][]byte
+	transfers        map[string]*Transfer
+	pendingTransfers map[string]*PendingTransfer
 }
 
 func (d *MockGovernorDB) StoreTransfer(t *Transfer) error {
+	if d.transfers == nil {
+		d.transfers = make(map[string]*Transfer)
+	}
+	d.transfers[t.MsgID] = t
 	return nil
 }
 
 func (d *MockGovernorDB) StorePendingMsg(k *PendingTransfer) error {
+	if d.pendingTransfers == nil {
+		d.pendingTransfers = make(map[string]*PendingTransfer)
+	}
+	d.pendingTransfers[k.Msg.MessageIDString()] = k
 	return nil
 }
 
 func (d *MockGovernorDB) DeleteTransfer(t *Transfer) error {
+	if d.transfers != nil {
+		delete(d.transfers, t.MsgID)
+	}
 	return nil
 }
 
 func (d *MockGovernorDB) DeletePendingMsg(pending *PendingTransfer) error {
+	if d.pendingTransfers != nil {
+		delete(d.pendingTransfers, pending.Msg.MessageIDString())
+	}
 	return nil
 }
 
 func (d *MockGovernorDB) GetChainGovernorData(logger *zap.Logger) (transfers []*Transfer, pending []*PendingTransfer, err error) {
-	return nil, nil, nil
+	if d.transfers != nil {
+		for _, t := range d.transfers {
+			transfers = append(transfers, t)
+		}
+	}
+	if d.pendingTransfers != nil {
+		for _, pt := range d.pendingTransfers {
+			pending = append(pending, pt)
+		}
+	}
+	return transfers, pending, nil
+}
+
+func (d *MockGovernorDB) GetSignedVAABytes(id VAAID) (b []byte, err error) {
+	v, exists := d.storedVAAs[id]
+	if !exists {
+		return nil, ErrVAANotFound
+	}
+	return v, nil
+}
+
+func (d *MockGovernorDB) StoreSignedVAA(v *vaa.VAA) error {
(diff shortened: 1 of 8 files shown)


node/pkg/db/governor.go
```diff
@@ -22,29 +22,81 @@ type GovernorDB interface {
 	DeleteTransfer(t *Transfer) error
 	DeletePendingMsg(k *PendingTransfer) error
 	GetChainGovernorData(logger *zap.Logger) (transfers []*Transfer, pending []*PendingTransfer, err error)
+	StoreSignedVAA(v *vaa.VAA) error
```
panoel, line 25: Are these the standard interface that we use, because they don't return errors?

node/pkg/db/governor.go
```diff
@@ -22,29 +22,81 @@ type GovernorDB interface {
...
+		delete(d.pendingTransfers, pending.Msg.MessageIDString())
+	}
 	return nil
 }
 
 func (d *MockGovernorDB) GetChainGovernorData(logger *zap.Logger) (transfers []*Transfer, pending []*PendingTransfer, err error) {
-	return nil, nil, nil
+	if d.transfers != nil {
+		for _, t := range d.transfers {
+			transfers = append(transfers, t)
```
panoel, line 68: Could this ever be too big and blow up the stack?  Or is transfers a pointer to an array?

panoel (dismissed):

node/pkg/db/governor.go
bruce-riley: The real versions in `node/pkg/db/db.go` can return errors. The versions here are only mocks for testing, so returning errors isn't important.

node/pkg/db/governor.go
bruce-riley: This mock is only used in testing, where we only ever insert a couple, so that is not an issue.

tbjump (changes requested):
I agree with the issue that this change is addressing, but I would solve it differently. With the current approach we are adding 
1. another code path that bypasses the governor
2. another place in the code where the guardian private key is being used

That's added threat surface in the two most security critical paths of the codebase, and I would like to avoid that. 

I'd prefer if the processor would check if the Guardian had made this observation in the past and if so broadcast the *existing* signature. That way, the guardian private key doesn't get involved in this new code path at all.

node/pkg/governor/governor.go
```diff
@@ -688,3 +715,54 @@ func (gov *ChainGovernor) HashFromMsg(msg *common.MessagePublication) string {
...
+		gov.logger.Warn("unable to approve previously stored VAA because the digest has changed",
+			zap.String("msgID", msg.MessageIDString()),
+			zap.String("oldDigest", oldDigest),
+			zap.String("newDigest", newDigest),
+		)
+		return false, nil
+	}
+
+	// Generate our signature for this VAA so we can see if we are listed as a signer.
+	sig, err := crypto.Sign(v.SigningDigest().Bytes(), gov.gk)
```
evan-gray, line 752: Right here I was expecting that we look through `v.Signatures` to find our index, then recover the public key from that signature (given that we just verified the hashes match above). If the public key matches our guardian's public key, we should be good to proceed. That would avoid needing the private key here, which would be best avoided.

evan-gray (changes requested):
Let's try without the additional private key usage.

evan-gray:
> # Threat Surface
> I agree with the issue that this change is addressing, but I would solve it differently. With the current approach we are adding
> 
> 1. another code path that bypasses the governor
> 2. another place in the code where the guardian private key is being used
> 
> That's added threat surface in the two most security critical paths of the codebase, and I would like to avoid that.
> 
> I'd prefer if the processor would check if the Guardian had made this observation in the past and if so broadcast the _existing_ signature. That way, the guardian private key doesn't get involved in this new code path at all.
> 
> # False Negatives
> The proposed change checking the stored VAA, which typically only has 13 signatures and even if Guardian `g` has signed a particular message in the past, `g`'s signature may not be in the locally stored VAA for that message, leading to false negatives in this check. To address this issue, all observations would need to be stored, probably separately from the VAAs.

On the threat surface front, see my comment and perhaps that mitigates the concerns. Ultimately it would be good to bypass the re-observation route that could trigger this issue (and the need for this fix) altogether. That said, I think it important to mitigate the issue affecting production today and leave observation collecting (and re-observation handling) to a new, breakout service in the future.

node/pkg/governor/governor.go
bruce-riley: I like this approach much better! I have changed it to use `Ecrecover` with the public key rather than the private key.

tbjump:
I want to get more clarity on the problem this PR is addressing. 
So right now if a Guardian `g0` would re-sync their Oasis node it would lead to that Guardian re-observing many messages, which would exceed the governor limits, leading to many transactions getting enqueued in the Governor. But that shouldn't affect any other Guardians. If `g0` has VAAs for those messages in store, he will not send any observation requests and the messages will just time-out. If on the other hand `g0` does not have any VAA for this message stored, this PR will not change any behavior surrounding that message. 

So this issue should only affect a the Guardians who are re-syncing their Oasis nodes and the issue goes away by itself ~48h after the re-sync is done, correct?

bruce-riley:
This problem was solved differently, so this PR is no longer necessary.

PR #3285, opened 2023-08-12
Setparams gov vaa

nik-suri:


evan-gray (changes requested):
I feel this could be made more generic, like `InPlaceUpgrade` or something with a payload  so if we ever have to do something like this again we don't need to make all these code changes all over again but just update a case statement where like `1` is `SetTokenFactoryPfmDefaultParams` but, you know, `2` could do the next thing we need to do in the future

bruce-riley (dismissed):

bruce-riley (approved):

wormchain/proto/wormhole/tx.proto
```diff
@@ -159,3 +163,15 @@ message MsgMigrateContractResponse {
   bytes data = 1;
 }
 // this line is used by starport scaffolding # proto/tx/message
+
+message MsgInPlaceUpgrade {
+  // Sender is the actor that signs the messages
+  string signer = 1;
+  // Possible set of in place upgrades to perform
+  MsgSetTokenFactoryPfmDefaultParams set_tokenfactory_pfm_default_params = 2;
```
misko9, line 171: Can `MsgSetTokenFactoryPfmDefaultParams set_tokenfactory_pfm_default_params = 2` be remove and replaced with `bytes vaa = 2`? Then, msgServer InPlaceUpgrade parses and verifies VAA, and performs the appropriate execution based on the Action?

evan-gray (changes requested):
okay, I realized we have an opportunity to save some lines of code here. can we make just one handler for `execute_gateway_governance_vaa` that will perform the verification and switch on `action`? 🙏 this could include handling the one from @misko9 's PR that was just merged, since it hasn't been released yet

evan-gray (dismissed):

misko9 (commented):
This should be 0

wormchain/x/wormhole/client/cli/tx_execute_gateway_governance_vaa.go
```diff
@@ -13,28 +13,26 @@ import (
...
+		Short: "Execute the provided Wormhole Gateway governance VAA",
+		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
 				return err
 			}
-			address := args[0]
 
 			vaaBz, err := hex.DecodeString(args[1])
```
misko9, line 28: args[1] should be args[0]

node/cmd/guardiand/admintemplate.go
```diff
@@ -147,6 +149,12 @@ func init() {
 	AdminClientIbcTranslatorUpdateChannelChainCmd.Flags().AddFlagSet(ibcUpdateChannelChainFlagSet)
 	TemplateCmd.AddCommand(AdminClientIbcReceiverUpdateChannelChainCmd)
 	TemplateCmd.AddCommand(AdminClientIbcTranslatorUpdateChannelChainCmd)
+
+	// GatewaySetTokenfactoryPfmDefaultParamsCmd doesn't have any flags
```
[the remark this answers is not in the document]
nik-suri: Updated naming of the governance messages to be consistent.

bruce-riley (approved):

evan-gray (approved):

PR #3287, opened 2023-08-15
AutoRelayer: remove forwards

JoeHowarth:
Forwards are currently broken due to charging the integrator's gas budget for temporary storage slots that can't be refunded properly to the integrator.

In this pr:
- remove all forwards public functions
- remove clearly forwards-only code paths
- remove or rework forwards specific tests

The pr will be followed by a pr allowing cheap cross-chain refunds to make up for the lack of forwards. This will be achieved by not verifying the VAA.


ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerDelivery.sol
```diff
@@ -282,19 +269,22 @@ abstract contract WormholeRelayerDelivery is WormholeRelayerBase, IWormholeRelay
...
+
 
         // Forces external call
         // In order to catch reverts
-        // (If the user's contract requests a forward
-        // and there ends up not being enough funds,
-        // then we will revert this call)
+        // (Previously needed for reverts, but waiting to remove since this is sensitive code 
+        //  and would need a real audit to redeploy)
         try
```
derpy-duck, line 288: correct me if I'm wrong, but executeInstruction should never revert right? Can we not remove the failure case of this

derpy-duck: as in, can we just change this into a normal function call rather than an external call

derpy-duck (commented):
tryDecodeExecuteInstructionError can be removed too probably? (since there will be no errors)

derpy-duck:
tryDecodeExecuteInstructionError can be removed too probably? (since there will be no errors)

sdk/js/scripts/compileAnchorIdls.js
```diff
@@ -5,38 +5,38 @@ const DST_IDL = __dirname + "/../src/anchor-idl";
...
+  if (!fs.existsSync(DST_IDL)) {
+    fs.mkdirSync(DST_IDL);
+  }
 
-    if (!fs.existsSync(TS)) {
-        fs.mkdirSync(TS);
-    }
+  if (!fs.existsSync(TS)) {
+    fs.mkdirSync(TS);
+  }
```
derpy-duck, line 21: do we remove this file's changes from this PR

derpy-duck (approved):

PR #3288, opened 2023-08-15
AutoRelayer: cheap cross-chain refunds

JoeHowarth:


derpy-duck (changes requested):
Ah there is a slight trickiness we have to think through - how do we quote the user the 'base price' of a cross chain refund (used to be an empty delivery):

`
        try deliveryProvider.quoteDeliveryPrice(
            refundChain,
            TargetNative.wrap(0),
            encodeEvmExecutionParamsV1(getEmptyEvmExecutionParamsV1())
        ) returns (LocalNative quote, bytes memory) {
            baseDeliveryPrice = quote;
        } catch (bytes memory) {
            return RefundStatus.CROSS_CHAIN_REFUND_FAIL_PROVIDER_NOT_SUPPORTED;
        }

        // If the refundAmount is not greater than the 'empty delivery price', the refund does not go through
        if (refundAmount <= getWormholeMessageFee() + baseDeliveryPrice) {
            return RefundStatus.CROSS_CHAIN_REFUND_FAIL_NOT_ENOUGH;
        }`
        
  This isn't true anymore - do we just make it free? do we make it a new endpoint on delivery providers? a constant doesn't make sense because it should be a fixed amount in the other chain's gas price.

JoeHowarth:
Like we discussed over slack, we can solve that problem by adding a new ExecutionParamters version and falling back to 0 cost if the DeliveryProvider does not support that version.

JoeHowarth:
Separately, there is a security issue with this pr because you can cause a Delivery event to be emitted based off unverified information. This means we either have to: 

1. not emit a Delivery event for cross-chain refunds and coordinate with DeliveryProvider's to update offchain code accordingly.
2. Somehow indicate in the Delivery event that this was a cross-chain refund IFF it takes the cross-chain refund code path

Regardless, this change is high risk and should be reviewed + audited carefully imo

derpy-duck:
> Separately, there is a security issue with this pr because you can cause a Delivery event to be emitted based off unverified information. This means we either have to:
> 
> 1. not emit a Delivery event for cross-chain refunds and coordinate with DeliveryProvider's to update offchain code accordingly.
> 2. Somehow indicate in the Delivery event that this was a cross-chain refund IFF it takes the cross-chain refund code path
> 
> Regardless, this change is high risk and should be reviewed + audited carefully imo

I don't actually think this change is ~that~ high risk, though I agree that we should emit a different event than 'Delivery' for cross chain refunds now to reduce risk

evan-gray:
Closing due to inactivity.

PR #3291, opened 2023-08-16
Add Gateway support for scheduled upgrades

misko9:


nik-suri (approved):
changes lgtm. Only suggestion would be to add some comments around the fact that only one upgrade can happen at a time, and scheduling a new upgrade implicitly cancel any existing scheduled upgrades.

proto/node/v1/node.proto
```diff
@@ -93,6 +93,10 @@ message GovernanceMessage {
     WormchainInstantiateContract wormchain_instantiate_contract = 15;
     WormchainMigrateContract wormchain_migrate_contract = 16;
 
+    // Gateway
+    GatewayScheduleUpgrade gateway_schedule_upgrade = 24;
```
bruce-riley, line 97: Is there a reason we're skipping 22 and 23?

misko9: To keep it consistent with the proposed changed to main in #3292

nik-suri (approved):

PR #3300, opened 2023-08-17
Node/gacct: Don't reobserve if enqueued in gov

bruce-riley:
When a transfer is enqueued in some of the guardians, but not all of them, the accountant contract reports it as missing. When this happens, the accountant code in the guardian requests a reobservation. This works well if a transaction was just missed by this guardian.

However, if the guardian has the transfer enqueued in the governor, this just generates needless reobservations. Additionally, if there are enough transfers enqueued in the governor, it can cause the reobservation channel to overflow with unnecessary requests.

This PR changes the global accountant to check to see if the transfer is enqueued in the governor before submitting it to the reobservation channel.

### DIFF
--- a/node/pkg/governor/governor_monitoring.go
+++ b/node/pkg/governor/governor_monitoring.go
@@ -75,6 +75,7 @@
 package governor
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"fmt"
 	"sort"
@@ -600,4 +601,22 @@ func (gov *ChainGovernor) publishStatus(hb *gossipv1.Heartbeat, sendC chan<- []b
 	}
 
 	sendC <- b
+}
+
+func (gov *ChainGovernor) IsTransactionEnqueued(emitterChainId vaa.ChainID, txHash ethCommon.Hash) bool {
+	gov.mutex.Lock()
+	defer gov.mutex.Unlock()
+
+	ce, exists := gov.chains[emitterChainId]
+	if !exists {
+		return false
+	}
+
+	for _, pe := range ce.pending {
+		if bytes.Equal(pe.dbData.Msg.TxHash.Bytes(), txHash.Bytes()) {
+			return true
+		}
+	}
+
+	return false
 }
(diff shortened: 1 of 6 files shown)


panoel (approved):

node/pkg/governor/governor_monitoring.go
```diff
@@ -601,3 +602,21 @@ func (gov *ChainGovernor) publishStatus(hb *gossipv1.Heartbeat, sendC chan<- []b
...
+	gov.mutex.Lock()
+	defer gov.mutex.Unlock()
+
+	ce, exists := gov.chains[emitterChainId]
+	if !exists {
+		return false
+	}
+
+	for _, pe := range ce.pending {
+		if bytes.Equal(pe.dbData.Msg.TxHash.Bytes(), txHash.Bytes()) {
```
tbjump, line 616: The TxHash does not uniquely identify a wormhole message because one transaction could create multiple wormhole messages.

tbjump (changes requested):
**Issue with TxHash**
TxHash does not uniquely identify a wormhole message because there could be multiple wormhole messages in a single transaction. Depending on message finality setting, the transaction could also be included in different blocks over time, leading to multiple different wormhole messages being observed by the guardian for the same transaction. 
If this is code gets merged this could yield to false positives in the `IsTransactionEnqueued` function, in turn leading to re-observation requests not being submitted when they actually should be submitted. 
Fixing this would require changes to the Global Accountant contracts, which may be a bigger lift. 
So I wouldn't be opposed merging this code since it's introducing one bug but fixing a bigger bug. The new bug should be properly documented in the code however. 

**Location where observation requests are filtered**
From a clean design perspective, I don't like that the Governor is becoming a dependency of the Accountant. Now the Accountant needs to understand the business logic of the Governor. 
I think it would be better if all incoming observation requests run through the processor and the processor checks the Governor for each observation request before forwarding them to the chain-specific observation request channels. This way it can also filter observation requests from other sources for which the message is pending in the Governor, not just observation requests that originated from the accountant.

PR #3301, opened 2023-08-17
Add tests for ibc-translator contract

misko9:


cosmwasm/contracts/ibc-translator/src/reply.rs
```diff
@@ -203,7 +203,7 @@ pub fn convert_cw20_to_bank_and_send(
 }
 
 // Base58 allows the subdenom to be a maximum of 44 bytes (max subdenom length) for up to a 32 byte address
-fn contract_addr_to_base58(deps: Deps, contract_addr: String) -> Result<String, anyhow::Error> {
+pub fn contract_addr_to_base58(deps: Deps, contract_addr: String) -> Result<String, anyhow::Error> {
```
nik-suri, line 206: nit - I don't think we need to make this public. Can we structure the tests similarly to https://github.com/everstake/sei-token-translator/blob/main/src/contract.rs#L432-L433?

nik-suri (approved):
Great work on these tests, thanks!

kcsongor (approved):

PR #3307, opened 2023-08-22
node/processor: Remove dead code

tbjump:
k.Timestamp is always equal to existing.Timestamp. Therefore `k.Timestamp.Sub(existing.Timestamp) > settlementTime` is always false. 
I am proposing to remove this dead code to avoid confusion. 

The intention of this code was to not make observations if there is a VAA already, but I think this is actually unwanted behavior because that would break re-observation requests and even if this was wanted, the check should've been done before accounting/governor and not after.

### DIFF
--- a/node/pkg/db/db.go
+++ b/node/pkg/db/db.go
@@ -128,6 +128,20 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
 	return nil
 }
 
+func (d *Database) HasVAA(id VAAID) (bool, error) {
+	err := d.db.View(func(txn *badger.Txn) error {
+		_, err := txn.Get(id.Bytes())
+		return err
+	})
+	if err == nil {
+		return true, nil
+	}
+	if err == badger.ErrKeyNotFound {
+		return false, nil
+	}
+	return false, err
+}
+
 func (d *Database) GetSignedVAABytes(id VAAID) (b []byte, err error) {
 	if err := d.db.View(func(txn *badger.Txn) error {
 		item, err := txn.Get(id.Bytes())
--- a/node/pkg/processor/cleanup.go
+++ b/node/pkg/processor/cleanup.go
@@ -81,7 +81,7 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 			// This occurs when we observed a message after the cluster has already reached
 			// consensus on it, causing us to never achieve quorum.
 			if ourVaa, ok := s.ourObservation.(*VAA); ok {
-				if _, err := p.getSignedVAA(*db.VaaIDFromVAA(&ourVaa.VAA)); err == nil {
+				if p.haveSignedVAA(*db.VaaIDFromVAA(&ourVaa.VAA)) {
 					// If we have a stored quorum VAA, we can safely expire the state.
 					//
 					// This is a rare case, and we can safely expire the state, since we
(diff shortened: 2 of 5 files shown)


tbjump:
* main
  * **PR #3307** <a href="https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3307" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/wormhole-foundation/wormhole/3307?utm_source=stack-comment).

node/pkg/db/db.go
```diff
@@ -128,6 +128,20 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
...
+func (d *Database) HasVAA(id VAAID) (bool, error) {
+	err := d.db.View(func(txn *badger.Txn) error {
+		_, err := txn.Get(id.Bytes())
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err == badger.ErrKeyNotFound {
+		return true, nil
```
bruce-riley, line 140: This looks backwards. Shouldn't it return false if the key is not found?

tbjump: Yes. Fixed.

node/pkg/db/db.go
```diff
@@ -128,6 +128,23 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
 	return nil
 }
 
+func (d *Database) HasVAA(id VAAID) (bool, error) {
+	err := d.db.View(func(txn *badger.Txn) error {
+		_, err := txn.Get(id.Bytes())
+		if err != nil {
```
bruce-riley, line 134: Could just return err here, right?

node/pkg/db/db.go
```diff
@@ -128,6 +128,23 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
...
+			return err
+		}
+		return nil
+	})
+	if err == nil {
+		return true, nil
+	}
+	if err == badger.ErrKeyNotFound {
+		return false, ErrVAANotFound
```
bruce-riley, line 143: Think this should be `return false, nil`.

node/pkg/processor/cleanup.go
```diff
@@ -81,7 +81,7 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 			// This occurs when we observed a message after the cluster has already reached
 			// consensus on it, causing us to never achieve quorum.
 			if ourVaa, ok := s.ourObservation.(*VAA); ok {
-				if _, err := p.getSignedVAA(*db.VaaIDFromVAA(&ourVaa.VAA)); err == nil {
+				if p.haveSignedVAA(*db.VaaIDFromVAA(&ourVaa.VAA)) {
```
bruce-riley, line 84: Check for err?

node/pkg/db/db.go
tbjump: yes, fixed

node/pkg/processor/cleanup.go
tbjump: If there is an error, p.haveSignedVAA() will just return false but the error is logged in p.haveSignedVAA now. What do you think?

bruce-riley (approved):

PR #3320, opened 2023-08-24
CCQ: Gossip split

bruce-riley:


node/pkg/p2p/p2p.go
```diff
@@ -351,6 +364,9 @@ func Run(
 					gst.Cleanup()
 				case <-ctx.Done():
 					return
+				case ccqErr := <-ccqErrC:
```
evan-gray, line 367: just curious, why does the err channel handler go here? it to make this stop on error?

evan-gray (approved):
looks like a effective start to segmenting this traffic.

node/pkg/p2p/p2p.go
bruce-riley: Fixed.

evan-gray (approved):

node/pkg/p2p/ccq_p2p.go
```diff
@@ -0,0 +1,296 @@
...
+
+		// Enable TLS security as the only security protocol.
+		libp2p.Security(libp2ptls.ID, libp2ptls.New),
+
+		// Enable QUIC transport as the only transport.
+		libp2p.Transport(libp2pquic.NewTransport),
+
+		// Let's prevent our peer from having too many
+		// connections by attaching a connection manager.
+		libp2p.ConnectionManager(components.ConnMgr), // TODO: Can we use the same connection manager?
```
kev1n-peters, line 110: I think that each p2p instance should be using its own connection manager otherwise it could prune each other's connections

bruce-riley: The code is actually creating a separate `components` object at line 87, which means the connection manager is not shared. I'll remove that todo with my next PR.

PR #3140, opened 2023-06-27
node: processor v2 (refactor)

tbjump:
This is the reincarnation of https://github.com/wormhole-foundation/wormhole/pull/1953, but better: 
* stacked on benchmarking code so we can compare the performance to pre-refactor
* instead of replacing the previous processor package, the refactored version is added as a new package `processor2`. Users can pick the old or new version interchangeably. 

With this refactor, the guardian message throughput is only up to ca. 3 messages/s (less with higher concurrency) vs. the previous ~30 messages/s on my host running 19 guardians. 

Suspected reasons for bad performance: 
* starts lot's of goroutines that are managed by the supervisor. 13% of execution time is spent in supervisor.processGC. 
* Too many goroutines in general 
  * 11.4% of time spent malloc'ing
  * 7.7% of time spent in runtime.schedule
  * 7.7% of time spent in runtime.findRunnable
  * lot's of time spent in golang GC

PR #3322, opened 2023-08-25
Node: Add StartRunnable to scissors

bruce-riley:
This PR adds a `StartRunnable` method to the `RunWithScissors` interface. This method allows you to control if panics should be caught or not, and is intended to eventually replace the `RunWithScissors` calls (which will be done with other PRs).

### DIFF
--- a/node/pkg/common/scissors.go
+++ b/node/pkg/common/scissors.go
@@ -24,37 +24,7 @@ var (
 
 // Start a go routine with recovering from any panic by sending an error to a error channel
 func RunWithScissors(ctx context.Context, errC chan error, name string, runnable supervisor.Runnable) {
-	ScissorsErrorsCaught.WithLabelValues(name).Add(0)
-	ScissorsPanicsCaught.WithLabelValues(name).Add(0)
-	go func() {
-		defer func() {
-			if r := recover(); r != nil {
-				var err error
-				switch x := r.(type) {
-				case error:
-					err = fmt.Errorf("%s: %w", name, x)
-				default:
-					err = fmt.Errorf("%s: %v", name, x)
-				}
-				// We don't want this to hang if the listener has already gone away.
-				select {
-				case errC <- err:
-				default:
-				}
-				ScissorsPanicsCaught.WithLabelValues(name).Inc()
-
-			}
-		}()
-		err := runnable(ctx)
-		if err != nil {
-			// We don't want this to hang if the listener has already gone away.
-			select {
-			case errC <- err:
-			default:
-			}
-			ScissorsErrorsCaught.WithLabelValues(name).Inc()
-		}
-	}()
+	StartRunnable(ctx, errC, true, name, runnable)
 }
 
 func WrapWithScissors(runnable supervisor.Runnable, name string) supervisor.Runnable {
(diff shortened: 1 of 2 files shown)


node/pkg/common/scissors.go
```diff
@@ -76,3 +76,50 @@ func WrapWithScissors(runnable supervisor.Runnable, name string) supervisor.Runn
...
+// StartRunnable starts a go routine with the ability to recover from errors by publishing them to an error channel. If catchPanics is true,
+// it will also catch panics and publish the panic message to the error channel. If catchPanics is false, the panic will be propagated upward.
+func StartRunnable(ctx context.Context, errC chan error, catchPanics bool, name string, runnable supervisor.Runnable) {
+	ScissorsErrorsCaught.WithLabelValues(name).Add(0)
+	if catchPanics {
+		ScissorsPanicsCaught.WithLabelValues(name).Add(0)
+	}
+	go func() {
+		if catchPanics {
+			defer func() {
```
evan-gray, line 89: can this func be named and shared with the one on line 30? unless there's a go thing I don't get (it's possible), seems like we could simply take `name` and `errC` as arguments.

panoel (dismissed):

node/pkg/common/scissors.go
```diff
@@ -76,3 +76,50 @@ func WrapWithScissors(runnable supervisor.Runnable, name string) supervisor.Runn
...
+
+				}
+			}()
+		}
+		startRunnable(ctx, errC, name, runnable)
+	}()
+}
+
+// startRunnable is used by StartRunnable. It is a separate function so we can call it directly from tests.
+func startRunnable(ctx context.Context, errC chan error, name string, runnable supervisor.Runnable) {
```
evan-gray, line 113: can line 48 use this function?

evan-gray (changes requested):

node/pkg/common/scissors.go
bruce-riley: I made `RunWithScissors` call `StartRunnable`.

node/pkg/common/scissors.go
bruce-riley: I made `RunWithScissors` call `StartRunnable`.

panoel (dismissed):

evan-gray (dismissed):

node/pkg/common/scissors.go
```diff
@@ -24,27 +24,64 @@ var (
...
+	func() {
```
tbjump, line 84: this `func() {` seems unnecessary

tbjump (dismissed):

panoel (approved):

tbjump:
It seems like this is going down the path of building an alternative to the supervisor. 
What is the motivation behind not using the supervisor for starting child routines? Using the supervisor has some benefits, e.g. having child loggers, being able to go through different states of the process (Healthy/Done), and automatically restarting on error. 
I'm wondering if it wouldn't be better to just add a feature to the supervisor to catch panics for certain sub-trees, like Hendrik originally mentioned: https://github.com/wormhole-foundation/wormhole/pull/2187#pullrequestreview-1235772285

evan-gray (approved):

evan-gray:
> It seems like this is going down the path of building an alternative to the supervisor.
> What is the motivation behind not using the supervisor for starting child routines? Using the supervisor has some benefits, e.g. having child loggers, being able to go through different states of the process (Healthy/Done), and automatically restarting on error.
> I'm wondering if it wouldn't be better to just add a feature to the supervisor to catch panics for certain sub-trees, like Hendrik originally mentioned: [#2187 (review)](https://github.com/wormhole-foundation/wormhole/pull/2187#pullrequestreview-1235772285)

I think we would want to consider / test how weighty the supervisor would be for some of these flows vs a simple go func catcher like this.

tbjump:
The supervisor has some overhead, due to supervisor.processGC. We saw that in https://github.com/wormhole-foundation/wormhole/pull/3140. 
But I think that only becomes and issue if there are 1000s of supervisor nodes created. I'm not aware of any use-case for that in the guardian today, and even if that will be a future use case, supervisor.processGC has probably some optimization potential.

bruce-riley:
I do not see this as "going down the path of building an alternative to the supervisor". I see this as building upon the existing `RunWithScissors` pattern to allow it to be used without catching panics. I think it is adding flexibility to an existing interface.

tbjump:
Yeah I'm wondering if the `RunWithScissors` pattern was that great to start with. I know that it was implemented with some urgency to respond to production problems, but before we add more features to it and use it more widely, I wanted to pause for a second and see if there was appetite for implementing that functionality inside of the supervisor instead.

PR #3328, opened 2023-08-27
aptos: use aptos_account::deposit_coins

kcsongor:
This allows transferring coins to accounts that don't exist, or exist but haven't opted in to the coins yet.
https://github.com/aptos-labs/aptos-core/blob/e71a5169217a2ebe23f39838ee3099981ed2d560/aptos-move/framework/aptos-framework/sources/aptos_account.move#L98-L115

In the tests, I removed all the code that creates accounts and registers them for the coins to ensure these are now done by the complete transfer function.


a5-pickle (approved):

evan-gray (approved):
exciting!

PR #3329, opened 2023-08-28
clients/js: fix aptos upgrade and migrate commands

kcsongor:
For some reason the bound variables in the lambda were underscored, unclear why. In turn, `yargs` referred to the package, resulting in runtime errors.


evan-gray (approved):
the diff really _underscores_ the bug

panoel (approved):

PR #3330, opened 2023-08-28
sdk/js: Aptos transfer with payload support

kev1n-peters:


evan-gray (approved):

derpy-duck (approved):
Looks good given the test passes!

PR #3336, opened 2023-08-29
Tokenfactory metadata symbol fix

misko9:
- if symbol is empty when registering an asset, populate tokenfactory's metadata symbol with the denom with exponent string so it is not empty.


nik-suri (dismissed):

kcsongor (approved):

nik-suri (approved):
LGTM and verified tests pass locally, thanks!

kcsongor (approved):

PR #3338, opened 2023-08-29
aptos: tooling for mainnet upgrades

kcsongor:
This PR adds support for mainnet aptos upgrades in the aptos upgrade script and the governance proposal generator.
Also added the missing testnet NFT bridge address.

Changed the default behaviour of the governance proposal generator script to fail if the directory is dirty. This can be overridden with the `—force` flag.

### DIFF
--- a/aptos/scripts/deploy
+++ b/aptos/scripts/deploy
@@ -11,21 +11,17 @@ EOF
 exit 1
 }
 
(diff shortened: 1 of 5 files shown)


evan-gray (approved):

kev1n-peters (approved):

a5-pickle (approved):
Reviewed aptos directory

PR #3344, opened 2023-08-31
sdk/js: change sei array conversion

panoel:


sdk/js/src/utils/array.ts
```diff
@@ -109,6 +109,10 @@ export const tryUint8ArrayToNative = (
   } else if (chainId === CHAIN_ID_XPLA) {
     return humanAddress("xpla", a.slice(-20));
   } else if (chainId === CHAIN_ID_SEI) {
+    const h = uint8ArrayToHex(a);
+    if (!isLikely20ByteCosmwasm(h)) {
+      return humanAddress("sei", a);
```
evan-gray, line 114: could make these a one liner like
```
return humanAddress("sei", isLikely20ByteCosmwasm(h) ? a.slice(-20) : a);
```

evan-gray (changes requested):
I think this change might be applicable for XPLA, Injective, and Wormchain too... I'm pretty sure the comment for wormchain is incorrect now 😞

sdk/js/src/utils/array.ts
panoel: Fixed.

evan-gray (approved):

sdk/js/src/utils/array.test.ts
```diff
@@ -33,10 +33,45 @@ test("wormchain address conversion", () => {
...
+
 test("injective address conversion", () => {
   const human = "inj180rl9ezc4389t72pc3vvlkxxs5d9jx60w9eeu3";
   const canonical = canonicalAddress(human);
   const lpadCanonical = zeroPad(canonical, 32);
   const native = tryUint8ArrayToNative(lpadCanonical, "injective");
   expect(native).toBe(human);
 });
+
+test("sei address conversion", () => {
```
kev1n-peters, line 71: Should we test a real sei address here?

panoel: 😅 How did that get in there???  I had an example from xLabs in there.  Fishing it back out.  Good catch!

evan-gray (approved):

kev1n-peters (approved):

PR #3347, opened 2023-09-01
CCQ: Server allow list

bruce-riley:
This PR adds a configurable allow list to the `ccq_server` so that we can specify which users are allowed to make queries and which RPC calls they are allowed to make.

### DIFF
--- a/node/cmd/ccq/devnet.config.json
+++ b/node/cmd/ccq/devnet.config.json
@@ -1,0 +1,64 @@
+{
+  "permissions": [
+    {
+      "userName": "Test User",
+      "apiKey": "my_secret_key",
+      "allowedCalls": [
+        {
+          "ethCall": {
+            "note:": "Name of WETH on Goerli",
+            "chain": 2,
+            "contractAddress": "B4FBF271143F4FBf7B91A5ded31805e42b2208d6",
+            "call": "0x06fdde03"
+          }
+        },
+        {
+          "ethCall": {
+            "note:": "Total supply of WETH on Goerli",
+            "chain": 2,
+            "contractAddress": "0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6",
+            "call": "0x18160ddd"
+          }
+        },
+        {
+          "ethCall": {
+            "note:": "Name of WETH on Devnet",
+            "chain": 2,
+            "contractAddress": "0xDDb64fE46a91D46ee29420539FC25FD07c5FEa3E",
+            "call": "0x06fdde03"
+          }
+        },
+        {
+          "ethCall": {
+            "note:": "Total supply of WETH on Devnet",
+            "chain": 2,
+            "contractAddress": "0xDDb64fE46a91D46ee29420539FC25FD07c5FEa3E",
+            "call": "0x18160ddd"
+          }
+        }
+      ]
+    },
+    {
+      "userName": "Test User Two",
+      "apiKey": "my_secret_key_2",
+      "allowedCalls": [
+        {
+          "ethCall": {
+            "note:": "Name of WETH on Goerli",
+            "chain": 2,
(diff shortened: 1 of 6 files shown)


node/cmd/ccq/utils.go
```diff
@@ -39,3 +49,151 @@ func FetchCurrentGuardianSet(rpcUrl, coreAddr string) (*common.GuardianSet, erro
...
+	ContractAddress string `json:"contractAddress"`
+	Call            string `json:"call"`
+}
+
+type Permissions map[string]*permissionEntry
+
+type permissionEntry struct {
+	userName     string
+	apiKey       string
+	allowedCalls allowedCallsForUser // Key is something like "thCall:2:000000000000000000000000b4fbf271143f4fbf7b91a5ded31805e42b2208d6:06fdde03"
```
kev1n-peters, line 78: Should this be `"ethCall"`

node/cmd/ccq/devnet.config.json
```diff
@@ -0,0 +1,64 @@
...
+  "permissions": [
+    {
+      "userName": "Test User",
+      "apiKey": "my_secret_key",
+      "allowedCalls": [
+        {
+          "ethCall": {
+            "note:": "Name of WETH on Goerli",
+            "chain": 2,
+            "contractAddress": "B4FBF271143F4FBf7B91A5ded31805e42b2208d6",
```
panoel, line 11: Do the contract addresses need to have leading "0x"?

node/cmd/ccq/http.go
```diff
@@ -38,16 +42,24 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// The API key should be in the header.
+	apiKey, exists := r.Header["X-Api-Key"]
+	if !exists || len(apiKey) != 1 {
```
panoel, line 47: The length of the api key is exactly 1?

node/cmd/ccq/utils.go
```diff
@@ -39,3 +49,151 @@ func FetchCurrentGuardianSet(rpcUrl, coreAddr string) (*common.GuardianSet, erro
...
+
+type EthCall struct {
+	Chain           int    `json:"chain"`
+	ContractAddress string `json:"contractAddress"`
+	Call            string `json:"call"`
+}
+
+type Permissions map[string]*permissionEntry
+
+type permissionEntry struct {
```
panoel, line 75: Isn't this just a User struct?

node/cmd/ccq/utils.go
```diff
@@ -39,3 +49,151 @@ func FetchCurrentGuardianSet(rpcUrl, coreAddr string) (*common.GuardianSet, erro
...
+	userName     string
+	apiKey       string
+	allowedCalls allowedCallsForUser // Key is something like "ethCall:2:000000000000000000000000b4fbf271143f4fbf7b91a5ded31805e42b2208d6:06fdde03"
+}
+
+type allowedCallsForUser map[string]struct{}
```
panoel, line 81: You can create a type with an empty struct?

node/cmd/ccq/devnet.config.json
bruce-riley: Nope. It handles both ways.

node/cmd/ccq/http.go
bruce-riley: That map look up returns a slice, so it is checking to make sure that there is only API key in the header. I updated the comment.

node/cmd/ccq/utils.go
bruce-riley: Yes, this is how you do a set in go.

node/cmd/ccq/utils.go
bruce-riley: Nope, `allowCalls` is different.

panoel (approved):

PR #3349, opened 2023-09-04
Updating sui contract address

barnjamin:
submitting this change to the SDK: https://github.com/wormhole-foundation/xdapp-book/pull/149


sebastianscatularo (approved):

kev1n-peters:
I don't think we want to make this change. `0xc57508ee0d4595e5a8728974a4a93a787d38f339757230d441e895422c07aba9` is the Sui object ID of the token bridge state module whereas `0xccceeb29348f71bdd22ffef43a2a19c1f5b5e17c5cca5411529120182672ade5` is the token bridge emitter address (the key insight is that for Sui the "contract address" and emitter address are different

PR #3350, opened 2023-09-05
Node/Gateway: Relay attestations

bruce-riley:
This PR adds the ability for the gateway relayer to forward attestation VAAs to the token bridge contract on Gateway / Wormchain.

### DIFF
--- a/node/pkg/gwrelayer/gwrelayer.go
+++ b/node/pkg/gwrelayer/gwrelayer.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/certusone/wormhole/node/pkg/common"
 	"github.com/certusone/wormhole/node/pkg/supervisor"
+	"github.com/wormhole-foundation/wormhole/sdk"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 
 	wasmdtypes "github.com/CosmWasm/wasmd/x/wasm/types"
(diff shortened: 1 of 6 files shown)


node/pkg/gwrelayer/gwrelayer.go
```diff
@@ -171,15 +267,32 @@ func shouldPublish(payload []byte, targetChain vaa.ChainID, targetAddress vaa.Ad
...
+func shouldPublishToTokenBridge(tokenBridges tokenBridgeMap, v *vaa.VAA) bool {
+	if _, exists := tokenBridges[tokenBridgeKey{emitterChainId: v.EmitterChain, emitterAddr: v.EmitterAddress}]; !exists {
+		return false
+	}
+
+	if len(v.Payload) == 0 {
+		return false
+	}
+
+	if v.Payload[0] != 2 {
```
nik-suri, line 280: Is there a const we can use here for VAA payload type instead of using `2` as a magic number?

bruce-riley: I am not aware of them being defined anywhere. I added a comment though.

node/pkg/gwrelayer/gwrelayer.go
```diff
@@ -171,15 +267,33 @@ func shouldPublish(payload []byte, targetChain vaa.ChainID, targetAddress vaa.Ad
...
+func shouldPublishToTokenBridge(tokenBridges tokenBridgeMap, v *vaa.VAA) bool {
+	if _, exists := tokenBridges[tokenBridgeKey{emitterChainId: v.EmitterChain, emitterAddr: v.EmitterAddress}]; !exists {
+		return false
+	}
+
+	if len(v.Payload) == 0 {
+		return false
+	}
+
+	// We only forward attestations (type two).
```
nik-suri, line 280: type two covers simple (non-payload3) token bridge messages as well, right?

Should we lock this down further to **only** forward attestations? I actually think it's okay to also forward non-payload3 messages, since people may accidentally initiate transfers directly to Gateway. cc @evan-gray

bruce-riley: No, type two is only attestations:

https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0003_token_bridge.md?plain=1#L185

evan-gray (approved):
reviewed non-node code

panoel (approved):

nik-suri (approved):

PR #3352, opened 2023-09-05
AutoRelayer: Simplify executeInstruction to internal call

JoeHowarth:


ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerDelivery.sol
```diff
@@ -368,16 +322,16 @@ abstract contract WormholeRelayerDelivery is WormholeRelayerBase, IWormholeRelay
             Gas postGas = Gas.wrap(gasleft());
 
             unchecked {
-                gasUsed = (preGas - postGas).min(gasLimit);
+                results.gasUsed = (preGas - postGas).min(gasLimit);
             }
         }
```
derpy-duck, line 327: does the above block need to be seperated out (with a '{ }') - maybe it was there for a stack too deep that no longer exists? (i'm referencing the block from line 298 to 327)

derpy-duck (approved):

PR #3356, opened 2023-09-05
Ccq/p2p with single host

bruce-riley:
This PR attempts to switch CCQ to use the same host / bootstrap parameters as regular gossip, but different pub/sub channels.

To test this:
- Bring up tilt with `tilt up -- --manual` (Filter for ccq in the guardian and you should see the ccq server initialializing.)
- Run the test by cd'ing to node/hack/query and doing `kubectl --namespace=wormhole exec -it spy-0 -- sh -c "cd node/hack/query/ && go run send_req.go"`.

You should see the request being received in the guardian and the response being published by the guardian. However, the test will hang waiting on the response (last message in stdout is "Waiting for message").

### DIFF
--- a/node/pkg/p2p/ccq_p2p.go
+++ b/node/pkg/p2p/ccq_p2p.go
@@ -16,16 +16,10 @@ import (
 
 	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
 
-	"github.com/libp2p/go-libp2p"
-	dht "github.com/libp2p/go-libp2p-kad-dht"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/peer"
-	"github.com/libp2p/go-libp2p/core/protocol"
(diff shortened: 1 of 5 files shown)


node/pkg/p2p/ccq_p2p.go
```diff
@@ -73,67 +65,23 @@ func newCcqRunP2p(
...
-				dht.BootstrapPeers(bootstrappers...),
-			)
-			return idht, err
-		}),
-	)
-
-	if err != nil {
-		return fmt.Errorf("failed to create p2p: %w", err)
+	if h == nil {
+		return fmt.Errorf("h is not initialized")
```
SEJeff, line 77: Can this be a more descriptive error? Preferably one that doesn't require reading the source and finding the definition of `h` in the function arg spec?

SEJeff (commented):
Some comments before you split this all out. I'd love to see more tests for the business logic, but this is great.

External facing services should ideally have:
* Unit test coverage of the positive and negative code paths
* Metrics for rpcs, errors, etc
* Logging of requests (middleware for the webserver).

SEJeff (commented):
These comments are ones I made several days ago, but forgot to submit. I'm going to submit them now and then try to reproduce your issue.

PR #3359, opened 2023-09-06
Client/js: Simplify edit-vaa command

bruce-riley:
This PR changes the `worm edit-vaa` command so that you can tell it to look up the signatures on wormscan without specifying the complete URL.

Instead of doing:

`worm edit-vaa -n mainnet --vaa $VAA --wormscanurl https://api.wormscan.io/api/v1/observations/1/0000000000000000000000000000000000000000000000000000000000000004/651169458827220885`

you can just do:

`worm edit-vaa -n mainnet --vaa $VAA --wormscan`

and it will use the sequence number from the VAA bytes to generate the URL. The old command is still supported.

### DIFF
--- a/clients/js/README.md
+++ b/clients/js/README.md
@@ -94,8 +94,8 @@ Options:
       --signatures, --sigs         comma separated list of signatures   [string]
       --wormscanurl, --wsu         url to wormscan entry for the vaa that
                                    includes signatures                  [string]
-      --wormscanfile, --wsf        json file containing wormscan entry for the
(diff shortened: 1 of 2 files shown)


evan-gray (changes requested):
Nice, but can this take chain and emitter from the bytes too? 🙏

kcsongor (approved):

evan-gray (approved):

kev1n-peters (approved):

PR #3361, opened 2023-09-07
solana: pattern match instead of label

kcsongor:
This makes it easier to see which fields of a struct are used. Will follow up with the same refactoring in other places where appropriate


a5-pickle (approved):
Beautiful

PR #3362, opened 2023-09-07
solana: make SigVerifyParameters canonical

kcsongor:
all messages are checked to be the same, so we just store one of them


a5-pickle (approved):
Sick changes

PR #3368, opened 2023-09-11
solana: fix core bridge IDL

a5-pickle:
This PR fixes the ordering of the last two accounts for the Post Message instruction. This change makes invoking this instruction consistent with how CPI is expected to call this instruction (see Token Bridge implementation as an example).


kev1n-peters (approved):

a5-pickle:
Not necessary anymore

PR #3373, opened 2023-09-11
CCQ: Query demo contract

bruce-riley:


ethereum/contracts/query/QueryDemo.sol
```diff
@@ -0,0 +1,103 @@
...
+
+        return ret;
+    }
+
+    // updateCounters takes the cross chain query response for the two other counters, stores the results for the other chains, and updates the counter for this chain.
+    function updateCounters(bytes memory response, IWormhole.Signature[] memory signatures) public {
+        uint256 adjustedBlockTime;
+        ParsedQueryResponse memory r = parseAndVerifyQueryResponse(address(wormhole), response, signatures);
+        require(r.responses.length == foreignChainIDs.length, "unexpected number of results");
+        for (uint idx=0; idx<r.responses.length; idx++) {
```
SEJeff, line 70: Is it possible for there to be enough records in r.responses for this to consume the block gas limit? Should there be a strict upper bounds on the number of responses? With so many storage reads and updates, this seems like it could be an expensive function.

bruce-riley: The number of responses is limited to 255, and enforced by the guardian. If that turns out to require too much gas, we could enforce something here, but for this demo, I don't think that should be an issue.

SEJeff: Ah totally thanks. I just see a lot of DoS / Gas griefing and always am leery of unbounded loops in solidity contracts with user supplied data.

bounding at 255 max would be way below the gas limits, thanks.

PR #3377, opened 2023-09-13
[BUG] save ethers as a dependency, not a devDependency

duxiaofeng-github:
Ethers should be a dependency of wormhole JS SDK, not a devDependency.

If we put ethers on devDependency, it doesn't have a compatible version calculation, so when I use ethers V6 in my project, and wormhole JS SDK uses the ethers V5, then npm won't install ethers v5 for wormhole JS SDK.

So, it will use the v6 directly, and ethers v5 and V6 are incompatible and will fail in the build phase.


evan-gray:
Thanks for the PR. The `package-lock.json` had conflicts so I rebased.

evan-gray:
One thing I'm concerned about is the type incompatibility with the Signer object that gets passed into the sdk methods. I might have considered solving with a peer dependency, but that would likely result in the same issue when you don't want to use ethers v5 in your project. 🤔

duxiaofeng-github:
> One thing I'm concerned about is the type incompatibility with the Signer object that gets passed into the sdk methods. I might have considered solving with a peer dependency, but that would likely result in the same issue when you don't want to use ethers v5 in your project. 🤔

Do you mean the v5 signer typing is incompatible with v6?

PR #3379, opened 2023-09-13
CCQ: Server allow unsigned requests

bruce-riley:


kev1n-peters (approved):

node/cmd/ccq/utils.go
```diff
@@ -197,3 +221,42 @@ func validateRequest(logger *zap.Logger, perms Permissions, apiKey string, qr *g
 
 	return nil
 }
+
+const (
+	GuardianKeyArmoredBlock = "WORMHOLE GUARDIAN PRIVATE KEY"
```
evan-gray, line 226: Ideally this would be something like `CCQ SERVER PRIVATE KEY`

bruce-riley: By using the above string we can generate a key by using the following existing command as follows:
`guardiand keygen <someKeyFileName`

SEJeff: Could we add a `--ccq` flag to guardiand keygen that simply changes that as @evan-gray suggested? Is it even worth it?

node/cmd/ccq/utils.go
```diff
@@ -197,3 +221,42 @@ func validateRequest(logger *zap.Logger, perms Permissions, apiKey string, qr *g
...
+		return nil, fmt.Errorf("failed to open file: %w", err)
+	}
+
+	p, err := armor.Decode(f)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read armored file: %w", err)
+	}
+
+	if p.Type != GuardianKeyArmoredBlock {
+		return nil, fmt.Errorf("invalid block type: %s", p.Type)
```
SEJeff, line 242: It might be nice (for the SRE / ops folks here) to put what is being expected in this error message. What do you think?
```suggestion
		return nil, fmt.Errorf("invalid block type: '%s', expecting '%s'", p.Type, GuardianKeyArmoredBlock)
```

node/cmd/ccq/utils.go
bruce-riley: Changed it to use a block type of "CCQ SERVER SIGNING KEY".

node/cmd/ccq/utils.go
bruce-riley: This code has been ripped out in favor of the recently added library function.

PR #3383, opened 2023-09-15
ethereum/relayer: fix storage constant comments

kcsongor:
The hash preimages for storage calculations were wrong for two storage slots. The new slots can be verified to be correct with e.g. the `chisel` command line tool from the foundry suite.


evan-gray (approved):

derpy-duck (approved):
Looks good to me

PR #3394, opened 2023-09-23
Node: Observations are incorrectly ignored when marked settled

bruce-riley:
A recent change to `handleObservation` caused it to ignore additional signatures if the transaction is marked settled. However, the observation gets marked settled after only 30 seconds, and it sometimes (like in the case of consistency level 201, safe blocks) observations can take more than 30 seconds to arrive.

This PR changes it to ignore additional signatures once the transaction is marked submitted, meaning we have submitted / published the VAA.

### DIFF
--- a/node/pkg/processor/observation.go
+++ b/node/pkg/processor/observation.go
@@ -83,8 +83,8 @@ func (p *Processor) handleObservation(ctx context.Context, obs *node_common.MsgW
 	m := obs.Msg
 	hash := hex.EncodeToString(m.Hash)
 	s := p.state.signatures[hash]
-	if s != nil && s.settled {
-		// already settled; ignoring additional signatures for it.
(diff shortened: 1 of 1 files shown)


bruce-riley:
Tested this PR by listening to mainnet gossip and mainnet eth, and doing a `postMessage` from etherscan. Before the fix, after seeing the message `observation considered settled`, the processor ignored all subsequent observations for this transaction. After the fix, it went to to generate `signed VAA with quorum` for the transaction.

evan-gray (approved):
Seems like https://github.com/wormhole-foundation/wormhole/commit/68f6cf8e6c4fd81d29960481b48f0e850b7d0f34 may have over-optimized 🙂 Should be safe to soften this restriction as the check didn't exist at all prior.

panoel (approved):

PR #3395, opened 2023-09-24
Deprecate truffle

derpy-duck:


ethereum/upgrade
```diff
@@ -62,8 +57,9 @@ if [ $ret = 0 ]; then
   exit
 fi
 
-deploy_output=$(npx truffle exec $SCRIPT --network "$truffle_network") || ( echo "✘ $chain $module: $deploy_output" && exit 1 )
-new_implementation=$(echo "$deploy_output" | grep "address:" | cut -d' ' -f3)
+forge script $SCRIPT --rpc-url $rpc_url --private-key $private_key --broadcast --via-ir || ( echo "✘ $chain $module" && exit 1 )
```
scnale, line 60: I don't think `--via-ir` should be set here. We should set the foundry profile appropriately and configure the build pipeline there.

scnale: Additionally, this makes it hard to use hardware wallets but I'm not sure there's an easy way around this other than editing the script or putting lots of conditionals :thinking:

derpy-duck: resolved the VIA IR issue!

derpy-duck: Yeah - you had mentioned this, but in any case this should be a strict improvement over the truffle scripts, right?

derpy-duck: by the way, is there a github issue mentioning the issue with the forge scripts + hardware wallets you're mentioning? (or, could you describe it again)

ethereum/package.json
```diff
@@ -5,45 +5,25 @@
...
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@certusone/wormhole-sdk": "^0.9.11",
     "@typechain/ethers-v5": "^10.2.0",
     "dotenv": "^10.0.0",
     "elliptic": "^6.5.2",
     "jsonfile": "^4.0.0",
-    "truffle-hdwallet-provider-klaytn": "^1.4.2",
+    "truffle-flattener": "^1.6.0",
```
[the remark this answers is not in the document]
derpy-duck: good catch, removed (and replaced with 'forge flatten') :)

ethereum/upgrade
scnale: Hm, I don't know if there is an issue for hardware wallet compatibility. I never opened one at the very least. The point here would be that this particular script uses `--private-key` option here without offering any way to use `--ledger` or `--trezor` for example. I think there are other key store mechanisms supported by foundry too but I personally only care about hardware wallets, at least for now.

devnet/eth-devnet.yaml
```diff
@@ -57,18 +54,12 @@ spec:
           command:
             - /bin/sh
             - -c
-            - "npm run migrate && npx truffle exec scripts/deploy_test_token.js && npx truffle exec scripts/register_all_chains.js && npm run deploy-relayers-evm1 && nc -lkn 2000"
+            - "CHAIN_ID=2 NETWORK=devnet ./forge-scripts/devnetInitialization.sh"
```
scnale, line 57: This `CHAIN_ID` would be the wormhole chain id, right? I got confused why it differs from the one above but that would make sense.

ethereum/Makefile
```diff
@@ -53,23 +52,15 @@ flattened/%.sol: contracts/%.sol node_modules
     # We remove the license (SPDX) lines and ABIEncoderV2 lines because they
     # break compilation in the flattened file if there are multiple conflicting
     # declarations.
-	npx truffle-flattener $< | grep -v SPDX | grep -v ABIEncoderV2 > $@
+	FOUNDRY_PROFILE=production forge flatten $< | grep -v SPDX | grep -v ABIEncoderV2 > $@
```
scnale, line 55: I wonder, do we actually need this? I think most block explorers support the standard json input which is the only input with reproducible builds.

ethereum/forge-scripts/DeployDummyContract.s.sol
```diff
@@ -0,0 +1,20 @@
...
+contract DeployDummyContract is Script {
+    function run(uint256 num) public {
+        vm.startBroadcast();
+        for(uint256 i=0; i<num; i++) {
+            deploy();
+        }
+        vm.stopBroadcast();
+    }
+    function deploy() internal {
+        Setup setup = new Setup();
```
scnale, line 18: This could be just a simple message call instead of a deploy.
In fact, if we're using this only in devnet, we should probably look at [cheatcodes](https://book.getfoundry.sh/cheatcodes/set-nonce) that let us set the current nonce instead of spamming transactions.

scnale:
>The Relayer contracts now use openzeppelin v4.3.1 instead of v4.8.1 - this is a downgrade to keep consistency with the Core, Token, and NFT bridge. This results in a bytecode change in the relayer contracts, modifying their address, hence the address change in devnet/CI

I think we should actually upgrade to the latest openzeppelin release instead. I'm not sure if any of the vulnerabilities found since that old version affects us but we could definitely benefit from improved docstrings and implementations at the very least.

scnale:
The same goes for the compiler version, I'd bump it to the latest one.

ethereum/upgrade
scnale: For a more concrete example you can take a look at how I use hardware wallets here with foundry [here](https://github.com/wormhole-foundation/example-circle-relayer#hardware-wallet-deployment). It's far from ideal because we would probably need to pipe the options through several scripts. I didn't check whether these could be passed as environment variables. :thinking:

SEJeff:
@scnale Re:
```
The same goes for the compiler version, I'd bump it to the latest one.
```
Solidity >= 0.8.20 include the new push0 opcode, which some Layer 2 chains [such as Arbitrum](https://docs.arbitrum.io/for-devs/concepts/differences-between-arbitrum-ethereum/solidity-support) do not support. If you build with that compiler and it uses that opcode, it will be impossible to deploy or run those contracts on those L2s.

scnale:
> Solidity >= 0.8.20 include the new push0 opcode, which some Layer 2 chains [such as Arbitrum](https://docs.arbitrum.io/for-devs/concepts/differences-between-arbitrum-ethereum/solidity-support) do not support. If you build with that compiler and it uses that opcode, it will be impossible to deploy or run those contracts on those L2s.

We're already handling this in the foundry config by setting the EVM version explicitly to `paris` (merge).

derpy-duck:
> > The Relayer contracts now use openzeppelin v4.3.1 instead of v4.8.1 - this is a downgrade to keep consistency with the Core, Token, and NFT bridge. This results in a bytecode change in the relayer contracts, modifying their address, hence the address change in devnet/CI
> 
> I think we should actually upgrade to the latest openzeppelin release instead. I'm not sure if any of the vulnerabilities found since that old version affects us but we could definitely benefit from improved docstrings and implementations at the very least.



> > The Relayer contracts now use openzeppelin v4.3.1 instead of v4.8.1 - this is a downgrade to keep consistency with the Core, Token, and NFT bridge. This results in a bytecode change in the relayer contracts, modifying their address, hence the address change in devnet/CI
> 
> I think we should actually upgrade to the latest openzeppelin release instead. I'm not sure if any of the vulnerabilities found since that old version affects us but we could definitely benefit from improved docstrings and implementations at the very least.

@scnale I think my main worry with that is it is a contract change to the Core, Token, and NFT bridge - which is maybe okay?

devnet/eth-devnet.yaml
derpy-duck: yup, wormhole chain id

ethereum/Makefile
derpy-duck: maybe not but I just wanted to emulate all existing functionality

derpy-duck: but yeah don't know if the flattened output is being used at all

ethereum/upgrade
derpy-duck: I see. Does the current upgrade script (using truffle) allow for hardware wallet compatibility? If not, then this PR doesn't add but doesn't remove that functionality either right

ethereum/forge-scripts/DeployDummyContract.s.sol
derpy-duck: I think the cheatcode solution is actually not possible - played around a little and am unable to get it to permanently modify the nonce value in the anvil devnet blockchain - I think the cheat-code is something that only works for a forge-test like environment (but not certain)

scnale:
What does change is the output of the compiler, but no changes to the contracts themselves should be required since there aren't any breaking changes within the 0.8 series as far as I recall. I'd say that we're more at risk of introducing bugs by downgrading than upgrading.

ethereum/forge-scripts/DeployDummyContract.s.sol
scnale: A shame :pensive:

ethereum/Makefile
scnale: I think I'll add a warning to use the verify plugin when possible instead. And to prefer the compiler standard input JSON in general.

evan-gray:
After #3877, everything in Tilt will have been ported to use Forge / Anvil instead of Truffle / Ganache.
- Relayer moved #3866
- Ganache -> Anvil #3869
- Forge deploy scripts #3862
- Tilt deployments with forge #3877

This leaves the only remaining ganache / truffle use as `make test-ganache` and `make test-upgrades`

evan-gray:
Closing, as this PR is quite stale - here is the status of the items in the description

✅ Use anvil instead of ganache

✅  ~~The Core, Token, and NFT bridge contracts now use openzeppelin v4.8.1 instead of v4.3.1~~ The relayer contracts were moved to `relayer/ethereum` so as not to have dependency conflicts.

✅ All deploys are done via forge scripts

✅ Bash scripts exist for devnet setup and relevant deploys and registrations.

❌ Truffle scripts / config / other remnants were not yet removed - The ethereum CI tests still use truffle and some various remnants remain.

✅ The simulate_upgrade bash script (works for Core, Token, NFT bridge) was modified to work with the new deployment process (though it still has a `truffle exec` for deployment instead of using the appropriate forge script)

✅ The upgrade bash script (works for Core, Token, NFT bridge) was modified to work with the new deployment process

✅ The verify bash script (checks that deployed bytecode matches local bytecode) (works for Core, Token, NFT bridge) was modified to work with the new deployment process

✅ Tests were modified to accommodate reaching anvil's `finalized` state

✅ build artifacts now live in build-forge/file.sol/file.json rather than build/contracts/file.json - some places were modified because of this

✅ Prune the package.json in the ethereum folder a little bit - including removing the @certusone/wormhole-sdk dependency

❌ All truffle tests were removed

❌ Removes 'generate-abi.sh' and 'generate-abi-celo.sh'

❌ Adds 61-ish forge tests for the Core Bridge, Token Bridge, NFT bridge that were originally written as truffle tests (and adds four TokenMigrator tests that were originally written as truffle tests) - These tests have not yet been adopted.

❌ There are a few tests [here](https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/test/upgrades/01_tokenbridge_feetoken_support.js) that weren't rewritten - input on if these need to be rewritten would help. - This has not been assessed

I have documented these in #4069

PR #3398, opened 2023-09-25
CI: run ethereum tests on ubuntu-latest

kcsongor:
ubuntu-20.04 ships with GLIBC-2.31, but the latest foundry nightly (which these CI actions use) link against 2.33+ (perhaps foundry should statically link against something like musl...?)

So instead we run these two tests on ubuntu-latest


evan-gray (approved):
🤞

derpy-duck (approved):

PR #3399, opened 2023-09-25
Client/js: adds key and rpc support for submit VAA on Sei mainnet.

scnale:
This adds the following for VAA submit on Sei:
- Gas estimation.
- Process/accept `rpc` option.
- Defines `SEI_KEY` as the environment variable for mainnet Sei keys.

### DIFF
--- a/clients/js/src/chains/sei/submit.ts
+++ b/clients/js/src/chains/sei/submit.ts
@@ -2,7 +2,7 @@ import { DirectSecp256k1HdWallet } from "@cosmjs/proto-signing";
 import { calculateFee } from "@cosmjs/stargate";
 import { MsgExecuteContractEncodeObject } from "@cosmjs/cosmwasm-stargate";
 import { toUtf8 } from "@cosmjs/encoding";
-import { MsgExecuteContract } from "cosmjs-types/cosmwasm/wasm/v1/tx"
+import { MsgExecuteContract } from "cosmjs-types/cosmwasm/wasm/v1/tx";
 import { getSigningCosmWasmClient } from "@sei-js/core";
 
 import { CONTRACTS } from "@certusone/wormhole-sdk/lib/esm/utils/consts";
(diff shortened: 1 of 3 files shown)


kev1n-peters (approved):

clients/js/src/consts/networks.ts
```diff
@@ -166,7 +166,7 @@ const MAINNET = {
   },
   sei: {
     rpc: undefined,
```
evan-gray, line 168: I think there's a public RPC now, yes?

clients/js/src/chains/sei/submit.ts
```diff
@@ -10,10 +13,13 @@ import { impossible, Payload } from "../../vaa";
 export const submit = async (
   payload: Payload,
   vaa: Buffer,
-  network: Network
+  network: Network,
+  rpc?: string
 ) => {
   const contracts = CONTRACTS[network].sei;
-  const { rpc, key } = NETWORKS[network].sei;
+  const networkInfo = NETWORKS[network].sei;
```
evan-gray, line 20: we should be able to keep the line that was here and avoid the extra parameter if we update the networks.ts to include the rpc.

clients/js/src/chains/sei/submit.ts
```diff
@@ -131,7 +137,26 @@ export const submit = async (
...
+  // For some reason, the simulation only provides the gas used but no events
+  // so we can't determine whether it worked unless we do away with cosmjs and query the node ourselves.
+  // See https://github.com/cosmos/cosmjs/issues/1148#issuecomment-1129259646
+  const gasUsed = await client.simulate(account.address, [executeContractMsg], undefined);
+  // It looks like the simulation is a bit lacking when it comes to estimating gas.
+  // See https://github.com/cosmos/cosmos-sdk/issues/4938
+  // That's why we multiply it by a factor of 1.3
+  const estimatedGas = Math.floor(gasUsed * 130 / 100);
+
+  const fee = calculateFee(estimatedGas, "0.1usei");
```
evan-gray, line 159: should the 0.1 usei be pulled from somewhere?

kcsongor (approved):

clients/js/src/chains/sei/submit.ts
kcsongor: agreed that using the public RPC would be nice. That said, there’s an argument to be made that all of these functions should propagate the RPC anyway to allow overriding for whatever reason

scnale: I think it's worth supporting switching out the RPC url though :thinking:

clients/js/src/consts/networks.ts
scnale: Yup, I took one from [here](https://docs.sei.io/develop/resources#public-rpcs) for `pacific-1`.

clients/js/src/chains/sei/submit.ts
scnale: Absolutely no idea. It just worked for me, but it could easily be something else for all I know :P

evan-gray: 🤷

evan-gray (changes requested):
Looks like `clients/js/src/chains/sei/submit.ts` doesn't pass the prettier check 😬

evan-gray (approved):

kev1n-peters (approved):

PR #3401, opened 2023-09-26
Node: Keygen changes

bruce-riley:
This PR restructures the code to generate, save and read guardian key files so that it can be shared with CCQ. Once this PR is merged, the code in the `ccq/integration` branch will be updated to use it, including changing the `ccq_server` to not use the guardian key label.

### DIFF
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/pflag"
 	"golang.org/x/crypto/sha3"
 
+	"github.com/certusone/wormhole/node/pkg/common"
 	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
 	publicrpcv1 "github.com/certusone/wormhole/node/pkg/proto/publicrpc/v1"
 	"github.com/wormhole-foundation/wormhole/sdk"
@@ -39,6 +40,7 @@ import (
 var (
 	clientSocketPath *string
 	shouldBackfill   *bool
+	unsafeDevnetMode *bool
 )
 
 func init() {
@@ -67,6 +69,10 @@ func init() {
 	PurgePythNetVaasCmd.Flags().AddFlagSet(pf)
 	SignExistingVaaCmd.Flags().AddFlagSet(pf)
 	SignExistingVaasFromCSVCmd.Flags().AddFlagSet(pf)
+
+	adminClientSignWormchainAddressFlags := pflag.NewFlagSet("adminClientSignWormchainAddressFlags", pflag.ContinueOnError)
+	unsafeDevnetMode = adminClientSignWormchainAddressFlags.Bool("unsafeDevMode", false, "Run in unsafe devnet mode")
+	AdminClientSignWormchainAddress.Flags().AddFlagSet(adminClientSignWormchainAddressFlags)
 
 	AdminCmd.AddCommand(AdminClientInjectGuardianSetUpdateCmd)
 	AdminCmd.AddCommand(AdminClientFindMissingMessagesCmd)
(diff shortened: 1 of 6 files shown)


SEJeff (dismissed):
I always wish there were more tests for refactors like this, but this seems like a very sensible approach. Nice job on the DRY.

node/cmd/guardiand/keygen.go
```diff
@@ -0,0 +1,49 @@
...
+
+var KeygenCmd = &cobra.Command{
+	Use:   "keygen [KEYFILE]",
+	Short: "Create guardian key at the specified path",
+	Run:   runKeygen,
+	Args:  cobra.ExactArgs(1),
+}
+
+func runKeygen(cmd *cobra.Command, args []string) {
+	common.LockMemory()
```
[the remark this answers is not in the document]
bruce-riley: This LockMemory call has been there since the beginning, I just moved the code to a new file. That being said, this command runs as a short lived, stand alone instance of guardiand, so I don't think this call will impact the long running guardian, or even badgerdb.

panoel (approved):

node/cmd/guardiand/keygen.go
```diff
@@ -0,0 +1,49 @@
...
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/spf13/cobra"
+)
+
+var keyDescription *string
+var blockType *string
+
+func init() {
+	keyDescription = KeygenCmd.Flags().String("desc", "", "Human-readable key description (optional)")
+	blockType = KeygenCmd.Flags().String("block-type", "", "block type of armored file (optional)")
```
evan-gray, line 19: Might have been more clear to give this a default of `common.GuardianKeyArmoredBlock` instead of lines 40-43

node/cmd/guardiand/adminclient.go
```diff
@@ -225,7 +226,7 @@ func runSignWormchainValidatorAddress(cmd *cobra.Command, args []string) error {
 	if !strings.HasPrefix(wormchainAddress, "wormhole") || strings.HasPrefix(wormchainAddress, "wormholeval") {
 		return fmt.Errorf("must provide a bech32 address that has 'wormhole' prefix")
 	}
-	gk, err := loadGuardianKey(guardianKeyPath)
+	gk, err := common.LoadGuardianKey(guardianKeyPath, *unsafeDevMode)
```
evan-gray, line 229: will `unsafeDevMode` be set before here? asking because I'm not sure if the node flags apply here as well 🤔

evan-gray (approved):
Couple questions / nits. Overall looks good though.

evan-gray (dismissed):
Couple questions / nits. Overall looks good though.

node/cmd/guardiand/adminclient.go
bruce-riley: Turns out it is not, and the existing code did not handle devnet. I fixed it.

node/cmd/guardiand/adminclient.go
```diff
@@ -68,6 +70,10 @@ func init() {
 	SignExistingVaaCmd.Flags().AddFlagSet(pf)
 	SignExistingVaasFromCSVCmd.Flags().AddFlagSet(pf)
 
+	adminClientSignWormchainAddressFlags := pflag.NewFlagSet("adminClientSignWormchainAddressFlags", pflag.ContinueOnError)
+	unsafeDevnetMode = adminClientSignWormchainAddressFlags.Bool("unsafedevnet", false, "Running in unsafe devnet mode")
```
evan-gray, line 74: why did this poor flag lose it's camel case? should be `unsafeDevMode` to match node?

panoel (dismissed):

node/cmd/guardiand/adminclient.go
evan-gray: also nit, but maybe the comment should be "Run in unsafe devnet mode" since it will be the helptext

evan-gray (approved):
Approving with one tiny nit comment left

node/cmd/guardiand/adminclient.go
```diff
@@ -85,6 +91,7 @@ func init() {
 	AdminCmd.AddCommand(SignExistingVaaCmd)
 	AdminCmd.AddCommand(SignExistingVaasFromCSVCmd)
 	AdminCmd.AddCommand(Keccak256Hash)
+	AdminCmd.AddCommand(AdminClientSignWormchainAddress)
```
evan-gray, line 94: just noticed, this is added twice

evan-gray (approved):

panoel (approved):

PR #3421, opened 2023-10-04
CCQ: Query SDK

bruce-riley:


sdk/js-query/src/query/ethCall.test.ts
```diff
@@ -0,0 +1,287 @@
...
+describe("eth call", () => {
+  test("serialize request", () => {
+    const toAddress = "0x0d500b1d8e8ef31e21c99d1db9a6444d3adf1270";
+    const nameCallData = createTestEthCallData(toAddress, "name", "string");
+    const totalSupplyCallData = createTestEthCallData(
+      toAddress,
+      "totalSupply",
+      "uint256"
+    );
+    const ethCall = new EthCallQueryRequest("0x28d9630", [
```
barnjamin, line 75: this is the method selector?

sdk/js-query/src/query/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./request";
+export * from "./response";
+export * from "./utils";
+export * from "./ethCall";
+export * from "./consts";
```
panoel, line 5: Is there any reason someone would want to use BinaryWriter outside of this directory?

sdk/js-query/src/query/ethCall.test.ts
barnjamin: jk this looks like a block num?

sdk/js-query/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "declaration": true,
+    "outDir": "./lib/esm",
+    "strict": true,
+    "esModuleInterop": true,
+    "downlevelIteration": true,
+    "allowJs": true,
+    "lib": ["dom", "es5", "scripthost", "es2020.bigint"]
+  },
+  "include": ["src"],
+  "exclude": ["node_modules", "**/*.test.*", "**/__tests__/*"]
```
panoel, line 15: Question: Do we normally exclude "lib/*"?  I'm fuzzy on tsconfig stuff.

sdk/js-query/src/query/BinaryWriter.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // Ensure the buffer has the capacity to write `size` bytes, otherwise allocate more memory
+  _ensure(size: number) {
+    const remaining = this._buffer.length - this._offset;
+    if (remaining < size) {
+      const oldBuffer = this._buffer;
+      const newSize = this._buffer.length * 2 + size;
+      this._buffer = Buffer.alloc(newSize);
+      oldBuffer.copy(this._buffer);
```
barnjamin, line 19: i think there is a `resize` but this seems fine

sdk/js-query/src/query/response.ts
```diff
@@ -0,0 +1,36 @@
+import { ChainQueryType } from "./request";
+
+// TODO: implement query response parsing
+
+export class QueryResponse {
```
panoel, line 5: This class doesn't do anything.  Do we want to remove it from the index.ts file until it does?

sdk/js-query/src/query/ethCall.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(public blockId: string, public callData: EthCallData[]) {}
+
+  type(): ChainQueryType {
+    return ChainQueryType.EthCall;
+  }
+
+  serialize(): Uint8Array {
+    const writer = new BinaryWriter()
+      .writeUint32(this.blockId.length)
+      .writeUint8Array(Buffer.from(this.blockId))
```
barnjamin, line 21: should this be a `Buffer.from(..., 'hex')`?

sdk/js-query/src/query/request.ts
```diff
@@ -0,0 +1,70 @@
+import { BinaryWriter } from "./BinaryWriter";
+import { Network } from "./consts";
+import { utils } from "web3";
```
barnjamin, line 3: I think we have web3 in the devDependencies, if we use it here we should promote it

sdk/js-query/src/query/BinaryWriter.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  writeUint8(value: number) {
+    this._ensure(1);
+    this._buffer.writeUint8(value, this._offset);
```
barnjamin, line 25: Node docs for Buffer have a scary-ish:
> Behavior is undefined when value is anything other than an unsigned 8-bit integer.

Maybe a check that the value is between 0 and 2**8-1 ?

sdk/js-query/src/query/ethCall.test.ts
```diff
@@ -0,0 +1,287 @@
...
+jest.setTimeout(60000);
+
+const CI = false;
+const ENV = "DEVNET";
+const ETH_NODE_URL = CI ? "ws://eth-devnet:8545" : "ws://localhost:8545";
+
+const CCQ_SERVER_URL = "http://localhost:6069/v1";
+const QUERY_URL = CCQ_SERVER_URL + "/query";
+const HEALTH_URL = CCQ_SERVER_URL + "/health";
+const PRIVATE_KEY =
```
panoel, line 30: Is this anything that we need to *really* keep private?

sdk/js-query/src/query/ethCall.test.ts
```diff
@@ -0,0 +1,287 @@
...
+        bytes: Buffer.from(serialized).toString("hex"),
+      },
+      { headers: { "X-API-Key": "my_secret_key" } }
+    );
+    expect(response.status).toBe(200);
+    const queryResponse = QueryResponse.fromBytes(
+      Buffer.from(response.data.bytes, "hex")
+    );
+    // TODO: verify signatures
+    // TOOD: verify query response
```
panoel, line 120: Are these worth doing?  Or wait until Response is fully implemented?

panoel (approved):

sdk/js-query/src/query/ethCall.test.ts
kev1n-peters: I think making the parameter more explicit like: 
```
// Can be a block number or a block hash
export type BlockTag = number | HexString;
```
would be better

sdk/js-query/src/query/index.ts
kev1n-peters: I don't think so, I think we should hide the implementation details

sdk/js-query/src/query/BinaryWriter.ts
kev1n-peters: Good idea

sdk/js-query/src/query/ethCall.test.ts
kev1n-peters: Nope, it's the devnet guardian secret

sdk/js-query/src/query/ethCall.test.ts
kev1n-peters: Nope, it's the devnet guardian secret

sdk/js-query/tsconfig.json
kev1n-peters: We're only going to include source files in the `src` directory here so we don't need to exclude `lib`

sdk/js-query/src/query/request.ts
kev1n-peters: Good catch

sdk/js-query/src/query/ethCall.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  type(): ChainQueryType {
+    return ChainQueryType.EthCall;
+  }
+
+  serialize(): Uint8Array {
+    const writer = new BinaryWriter()
+      .writeUint32(this.blockTag.length)
+      .writeUint8Array(Buffer.from(this.blockTag))
```
barnjamin, line 37: This wont error but is this def what we want? vs creating a buffer from the hex value
```
> const x = "0xdeadbeef"
undefined
> Buffer.from(x)
<Buffer 30 78 64 65 61 64 62 65 65 66>
> const y = "deadbeef"
undefined
> Buffer.from(y, 'hex')
<Buffer de ad be ef>
```

kev1n-peters: Yeah, this is because the block tag can be a string like "latest" for example in the future

sdk/js-query/src/query/ethCall.test.ts
kev1n-peters: Not worth doing at the moment

sdk/js-query/src/query/response.ts
kev1n-peters: Removed

panoel (approved):

kev1n-peters (approved):

barnjamin (approved):

evan-gray (approved):
I think `types` might be missing from the `package.json` and we should add a note to the `sdk` folder readme (because it has a list of it's contents and now this will be missing) but feel free to make an issue from this.

PR #3422, opened 2023-10-04
CCQ: Query Server

bruce-riley:
This PR adds the CCQ (Cross Chain Query) REST server. This PR is dependent on PR #3423.

For background on this feature see the whitepaper (PR #3418).


node/cmd/ccq/query_server.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if *permFile == "" {
+		logger.Fatal("Please specify --permFile")
+	}
+	if *ethRPC == "" {
+		logger.Fatal("Please specify --ethRPC")
+	}
+	if *ethContract == "" {
+		logger.Fatal("Please specify --ethContract")
+	}
```
SEJeff, line 119: Does it make sense to have a check along the lines of:

```go
if *envStr != "dev" && *p2pNetworkID == "/wormhole/dev" {
	logger.Fatal("Please specify either --env=dev or a different --p2pNetworkID")
}
```

At the very minimum there is likely never a case where we want `envStr="prod"` and `p2pNetworkID="/wormhole/dev"` I'd assume. This is a footgun for operators and should be more forgiving. Perhaps we want to hardcode the options for p2pNetworkID?

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	gs, err := caller.GetGuardianSet(&ethBind.CallOpts{Context: ctx}, currentIndex)
+	if err != nil {
+		return nil, fmt.Errorf("error requesting current guardian set value: %w", err)
+	}
+	return &common.GuardianSet{
+		Keys:  gs.Keys,
+		Index: currentIndex,
+	}, nil
+}
```
SEJeff, line 53: I could see this making a ton of sense to put in the go sdk. What do you think?

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+					return nil, fmt.Errorf(`invalid eth call "%s" for API key "%s" in permissions file "%s"`, ac.EthCall.Call, apiKey, fileName)
+				}
+				if len(call) != 4 {
+					return nil, fmt.Errorf(`eth call "%s" for API key "%s" in permissions file "%s" has an invalid length, must be four bytes`, ac.EthCall.Call, apiKey, fileName)
+				}
+
+				// The permission key is the chain, contract address and call formatted as a colon separated string.
+				callKey = fmt.Sprintf("ethCall:%d:%s:%s", ac.EthCall.Chain, contractAddress, hex.EncodeToString(call))
+			} else {
+				return nil, fmt.Errorf(`unsupported call type for API key "%s" in permissions file "%s"`, apiKey, fileName)
```
SEJeff, line 135: If you flipped the conditional at the top of this to:
```go
if ac.EthCall == nil {
    return nil, fmt.Errorf(`unsupported call type for API key "%s" in permissions file "%s"`, apiKey, fileName)
}
# no else block since there is a return in the primary path
```

You could left indent the non-error condition logic one and make this code easier to read.

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+	userName      string
+	apiKey        string
+	allowUnsigned bool
+	allowedCalls  allowedCallsForUser // Key is something like "ethCall:2:000000000000000000000000b4fbf271143f4fbf7b91a5ded31805e42b2208d6:06fdde03"
+}
+
+type allowedCallsForUser map[string]struct{}
+
+// parseConfig parses the permissions config file into a map keyed by API key.
+func parseConfig(fileName string) (Permissions, error) {
```
SEJeff, line 88: I realize it is a touch more work on the caller, but if you would be so kind as to make this instead `parseConfig(reader io.Reader) (Permissions, error) {`, it would be [much easier to unit test](https://endler.dev/2018/go-io-testing/).

If you want help writing some of these tests I'd be willing to help you do it.

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return ret, nil
+}
+
+// validateRequest verifies that this API key is allowed to do all of the calls in this request.
+func validateRequest(logger *zap.Logger, env common.Environment, perms Permissions, signerKey *ecdsa.PrivateKey, apiKey string, qr *gossipv1.SignedQueryRequest) error {
+	apiKey = strings.ToLower(apiKey)
+	permsForUser, exists := perms[strings.ToLower(apiKey)]
+	if !exists {
+		logger.Debug("invalid api key", zap.String("apiKey", apiKey))
```
SEJeff, line 163: Logging the api key, even in debug mode only, is a way to potentially leak secrets and is quite cringey. Perhaps you could just sha256sum the api key or something? That way someone with access to the perms file could still figure out which api key it is without accidentally leaking credentials when the logs are shared.

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("invalid api key")
+	}
+
+	// TODO: Should we verify the signatures?
+
+	if len(qr.Signature) == 0 {
+		if !permsForUser.allowUnsigned || signerKey == nil {
+			logger.Debug("request not signed and unsigned requests not supported for apiKey",
+				zap.String("apiKey", apiKey),
```
SEJeff, line 172: Same as before. Do we really want to log the raw apiKey? it is possible and even likely these logs are shared in slack or elsewhere with parties who might not have direct access to the permissions file and shouldn't have access to ccq.

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+	for _, pcq := range queryRequest.PerChainQueries {
+		switch q := pcq.Query.(type) {
+		case *query.EthCallQueryRequest:
+			for _, callData := range q.CallData {
+				contractAddress, err := vaa.BytesToAddress(callData.To)
+				if err != nil {
+					logger.Debug("failed to parse contract address", zap.String("apiKey", apiKey), zap.String("contract", hex.EncodeToString(callData.To)), zap.Error(err))
+					return fmt.Errorf("failed to parse contract address: %w", err)
+				}
+				if len(callData.Data) < 4 {
```
SEJeff, line 212: consider using a constant such as this for the few places where you validate this:

```go
const ETH_SELECTOR_LENGTH = 4
```

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+					logger.Debug("eth call data must be at least four bytes", zap.String("apiKey", apiKey), zap.String("data", hex.EncodeToString(callData.Data)))
+					return fmt.Errorf("eth call data must be at least four bytes")
+				}
+				call := hex.EncodeToString(callData.Data)
+				callKey := fmt.Sprintf("ethCall:%d:%s:%s", int(pcq.ChainId), contractAddress, call)
```
SEJeff, line 217: Why cast the uint16 ChainId to int? I don't think that is strictly necessary? The `%d` format string converts a uint16 the same as an int (aka int64).

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+				if _, exists := permsForUser.allowedCalls[callKey]; !exists {
+					logger.Debug("requested call not authorized", zap.String("apiKey", apiKey), zap.String("callKey", callKey))
+					return fmt.Errorf(`call "%s" not authorized`, callKey)
+				}
+			}
+		default:
+			logger.Debug("unsupported query type", zap.String("apiKey", apiKey), zap.Any("type", pcq.Query))
```
SEJeff, line 224: I'm not sure that you need to have any logs for any of these errors if you're returning the errors. Just log in the caller.

That's an easy way to not print the apiKey secret over and over to the logs, even though it requires debug mode.

SEJeff (commented):
A lot of nits, but mainly I'm worried about productionizing the webserver and the p2p code that is public facing.

More unit tests for those components would be massively appreciated.

node/cmd/ccq/utils.go
```diff
@@ -0,0 +1,231 @@
...
+				call, err := hex.DecodeString(strings.TrimPrefix(ac.EthCall.Call, "0x"))
+				if err != nil {
+					return nil, fmt.Errorf(`invalid eth call "%s" for API key "%s" in permissions file "%s"`, ac.EthCall.Call, apiKey, fileName)
+				}
+				if len(call) != 4 {
+					return nil, fmt.Errorf(`eth call "%s" for API key "%s" in permissions file "%s" has an invalid length, must be four bytes`, ac.EthCall.Call, apiKey, fileName)
+				}
+
+				// The permission key is the chain, contract address and call formatted as a colon separated string.
+				callKey = fmt.Sprintf("ethCall:%d:%s:%s", ac.EthCall.Chain, contractAddress, hex.EncodeToString(call))
```
evan-gray, line 133: should only be `[0:8]` of the call

evan-gray (changes requested):

node/cmd/ccq/utils.go
bruce-riley: Fixed.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+		return
+	}
+
+	signedQueryRequest := &gossipv1.SignedQueryRequest{
+		QueryRequest: queryRequestBytes,
+		Signature:    signature,
+	}
+
+	if err := validateRequest(s.logger, s.env, s.permissions, s.signerKey, apiKey[0], signedQueryRequest); err != nil {
+		s.logger.Debug("invalid request", zap.String("api_key", apiKey[0]), zap.Error(err))
```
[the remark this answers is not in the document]
bruce-riley: Changed it to log the user name rather than the api key. Removed the above debug since every error case logs a debug in `validateRequest`.

node/cmd/ccq/utils.go
bruce-riley: Changed it to log user name rather than api key. Also these are only debugs, which will not routinely be enabled.

node/cmd/ccq/p2p.go
```diff
@@ -0,0 +1,255 @@
...
+	go func() {
+		// Maps the request signature to a map of response digests which maps to a list of guardian signatures.
+		// A request could have responses with different digests, because the guardians could have
+		// different results returned for the query in the event of a rollback.
+		responses := make(map[string]map[ethCommon.Hash][]GuardianSignature)
+		for {
+			envelope, err := sub.Next(ctx)
+			if err != nil {
+				logger.Error("Failed to read next pubsub message", zap.Error(err))
+				break
```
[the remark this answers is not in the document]
bruce-riley: Changed it to a `logger.Fatal` so the process will be restarted.

node/cmd/ccq/p2p.go
```diff
@@ -0,0 +1,255 @@
...
+				bootstrappers = append(bootstrappers, *pi)
+			}
+			idht, err := dht.New(ctx, h, dht.Mode(dht.ModeServer),
+				// This intentionally makes us incompatible with the global IPFS DHT
+				dht.ProtocolPrefix(protocol.ID("/"+networkID)),
+				dht.BootstrapPeers(bootstrappers...),
+			)
+			return idht, err
+		}),
+	)
```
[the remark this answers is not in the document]
bruce-riley: Created issue #3433 to address adding prometheus metrics for CCQ, both in the `ccq_server` and the guardian. That will be addressed outside this PR.

node/cmd/ccq/utils.go
bruce-riley: I prefer these more detailed messages. Also, since they are debugs, they will rarely be logged.

node/cmd/ccq/query_server.go
bruce-riley: Added a check that the network may not be dev unless the env is also dev. Not putting the check the other way because I'm not sure that wouldn't happen if we are running some sort of ad hoc tool.

node/cmd/ccq/utils.go
bruce-riley: It was simple to create two separate methods `parseConfig(buf []byte)` and `parseConfigFile(fileName string)` where the second just calls the first. That made it simple to add parsing unit tests, which I did.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+	}
+
+	s.pendingResponses.Remove(pendingResponse)
+}
+
+func (s *httpServer) handleHealth(w http.ResponseWriter, r *http.Request) {
+	s.logger.Debug("health check")
```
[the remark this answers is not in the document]
bruce-riley: Yes, it's just a debug that won't normally be enabled.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	r := mux.NewRouter()
+	r.HandleFunc("/v1/query", s.handleQuery).Methods("PUT", "OPTIONS")
+	r.HandleFunc("/v1/health", s.handleHealth).Methods("GET")
+	return &http.Server{
+		Addr:              addr,
+		Handler:           r,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+}
```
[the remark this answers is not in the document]
bruce-riley: Created issue #3433 to address adding prometheus metrics as a separate PR.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	pendingResponse := NewPendingResponse(signedQueryRequest)
+	added := s.pendingResponses.Add(pendingResponse)
+	if !added {
+		http.Error(w, "Duplicate request", http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
kev1n-peters: Maybe we could return `http.StatusBadRequest` instead

node/cmd/ccq/p2p.go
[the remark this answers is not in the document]
bruce-riley: This code was replaced with a single call to the new `p2p.NewHost` function.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+			SignedQueryRequest: signedQueryRequest,
+		},
+	}
+
+	b, err := proto.Marshal(&m)
+	if err != nil {
+		s.logger.Debug("failed to marshal gossip message", zap.Error(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
kev1n-peters: I think that we should definitely log as `Error`, but maybe we can keep `http.StatusInternalServerError` since `validateRequest` will return `false` if we can't `Unmarshal` the `signedQueryRequest` - we should be able to `Marshal` `m` at this point (I think failure here would be a very rare occurrance)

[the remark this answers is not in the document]
SEJeff: Oh totally agree here @kev1n-peters.

node/cmd/ccq/http.go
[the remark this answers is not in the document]
SEJeff: Would you be ok with opening one to add http request logging as well?

node/cmd/ccq/query_server.go
SEJeff: Perfect sir, thanks!

node/cmd/ccq/utils.go
SEJeff: Legend. Thanks so much.

node/cmd/ccq/utils.go
SEJeff: That works. I almost wish we had a trace log level that was more just for devs for stuff like this.

node/cmd/ccq/http.go
```diff
@@ -0,0 +1,175 @@
...
+	s := &httpServer{
+		topic:            t,
+		permissions:      permissions,
+		signerKey:        signerKey,
+		pendingResponses: p,
+		logger:           logger,
+		env:              env,
+	}
+	r := mux.NewRouter()
+	r.HandleFunc("/v1/query", s.handleQuery).Methods("PUT", "OPTIONS")
```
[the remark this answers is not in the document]
SEJeff: I added this in 45035e5ad248d494c2186bb6e2fa1ae990b5ab6a

node/cmd/ccq/http.go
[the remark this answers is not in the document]
SEJeff: I'll just file one for you.

node/cmd/ccq/pending_request.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return nil
+}
+
+func (p *PendingResponses) Remove(r *PendingResponse) {
+	signature := hex.EncodeToString(r.req.Signature)
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	delete(p.pendingResponses, signature)
+}
```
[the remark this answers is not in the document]
bruce-riley: @SEJeff added post support.

SEJeff (approved):

kev1n-peters (approved):

SEJeff (approved):

kev1n-peters (approved):

evan-gray (approved):

PR #3429, opened 2023-10-09
Deploy scroll to testnet

bruce-riley:


panoel (dismissed):
I only looked at the node/* files

evan-gray (commented):
If Scroll supports `finalized` and `safe` can we add it to the list in `getFinality`?

evan-gray (approved):

panoel (approved):

kcsongor (approved):

a5-pickle (approved):
rust/core looks good

kev1n-peters (approved):

devnet/node.yaml
```diff
@@ -103,6 +103,8 @@ spec:
             - ws://eth-devnet:8545
             - --baseRPC
             - ws://eth-devnet:8545
+            - --scrollRPC
+            - ws://eth-devnet:8545
```
SEJeff, line 107: Won't this fail the guardian start due to [this code](https://github.com/wormhole-foundation/wormhole/pull/3429/files#diff-7f43e1a5bbb6083b5e470710e19e804d778f4a411b23fdbef1b2955f56b67bbdR637-R639) added in this PR? Seems like a contract needs to be specified and I'm clearly missing some magic here.

SEJeff (approved):

devnet/node.yaml
bruce-riley: No, this is not a problem. In tilt (unsafeDevMode) the contract address for all the EVM chains are set to the one deployed locally. See line node.go, line 498.

PR #3431, opened 2023-10-10
Node: QUIC Cut Over

bruce-riley:


node/pkg/p2p/p2p.go
```diff
@@ -559,7 +565,7 @@ func Run(
 				continue
 			}
 
-			logger.Debug("received message",
+			logger.Info("received message",
```
SEJeff, line 568: Did you purposefully change this log level? I'm assuming yes, but it is totally unrelated to this PR.

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,109 @@
...
+// we need to start up with the existing bootstrap string and cut over to the new quic-v1 at the specified time.
+func checkForCutOver(logger *zap.Logger, bootstrapPeers string, ccqBootstrapPeers string, components *Components) (newBootstrapPeers string, newCcqBootstrapPeers string, err error) {
+	newBootstrapPeers, newCcqBootstrapPeers, delay, err := checkForCutOverImpl(logger, bootstrapPeers, ccqBootstrapPeers, components, cutOverTimeStr, time.Now())
+
+	if delay != time.Duration(0) {
+		// Wait for the cut over time and then panic so we restart with the new quic-v1. TODO: Can we just restart p2p??
+		go func() {
+			time.Sleep(delay)
+			logger.Info("time to cut over to new quic-v1", zap.String("cutOverTime", cutOverTimeStr), zap.String("component", "p2pco"))
+			panic("p2pco: time to cut over to new quic-v1")
```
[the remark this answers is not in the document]
evan-gray: My concerns with executing this via governance are

1. Not every peer is guaranteed to see the signed VAA
2. Anyone offline would not know it happened without storing the governance VAA somewhere
3. Where would you store the governance VAA where we could guarantee access to all clients?
4. That storage location would need to be accessible to all clients until all clients have been upgraded to a release dropping the old QUIC support, otherwise clients could suffer a reversion.

I don't feel the added complexity of governance for a simple (and required) protocol switch adds more to the execution and guarantees than it subtracts. This is fundamentally different from a blockchain where guaranteed data synchronization is a part of the protocol (whereas this is not part of the guardian network)

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,109 @@
...
+	newCcqBootstrapPeers = ccqBootstrapPeers
+
+	if cutOverTimeStr == "" {
+		return
+	}
+
+	if !strings.Contains(bootstrapPeers, "quic") {
+		logger.Error(`bootstrap peers string has an unexpected value, doesn't contain "quic"`, zap.String("bootstrapPeers", bootstrapPeers), zap.String("component", "p2pco"))
+		err = fmt.Errorf("unexpected format of bootstrap peers")
+		return
```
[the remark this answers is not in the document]
bruce-riley: What you are suggesting does not work, instead it has to be this:
`return "", "", 0, fmt.Errorf("unexpected format of bootstrap peers")`
In which case, I prefer it the way it is.

node/pkg/p2p/p2p.go
bruce-riley: Nope. Undid it.

node/pkg/p2p/cutover_test.go
```diff
@@ -0,0 +1,259 @@
...
+// The time parsing function is very picky. This test allows an easy way to test out a new time string.
+func TestParseTime(t *testing.T) {
+	ts, err := time.Parse(cutOverFmtStr, "2023-10-06T18:12:05-0000")
+	require.NoError(t, err)
+	assert.Equal(t, 2023, ts.Year())
+	assert.Equal(t, time.Month(10), ts.Month())
+	assert.Equal(t, 6, ts.Day())
+	assert.Equal(t, 18, ts.Hour())
+	assert.Equal(t, 12, ts.Minute())
+	assert.Equal(t, 05, ts.Second())
```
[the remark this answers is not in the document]
bruce-riley: This test is no longer necessary since we already have `TestVerifyCutOverTime`. I removed this one.

node/pkg/p2p/cutover_test.go
```diff
@@ -0,0 +1,259 @@
...
+		reflect.DeepEqual(lhs.ProtectedHostByGuardianKey, rhs.ProtectedHostByGuardianKey) &&
+		lhs.SignedHeartbeatLogLevel == rhs.SignedHeartbeatLogLevel
+}
+
+func TestCutOverDisabled(t *testing.T) {
+	logger := zap.NewNop()
+
+	cutOverTimeStr := ""
+	now, err := time.Parse(cutOverFmtStr, "2023-10-06T18:19:00-0000")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
bruce-riley: No, this is ensuring that the parameter it is set to is valid, which is why I have the require. I prefer to leave it.

node/pkg/p2p/cutover.go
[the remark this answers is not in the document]
SEJeff: @evan-gray works for me! It was more of a thought that I wondered if we couldn't do it for things in the future as well that are more or less "hard forks".

This is a hard change that is "simple", but there will be more in the future. Especially as we start to tease apart the guardian into clean components. Just trying to see where we can reuse stuff.

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,109 @@
...
+			time.Sleep(delay)
+			logger.Info("time to cut over to new quic-v1", zap.String("cutOverTime", cutOverTimeStr), zap.String("component", "p2pco"))
+			panic("p2pco: time to cut over to new quic-v1")
+		}()
+	}
+
+	return newBootstrapPeers, newCcqBootstrapPeers, err
+}
+
+func checkForCutOverImpl(logger *zap.Logger, bootstrapPeers string, ccqBootstrapPeers string, components *Components, cutOverTimeStr string, now time.Time) (newBootstrapPeers string, newCcqBootstrapPeers string, delay time.Duration, err error) {
```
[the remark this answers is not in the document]
bruce-riley: I renamed the local variable.

As far as having the `cutOverTimeStr` be a time.Time rather than a string, go does not allow const structs, so this seemed like the most straight forward way to do this. It allows someone to easily update the config time, run the tests to verify they have the format right, and commit. I would like to leave it as a string.

SEJeff (approved):

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if ccqBootstrapPeers != "" && !strings.Contains(ccqBootstrapPeers, "quic") {
+		logger.Error(`ccq bootstrap peers string has an unexpected value, doesn't contain "quic"`, zap.String("ccqBootstrapPeers", ccqBootstrapPeers), zap.String("component", "p2pco"))
+		err = fmt.Errorf("unexpected format of ccq bootstrap peers")
+		return
+	}
+
+	if strings.Contains(bootstrapPeers, "quic-v1") {
+		if ccqBootstrapPeers != "" && !strings.Contains(ccqBootstrapPeers, "quic-v1") {
```
panoel, line 54: Is it okay for the ccq bootstrap peers to be v1 but the bootstrap peers are not?

panoel (dismissed):

node/pkg/p2p/cutover.go
bruce-riley: I beefed up the validation checks.

SEJeff (dismissed):

panoel (approved):

SEJeff (approved):

node/pkg/p2p/p2p.go
```diff
@@ -268,6 +268,12 @@ func Run(
 
 		logger := supervisor.Logger(ctx)
 
+		var coerr error
+		bootstrapPeers, ccqBootstrapPeers, coerr = checkForCutOver(logger, bootstrapPeers, ccqBootstrapPeers, components)
```
evan-gray, line 272: might be worth noting here that this side-effects `components.ListeningAddressesPatterns`. in that way, it does a bit more than "check"

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,138 @@
...
+const cutOverFmtStr = "2006-01-02T15:04:05-0700"
+
+// checkForCutOver checks to see if / when we need to cut over to the new quic-v1 based on a preset time. If cutOverTimeStr is set, then this feature is enabled.
+// If the current time is after the configured time, we should start up using the new quic-v1 bootstrap peers immediately. If the configured time is in the future,
+// we need to start up with the existing bootstrap string and cut over to the new quic-v1 at the specified time.
+func checkForCutOver(logger *zap.Logger, bootstrapPeers string, ccqBootstrapPeers string, components *Components) (newBootstrapPeers string, newCcqBootstrapPeers string, err error) {
+	newBootstrapPeers, newCcqBootstrapPeers, delay, err := checkForCutOverImpl(logger, bootstrapPeers, ccqBootstrapPeers, components, cutOverTimeStr, time.Now())
+
+	if delay != time.Duration(0) {
+		// Wait for the cut over time and then panic so we restart with the new quic-v1. TODO: Can we just restart p2p??
```
evan-gray, line 22: Judging by your PR comment, sounds like this TODO has been resolved?

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// The format of this time is very picky. Please use the exact format specified by cutOverFmtStr!
+const cutOverTimeStr = "2024-12-31T23:59:59-0000"
```
evan-gray, line 12: can we create separate cutover strings per environment? that way we can set testnet to something in the near future and leave mainnet `""` until the testnet cutover completes successfully. this will be necessary to support testnet and mainnet guardians and spies with the same codebase concurrently.

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,138 @@
...
+			)
+			err = fmt.Errorf("quic version mismatch between bootstrap peers and ccq bootstrap peers")
+			return
+		}
+	}
+
+	if bootstrapIsV1 {
+		for _, la := range components.ListeningAddressesPatterns {
+			if strings.Contains(la, "quic") && !strings.Contains(la, "quic-v1") {
+				err = fmt.Errorf("bootstrapPeers has been updated to quic-v1, but components.ListeningAddressesPatterns has not: %s", la)
```
evan-gray, line 72: just want to confirm my understanding of this condition, `ListeningAddressesPatterns` has not yet been changed before this point, so this would trip if a guardian / spy attempts to pass in a v1 url right now? is the thought that `ListeningAddressesPatterns` would be changed in the future (in which case this would need to be reversed?) or is `ListeningAddressesPatterns` supposed to be conditionally set before this and I'm missing something

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,138 @@
...
+	cutOverTime, err := time.Parse(cutOverFmtStr, coTimeStr)
+	if err != nil {
+		logger.Error("failed to parse p2p cut over time", zap.String("cutOverTimeStr", coTimeStr), zap.Error(err), zap.String("component", "p2pco"))
+		err = fmt.Errorf("failed to parse cut over time: %w", err)
+		return
+	}
+
+	if cutOverTime.Before(now) {
+		// We should already be using the new quic-v1.
+		newBootstrapPeers = strings.ReplaceAll(bootstrapPeers, "quic", "quic-v1")
```
evan-gray, line 90: should we add `/`s to all these? has a bit of a smell without them. what if someone's bootstrap domain has `quic` in it?

node/pkg/p2p/cutover.go
evan-gray: this might work in the tests because you provide different components, but it looks like the `p2p.go` code always uses hardcoded `/quic` ones. perhaps it would be easier to deal with the default string with a ternary there and not have the awkward side effect here.

evan-gray (changes requested):

node/pkg/p2p/p2p.go
```diff
@@ -268,6 +268,12 @@ func Run(
 
 		logger := supervisor.Logger(ctx)
 
+		var coerr error
```
evan-gray, line 271: upon further review, I'm fairly confident that the bootstrap peer replacement code will not adequately catch all p2p use cases. perhaps it should be moved to `NewHost` instead. as written, this cutover check will not be performed for `cmd/ccq` or `hack/query/test` (which would suddenly break CI after the cutover) as they do not use `p2p.Run` and therefore would not hit this code path.

evan-gray (changes requested):

node/pkg/p2p/cutover.go
evan-gray: i.e. bootstrap peers look like 

```
/dns4/wormhole-testnet-v2-bootstrap.certus.one/udp/8999/quic/p2p/12D3KooWAkB9ynDur1Jtoa97LBUp8RXdhzS5uHgAfdTquJbrbN7i,/dns4/t-guardian-01.nodes.stable.io/udp/8999/quic/p2p/12D3KooWCW3LGUtkCVkHZmVSZHzL3C4WRKWfqAiJPz1NR7dT9Bxh,/dns4/t-guardian-02.nodes.stable.io/udp/8999/quic/p2p/12D3KooWJXA6goBCiWM8ucjzc4jVUBSqL9Rri6UpjHbkMPErz5zK
```

so we wouldn't want to replace `/dns4/testnet-quic.certus.one/...` with `/dns4/testnet-quic-v1.certus.one/...`.

including the separators should be safer - `"/quic/","/quic-v1/"`

node/pkg/p2p/p2p.go
bruce-riley: @evan-gray and I reworked the design to address these issues.

node/pkg/p2p/cutover.go
bruce-riley: Reworked the design.

node/pkg/p2p/p2p.go
```diff
@@ -352,7 +357,7 @@ func Run(
 		if ccqEnabled {
 			ccqErrC := make(chan error)
 			ccq := newCcqRunP2p(logger, ccqAllowedPeers)
-			if err := ccq.run(ctx, priv, gk, networkID, ccqBootstrapPeers, ccqPort, signedQueryReqC, queryResponseReadC, ccqErrC); err != nil {
+			if err := ccq.run(ctx, priv, gk, networkID, ccqBootstrapPeers, ccqPort, components.ListeningAddressesPatterns, signedQueryReqC, queryResponseReadC, ccqErrC); err != nil {
```
evan-gray, line 360: can you remove this now?

node/pkg/p2p/ccq_p2p.go
```diff
@@ -72,6 +72,7 @@ func (ccq *ccqP2p) run(
 	networkID string,
 	bootstrapPeers string,
 	port uint,
+	listeningAddressesPatterns []string,
```
evan-gray, line 75: seems like this isn't needed anymore

node/pkg/p2p/ccq_p2p.go
```diff
@@ -84,6 +85,9 @@ func (ccq *ccqP2p) run(
 	}
 	components.Port = port
 
+	// If we have already cut over, the default listen addresses may be out of date. Use the updated ones passed in.
+	copy(components.ListeningAddressesPatterns, listeningAddressesPatterns)
```
evan-gray, line 89: likewise

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,131 @@
...
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// The format of this time is very picky. Please use the exact format specified by cutOverFmtStr!
+const mainnetCutOverTimeStr = ""
+const testnetCutOverTimeStr = "2024-12-31T23:59:59-0000"
+const devnetCutOverTimeStr = "2022-12-31T23:59:59-0000"
```
evan-gray, line 14: just to confirm, this means devnet is using quic-v1 after this PR?

node/pkg/p2p/cutover.go
```diff
@@ -0,0 +1,131 @@
...
+	return pattern
+}
+
+// validateBootstrapPeers ensures that the quic parameters in a bootstrap peers string are consistent. It is a separate function for testing purposes.
+func validateBootstrapPeers(bootstrapPeers string) error {
+	if !strings.Contains(bootstrapPeers, "/quic/") && !strings.Contains(bootstrapPeers, "/quic-v1/") {
+		return fmt.Errorf(`unexpected format, does not contain "/quic/" or  "/quic-v1/"`)
+	}
+
+	if strings.Contains(bootstrapPeers, "/quic/") && strings.Contains(bootstrapPeers, "/quic-v1/") {
```
evan-gray, line 126: why does this matter? `cutOverBootstrapPeers` can safely handle either of these cases

evan-gray (changes requested):
Really like this new design, much lower footprint and simplified code paths.

node/pkg/p2p/cutover.go
bruce-riley: Removed this function.

node/pkg/p2p/cutover_test.go
```diff
@@ -0,0 +1,97 @@
...
+const newBootstrapPeers = "/dns4/guardian-0.guardian/udp/8999/quic-v1/p2p/12D3KooWL3XJ9EMCyZvmmGXL2LMiVBtrVa2BuESsJiXkSj7333Jw,/dns4/guardian-0.guardian/udp/8999/quic-v1/p2p/12D3KooWL3XJ9EMCyZvmmGXL2LMiVBtrVa2BuESsJiXkSj7333Jx"
+
+func TestGetCutOverTimeStr(t *testing.T) {
+	assert.Equal(t, mainnetCutOverTimeStr, getCutOverTimeStr("blah/blah/mainnet/blah"))
+	assert.Equal(t, testnetCutOverTimeStr, getCutOverTimeStr("blah/blah/testnet/blah"))
+	assert.Equal(t, devnetCutOverTimeStr, getCutOverTimeStr("blah/blah/devnet/blah"))
+}
+
+func TestValidateBootstrapPeers(t *testing.T) {
+	assert.NoError(t, validateBootstrapPeers(oldBootstrapPeers)) // All old quic should work.
```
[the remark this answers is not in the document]
bruce-riley: Added tests for `bootstrapAddrs` and `ListeningAddresses` assuming we are cutting over.

node/pkg/p2p/cutover.go
bruce-riley: Right. Had an offline discussion with @evan-gray and we agreed this was okay.

evan-gray (approved):

panoel (approved):

SEJeff (approved):

PR #3443, opened 2023-10-13
CCQ: Server should check api key first

bruce-riley:
The CCQ server should make sure the API key is valid before decoding the request body. Also it should limit the request body to 5MB.

There are two new integration tests for this PR that can't be merged until PR #3421 lands:

`valid api key but payload too large should fail based on size`
`invalid api key with payload too large should fail based on api key`

### DIFF
--- a/node/cmd/ccq/http.go
+++ b/node/cmd/ccq/http.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/certusone/wormhole/node/pkg/common"
@@ -17,6 +18,8 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 )
+
+const MAX_BODY_SIZE = 5 * 1024 * 1024
 
 type queryRequest struct {
 	Bytes     string `json:"bytes"`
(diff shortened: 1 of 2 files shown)


claudijd (approved):

panoel (dismissed):

bruce-riley:
PR https://github.com/wormhole-foundation/wormhole/pull/3421 has landed, so the two integration tests mentioned above have been added.

claudijd (approved):

node/cmd/ccq/http.go
```diff
@@ -43,27 +46,37 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
...
+	apiKeys, exists := r.Header["X-Api-Key"]
+	if !exists || len(apiKeys) != 1 {
+		s.logger.Debug("received a request with the wrong number of api keys", zap.Stringer("url", r.URL), zap.Int("numApiKeys", len(apiKeys)))
 		http.Error(w, "api key is missing", http.StatusUnauthorized)
 		return
 	}
+	apiKey := apiKeys[0]
+
+	// Make sure the user is authorized before we go any farther.
+	_, exists = s.permissions[strings.ToLower(apiKey)]
```
evan-gray, line 66: we do this `ToLower` in two spots now (here and at the top of validate). perhaps just do that once in the line above?

bruce-riley: Will do this in the next PR.

PR #3447, opened 2023-10-13
CODEOWNERS: sdk/js-query

bruce-riley:


evan-gray (approved):
Assuming the switch is due to the fact that these are mostly js and there's only one go one 😅

SEJeff (approved):
I'm happy to do the go sdk if you want as well.

bruce-riley:
Put myself back and Added @SEJeff as a backup for the go stuff. It would be nice if the go stuff were in a sub directory so we could just be owners of that.

SEJeff (dismissed):

evan-gray (approved):

SEJeff (approved):

PR #3466, opened 2023-10-26
Node/EVM: Shouldn't publish safe block in heartbeats

bruce-riley:
The EVM watcher is incorrectly publishing the safe block height in the heartbeat messages. It should be publishing the finalized block.


panoel (approved):

evan-gray (approved):
Thank you! This was bothering me. :sweat_smile:

PR #3470, opened 2023-10-27
Node/EVM: Track latest block

bruce-riley:
This PR changes the EVM watcher so that it publishes latest, safe and finalized blocks for all chains. These are published in the gossip heartbeats and in Prometheus metrics.

This PR does the following:

- Adds a new `BatchPollConnector` which uses batch RPC calls to query chains that support querying for `finalized` and `safe`.
- For chains that support instant finality, the `EthereumConnector` and `CeloConnector` publish each block as `latest`, `safe` and `finalized`.
- For Polygon, in addition to using the root contract on Ethereum to read finalized blocks (which are also published as safe), it also subscribes to block events to be published as latest.
- For chains that do not have instant finality and do not support querying for the `finalized` tag (currently only Neon), the existing `BlockPollConnector` that uses a finalizer was modified to also publish latest and safe, in addition to the current finalized.

The result of this PR is that every chain now publishes all three finality types, where if the chain does not provide a query for `safe`, it is published when the block is finalized.


node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -0,0 +1,285 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-timer.C:
+			lastBlocks, err = b.pollBlocks(ctx, logger, lastBlocks)
+			if err != nil {
+				errCount++
+				logger.Error("batch polling encountered an error", zap.Int("errCount", errCount), zap.Error(err))
+				if errCount > 3 {
+					b.errFeed.Send(fmt.Sprint("polling encountered an error: ", err))
```
panoel, line 120: Do you want to clear the error count so you send on every 3rd or just every one after 3?

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -53,186 +46,169 @@ func NewBlockPollConnector(ctx context.Context, baseConnector Connector, finaliz
...
+		newLatest = prevLatest
 	}
 
-	return
-}
+	newFinalized = prevFinalized
+	if newLatest.Number.Cmp(prevFinalized.Number) > 0 {
+		var finalized bool
+		// If there is a gap between prev and new, we have to look up the transaction hashes for the missing ones. Do that in batches.
+		newBlockNum := newLatest.Number.Uint64()
```
panoel, line 159: This looks almost the same as lines 134 - 149.  Why are we doing it again?

node/pkg/watchers/evm/connectors/polygon.go
```diff
@@ -128,6 +136,20 @@ func (c *PolygonConnector) SubscribeForBlocks(ctx context.Context, errC chan err
...
+					c.logger.Error("new latest header event is nil")
+					continue
+				}
+				if ev.Number == nil {
+					c.logger.Error("new latest header block number is nil")
+					continue
+				}
+				sink <- &NewBlock{
+					Number:   ev.Number,
+					Hash:     ev.Hash(),
```
panoel, line 150: Do you need to error check if hash is nil?

node/pkg/watchers/evm/watcher.go
```diff
@@ -595,36 +608,42 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+					currentEthSafeHeight.WithLabelValues(w.networkName).Set(float64(blockNumberU))
 				} else {
-					p2p.DefaultRegistry.SetNetworkStats(w.chainID, &gossipv1.Heartbeat_Network{
-						Height:          ev.Number.Int64(),
-						ContractAddress: w.contract.Hex(),
-					})
 					atomic.StoreUint64(&currentBlockNumber, blockNumberU)
 					atomic.StoreUint64(&w.latestFinalizedBlockNumber, blockNumberU)
-					currentEthHeight.WithLabelValues(w.networkName).Set(float64(ev.Number.Int64()))
+					currentEthFinalizedHeight.WithLabelValues(w.networkName).Set(float64(blockNumberU))
```
panoel, line 631: Why a float64?  Is that just what gauges take?

panoel (dismissed):

node/pkg/watchers/evm/connectors/poller.go
bruce-riley: Above is for latest. This is for finalized. I added comments.

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+)
+
+const MAX_GAP_BATCH_SIZE uint64 = 5
+
+func NewBatchPollConnector(ctx context.Context, baseConnector Connector, delay time.Duration) (*BatchPollConnector, error) {
+	// Create the batch data in the order we want to report them to the watcher, so finalized is most important, latest is least.
+	batchData := []BatchEntry{
+		{tag: "finalized", finality: Finalized},
+		{tag: "safe", finality: Safe},
```
evan-gray, line 48: This looks like it will always query for `safe` though the spot where it is used has a boolean flag for if safe is supported. Does something more flexible need to be implemented?

node/pkg/watchers/evm/connectors/polygon.go
bruce-riley: I don't think hash can be nil geth always allocates the bytes for it.

node/pkg/watchers/evm/watcher.go
```diff
@@ -225,7 +238,7 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
 			return fmt.Errorf("dialing eth client failed: %w", err)
 		}
-		w.ethConn, err = connectors.NewBlockPollConnector(ctx, baseConnector, finalizers.NewDefaultFinalizer(), 250*time.Millisecond, true, safeBlocksSupported)
+		w.ethConn, err = connectors.NewBatchPollConnector(ctx, baseConnector, 250*time.Millisecond)
```
evan-gray, line 241: The preview poll connector looks like it optionally supported safe blocks based on `safeBlocksSupported` but `NewBatchPollConnector` will always try to query for finalized, safe, and latest. Is that an issue?

evan-gray (changes requested):

panoel (dismissed):

node/pkg/watchers/evm/watcher.go
evan-gray: Yep https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gauge

node/pkg/watchers/evm/watcher.go
```diff
@@ -202,7 +213,7 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 	defer watcherContextCancelFunc()
 
 	var useFinalizedBlocks bool
-	useFinalizedBlocks, w.safeBlocksSupported, err = w.getFinality(ctx)
+	useFinalizedBlocks, err = w.getFinality(ctx)
```
evan-gray, line 216: I think not setting `w.safeBlocksSupported` will break ccq.go in that it will always be false even when a chain supports `safe`

evan-gray (changes requested):
I think there's one case of not setting safe + ccq that I'm concerned about.

node/pkg/watchers/evm/watcher.go
evan-gray: ln 479 in ccq.go is where it's still used.

panoel (approved):

node/pkg/watchers/evm/watcher.go
evan-gray: Confirming that this removal is sane, because the watcher will have safe and finalized block heights populated for all chains.

evan-gray (approved):

PR #3491, opened 2023-11-02
CCQ/Node/EVM: Cache timestamps

bruce-riley:
This PR is the first phase of caching the timestamp to block ID mapping in the EVM watcher for use by the CCQ `eth_call_by_timestamp` request.

This first phase only adds latest blocks to the cache, without any back filling of the cache. It also only supports three chains, Ethereum, Base and Optimism.

The next phase after this PR will add the ability to back fill the cache by doing batch queries to read old blocks.

Future phases will add support for more EVM chains, which should really only be a testing effort.

### DIFF
--- a/node/pkg/watchers/evm/blocks_by_timestamp.go
+++ b/node/pkg/watchers/evm/blocks_by_timestamp.go
@@ -46,18 +46,22 @@ func (bts *BlocksByTimestamp) AddLatest(logger *zap.Logger, timestamp uint64, bl
 	defer bts.mutex.Unlock()
 	l := len(bts.cache)
 	if l > 0 && (blockNum <= bts.cache[l-1].BlockNum || timestamp < bts.cache[l-1].Timestamp) {
-		for idx := len(bts.cache) - 1; idx >= 0; idx-- {
+		// The cache is in order of both timestamp and block number. Search backwards until we find the entry where the block number is less than the one
+		// passed in and the timestamp is less than or equal to the one passed in. We then truncate everything after that before adding the new one.
+		idx := l - 1
+		for ; idx >= 0; idx-- {
 			if bts.cache[idx].BlockNum < blockNum && bts.cache[idx].Timestamp <= timestamp {
-				logger.Warn("rollback detected in timestamp cache",
-					zap.Uint64("oldLatestBlockNum", bts.cache[len(bts.cache)-1].BlockNum),
-					zap.Uint64("oldLatestTimestamp", bts.cache[len(bts.cache)-1].Timestamp),
-					zap.Uint64("newLatestBlockNum", blockNum),
-					zap.Uint64("newLatestTimestamp", timestamp),
-				)
-				bts.cache = bts.cache[:idx+1]
 				break
 			}
 		}
+
+		logger.Warn("rollback detected in timestamp cache",
+			zap.Uint64("oldLatestBlockNum", bts.cache[l-1].BlockNum),
+			zap.Uint64("oldLatestTimestamp", bts.cache[l-1].Timestamp),
+			zap.Uint64("newLatestBlockNum", blockNum),
+			zap.Uint64("newLatestTimestamp", timestamp),
+		)
+		bts.cache = bts.cache[:idx+1]
 	}
 
 	bts.cache = append(bts.cache, Block{Timestamp: timestamp, BlockNum: blockNum})
--- a/node/pkg/watchers/evm/blocks_by_timestamp.go
+++ b/node/pkg/watchers/evm/blocks_by_timestamp.go
@@ -1,0 +1,149 @@
+package evm
+
+import (
+	"fmt"
+	"sort"
+	"sync"
+)
+
+const (
+	BTS_MAX_BLOCKS = 10000
+)
+
+type (
+	BlocksByTimestamp struct {
+		// cache is ordered by timestamp, blockNum. There may be multiple entries for the same timestamp, but not the same block.
+		cache Blocks
+
+		// maxCacheSize is used to trim the cache.
+		maxCacheSize int
+
+		// mutex is used to protect the cache.
+		mutex sync.Mutex
+	}
+
+	Blocks []Block
+
+	Block struct {
+		Timestamp uint64
+		BlockNum  uint64
+	}
+)
+
+// NewBlocksByTimestamp creates an empty cache of blocks by timestamp.
+func NewBlocksByTimestamp(maxCacheSize int) *BlocksByTimestamp {
+	return &BlocksByTimestamp{
+		cache:        Blocks{},
+		maxCacheSize: maxCacheSize,
+	}
+}
+
+// AddLatest adds a block to the end of the cache. This is meant to be used in the normal scenario when a new latest block is received. It will return an error if the block should not go at the end.
+func (bts *BlocksByTimestamp) AddLatest(timestamp uint64, blockNum uint64) error {
+	bts.mutex.Lock()
+	defer bts.mutex.Unlock()
+	if len(bts.cache) > 0 {
+		last := &bts.cache[len(bts.cache)-1]
+		if timestamp < last.Timestamp {
+			return fmt.Errorf("timestamp %d must be greater than or equal to the previous latest %d", timestamp, last.Timestamp)
+		}
+		if blockNum <= last.BlockNum {
+			return fmt.Errorf("block number %d must be greater than the previous latest %d", blockNum, last.BlockNum)
+		}
+	}
+
+	bts.cache = append(bts.cache, Block{Timestamp: timestamp, BlockNum: blockNum})
+
+	if len(bts.cache) > bts.maxCacheSize {
+		bts.cache = bts.cache[1:]
+	}
+	return nil
+}
+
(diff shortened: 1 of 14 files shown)


node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// First build a map of all the existing blocks so we can avoid duplicates.
+	blockMap := make(map[uint64]uint64)
+	for _, block := range bts.cache {
+		blockMap[block.BlockNum] = block.Timestamp
+	}
+
+	// Now add the new blocks to the map, overwriting any duplicates. (Maybe there was a reorg. . .)
+	for _, block := range blocks {
+		blockMap[block.BlockNum] = block.Timestamp
```
panoel, line 76: Do you need to check and have only the most recent timestamp in an overwrite?

panoel (dismissed):

node/pkg/watchers/evm/blocks_by_timestamp.go
evan-gray: I would think this is intentional, I'm not sure in a reorg the timestamp would necessarily be later, but we could expect it would come in second (at least when streaming via websocket)

evan-gray: Though, I am wondering if we should correct them on Safe and Finalized (to have a Safe value overwrite a Latest value and a Finalized value overwrite either)

evan-gray: (now seeing that this is within Add Batch, I'm not even sure about the above comments)

node/pkg/watchers/evm/ccq.go
```diff
@@ -211,6 +211,61 @@ func (w *Watcher) ccqHandleEthCallByTimestampQueryRequest(logger *zap.Logger, ct
...
+		if w.ccqTimestampCache == nil {
+			logger.Error("error in block id hints in eth_call_by_timestamp query request, they are unset and chain does not support timestamp caching")
+			w.ccqSendQueryResponseForError(logger, queryRequest, query.QueryFatalError)
+			return
+		}
+
+		// Look the timestamp up in the cache. Note that the cache uses native EVM time, which is seconds, but CCQ uses milliseconds, so we have to convert.
+		blockNum, nextBlockNum, found := w.ccqTimestampCache.LookUp(req.TargetTimestamp / 1000000)
+		if !found {
+			if blockNum == 0 {
```
evan-gray, line 234: this is also be the case when the cache is empty though, right? which might be helped by waiting? guess that would only be a startup issue so it's likely not a big deal and would only be for the first second of the watcher lifecycle

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -0,0 +1,149 @@
...
+	bts.cache = cache
+}
+
+// LookUp searches the cache for the specified timestamp and returns the blocks surrounding that timestamp. It also returns true if the results are complete or false if they are not.
+// The following rules apply:
+// - If timestamp is less than the first timestamp in the cache, it returns (0, <theFirstBlockInTheCache>, false)
+// - If timestamp is greater than or equal to the last timestamp in the cache, it returns (<theLastBlockInTheCache>, 0, false)
+// - If timestamp exactly matches one in the cache, it returns (<theLastBlockForThatTimestamp>, <theFirstBlockForTheNextTimestamp>, true)
+// - If timestamp is not in the cache, but there are blocks around it, it returns (<theLastBlockForThePreviousTimestamp>, <theFirstBlockForTheNextTimestamp>, false)
+func (bts *BlocksByTimestamp) LookUp(timestamp uint64) (uint64, uint64, bool) {
```
evan-gray, line 105: this function is a little confusing but I think I get it :joy:

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	bts.cache = cache
+}
+
+// LookUp searches the cache for the specified timestamp and returns the blocks surrounding that timestamp. It also returns true if the results are complete or false if they are not.
+// The following rules apply:
+// - If timestamp is less than the first timestamp in the cache, it returns (0, <theFirstBlockInTheCache>, false)
+// - If timestamp is greater than or equal to the last timestamp in the cache, it returns (<theLastBlockInTheCache>, 0, false)
+// - If timestamp exactly matches one in the cache, it returns (<theLastBlockForThatTimestamp>, <theFirstBlockForTheNextTimestamp>, true)
```
evan-gray, line 103: it seems like this is what it returns if it matches one or more in the cache, right?

evan-gray (commented):
Commenting for now as I feel slightly confused about some edge cases.

evan-gray:
Leaving a comment to follow up - I think this will have to handle correcting the timestamps on rollbacks since they can change.
Here's a block that rolled back on Ethereum
https://etherscan.io/block/18520556/f 2:04:47
https://etherscan.io/block/18520556 2:04:59

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -0,0 +1,149 @@
...
+		// mutex is used to protect the cache.
+		mutex sync.Mutex
+	}
+
+	Blocks []Block
+
+	Block struct {
+		Timestamp uint64
+		BlockNum  uint64
+	}
```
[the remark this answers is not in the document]
bruce-riley: For now we are only storing latest blocks in the cache.

node/pkg/watchers/evm/blocks_by_timestamp.go
bruce-riley: `AddBatch` will only be used for back filling with the latest data received from the RPC, so it is assumed to be correct. That's why it overwrites things.

node/pkg/watchers/evm/blocks_by_timestamp.go
bruce-riley: A timestamp can only match one timestamp. :grin:

node/pkg/watchers/evm/watcher.go
```diff
@@ -616,6 +622,7 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 					currentEthHeight.WithLabelValues(w.networkName).Set(float64(blockNumberU))
 					stats.Height = int64(blockNumberU)
 					w.updateNetworkStats(&stats)
+					w.ccqAddLatestBlock(logger, ev)
```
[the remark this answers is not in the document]
bruce-riley: After much offline discussion, the decision was made to switch the batch poller to subscribe for `newHead` for latest so that rollbacks will be published as new blocks. Also `AddLatest` was reworked to truncate the cache and apply the new block rather than throwing an error on rollback.

node/pkg/watchers/evm/ccq.go
bruce-riley: Changed it to retry.

panoel (dismissed):

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -0,0 +1,157 @@
...
+	if l > 0 && (blockNum <= bts.cache[l-1].BlockNum || timestamp < bts.cache[l-1].Timestamp) {
+		for idx := len(bts.cache) - 1; idx >= 0; idx-- {
+			if bts.cache[idx].BlockNum < blockNum && bts.cache[idx].Timestamp <= timestamp {
+				logger.Warn("rollback detected in timestamp cache",
+					zap.Uint64("oldLatestBlockNum", bts.cache[len(bts.cache)-1].BlockNum),
+					zap.Uint64("oldLatestTimestamp", bts.cache[len(bts.cache)-1].Timestamp),
+					zap.Uint64("newLatestBlockNum", blockNum),
+					zap.Uint64("newLatestTimestamp", timestamp),
+				)
+				bts.cache = bts.cache[:idx+1]
```
evan-gray, line 57: I think in the case where the cache is fresh and a rollback is greater than `l-1`, the cache won't get pruned and will no longer be sorted appropriately. :cry:

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -70,6 +72,15 @@ func (b *BatchPollConnector) SubscribeForBlocks(ctx context.Context, errC chan e
 	innerErrSink := make(chan string, 10)
 	innerErrSub := b.errFeed.Subscribe(innerErrSink)
 
+	// Use the standard geth head sink to get latest blocks.
```
evan-gray, line 75: Should we mention that the intent is to capture every block including rollback corrections? And that this should only be replaced with polling if the replacement code also sufficiently handles replaying rollback replacement blocks?

evan-gray (changes requested):

node/pkg/watchers/evm/blocks_by_timestamp.go
bruce-riley: Fixed and wrote a test to confirm it.

evan-gray (approved):

panoel (approved):

PR #3495, opened 2023-11-03
Node/EVM: Suppress unnecessary polling errors

bruce-riley:
The new batch poller is generating a lot of errors when running against certain public endpoints, due to load balancing. This PR suppresses those errors and makes the batch poller more robust.

### DIFF
--- a/node/pkg/watchers/evm/connectors/batch_poller.go
+++ b/node/pkg/watchers/evm/connectors/batch_poller.go
@@ -143,38 +143,43 @@ func (b *BatchPollConnector) pollBlocks(ctx context.Context, logger *zap.Logger,
 
 	for idx, newBlock := range newBlocks {
 		if newBlock.Number.Cmp(prevBlocks[idx].Number) > 0 {
-			// If there is a gap between prev and new, we have to look up the transaction hashes for the missing ones. Do that in batches.
+			// If there is a gap between prev and new, we have to look up the hashes for the missing ones. Do that in batches.
 			newBlockNum := newBlock.Number.Uint64()
 			blockNum := prevBlocks[idx].Number.Uint64() + 1
-			for blockNum < newBlockNum {
+			errorFound := false
+			lastPublishedBlock := prevBlocks[idx]
+			for blockNum < newBlockNum && !errorFound {
 				batchSize := newBlockNum - blockNum
 				if batchSize > MAX_GAP_BATCH_SIZE {
 					batchSize = MAX_GAP_BATCH_SIZE
 				}
 				gapBlocks, err := b.getBlockRange(ctx, logger, blockNum, batchSize, b.batchData[idx].finality)
 				if err != nil {
-					return prevBlocks, fmt.Errorf("failed to get gap blocks: %w", err)
-				}
-				prevBlock := prevBlocks[idx]
-				errorFound := false
-				for _, block := range gapBlocks {
-					if block.Number.Uint64() == 0 {
-						errorFound = true
-						break
+					// We don't return an error here because we want to go on and check the other finalities.
+					logger.Error("failed to get gap blocks", zap.Stringer("finality", b.batchData[idx].finality), zap.Error(err))
+					errorFound = true
+				} else {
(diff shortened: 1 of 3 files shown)


node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -204,12 +217,13 @@ func (b *BatchPollConnector) getBlocks(ctx context.Context, logger *zap.Logger)
 			return nil, err
 		}
 
+		var n big.Int
 		m := &result.result
 		if m.Number == nil {
-			logger.Error("failed to unmarshal block: Number is nil", zap.Stringer("finality", finality), zap.String("tag", b.batchData[idx].tag))
-			return nil, fmt.Errorf("failed to unmarshal block: Number is nil")
+			logger.Debug("number is nil, treating as zero", zap.Stringer("finality", finality), zap.String("tag", b.batchData[idx].tag))
```
evan-gray, line 223: should we be publishing `NewBlock`s with zero height?

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -155,15 +155,28 @@ func (b *BatchPollConnector) pollBlocks(ctx context.Context, logger *zap.Logger,
...
+						errorFound = true
+						break
+					}
+
 					b.blockFeed.Send(block)
+					prevBlock = block
+				}
+
+				if errorFound {
+					newBlocks[idx] = prevBlock
```
evan-gray, line 171: can you add a comment with what this code effectively does on errors?

node/pkg/watchers/evm/connectors/batch_poller.go
evan-gray: ah, I think I see - this is internal only and this `0` will cause it to skip on line 149

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -262,12 +276,13 @@ func (b *BatchPollConnector) getBlockRange(ctx context.Context, logger *zap.Logg
 			return nil, err
 		}
 
+		var n big.Int
 		m := &result.result
 		if m.Number == nil {
-			logger.Error("failed to unmarshal block: Number is nil")
-			return nil, fmt.Errorf("failed to unmarshal block: Number is nil")
+			logger.Debug("number is nil, treating as zero", zap.Stringer("finality", finality), zap.String("tag", b.batchData[idx].tag))
```
evan-gray, line 282: similarly this would be skipped on line 161?

node/pkg/watchers/evm/connectors/batch_poller.go
evan-gray: if I'm reading it correctly, it appears to not move that block (tag) forward if it hits a block which had an error. but should it just break instead of setting it back to the previous block (before any of these gap blocks) and continuing. maybe it's worth detailing an example

evan-gray (changes requested):

panoel (approved):

node/pkg/watchers/evm/connectors/batch_poller.go
bruce-riley: Right, nil means the block is not available yet. We don't want to treat it as an error, so we return a block number of zero, which signals the caller that we've reached a point where the block is not available, so we should stop.

node/pkg/watchers/evm/connectors/batch_poller.go
bruce-riley: It sets it to the last successfully read block so that the next iteration will pick up after that. We have already published blocks up to that point, so I don't think we want to revert back before blocks we already published.

evan-gray (approved):
I think this makes sense

node/pkg/watchers/evm/connectors/batch_poller.go
bruce-riley: Made the revisions we discussed offline.

PR #3502, opened 2023-11-06
CCQ: Query server response channel too small

bruce-riley:


evan-gray (dismissed):
Approving as this fixes a bug we were seeing where two responses at the same time for a request which reaches quorum with one more response will cause an infinite hand in the ccq_resp processor.

node/cmd/ccq/p2p.go
```diff
@@ -176,6 +176,12 @@ func runP2P(ctx context.Context, priv crypto.PrivKey, port uint, networkID, boot
 						}
 						delete(responses, requestSignature)
 						pendingResponse.ch <- s
+						select {
+						case pendingResponse.ch <- s:
```
evan-gray, line 180: does this happen twice now?

bruce-riley: Fixed.

evan-gray (approved):

kev1n-peters (approved):

PR #3503, opened 2023-11-06
CCQ: Query server metrics

bruce-riley:
This server adds Prometheus metrics to the CCQ query_server. It adds a new `statusAddr` config parameter which specifies the port on which the query_server should publish metrics and listen for health check requests. This defaults to `localhost:6068`.

In tilt, the following are useful endpoints:

- http://localhost:6068/metrics
- http://localhost:6068/health

### DIFF
--- a/devnet/query-server.yaml
+++ b/devnet/query-server.yaml
@@ -62,5 +62,5 @@ spec:
               protocol: TCP
           readinessProbe:
             httpGet:
-              port: 6068
+              port: status
               path: /health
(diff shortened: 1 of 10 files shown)


devnet/query-server.yaml
```diff
@@ -55,6 +57,10 @@ spec:
...
               name: rest
               protocol: TCP
+            - containerPort: 6068
+              name: status
+              protocol: TCP
           readinessProbe:
-            tcpSocket:
-              port: rest
+            httpGet:
+              port: 6068
```
panoel, line 65: Do you want to call this "status" instead of "6068" in case you change the port number later on?

panoel (dismissed):

evan-gray (approved):

panoel (approved):

PR #3504, opened 2023-11-07
SDK: Add core/base files and  unit tests

barnjamin:
Initial PR to start bringing the connect-sdk into Wormhole monorepo

See the DESIGN.md file for an overview of the intended structure.

The core/base module provides constants and a convenient way to access them as well as utilities for higher level packages.


sdk/js-connect/README.md
```diff
@@ -0,0 +1,338 @@
...
+const txGenerator = tb.createAttestation(token) // => AsyncGenerator<UnsignedTransaction, ...> 
+const txids = await signSendWait(srcChain, txGenerator, src.signer) // => TxHash[]
+
+```
+
+Supported protocols are defined in the [definitions module](https://github.com/wormhole-foundation/connect-sdk/tree/develop/core/definitions/src/protocols)
+
+### Signers
+
+In order to sign transactions, an object that fulfils the `Signer` interface is required.  This is a simple interface that can be implemented by wrapping a web wallet or other signing mechanism.
```
SEJeff, line 258: Down the line, it would be super cool to see a tutorial on hooking this up with [aws kms based ethereum signers](https://aws.amazon.com/blogs/database/part1-use-aws-kms-to-securely-manage-ethereum-accounts/).

sdk/js-connect/README.md
```diff
@@ -0,0 +1,338 @@
...
+const sender: Signer =  // ...
+const receiver: Signer = // ...
+
+```
+
+### Addresses
+
+Within the Wormhole context, addresses are [normalized](https://docs.wormhole.com/wormhole/blockchain-environments/evm#addresses) to 32 bytes and referred to in this SDK as a `UniversalAddresses`.
+
+Each platform comes with an address type that understands the native address formats, unsuprisingly referred to a NativeAddress. This abstraction allows the SDK to work with addresses in a consistent way regardless of the underlying chain.
```
SEJeff, line 311:
```suggestion
Each platform comes with an address type that understands the native address formats, unsuprisingly referred to a `NativeAddress`. This abstraction allows the SDK to work with addresses in a consistent way regardless of the underlying chain. 
```

sdk/js-connect/core/base/src/constants/chains.ts
```diff
@@ -0,0 +1,99 @@
...
+  ["Base", 30],
+  ["Sei", 32],
+  ["Rootstock", 33],
+  ["Wormchain", 3104],
+  ["Cosmoshub", 4000],
+  ["Evmos", 4001],
+  ["Kujira", 4002],
+  // holy cow, how ugly of a hack is that?! - a chainId that's exclusive to a testnet!
+  ["Sepolia", 10002],
+] as const satisfies MapLevel<string, number>;
```
SEJeff, line 47: Does this mean we are going to have yet another place to maintain this list?

Thinking outloud, does it make sense to have a clearly namespaced singular wormhole typescript lib that is solely the constants that can be imported by this new sdk and the "legacy" sdk, or is the legacy one going to be fully deprecated?

testing/Dockerfile.connectsdk.test
```diff
@@ -0,0 +1,24 @@
+FROM node:19.6.1-slim@sha256:a1ba21bf0c92931d02a8416f0a54daad66cb36a85d2b73af9d73b044f5f57cfc
+
+RUN apt-get update && apt-get -y install \
+  git python3 make curl netcat
```
SEJeff, line 4:
```suggestion
RUN apt-get update && apt-get -y install --no-install-recommends \
  git python3 make curl netcat && \
  apt-get clean && \
  rm -rf /var/lib/apt/lists/*
```

Put those images on a diet since you're not using a multi-stage build! Feel free to confirm the before and after with this trick:

```bash
docker image ls --format "{{ .Size }}" $IMAGE_NAME
```

testing/connectsdk.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# num=${NUM_GUARDIANS:-1} # default value for NUM_GUARDIANS = 1
+# for ((i=0; i<num; i++)); do
+#     while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' guardian-$i.guardian:6060/readyz)" != "200" ]]; do sleep 5; echo "waiting for guardian $i"; done
+# done
+# 
+# while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' spy:6060/metrics)" != "200" ]]; do sleep 5; done
+# while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' ibc-relayer:7597/debug/pprof/)" != "200" ]]; do sleep 5; done
+# while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' relayer-engine:3000/metrics)" != "200" ]]; do sleep 5; done
+#CI=true npm --prefix ../sdk/js-connect run test-ci
```
SEJeff, line 14: Unless there is a really good reason, don't commit commented out dead code. If there is, commit it and then remove it in the next commit. That way it is in version control and can be dug up trivially using some git foo that first requires sacrificing the a goat to the git gods.
```suggestion
```

SEJeff (approved):
All nits, very very nice. Feel free to ignore all of them :)

I did not review the typescript bits indepth or the tests at all. The docs are super nice and the sdk is obviously really nice given the magic behind the scenes that connect gives you.

sdk/js-connect/core/base/src/constants/chains.ts
barnjamin: I'm hopeful we end up deriving the "legacy" config/consts from this one so we only have to update it/reference it from one spot. 

If we do have a single Wormhole ts lib, this one is its own module so should be importable to access the config

sdk/js-connect/README.md
```diff
@@ -0,0 +1,338 @@
...
+```bash
+npm install @wormhole-foundation/wormhole-sdk-evm-core
+npm install @wormhole-foundation/wormhole-sdk-evm-tokenbridge
+npm install @wormhole-foundation/wormhole-sdk-solana-core
+npm install @wormhole-foundation/wormhole-sdk-solana-tokenbridge
+```
+
+## Usage
+
+A developer would use the core wormhole-sdk package in conjunction with 1 or more of the chain context packages. Most developers don't use every single chain and may only use a couple, this allows developers to import only the dependencies they actually need.
```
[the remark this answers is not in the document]
barnjamin: There is the `connect-sdk` package that imports/re-exports the core stuff but no kitchen-sink package yet.

I'm unsure how best to handle that but I agree its a PITA to list out every single chain/protocol as a dep if you actually want all of them

sdk/js-connect/core/base/src/constants/chains.ts
artursapek: the plan is that this SDK is the one to rule them all. the legacy one in the monorepo, as well as the `sdk` directory inside `wormhole-connect`, will both be replaced by this.

sdk/js-connect/core/base/src/constants/contracts/relayer.ts
```diff
@@ -0,0 +1,37 @@
+import { MapLevel } from "../../utils";
+import { Network } from "../networks";
+import { Chain } from "../chains";
+
+export const relayerContracts = [[
+  "Mainnet", [
+    ["Ethereum",  "0x27428DD2d3DD32A4D7f7C497eAaa23130d894911"],
```
artursapek, line 7: I'm guessing you copy pastad all this and they don't need to be individually verified?

artursapek:
I would change the `js-connect` directory name to `typescript` or `js-v2` or something like that. make it clear this is more than just a Connect SDK.

sdk/js-connect/core/base/src/constants/contracts/relayer.ts
barnjamin: yes copy/pasta'd but probably makes sense to check a few

sdk/js-connect/core/base/src/utils/metaprogramming.ts
```diff
@@ -0,0 +1,59 @@
+export type Extends<T, U> = [T] extends [U] ? true : false;
+
+export type Function<P extends any[] = any[], R = any> = (...args: P) => R;
```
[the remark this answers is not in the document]
barnjamin: it does https://www.typescriptlang.org/docs/handbook/2/functions.html#function

cc @nonergodic do we need this definition?

kev1n-peters:
I think that we should add a `.prettierrc` file in the root directory and potentially a CI job that enforces formatting

nonergodic:
I strongly object against the prettier. What's currently not pretty? I can point to loads of places that will be less readable, once a "prettier" is done with disfiguring the code that I wrote.

nonergodic:
The compromise that Ben and I originally came up with was that "my" part of the code would be exempt (i.e. get a prettier ignore entry). This way, whoever has the most skin in the games in various parts of the code gets to call the shots, rather than having some external "authority" come up with "sensible regulations". Uniformity and standardization is necessary when you run a fast-food franchise, but you'll not get a five star restaurant that way.

barnjamin:
> The compromise that Ben and I originally came up with was that "my" part of the code would be exempt (i.e. get a prettier ignore entry). This way, whoever has the most skin in the games in various parts of the code gets to call the shots, rather than having some external "authority" come up with "sensible regulations". Uniformity and standardization is necessary when you run a fast-food franchise, but you'll not get a five star restaurant that way.

fwiw I think I borked the formatting again, will have to make it pretty once again. I think we should use the formatting everywhere _but_ cordon off certain things like const declarations with an ignore directive https://prettier.io/docs/en/ignore.html#javascript

sdk/js-connect/core/base/src/utils/metaprogramming.ts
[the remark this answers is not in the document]
nonergodic: As the documentation says:
> This is an untyped function call and is generally best avoided because of the unsafe any return type.
I.e. another thing that TypeScript has and says "oh, it's there but, you shouldn't really use it. Think of it like a rake that we put in tall grass so you can accidentally step on it and have it smack you in the face. We put it there because we thought it was funny."

Ultimately, both `Function` and `Extends`, just like e.g. `RoArrayf, are simply there to make stuff more readable (in my book - reasonable people can disagree here I think :man_shrugging: )

nonergodic:
> fwiw I think I borked the formatting again, will have to make it pretty once again. Can we use the formatting everywhere _but_ cordon off certain things like const declarations with an ignore directive https://prettier.io/docs/en/ignore.html#javascript

Making my point for me.

Example taken from mapping.ts:
Before "prettier":
```
type TransformMapping<M extends MappingEntries, S extends Shape | void = void> =
  //check that M has a valid structure for mapping entries
  CartesianRightRecursive<M> extends infer CRR extends RoArray2D
  ? IsRectangular<CRR> extends true
    //ensure CRR is not empty  
    ? CRR extends readonly [RoArray, ...RoArray2D]
      ? S extends Shape
        ? SplitAndReorderKeyValueColumns<CRR, S> extends [
            infer KC extends CartesianSet<MappableKey>,
            infer VC extends CartesianSet
          ]
          ? KC["length"] extends Depth[number]
            ? CombineValueColumnsToLeafValues<VC> extends infer VR extends RoArray<LeafValue>
              ? ProcessNextKeyColmn<KC, VR> extends infer TM extends MappingEntries
                ? [MaybeUnwrapValuesIfAllAreSingletons<TM, KC["length"]>, KC["length"]]
                : never
              : never
            : never
          : never
        //if we don't have an explicit shape, take the first row and subtract 1 (for the value
        //  column) to determine the count of key columns
        : CRR[0] extends readonly [...infer KC extends RoArray, unknown]
        ? KC["length"] extends Depth[number]
          ? [M, KC["length"]]
          : never
        : never
      : never
    : never
  : never;
  ```
After "prettier":
```
type TransformMapping<M extends MappingEntries, S extends Shape | void = void> =
  //check that M has a valid structure for mapping entries
  CartesianRightRecursive<M> extends infer CRR extends RoArray2D
  ? IsRectangular<CRR> extends true
  //ensure CRR is not empty  
  ? CRR extends readonly [RoArray, ...RoArray2D]
  ? S extends Shape
  ? SplitAndReorderKeyValueColumns<CRR, S> extends [
    infer KC extends CartesianSet<MappableKey>,
    infer VC extends CartesianSet
  ]
  ? KC["length"] extends Depth[number]
  ? CombineValueColumnsToLeafValues<VC> extends infer VR extends RoArray<LeafValue>
  ? ProcessNextKeyColmn<KC, VR> extends infer TM extends MappingEntries
  ? [MaybeUnwrapValuesIfAllAreSingletons<TM, KC["length"]>, KC["length"]]
  : never
  : never
  : never
  : never
  //if we don't have an explicit shape, take the first row and subtract 1 (for the value
  //  column) to determine the count of key columns
  : CRR[0] extends readonly [...infer KC extends RoArray, unknown]
  ? KC["length"] extends Depth[number]
  ? [M, KC["length"]]
  : never
  : never
  : never
  : never
  : never;
  ```

I rest my case.

kev1n-peters:
Ah, prettier does make the code pretty unreadable in that case. Maybe we could compromise like previously suggested and ignore formatting in cases where it hurts readability.

artursapek:
it's worth trying this `prettier` option: https://prettier.io/docs/en/options.html#experimental-ternaries

and just generally seeing if we can configure it to suit your fancy. it's highly configurable.

nonergodic:
> it's worth trying this `prettier` option: https://prettier.io/docs/en/options.html#experimental-ternaries
> 
> and just generally seeing if we can configure it to suit your fancy. it's highly configurable.

We have a fundamental, philosophical difference here.

You can't nail down what makes readable code, just like you can't nail down virtue and what it means to be virtuous.

Sure, you can give guidelines (be honest, be charitable, and so on) but you can't put this "into code". Trying to replace wisdom with intelligence is exactly the sort of trap that intelligent nerds fall into (I used to be one of them). I now reject this entire line of thinking.

You can define some approximation that you keep refining over time, but I don't see the purpose.

Be a virtuous person. Write readable code.

artursapek:
I think people who use formatters usually do it to be more efficient. I don't like spending my time formatting my code. it's one of many chores in my life which I prefer to automate as much as I can.

sdk/js-connect/core/base/src/utils/layout/index.ts
```diff
@@ -0,0 +1,36 @@
+export {
```
artursapek, line 1: this `utils/layout` is quite a piece of work. I don't see unit tests in this directory. is this unit tested somehow elsewhere in this large diff?

barnjamin: There are some tests that use this downstream https://github.com/wormhole-foundation/connect-sdk/blob/develop/core/definitions/__tests__/transferVaa.ts

barnjamin: Some tests here https://github.com/wormhole-foundation/wormhole/blob/3926dc88eba6d3e9ede3eed4864796a99a9249fc/sdk/js-connect/core/base/__tests__/layout.ts

PR #3509, opened 2023-11-08
node/watchers/solana: additional safety check

evan-gray:
This PR adds a safety check to ensure that, on Solana, message publication account fields which were already zeroed out, are definitely zeroed out. This becomes important in the upcoming rewrite when the account can be written to across multiple instructions and one of these fields contains the `status`, which is set to zero when the account can no longer be written to.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -1,6 +1,7 @@
 package solana
 
 import (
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):

a5-pickle (approved):
Clean

PR #3512, opened 2023-11-09
node/watchers/solana: pythnet safety carve out

evan-gray:
As of 2023-11-09, Pythnet has a bug which is not zeroing out these fields appropriately. This carve out should be removed after a fix is deployed.

This bug was identified as this security check was triggered in testnet.


bruce-riley (approved):

a5-pickle (approved):
Looking forward to when this can be removed.

PR #3516, opened 2023-11-10
CCQ: Prevent arbitrary external calls, fuzzing and extra safety checks for QueryResponse contract

djb15:
EDIT: Also restricted the visibility of a couple of methods to prevent arbitrary external calls from the context of the integrating/inheriting contract. My preference would be to make all the other parsing methods internal as well since I don't think there is any real need for them to be public facing, but let me know if you disagree.

Added fuzz tests to `parseAndVerifyQueryResponse` and `verifyQueryResponseSignatures` as I think these are the most important entrypoints. This required a small refactor to split the hex string into its constituent parts so we can construct the payloads on the fly.

Assuming you could get the guardians to sign the CCQ, there was a potential to manipulate the byte parsing in `parseAndVerifyQueryResponse` and double parse the same bytes. I have added two new checks to guard against this.

The other parsing functions don't perform much validation and thus don't have many permutations. The `checkLength` calls at the end of those methods should guard against weird input bytes but ultimately most of the validation of the data is left to the consuming contract that's inheriting the library. Hence no fuzz testing for those since I think it would add very little value.

Feel free to rip this to shreds, but hopefully the extra tests are useful 😄

### DIFF
--- a/ethereum/contracts/query/QueryDemo.sol
+++ b/ethereum/contracts/query/QueryDemo.sol
@@ -19,7 +19,7 @@ error UnexpectedCallData();
 error UnexpectedResultLength();
 error UnexpectedResultMismatch();
 
-/// @dev QueryDemo is a library that implements the parsing and verification of Cross Chain Query (CCQ) responses.
+/// @dev QueryDemo is an example of using the QueryResponse library to parse and verify Cross Chain Query (CCQ) responses.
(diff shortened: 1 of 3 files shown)


djb15:
Don't current have permissions to update the list of reviewers for this so sorry if you get pinged! Will update as soon as I can

Issue #783, opened 2022-01-26
solana: handle multiple Wormhole instructions in a single tx 


leoluk:
This isn't currently supported.

bruce-riley:
This issue was fixed by PR #3527.

PR #3527, opened 2023-11-14
Node/Solana: Process multiple wormhole instructions

bruce-riley:
This PR addresses issue #783.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -821,7 +821,7 @@ func (s *SolanaWatcher) processMessageAccount(logger *zap.Logger, data []byte, a
 			zap.Error(err))
 		return
 	}
-	if commitment != s.commitment {  
+	if commitment != s.commitment {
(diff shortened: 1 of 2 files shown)


node/pkg/watchers/solana/client.go
```diff
@@ -598,14 +599,20 @@ OUTER:
```
barnjamin, line 599: Change looks good but I do think we might be able to remove the above RPC call to get inners:
https://github.com/gagliardetto/solana-go/blob/main/rpc/getTransaction.go#L40-L94

barnjamin (approved):

node/pkg/watchers/solana/client.go
bruce-riley: I made this change.

barnjamin (dismissed):

a5-pickle (dismissed):

barnjamin (dismissed):

a5-pickle (approved):

evan-gray (approved):

PR #3528, opened 2023-11-14
node: Add cache buster for CoinGecko price queries

djb15:
Probably overkill but my thinking was that since price queries are every 15 minutes, someone could predict the cache key if just the Unix timestamp was used; hence concatenating the unix timestamp and a random number.

Tested cache hits/misses with Postman


SEJeff (approved):
Nice!

SEJeff (approved):

bruce-riley (approved):

PR #3529, opened 2023-11-14
Algorand/upgrade

scnale:
This PR installs `goal` into the build docker image and uses it to compile the contracts instead of building against a public third party node.

### DIFF
--- a/algorand/Dockerfile.build
+++ b/algorand/Dockerfile.build
@@ -1,15 +1,16 @@
 FROM docker.io/python:3.10@sha256:eeed7cac682f9274d183f8a7533ee1360a26acb3616aa712b2be7896f80d8c5f as algorand-contracts-build
 
-RUN python3 -m pip install virtualenv
-COPY Pipfile.lock Pipfile.lock
-COPY Pipfile Pipfile
+RUN wget https://github.com/algorand/go-algorand/releases/download/v3.19.0-stable/node_stable_linux-amd64_3.19.0.tar.gz
+RUN tar --extract --file node_stable_linux-amd64_3.19.0.tar.gz bin/goal && rm node_stable_linux-amd64_3.19.0.tar.gz
+
+RUN mkdir teal
+COPY Pipfile Pipfile.lock ./
 
 RUN python3 -m pip install pipenv
 RUN pipenv install
-RUN mkdir teal
-COPY *.py .
 
-RUN pipenv run python3 admin.py --mainnet --genTeal 
+COPY *.py ./
+RUN pipenv run python3 admin.py --mainnet --genTeal
 
 FROM scratch AS algorand-contracts-export
 COPY --from=algorand-contracts-build /teal/* .
--- a/algorand/test/test.py
+++ b/algorand/test/test.py
@@ -661,8 +661,8 @@ class AlgoTest(PortalCore):
             try:
                 self.submitVAA(vaa, client, player, self.tokenid)
             except algosdk.error.AlgodHTTPError as e:
-                # should fail right at line 963
-                if "opcodes=pushint 963" in str(e):
+                # should fail right at line 935
+                if "opcodes=pushint 935" in str(e):
                     return True, vaa, None
                 return False, vaa, e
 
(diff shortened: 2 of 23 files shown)


algorand/Dockerfile.build
```diff
@@ -1,15 +1,16 @@
 FROM docker.io/python:3.10@sha256:eeed7cac682f9274d183f8a7533ee1360a26acb3616aa712b2be7896f80d8c5f as algorand-contracts-build
 
-RUN python3 -m pip install virtualenv
-COPY Pipfile.lock Pipfile.lock
-COPY Pipfile Pipfile
+RUN mkdir algosdk
+RUN --mount=type=cache,target=/algosdk wget https://github.com/algorand/go-algorand/releases/download/v3.19.0-stable/node_stable_linux-amd64_3.19.0.tar.gz --output-document=algosdk/node-stable-linux.tar.gz
```
kcsongor, line 4: does this cache mount do anything? the file is deleted in the next run anyway, and even if it wasn't, this doesn't provide any better caching than what docker would do anyway by just caching the output of the `RUN` layer without the cache mount. applies to the line below too

algorand/Dockerfile.build
```diff
@@ -1,15 +1,16 @@
...
+RUN --mount=type=cache,target=/algosdk tar --extract --file algosdk/node-stable-linux.tar.gz bin/goal && rm algosdk/node-stable-linux.tar.gz
+
+RUN mkdir teal
+COPY Pipfile Pipfile.lock *.py ./
```
kcsongor, line 8: why move the `COPY *.py` before the `pipenv install` command? this way, each time any of the python files are modified, `pipenv install` will be run again. The old version only reran `pipenv install` when the `Pipfile` changed

kcsongor (changes requested):
The are other places where `client.compile` is used directly:
https://github.com/wormhole-foundation/wormhole/blob/22bf5f698749dbfceadf81392595ab4b01268499/algorand/admin.py#L283
https://github.com/wormhole-foundation/wormhole/blob/22bf5f698749dbfceadf81392595ab4b01268499/algorand/test_contract.py#L40
https://github.com/wormhole-foundation/wormhole/blob/22bf5f698749dbfceadf81392595ab4b01268499/algorand/token_bridge.py#L61

at a quick glance it looks like the token bridge and wormhole core `fullyCompileContract` functions are identical, so maybe we could just unify them in a separate utility module

algorand/Dockerfile.build
scnale: It probably doesn't. I thought I needed it to cache locally but I realized I was modifying a previous layer afterwards invalidating everything. I'll try removing it and see how it behaves.

algorand/Dockerfile.build
scnale: Hm, you're right. Once again, I was running into something that was triggering a rebuild always but it was probably me modifying the dockerfile. I can no longer reproduce it after reintroducing the copy command at least.

algorand/test/test.py
```diff
@@ -672,7 +662,7 @@ def double_submit_transfer_vaa_fails(seq):
                 self.submitVAA(vaa, client, player, self.tokenid)
             except algosdk.error.AlgodHTTPError as e:
                 # should fail right at line 963
```
kcsongor, line 664: comment outdated now

algorand/txs.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if pending_txn.get("confirmed-round", 0) > 0:
+            return PendingTxnResponse(pending_txn)
+
+        if pending_txn["pool-error"]:
+            raise Exception(f"Pool error: {pending_txn['pool-error']}")
+
+        lastStatus = client.status_after_block(lastRound + 1)
+
+        lastRound += 1
```
kcsongor, line 55: shouldn't we sleep here a bit?

scnale: Could be :thinking: I just moved this logic from elsewhere but didn't attempt to look into its correctness :P

PR #3537, opened 2023-11-15
CCQ: High level validation helpers + slight refactor

djb15:
As discussed I've added some high level helpers that should hopefully make validating a parsed response easier for integrators rather than having to write validation logic from scratch every time.

I've also slightly refactored the passing of the wormhole address to avoid the potential for an integrator accidentally allowing the caller to specify this value.

### DIFF
--- a/ethereum/contracts/query/QueryDemo.sol
+++ b/ethereum/contracts/query/QueryDemo.sol
@@ -10,12 +10,11 @@ import "./QueryResponse.sol";
 error InvalidOwner();
 // @dev for the onlyOwner modifier
 error InvalidCaller();
-error InvalidContractAddress();
+error InvalidCalldata();
 error InvalidWormholeAddress();
 error InvalidForeignChainID();
 error ObsoleteUpdate();
 error StaleUpdate();
-error UnexpectedCallData();
 error UnexpectedResultLength();
 error UnexpectedResultMismatch();
 
@@ -32,14 +31,13 @@ contract QueryDemo is QueryResponse {
     }
 
     address private immutable owner;
-    address private immutable wormhole;
     uint16 private immutable myChainID;
     mapping(uint16 => ChainEntry) private counters;
     uint16[] private foreignChainIDs;
 
     bytes4 GetMyCounter = bytes4(hex"916d5743");
 
-    constructor(address _owner, address _wormhole, uint16 _myChainID) {
+    constructor(address _owner, address _wormhole, uint16 _myChainID) QueryResponse(_wormhole) {
         if (_owner == address(0)) {
             revert InvalidOwner();
         }
@@ -48,7 +46,7 @@ contract QueryDemo is QueryResponse {
         if (_wormhole == address(0)) {
             revert InvalidWormholeAddress();
         }
-        wormhole = _wormhole;  
+
         myChainID = _myChainID;
         counters[_myChainID] = ChainEntry(_myChainID, address(this), 0, 0, 0);
     }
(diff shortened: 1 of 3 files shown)


ethereum/contracts/query/QueryResponse.sol
```diff
@@ -341,13 +357,80 @@ abstract contract QueryResponse {
...
+    }
+
+    /// @dev validateBlockNum validates that the parsed blockNum is in an acceptable range
+    function validateBlockNum(uint64 _blockNum, uint256 _minBlockNum, uint256 _maxBlockNum) public pure {
+        if (_blockNum < _minBlockNum || _blockNum > _maxBlockNum) {
+            revert InvalidBlockNum();
+        }
+    } 
+
+    /// @dev validateChainId validates that the parsed chainId is one of an array of chainIds we expect
```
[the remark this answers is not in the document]
djb15: Yup 🤦 nice spot, will fix when I move on to writing tests

ethereum/contracts/query/QueryResponse.sol
```diff
@@ -73,18 +74,33 @@ error RequestTypeMismatch();
...
+error InvalidFunctionSignature();
+error InvalidChainId();
+error InvalidBlockNum();
+error InvalidBlockTime();
 
 // @dev QueryResponse is a library that implements the parsing and verification of Cross Chain Query (CCQ) responses.
 abstract contract QueryResponse {
     using BytesParsing for bytes;
 
+    IWormhole public immutable wormhole;
```
evan-gray, line 87: I don't think we should have state defined in a helper contract unless it is stored at a deterministic address. @kcsongor for best practice recommendations here.

evan-gray (changes requested):

ethereum/contracts/query/QueryResponse.sol
djb15: It's immutable so it's not really state (it's inlined into the bytecode at compile time), plus if you have to inherit the helper contract anyway (because it's abstract) I'm not sure why declaring this in the helper is a problem? But intrigued to hear the argument against this

evan-gray: Ah, fair. Def right considering the immutable state. That said, I'm wondering if it _should_ be immutable - it's (very remotely) possible that an integrator may want / need to switch this.

evan-gray: Sounds like @kcsongor is in favor of immutable use here. So I'll close this.

SEJeff (dismissed):

gator-boi (changes requested):

ethereum/contracts/query/QueryDemo.sol
```diff
@@ -101,33 +99,24 @@ contract QueryDemo is QueryResponse {
...
-                revert ObsoleteUpdate();
-            }
 
-            // wormhole time is in microseconds, timestamp is in seconds
-            adjustedBlockTime = eqr.blockTime / 1_000_000;
-            if (adjustedBlockTime <= block.timestamp - 300) {
-                revert StaleUpdate();
-            }
+            // Validate that update is not obsolete
+            validateBlockNum(eqr.blockNum, chainEntry.blockNum, block.number);
```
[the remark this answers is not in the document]
djb15: Good point. We only really care if the updates are stale so I've updated as suggested

ethereum/contracts/query/QueryResponse.sol
```diff
@@ -341,13 +357,85 @@ abstract contract QueryResponse {
...
+        if (_blockNum < _minBlockNum || _blockNum > _maxBlockNum) {
+            revert InvalidBlockNum();
+        }
+    } 
+
+    /// @dev validateChainId validates that the parsed chainId is one of an array of chainIds we expect
+    function validateChainId(uint16 chainId, uint16[] memory _validChainIds) public pure {
+        bool validChainId = false;
+        
+        for (uint256 i = 0; i < _validChainIds.length; ++i) {
```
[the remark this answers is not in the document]
djb15: Good spot, thanks. Have updated this and also used "idx" instead of "i" to match existing code

gator-boi (approved):
lgtm

SEJeff (approved):

PR #3541, opened 2023-11-16
ci: Update tilt timeout to 60m

djb15:
Tilt CI has a default timeout of 30m, so it times out before the github action step. This change should fix the intermittent build failures.


djb15:
Closing to debug some more

PR #3545, opened 2023-11-16
clients/js: Fix `maybeUnsupported` function

kev1n-peters:
Previously this function would not actually catch an exception and cause the program to crash. This commit fixes that.


kev1n-peters:
I was testing the `simulate_upgrades.sh` script for some other EVM chains and encountered this issue.

kcsongor (approved):

evan-gray (approved):

PR #3363, opened 2023-09-07
ethereum: fix some intermittently failing tests

SEJeff:
Fix the assumptions to match the storage layout correctly. This fixes some fallout from #2956

PR #3548, opened 2023-11-17
ethereum: Fix flaky tests

djb15:
Turns out that the recipient address for the failure cases was the same as the WH proxy, so I've added that as another address to avoid

### DIFF
--- a/ethereum/forge-test/Governance.t.sol
+++ b/ethereum/forge-test/Governance.t.sol
@@ -792,8 +792,10 @@ contract TestGovernance is TestUtils {
         return
             // Avoid precompiled contracts
             addr <= address(0x9) ||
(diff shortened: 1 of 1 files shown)


SEJeff (approved):
My intuition tells me this is akin to #3363 where some of the calculations were a bit off. This makes sense, thanks!

djb15:
Looks like there's another flaky one 😬 are we better off finding a fix rather than commenting out more tests? I don't know how useful these tests are yet but feels bad to comment more things out that I assume are there for a reason

SEJeff (approved):

evan-gray (approved):
thanks!

gator-boi (approved):
lgtm

PR #3552, opened 2023-11-21
Node/EVM: Celo TimeOfBlockByHash is wrong

bruce-riley:


evan-gray (approved):
I expect this should fix the "transaction type not supported" error caused by another transaction in the same block tripping this up. Bonus, not it's consistent with the Eth one.

panoel (approved):

PR #3558, opened 2023-11-27
CCQ/Node/EVM: Backfill cache

bruce-riley:


node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,303 @@
...
+)
+
+const CCQ_MAX_BATCH_SIZE = int64(1000)
+const CCQ_TIMESTAMP_RANGE_IN_MS = uint64(30 * 60)
+
+type ccqBackfillRequest struct {
+	timestamp uint64
+}
+
+// ccqRequestBackfill submits a request to backfill a gap in the timestamp cache.
```
panoel, line 24: Could you add to this comment saying the timestamp is in microseconds?

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,303 @@
...
+		return fmt.Errorf("failed to backfill timestamp cache: %w", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case evt := <-w.ccqBackfillChannel:
+			w.ccqPerformBackfill(ctx, evt)
+		}
```
panoel, line 47: Do you need a default case?

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,303 @@
...
+		w.ccqBatchSize, blocks, err = ccqBackFillDetermineMaxBatchSize(ctx, w.ccqLogger, w.ethConn, latestBlockNum, 100*time.Millisecond)
+		if err != nil {
+			return fmt.Errorf("failed to determine max batch size: %w", err)
+		}
+	} else {
+		blocks = append(blocks, Block{BlockNum: latestBlock.Number.Uint64(), Timestamp: latestBlock.Time})
+		w.ccqLogger.Info("using existing batch size for timestamp cache", zap.Int64("batchSize", w.ccqBatchSize))
+	}
+
+	if len(blocks) == 0 {
```
panoel, line 87: Is this even possible?  Wouldn't you have, at least, the latest block in here, if nothing else?

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,303 @@
...
+
+	"go.uber.org/zap"
+
+	ethHexUtil "github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/rpc"
+	ethRpc "github.com/ethereum/go-ethereum/rpc"
+)
+
+const CCQ_MAX_BATCH_SIZE = int64(1000)
+const CCQ_TIMESTAMP_RANGE_IN_MS = uint64(30 * 60)
```
panoel, line 18: Isn't this 30 minutes in seconds?

panoel (approved):

node/pkg/watchers/evm/ccq_backfill.go
bruce-riley: Actually it's in seconds. I added a comment.

node/pkg/watchers/evm/ccq_backfill.go
bruce-riley: I don't think so. We want it to just block if there's nothing to do.

node/pkg/watchers/evm/ccq_backfill.go
bruce-riley: It shouldn't but the for loop would panic if it did, so I would like to keep it. I added a comment.

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,303 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get batch starting at %d: %w", blocks[len(blocks)-1].BlockNum-1, err)
+		}
+
+		if len(newBlocks) == 0 {
+			w.ccqLogger.Error("failed to read any more blocks, giving up on the backfill")
+			break
+		}
+
+		blocks = append(blocks, newBlocks...)
```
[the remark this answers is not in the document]
bruce-riley: Actually they are in descending order. I added a comment saying so.

panoel (approved):

bruce-riley:
Should add a feature flag to enable this. Initially it can default to disabled, so we can ask a single guardian to enable it. Once it is working smoothly we would probably want to make it default to enabled.

bruce-riley:
There is a guardian flag to make back filling configurable. It defaults to true. This is okay because if the back fill fails, it does not kill the watcher, it just disables back filling and continues.

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -159,3 +159,9 @@ func (lhs Block) Cmp(rhs Block) int {
 
 	return 0
 }
+
+func (bts *BlocksByTimestamp) GetStats() (int, Block, Block) {
```
bingyuyap, line 163: is this still used anywhere? only found the definition here but no references

node/pkg/watchers/evm/blocks_by_timestamp.go
```diff
@@ -159,3 +159,9 @@ func (lhs Block) Cmp(rhs Block) int {
...
+	bts.mutex.Lock()
+	defer bts.mutex.Unlock()
+	return len(bts.cache), bts.cache[0], bts.cache[len(bts.cache)-1]
```
bingyuyap, line 166: can `len(bts.cache)` be `0`? should we check just in case `bts.cache[0]` gives `index out of range` error

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// ccqBackfillInit determines the maximum batch size to be used for backfilling the cache. It also loads the initial batch of timestamps.
+func (w *Watcher) ccqBackfillInit(ctx context.Context) error {
+	// Get the latest block so we can use that as the starting point in our cache.
+	latestBlock, err := connectors.GetLatestBlock(ctx, w.ccqLogger, w.ethConn)
+	if err != nil {
+		return fmt.Errorf("failed to look up latest block: %w", err)
+	}
+	latestBlockNum := int64(latestBlock.Number.Uint64())
```
bingyuyap, line 81: is there a reason why we use `int64(latestBlock.Number.Uint64())` instead of `latestBlock.Number.Int64()`?  
if the value is greater than `9223372036854775807` I think both gives us issue so wondering if there's a reason

node/pkg/watchers/evm/ccq_backfill.go
```diff
@@ -0,0 +1,316 @@
...
+	w.ccqLogger.Info("looked up latest block", zap.Int64("latestBlockNum", latestBlockNum), zap.Uint64("timestamp", latestBlock.Time))
+
+	var blocks Blocks
+	if w.ccqBatchSize == 0 {
+		// Determine the max supported batch size and get the first batch which will start with the latest block and go backwards.
+		var err error
+		w.ccqBatchSize, blocks, err = ccqBackFillDetermineMaxBatchSize(ctx, w.ccqLogger, w.ethConn, latestBlockNum, 100*time.Millisecond)
```
bingyuyap, line 88: would be nice if the delay is defined as a constant!

bingyuyap (approved):
thanks for the hard work!

node/pkg/watchers/evm/blocks_by_timestamp.go
bruce-riley: Deleted it.

node/pkg/watchers/evm/ccq_backfill.go
bruce-riley: Changed it.

node/pkg/watchers/evm/ccq.go
```diff
@@ -213,15 +213,48 @@ func (w *Watcher) ccqHandleEthCallByTimestampQueryRequest(ctx context.Context, q
...
-				zap.Uint64("timestamp", req.TargetTimestamp),
-				zap.String("block", block),
-				zap.String("nextBlock", nextBlock),
-				zap.Uint64("blockNum", blockNum),
-				zap.Uint64("nextBlockNum", nextBlockNum),
-			)
-			w.ccqSendQueryResponse(queryRequest, query.QueryRetryNeeded, nil)
+			status := query.QueryRetryNeeded
+			if nextBlockNum == 0 {
+				w.ccqLogger.Warn("block look up failed in eth_call_by_timestamp query request, timestamp beyond the end of the cache, will wait and retry",
```
panoel, line 218: "timestamp beyond end of cache" means it is too old?

evan-gray (approved):

panoel (approved):

PR #3562, opened 2023-11-29
Node: Add cosmos chains over IBC

bruce-riley:
This PR adds guardian support for the following cosmos chains over IBC:

- cosmoshub 
- evmos
- kujira
- celestia
- osmosis

Support for these chains was previously added to the SDK. They are defined here:
https://github.com/wormhole-foundation/wormhole/blob/main/sdk/js/src/utils/consts.ts#L36-L39


node/pkg/watchers/ibc/watcher.go
```diff
@@ -63,7 +63,13 @@ func setFeatures(f string) {
 
 var (
 	// Chains defines the list of chains to be monitored by IBC. Add new chains here as necessary.
-	Chains = []vaa.ChainID{vaa.ChainIDSei}
+	Chains = []vaa.ChainID{
+		vaa.ChainIDSei,
+		vaa.ChainIDCosmoshub,
```
SEJeff, line 68: We should change this from Cosmos Hub to Cosmos. Hub is the name of the mainnet and gaia is the name of the testnet. Is it possible to change the sdk at this point still?

Unless we want to add a CosmosGaia as well, but that doesn't fit how most all of the rest of the chains are.

SEJeff (commented):
The change is totally reasonable, but I feel naming it cosmos hub after the mainnet network explicitly is poor form. We should change it to just `cosmos` like the rest of the networks.

node/pkg/watchers/ibc/watcher.go
evan-gray: @SEJeff not sure your comment is quite correct. It looks to me like both [Mainnet](https://github.com/cosmos/mainnet) and [Testnets](https://github.com/cosmos/testnets) are referred to as Cosmos Hub and [Gaia](https://hub.cosmos.network/main/getting-started/what-is-gaia.html) is the Cosmos SDK application.

node/pkg/watchers/ibc/watcher.go
```diff
@@ -63,7 +63,13 @@ func setFeatures(f string) {
 
 var (
 	// Chains defines the list of chains to be monitored by IBC. Add new chains here as necessary.
-	Chains = []vaa.ChainID{vaa.ChainIDSei}
+	Chains = []vaa.ChainID{
```
evan-gray, line 66: Just want to clarify that with this change, these chains will be able to emit messages as soon as the channel is allow-listed with governance - is that the idea?

evan-gray (approved):

node/pkg/watchers/ibc/watcher.go
SEJeff: I stand corrected sir. They've changed this from when I last worked on cosmos. Perfect!

SEJeff (dismissed):

node/pkg/watchers/ibc/watcher.go
bruce-riley: Yes. When the watcher receives a message for a channel it does not know about, it queries the ibc-receiver contract for the channel to chain mapping. If for some reason, the contract doesn't have the channel to chain mapping (meaning governance hasn't happened yet, but if governance hasn't happened yet, the contract shouldn't publish the message), the watcher generates an error message like "received message from unknown IBC channel" and drops the message.

evan-gray (approved):

nik-suri (approved):
LGTM, thanks!

Only nit thought is we might as well as Osmosis (chain id 20) in here as well, since it was defined in the SDK a long time ago and isn't in the guardian codebase.

nik-suri (approved):
LGTM

evan-gray (approved):

SEJeff (approved):

PR #3566, opened 2023-11-30
Node/Gov: Update token list

bruce-riley:
This is just a routine update of the governor token list.


djb15 (approved):
Actually just been having a think about how to do this in a more automated manner + make it easier to review changes. I'll put some thoughts in a Notion doc

Issue #3190, opened 2023-07-12
sdk/js: catch-all catch blocks swallow exceptions


kcsongor:
## Description and context
In several places, such as integration tests
https://github.com/wormhole-foundation/wormhole/blob/b09a644dac97fa8e037a16765728217ff3a1d057/sdk/js/src/token_bridge/__tests__/aptos-integration.ts#L108-L112
but even in core sdk functions
https://github.com/wormhole-foundation/wormhole/blob/b09a644dac97fa8e037a16765728217ff3a1d057/sdk/js/src/nft_bridge/getIsTransferCompleted.ts#L55-L58

the catch blocks just swallow the exceptions. This is in most cases anticipating a specific exception, so the catch block should just check for that, and rethrow in any other case.

A good example of a catch block that does the correct thing is
https://github.com/wormhole-foundation/wormhole/blob/b09a644dac97fa8e037a16765728217ff3a1d057/sdk/js/src/token_bridge/getIsTransferCompleted.ts#L72-L79

## Definition of done
No catch-all catch blocks in the sdk.

AlberErre:
Happy to help with this 👀 @kcsongor 

could you assign me the issue? 🏄‍♂️

ritvij14:
@AlberErre are you still working on this or can I take it up?

AlberErre:
Hey @ritvij14 feel free to pick this up 👍, I think @kcsongor could assign you the issue 👀

ritvij14:
@kcsongor could you please assign the issue to me, I will get started on it

Abhishekkochar:
@kcsongor If no-one is actively working on this. I can pick this up from here. Cheers

kcsongor:
hey @Abhishekkochar, thanks! I don't believe anyone's actively working on this, so I'll assign the issue to you if you're interested in giving it a go

Abhishekkochar:
@kcsongor double checking, are non-org people allowed to contribute? Currently getting error 403 when pushing the branch. I have already check the credentials.  Thanks heaps.

kcsongor:
You might have to fork the repo and push the branch to your own fork, then open a PR from there!

PR #3590, opened 2023-12-09
sdk/js: handling valid exceptions

Abhishekkochar:
fix: #3190


Abhishekkochar:
@kcsongor, Please have a look.

sdk/js/src/token_bridge/__tests__/algorand-integration.ts
```diff
@@ -127,7 +127,11 @@ describe("Algorand tests", () => {
             vaaBytes
           );
         } catch (e) {
-          success = false;
+          if(e instanceof Error && e.message.includes("wrapped asset already exists")){
+            success = true;
```
kcsongor, line 131: shouldn't this be the other way around?

sdk/js/src/token_bridge/__tests__/algorand-integration.ts
```diff
@@ -370,7 +374,12 @@ describe("Algorand tests", () => {
             vaaBytes
           );
         } catch (e) {
-          success = false;
+          //@notice
+          if(e instanceof Error && e.message.includes("wrapped asset already exists")){
+            success = true;
```
kcsongor, line 379: this one too

sdk/js/src/token_bridge/__tests__/aptos-integration.ts
```diff
@@ -420,6 +425,11 @@ describe("Aptos SDK tests", () => {
       await createWrappedOnEth(ethTokenBridge, recipient, attestVAA);
     } catch (e) {
       // this could fail because the token is already attested (in an unclean env)
+      if(e instanceof Error && e.message.includes("wrapped asset already exists")){
```
kcsongor, line 428: does the error message actually contain this?

kcsongor (changes requested):
Did you run these tests locally?

sdk/js/src/token_bridge/__tests__/algorand-integration.ts
Abhishekkochar: Oh yes. Fixed. Will push soon.

Abhishekkochar:
@kcsongor, Can you please confirm if this is correct [algogrand/package](https://github.com/wormhole-foundation/wormhole/blob/main/algorand/package.json#L14)?
There is no test.ts file in test dir.

Abhishekkochar:
> Did you run these tests locally?

I have tried to run them. There is an error when running alogrand tests?

evan-gray:
This error case is only hit when the tests are run a second time on an existing tilt instance. This never happens in CI and is only a hinderance to a developer trying to run the tests without cycling the corresponding chains. Given that the subsequent test cannot pass if the token was not properly attested, the existing code is adequate for CI. So, for the benefit of devs in this repo, it seems a more prudent thing would be to skip the test step altogether if the token was already attested.

PR #3138, opened 2023-06-27
terra: migrate to cosmwasm 1.1.0

kcsongor:
Terra classic upgraded to CosmWasm 1.0+ (https://classic-agora.terra.money/t/wasmvm-v0-16-6-v1-2-0/50070/73), which introduced a number of breaking changes across the whole ecosystem, ranging from tooling to smart contracts.

This PR upgrades the terra classic contracts to the new CosmWasm stdlib version, along with upgrades to the testing pipeline to support the new version. Some contract changes had to be made to accommodate the new 32 byte addressing scheme, these are documented in #3597.

## Reviewing the PR
The PR is best reviewied commit-by-commit. The upgrade to CW1 necessitated vendoring the `cw20-legacy` library. To keep things tidy, I forked it as a separate repo, the relevant commit is here https://github.com/wormhole-foundation/cw20-legacy/commit/d12724701ffb9920cb1fdf1eee49efdcc20048e5.

PR #3597, opened 2023-12-11
terra/doc: add some implementation notes

kcsongor:
Additional docs following #3138

### DIFF
--- a/terra/docs/token_id.md
+++ b/terra/docs/token_id.md
@@ -1,0 +1,58 @@
+# Token ID encoding in the Token bridge
+
+## Background
+
+Terra classic (being a Cosmos chain with a CosmWasm runtime) supports two types of tokens: "native Bank tokens" and "CW20 tokens".
+The Bank module is a native Cosmos SDK module that supports the transfer of tokens between accounts. These tokens are identified by their denomination (e.g. `uluna` or `uusd`), and are analogous to the native tokens on other chains, such as eth on Ethereum or sol on Solana.
+CW20 tokens on the other hand are smart contracts that implement the CW20 interface, which is analogous to the ERC20 interface on EVM chains. These tokens are identified by their contract address.
+The Terra token bridge supports both types of tokens directly (i.e. without wrapping them in a synthetic token like the EVM token bridge wraps eth into the canonical Wrapped ETH ERC20 contract). This means that the token bridge needs to be able to distinguish between the two types of tokens.
+
+Addresses (both account and CosmWasm contract) on Terra used to fit into 20 bytes.
+This changed when the chain underwent a hard fork to upgrade the runtime to CosmWasm 1.1.0 in June 2023. New contract addresses and account addresses are now 32 bytes long.
+
+The initial design was made with the assumption that addresses are 20 bytes long. We first discuss that original version, and the adjustments that were made to support 32 byte addresses.
+
+## Token ID encoding before CosmWasm 1.1.0
+
+In the [Wormhole Token Bridge wire format](../../whitepapers/0003_token_bridge.md), token addresses are encoded as 32 bytes. Since CW20 addresses were 20 bytes long, the first 12 bytes were set to zero. The decision was also made to limit the length of native denom strings to 20 bytes also. This meant that the first 12 bytes of both CW20 addresses and native denoms were always zero.
+The way the token bridge would then distinguish between the two is by writing a `0x01` byte in the first byte position of native denoms. Then, if the first byte of the token address is `0x01`, the token is a native denom, and if it is `0x00`, the token is a CW20 token. If it is anything else, the token is invalid.
+
+```rust
+    let marker_byte = transfer_info.token_address.as_slice()[0];
+    if transfer_info.token_chain == CHAIN_ID {
+        match marker_byte {
+            1 => handle_complete_transfer_token_native(...),
+            0 => handle_complete_transfer_token(...),
+            b => Err(StdError::generic_err(format!("Unknown marker byte: {b}"))),
+        }
+    } else {
+        handle_complete_transfer_token(...)
+    }
+```
+[wormhole/terra/contracts/token-bridge/src/contract.rs#L734-L770](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L734-L770)
+
+## Token ID encoding after CosmWasm 1.1.0
+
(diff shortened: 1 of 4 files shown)


terra/docs/token_id.md
```diff
@@ -0,0 +1,58 @@
...
+[wormhole/terra/contracts/token-bridge/src/contract.rs#L734-L770](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L734-L770)
+
+## Token ID encoding after CosmWasm 1.1.0
+
+After the hard fork, addresses can now be 32 bytes long. Theoretically this would mean that new (32 byte addressed) tokens bridged out then back could collide with that check above. However, on the way out the token's address was checked to fit into 20 bytes, so no 32 byte addressed CW20 could be bridged out. That is, prior to upgrading the contracts to CW 1.1.0, the token bridge would not allow bridging out 32 byte addressed CW20 tokens.
+
+In order to support 32 byte addresses, we simply change the above check so instead of just checking that the first byte is `0x01` for native denoms, we check that the first byte is `0x01` and the next 11 bytes are `0x00`:
+```rust
+fn is_native_id(address: &[u8]) -> bool {
+    address[0] == 1 && address[1..12].iter().all(|&x| x == 0)
```
SEJeff, line 41: This document explains it well, but that code doesn't explain why those 11 bytes are important.

terra/docs/token_upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+        })
+```
+[wormhole/terra/contracts/token-bridge/src/contract.rs#L458-L477](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L458-L477)
+
+
+## Upgrading the wrapped CW20 contract
+
+When upgrading the token contract, two steps need to be taken:
+
+1. update the code ID in the state so future wrapped assets are created from the new ID:
```
SEJeff, line 44:
```suggestion
1. Update the code ID in the state so future wrapped assets are created from the new ID:
```

terra/docs/token_upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+
+```rust
+    let mut c = config(deps.storage).load()?;
+    c.wrapped_asset_code_id = new_code_id;
+    config(deps.storage).save(&c)?;
+```
+[wormhole/terra/contracts/token-bridge/src/contract.rs#L79-L81](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L79-L81)
+
+2. migrate all existing contracts to the new code ID. A simple implementation of this function can be seen [here](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L123-L147). Unfortunately, while this simple approach works in local testing, it does not scale to mainnet, the contract sends too many migrate messages (one to each wrapped asset), and exceeds the gas counter.
```
SEJeff, line 53:
```suggestion
2. Migrate all existing contracts to the new code ID. A simple implementation of this function can be seen [here](https://github.com/wormhole-foundation/wormhole/blob/dee0d1532b4a4ab6657dbdd1f0b8d19eadd90ec9/terra/contracts/token-bridge/src/contract.rs#L123-L147). Unfortunately, while this simple approach works in local testing, it does not scale to mainnet, the contract sends too many migrate messages (one to each wrapped asset), and exceeds the gas counter.
```

terra/docs/token_upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+
+Currently there is no implementation of a token contract migration function that works on mainnet. If the need arises, one would have to be designed. If the upgrade does not require the migration to happen atomically, then it could be handled with a new permissionless entrypoint that can handle a range of contracts, and it could be done in multiple transactions. If for some reason atomicity is required, then extra care must be taken to pause all other bridge operation and perform the ugprade over multiple transactions.
```
SEJeff, line 55:
```suggestion
Currently there is no implementation of a token contract migration function that works on mainnet. If the need arises, one would have to be designed. If the upgrade does not require the migration to happen atomically, then it could be handled with a new permissionless entrypoint that can handle a range of contracts, and it could be done in multiple transactions. If for some reason atomicity is required, then extra care must be taken to pause all other bridge operation and perform the upgrade over multiple transactions.
```

terra/docs/token_upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Historical note:
+
+In Aug 2022, the token contracts were attempted to be upgraded, but failed due to the gas error:
```
SEJeff, line 59:
```suggestion
In Aug 2022, an attempt to upgrade the the token contracts was made, but failed due to the gas error:
```

terra/docs/token_upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+https://finder.terra.money/classic/tx/FE39E9549770F59E2AAA1C6B0B86DDF36A4C56CED0CFB0CA4C9D4CC9FBE1E5BA. A subsequent ugprade changed the code id of *new* wrapped contracts to 767, but did not perform the migration for old contracts. This means that currently (as of Dec 2023), some wrapped tokens are still on the old code id, and some (the ones deployed after Aug 2022) are on the new code. This discrepancy is fine in the current case, because it only affects the rendering of the token name (https://github.com/wormhole-foundation/wormhole/commit/c832b123fcfb017d55086cb4d71241370ed270c6).
```
SEJeff, line 60: Bare urls fail in all markdown linters btw.

```suggestion
A [subsequent upgrade](https://finder.terra.money/classic/tx/FE39E9549770F59E2AAA1C6B0B86DDF36A4C56CED0CFB0CA4C9D4CC9FBE1E5BA) changed the code id of *new* wrapped contracts to 767, but did not perform the migration for old contracts. This means that currently (as of Dec 2023), some wrapped tokens are still on the old code id, and some (the ones deployed after Aug 2022) are on the new code. This discrepancy is fine in the current case, because it only affects the rendering of the token name (https://github.com/wormhole-foundation/wormhole/commit/c832b123fcfb017d55086cb4d71241370ed270c6).
```

I'm not sure what additional context to add to the url at the very end in parenthesis, but it should be turned into more context if possible.

SEJeff (approved):
A couple of nits and grammar / spelling tyops. This is fabulous and I wish more parts of wormhole had this.

evan-gray (approved):

a5-pickle (approved):

PR #3598, opened 2023-12-11
Node/CCQ/Server: Clean restart

bruce-riley:
This PR changes the proxy server to handle restarts more cleanly, hopefully without dropping query requests.

It adds two new config parameters (with what should be reasonable default values): `shutdownDelay1` and `shutdownDelay2`. When a sigterm is received, it will do the following:

- Set a flag to stop responding to health checks
- Sleep for `shutdownDelay1` seconds to allow the load balancer to detect that the server is down and stop routing requests to it.
- Sleep for up to `shutdownDelay2` seconds, but checking every second to see if there are no more pending requests.
- When there are no more pending requests (or after `shutdownDelay2` seconds), the server exits.

The default values for the two config parameters are based on the default health check parameter values. We may want to tweak those parameters, in which case these parameters can be reduced.

The length of time that Linux will wait between a sigterm and a sigkill is controlled by the `TimeoutStopSec` parameter in the systemd config file. The default is 90s (seconds).

In the tilt environment, `shutdownDelay1` is set to zero by default, meaning just shutdown immediately.

### DIFF
--- a/devnet/query-server.yaml
+++ b/devnet/query-server.yaml
@@ -56,6 +56,8 @@ spec:
             - --bootstrap
             - /dns4/guardian-0.guardian/udp/8996/quic/p2p/12D3KooWL3XJ9EMCyZvmmGXL2LMiVBtrVa2BuESsJiXkSj7333Jw
             - --logLevel=info
+            - --shutdownDelay1
+            - "0"
           ports:
             - containerPort: 6069
               name: rest
--- a/node/cmd/ccq/query_server.go
+++ b/node/cmd/ccq/query_server.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/certusone/wormhole/node/pkg/common"
 	"github.com/certusone/wormhole/node/pkg/telemetry"
@@ -41,6 +42,8 @@ var (
 	telemetryNodeName *string
 	statusAddr        *string
 	promRemoteURL     *string
+	shutdownDelay1    *int
+	shutdownDelay2    *int
 )
 
 const DEV_NETWORK_ID = "/wormhole/dev"
@@ -61,6 +64,12 @@ func init() {
 	telemetryNodeName = QueryServerCmd.Flags().String("telemetryNodeName", "", "Node name used in telemetry")
 	statusAddr = QueryServerCmd.Flags().String("statusAddr", "[::]:6060", "Listen address for status server (disabled if blank)")
 	promRemoteURL = QueryServerCmd.Flags().String("promRemoteURL", "", "Prometheus remote write URL (Grafana)")
+
+	// The default health check monitoring is every five seconds, with a five second timeout, and you have to miss two, for 20 seconds total.
+	shutdownDelay1 = QueryServerCmd.Flags().Int("shutdownDelay1", 25, "Seconds to delay after disabling health check on shutdown")
+
+	// The guardians will wait up to 60 seconds before giving up on a request.
+	shutdownDelay2 = QueryServerCmd.Flags().Int("shutdownDelay2", 65, "Seconds to wait after delay1 for pending requests to complete")
 }
 
 var QueryServerCmd = &cobra.Command{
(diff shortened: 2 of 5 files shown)


node/cmd/ccq/status.go
```diff
@@ -14,26 +15,38 @@ import (
...
+}
+
+func (s *statusServer) disableHealth() {
+	s.healthEnabled.Store(false)
 }
 
 func (s *statusServer) handleHealth(w http.ResponseWriter, r *http.Request) {
+	if !s.healthEnabled.Load() {
+		s.logger.Info("ignoring health check")
+		return
```
evan-gray, line 48: Should this write a 503 or something?

node/cmd/ccq/query_server.go
```diff
@@ -209,7 +219,21 @@ func runQueryServer(cmd *cobra.Command, args []string) {
...
+			statServer.disableHealth()
+			time.Sleep(time.Duration(*shutdownDelay1) * time.Second)
+			logger.Info("Waiting for outstanding requests to complete before shutting down.")
+			for count := 0; count < int(*shutdownDelay2); count++ {
+				time.Sleep(time.Second)
+				if pendingResponses.Empty() {
+					break
+				}
+			}
+			logger.Info("Done waiting. shutting down.")
```
evan-gray, line 233: should we warn here if the requests aren't empty when we start shutting down anyway?

evan-gray (commented):
generally in favor, just a couple small comments

devnet/query-server.yaml
```diff
@@ -56,6 +56,8 @@ spec:
             - --bootstrap
             - /dns4/guardian-0.guardian/udp/8996/quic/p2p/12D3KooWL3XJ9EMCyZvmmGXL2LMiVBtrVa2BuESsJiXkSj7333Jw
             - --logLevel=info
+            - --shutdownDelay1
```
panoel, line 59: do you need to set a value for shutdownDelay2?

bruce-riley: Don't need to. If delay1 is zero, it never checks delay2.

node/cmd/ccq/query_server.go
```diff
@@ -209,7 +219,27 @@ func runQueryServer(cmd *cobra.Command, args []string) {
 	signal.Notify(sigterm, syscall.SIGTERM)
 	go func() {
 		<-sigterm
-		logger.Info("Received sigterm. exiting.")
+		if statServer != nil && *shutdownDelay1 != 0 {
```
panoel, line 222: Should this check be >0 because shutdownDelay1 is signed?

devnet/query-server.yaml
panoel: If shutdownDelay1 == 0, shutdownDelay2 isn't used, correct?

evan-gray (approved):
Looking forward to this

node/cmd/ccq/query_server.go
panoel: General comment: Do you need to make sure the delay values are non-negative?

node/cmd/ccq/query_server.go
```diff
@@ -209,7 +219,27 @@ func runQueryServer(cmd *cobra.Command, args []string) {
...
+				time.Sleep(time.Second)
+				numPending = pendingResponses.NumPending()
+				if numPending == 0 {
+					break
+				}
+			}
+			if numPending == 0 {
+				logger.Info("Done waiting. shutting down.")
+			} else {
+				logger.Error("Gave up waiting for pending requests to finish. shutting down anyway.", zap.Int("numStillPending", numPending))
```
panoel, line 238: Is it necessary to walk the pending requests and log that they are getting dropped?

panoel (approved):

node/cmd/ccq/query_server.go
bruce-riley: I don't think that is necessary at this time.

node/cmd/ccq/query_server.go
bruce-riley: Changed it to be a uint.

PR #3600, opened 2023-12-12
Node/EVM: Minor code cleanup

bruce-riley:
This PR removes some dead code from the EVM watcher in an attempt to make the code more readable.

Since the watcher has been upgraded to use `finalized` for all chains, the `waitForConfirmations` config parameter was always false, so I've removed it. This meant that the `expectedConfirmations` variable was always zero, so I also removed that.

Additionally, the `maxWaitConfirmations` config parameter is always 60, so I made it a constant.


panoel (approved):

SEJeff (approved):
Nice

PR #3611, opened 2023-12-20
Node/CCQ: Server auto reconnect

bruce-riley:
In testnet, the CCQ proxy server is not detecting when a guardian restarts. As a work around, this PR adds the ability for the server to attempt to automatically reconnect when it detects that a bootstrap peer has gone away.

This feature is only intended to be used in testnet, and is controlled by a config parameter, `monitorPeers`.

This PR also changes some CCQ log messages from error to debug so that routine conditions (like block not available yet) don't flood the logs.


evan-gray (approved):
Might need to switch some of those debugs back later, but such errors have been limited in testing (or can be determined by testing)

panoel (approved):

PR #3617, opened 2023-12-25
update some third party links

hattizai:


evan-gray:
The existing links seem to get viewers to the right place.

PR #3618, opened 2023-12-26
CI: Solana -- Allow running on x86 with no avx

barnjamin:
In order to run Tilt on a M1/M2 architecture, we can rely on Rosetta for virtualization but the binaries must be compiled without support for AVX/AVX2 (and SIMD for Reed Solomon crate).

This is probably slower but I'm not sure by how much or if it will affect CI/testing.

We should also probably make this not the default but add some check during `tilt up` to see if the host is 1) on a mac and 2) has rosetta enabled, then switch to the non-avx base docker image. 

This does build the contracts and run `solana-test-validator` successfully but currently hitting an issue during the setup process: 

image

### DIFF
--- a/solana/Dockerfile
+++ b/solana/Dockerfile
@@ -1,5 +1,5 @@
 #syntax=docker/dockerfile:1.2@sha256:e2a8561e419ab1ba6b2fe6cbdf49fd92b95912df1cf7d313c3e2230a333fdbcc
-FROM ghcr.io/wormhole-foundation/solana:1.10.31@sha256:d31e8db926a1d3fbaa9d9211d9979023692614b7b64912651aba0383e8c01bad AS solana
+FROM ghcr.io/wormhole-foundation/solana-no-avx:1.16.24@sha256:ce400089cca28ac353cc550ee488ea0c6fea2bb07936eca9b113aa487ea1dc56 AS solana
 
 # Add bridge contract sources
 WORKDIR /usr/src/bridge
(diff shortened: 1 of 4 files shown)


barnjamin:
Attempting to set this to tag 1.10.31 to match the current version fails after the main compilation with:

```
373.6     Finished release [optimized] target(s) in 5m 50s
374.9 + [[ -z '' ]]
374.9 + /solana-1.10.31/cargo install --locked spl-token-cli --root /root/.local/share/solana/install/releases/1.10.31/solana-release
374.9 + exec cargo +1.59.0 install --locked spl-token-cli --root /root/.local/share/solana/install/releases/1.10.31/solana-release
374.9     Updating crates.io index
375.6  Downloading crates ...
376.0   Downloaded spl-token-cli v3.3.0
376.0   Installing spl-token-cli v3.3.0
376.2 error: failed to select a version for the requirement `spl-token-2022 = "^1.0"`
376.2 candidate versions found which didn't match: 0.7.0, 0.6.1, 0.6.0, ...
376.2 location searched: crates.io index
376.2 required by package `spl-token-cli v3.3.0`
```

https://github.com/solana-labs/solana/blob/v1.10.31/scripts/cargo-install-all.sh#L133-L137


To make this version work we'd have to make some additional change to force the install of the spl-token-cli necessary

https://github.com/wormhole-foundation/wormhole/blob/main/solana/Dockerfile.client#L20-L21

evan-gray:
Seems like this has been stalled for a bit. It might be more fruitful to explore alternative methods of attaching Solana for ARM Mac users. Maybe using the [devnet accounts directly loaded](https://github.com/wormholelabs-xyz/example-queries-solana-verify/blob/main/Anchor.toml#L26) into a solana-test-validator using the new `solana_watcher` flag akin to what the [NTT example repo is doing](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/Tiltfile). 

Alternatively, I wonder if the pinned SHA is messing with the build if docker is selecting the amd64 image instead of the arm64 one. Perhaps there's a clean-ish way to provide alternate dockerfiles for arm.

I'll close this for now, but feel free to reopen with advancements.

Issue #3612, opened 2023-12-21
Node: Document the RPC / WS parameters


evan-gray:
## Description and context
The guardian node includes multiple formats and requirements for URLs passed as CLI parameters.
e.g. https://github.com/wormhole-foundation/wormhole/blob/main/devnet/node.yaml#L72-L182

Some of them must be ws:// or http:// or either or not have the protocol.
These nuances should be documented in the flag descriptions.
https://github.com/wormhole-foundation/wormhole/blob/main/node/cmd/guardiand/node.go#L231-L407

## Definition of done
All URL (RPC / WS) flags in `node.go` should have examples of their accepted formats in their description.

PR #3622, opened 2024-01-03
node: Add accepted formats and examples for rpc and ws parameter desc

0xluffyb:
PR for https://github.com/wormhole-foundation/wormhole/issues/3612

### DIFF
--- a/node/hack/governor/src/index.ts
+++ b/node/hack/governor/src/index.ts
@@ -164,7 +164,7 @@ axios
             }
 
             // This is a new token
-            if (!existingTokenPrices[chain][wormholeAddr]) {
+            if (existingTokenPrices[chain] == undefined || existingTokenPrices[chain][wormholeAddr] == undefined) {
               addedTokens.push(chain + "-" + wormholeAddr + "-" + data.Symbol);
             }
             // This is an existing token


node/cmd/guardiand/node.go
```diff
@@ -244,130 +244,129 @@ func init() {
 	guardianKeyPath = NodeCmd.Flags().String("guardianKey", "", "Path to guardian key (required)")
 	solanaContract = NodeCmd.Flags().String("solanaContract", "", "Address of the Solana program (required)")
 
-	ethRPC = NodeCmd.Flags().String("ethRPC", "", "Ethereum RPC URL")
+	ethRPC = NodeCmd.Flags().String("ethRPC", "", "Ethereum RPC URL.\nFormat: 'ws://'. Example: 'ws://eth-devnet:8545'")
```
panoel, line 247: Does it matter if the format is 'ws://' or 'wss://'?  If we only show 'ws://' are we implying 'wss://' is not allowed?  For one testnet guardian 'wss://' was used here.
This comment applies everywhere.

panoel (approved):

bruce-riley:
Will there be a follow up to this PR? Because I think we should have some verification to enforce that the parameter is the right format (with ws/wss vs. http/https, etc).

node/cmd/guardiand/node.go
```diff
@@ -244,130 +244,129 @@ func init() {
...
 	algorandAppID = NodeCmd.Flags().Uint64("algorandAppID", 0, "Algorand app id")
 
-	nearRPC = NodeCmd.Flags().String("nearRPC", "", "near RPC URL")
-	nearContract = NodeCmd.Flags().String("nearContract", "", "near contract")
-
-	wormchainURL = NodeCmd.Flags().String("wormchainURL", "", "wormhole-chain gRPC URL")
+	nearRPC = NodeCmd.Flags().String("nearRPC", "", "Near RPC URL.\nFormat: HTTP (http://) or HTTPS (https://). Example: 'http://near:3030'")
+	nearContract = NodeCmd.Flags().String("nearContract", "", "Near contract")
 
+	wormchainURL = NodeCmd.Flags().String("wormchainURL", "", "Wormhole-chain gRPC URL.\nFormat: `<host>:<port>`. Example: 'wormchain:9090'")
```
panoel, line 315: Is this correct?  It doesn't have an http or ws prefix?  It is IP:port?

bruce-riley: Yes, this is correct. The GRPC library doesn't expect the prefix.

bruce-riley (approved):

PR #3631, opened 2024-01-06
Upgrade dance for go-libp2p-kad-dht

SEJeff:
This was supposed to be an "easy" upgrade of a single package, but the open telemetry mess made this super challenging to only update a few packages. I ended up just doing `go mod tidy` to update quite a few of the packages unfortunately.

This is to fix some known bugs in `go-libp2p-kad-dht` that @barnjamin theorizes is causing issues with some users running guardiand spy that lose connections and have to restart to regain connections.


bruce-riley:
Upgraded libp2p to `v0.32.2` so this PR can replace #3716.

evan-gray (approved):

barnjamin (approved):

bruce-riley (approved):

PR #3705, opened 2024-01-11
[Solana rewrite] make some stuff public

guibescos:
I'd like to make `WriteEncodedVaaArgs` and `VAA_START` public so people can import them.
These two fields are useful for client-side Rust code that will interact with this contract:
- `WriteEncodedVaaArgs` is useful for calling `write_encoded_vaa` https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/solana/cli/src/main.rs#L419
- `VAA_START` is useful for initializing the VAA account (the user needs to call this before calling wormhole) https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/solana/cli/src/main.rs#L388

### DIFF
--- a/solana/programs/core-bridge/src/lib.rs
+++ b/solana/programs/core-bridge/src/lib.rs
@@ -20,7 +20,6 @@ pub mod legacy;
 
 mod processor;
 pub(crate) use processor::*;
-pub use processor::WriteEncodedVaaArgs;
 
 pub mod sdk;
 
--- a/solana/programs/core-bridge/src/lib.rs
+++ b/solana/programs/core-bridge/src/lib.rs
@@ -20,6 +20,7 @@ pub mod legacy;
 
 mod processor;
 pub(crate) use processor::*;
+pub use processor::WriteEncodedVaaArgs;
 
 pub mod sdk;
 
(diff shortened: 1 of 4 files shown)


solana/programs/core-bridge/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod legacy;
 
 mod processor;
 pub(crate) use processor::*;
+pub use processor::WriteEncodedVaaArgs;
```
a5-pickle, line 23: Can we move this to the sdk submodule where the other arg structs are exported?

a5-pickle (changes requested):

solana/programs/core-bridge/src/lib.rs
a5-pickle: Along with that, can we put the other cpi methods you're using in this export, too? https://github.com/guibescos/wormhole/blob/solana-rewrite/make-some-stuff-public/solana/programs/core-bridge/src/sdk/mod.rs#L40

a5-pickle: I just meant that the sdk module aggregates all of the cpi calls that an integrator needs, so we can re-export cpi::write_encoded_vaa in here.

I can do that in the next alpha release. Everything looks good thank you

Issue #2393, opened 2023-02-17
Allow using a configuration file instead of only flags for configuration


SEJeff:
## Expected Behavior

Support a configuration file instead of only supporting cli flags for configuration.

## Current Behavior

Only cli flags are supported. Currently, there are around 75+ flags required for running a guardian.

## Possible Solution

Since the guardian already uses [cobra](https://github.com/spf13/cobra), use [viper](https://github.com/spf13/viper) from the same author to unify the config handling and clean up the hand rolled flag parsing code.

PR #3710, opened 2024-01-12
node: Add config file support

0xluffyb:
PR for https://github.com/wormhole-foundation/wormhole/issues/2393

Adds config file and env var functionality. Configs are set according to the precedence of:
1. Command line flags
2. Env var
3. Config file
4. Cobra default values

### DIFF
--- a/devnet/node.yaml
+++ b/devnet/node.yaml
@@ -104,8 +104,6 @@ spec:
             - ws://eth-devnet:8545
             - --optimismRPC
             - ws://eth-devnet:8545
-            - --neonRPC
-            - ws://eth-devnet:8545
             - --baseRPC
             - ws://eth-devnet:8545
             - --scrollRPC
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ devnet-consts.json
 sui.log.*
 sui/examples/wrapped_coin
 *.prof
+# Only used for internal testing
+*.testnet.yaml
--- a/Tiltfile
+++ b/Tiltfile
@@ -188,7 +188,7 @@ def build_node_yaml():
                     "--suiMoveEventType",
                     "0x7f6cebb8a489654d7a759483bd653c4be3e5ccfef17a8b5fd3ba98bd072fabc3::publish_message::WormholeMessage",
                     "--suiWS",
-                    "sui:9000",
+                    "ws://sui:9000",
                 ]
 
             if evm2:
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -283,7 +283,7 @@ func init() {
 	moonbeamRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "moonbeamRPC", "Moonbeam RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	moonbeamContract = NodeCmd.Flags().String("moonbeamContract", "", "Moonbeam contract address")
 
-	neonRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "neonRPC", "Neon RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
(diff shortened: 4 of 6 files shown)


devnet/node.yaml
```diff
@@ -68,11 +71,13 @@ spec:
...
             - mountPath: /tmp/mounted-keys/wormchain
               name: node-wormchain-key
+            - mountPath: /app/node/config
+              name: node-config
           command:
             - /guardiand
             - node
-            - --ethRPC
-            - ws://eth-devnet:8545
+            # - --ethRPC
```
panoel, line 79: Why did this get commented out?

SEJeff: Pretty sure because it is already specified in the config file here:

https://github.com/wormhole-foundation/wormhole/pull/3710/files#diff-9ca9c1f4adddc2b4c6064f90ce44be7c99f78d4b4586c8349d62f873e8637a3eR1

node/pkg/node/config_file_reader_test.go
```diff
@@ -0,0 +1,106 @@
...
+	ethRPC = rootCmd.Flags().String("ethRPC", "", "Ethereum RPC URL")
+	solRPC = rootCmd.Flags().String("solRPC", "", "Solana RPC URL")
+
+	return rootCmd
+}
+
+func TestInitFileConfig(t *testing.T) {
+	// Set ethRPC with config file
+	// Tests that the config file is read and the default value is set
+	t.Run("config file", func(t *testing.T) {
```
SEJeff, line 50: This is entirely fine, but is not generally idiomatic go. Generally speaking, you'll use `t.Run()` when you're doing [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests).

For this, it likely would be cleaner to make separate test functions instead of `TestInitFileConfig` that shares nothing.

node/pkg/node/testdata/test.yaml
```diff
@@ -0,0 +1,2 @@
+ethRPC: "ws://eth-config-file:8545"
```
SEJeff, line 1: How about putting the eth and solana contracts in this configuration file as well?

A comment at the top of this denoting you can use flags, env vars, or this config would also be nice for those not fully familiar with wormhole.

SEJeff (approved):
A few small nits. Looks good!

docs/operations.md
```diff
@@ -300,3 +300,48 @@ docker run \
...
 spy --nodeKey /node.key --spyRPC "[::]:7073" --network /wormhole/mainnet/2 --bootstrap /dns4/wormhole-v2-mainnet-bootstrap.xlabs.xyz/udp/8999/quic/p2p/12D3KooWNQ9tVrcb64tw6bNs2CaNrUGPM7yRrKvBBheQ5yCyPHKC,/dns4/wormhole.mcf.rocks/udp/8999/quic/p2p/12D3KooWDZVv7BhZ8yFLkarNdaSWaB43D6UbQwExJ8nnGAEmfHcU,/dns4/wormhole-v2-mainnet-bootstrap.staking.fund/udp/8999/quic/p2p/12D3KooWG8obDX9DNi1KUwZNu9xkGwfKqTp2GFwuuHpWZ3nQruS1
 ```
+
+## Guardian Configurations
+
+Configuration files, environment variables and flags are all supported.
+
+### Config File
+
+**Location/Naming**: By default, the config file is expected to be in the `node/config` directory. The standard name for the config file is `guardian.yaml`. Currently there's no support for custom directory or filename yet.
```
bruce-riley, line 310: Again, `guardiand.yaml` would be better.

SEJeff (approved):
Fantastic change! I do agree with @bruce-riley on the name for the config.

bruce-riley (approved):

SEJeff (approved):

evan-gray (approved):
reviewed devnet changes

PR #3711, opened 2024-01-12
Node/CCQ: Unmarshal should check for extra bytes

bruce-riley:
The CCQ `Unmarshal` methods should fail if someone slips extra bytes at the end of the message. Not sure if this is really an exploitable issue, but it seems like a reasonable safety.


evan-gray (approved):

SEJeff (approved):
Very good defense.

djb15 (approved):
Nice! Lgtm

PR #3715, opened 2024-01-16
Node: Always cut over to quic-v1

bruce-riley:
Now that mainnet has officially cut over to quic-v1, we no longer need to check the cutover time on startup. Instead we will just always convert the bootstrap peers and listen addresses to use `/quic-v1/`.

We want to keep doing that so that the guardians and other users don't have to change their configs.

Question: Should I update `operations.md` to reference `/quic-v1`, or leave it as is?

### DIFF
--- a/node/pkg/p2p/cutover.go
+++ b/node/pkg/p2p/cutover.go
@@ -1,112 +1,19 @@
 package p2p
 
 import (
-	"fmt"
 	"strings"
-	"time"
-
-	"go.uber.org/zap"
 )
 
-// The format of this time is very picky. Please use the exact format specified by cutOverFmtStr!
-const mainnetCutOverTimeStr = "2024-01-16T15:00:00-0000"
-const testnetCutOverTimeStr = "2023-11-07T14:00:00-0000"
-const devnetCutOverTimeStr = "2022-12-31T23:59:59-0000"
-const cutOverFmtStr = "2006-01-02T15:04:05-0700"
-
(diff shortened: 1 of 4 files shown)


node/pkg/p2p/cutover.go
```diff
@@ -1,112 +1,19 @@
...
+// cutOverAddressPattern updates the address patterns. It assumes that the string is valid.
 func cutOverAddressPattern(pattern string) string {
-	if shouldCutOver() {
-		if !strings.Contains(pattern, "/quic-v1") {
-			// These patterns are hardcoded so we are not worried about invalid values.
-			pattern = strings.ReplaceAll(pattern, "/quic", "/quic-v1")
-		}
+	if !strings.Contains(pattern, "/quic-v1") {
+		// These patterns are hardcoded so we are not worried about invalid values.
+		pattern = strings.ReplaceAll(pattern, "/quic", "/quic-v1")
```
SEJeff, line 16: Seems like we should error out now that the cut is official or at a minimum log that the operator has an invalid configuration. What do you think?

SEJeff (approved):

node/pkg/p2p/cutover.go
evan-gray: I don't think we should error on these. We know they should have been cutover, we can be smart enough to not break a production system. I could go either way on even logging it.

evan-gray (approved):

PR #3716, opened 2024-01-16
Node upgrade libp2p

bruce-riley:
This PR upgrades the version of `libp2p` used by the guardian in preparation for upgrading to go 1.21.

Note that the `Pretty` method in `peer.ID` has be removed and the docs say to use `String` instead.


djb15 (approved):
I'm assuming the old `Compare`/`SortFunc` logic worked differently and that it wasn't a bug? New changes look good

SEJeff:
This makes me think of #3631. Perhaps we can do both of these PRs? I had to unfortunately update almost everything due to that one dep requiring updating all of the open telemetry stuff which is an almost impossible maze of deps.

SEJeff (approved):

bruce-riley:
Closing this PR in favor of #3631.

PR #3731, opened 2024-01-23
Set Sui package manifest to pass source verification

unmaykr-aftermath:
Update the Sui package manifest (`Move.toml`) so that it passes source verification against the package published on Sui's testnet.

This ensures that downstream Sui Move packages that have this version of the `Wormhole` package as a dependency can verify that exact same code is published on the Sui Testnet via `sui client verify-source --skip-source --verify-deps`.

Instructions for verifying the source code have been added to the README.

If the PR is approved, it would be nice to tag the resulting commit so that other Sui Move packages can declare this as a dependency as follows.
```
[dependencies.Wormhole]
git = "https://github.com/wormhole-foundation/wormhole.git"
subdir = "sui/wormhole"
rev = "<tag-name>"
```

I have ensured that the package still builds and tests pass:
```
sui move build
sui move test
```


sui/wormhole/Move.lock
```diff
@@ -2,19 +2,26 @@
...
+deps_digest = "F8BBB0CCB2491CA29A3DF03D6F92277A4F3574266507ACD77214D37ECA3F3082"
 
 dependencies = [
   { name = "Sui" },
 ]
 
 [[move.package]]
 name = "MoveStdlib"
-source = { git = "https://github.com/MystenLabs/sui.git", rev = "09b2081498366df936abae26eea4b2d5cafb2788", subdir = "crates/sui-framework/packages/move-stdlib" }
+source = { git = "https://github.com/MystenLabs/sui.git", rev = "framework/testnet", subdir = "crates/sui-framework/packages/move-stdlib" }
```
evan-gray, line 14: might be best to keep these dependencies pinned to a commit hash

sui/wormhole/Move.toml
```diff
@@ -1,14 +1,12 @@
 [package]
 name = "Wormhole"
 version = "0.2.0"
+published-at = "0xcc029e2810f17f9f43f52262f40026a71fbdca40ed3803ad2884994361910b7e"
```
evan-gray, line 4: The address I see in the testnet branch is `0xf47329f4344f3bf0f8e436e2f7b485466cff300f12a166563995d3888c296a94`

sui/wormhole/README.md
```diff
@@ -23,3 +23,10 @@ TODO: up to date notes on where and how we use dynamic fields.
 
 Sui currently does not have fine-grained timestamps, so we use
 `tx_context::epoch(ctx)` in place of on-chain time in seconds.
+
+# Source verification
+
+Verify that the Move sources here match the package published on-chain by running
+`sui client verify-source --verify-deps`
+
+Make sure that `sui client active-env` is connected to a Sui testnet RPC.
```
evan-gray, line 32: Is this only for testnet and not mainnet?

evan-gray (changes requested):
Source verification seems like a positive step, but I'd want to make sure all of these fields are set correctly. There was also recent work on updating the dependencies in #3803 so perhaps this could be revisited after that is merged and the `sui/mainnet` and `sui/testnet` branches are updated similarly.

Testing branches for those are here:
https://github.com/wormhole-foundation/wormhole/tree/sui-upgrade-mainnet
https://github.com/wormhole-foundation/wormhole/tree/sui-upgrade-testnet

sui/wormhole/README.md
unmaykr-aftermath: No, the procedure is general whether you're verifying mainnet or testnet code. But ofc, if one checked-out the testnet branch, a testnet RPC should be used. I may update the description here to be more general

sui/wormhole/Move.lock
unmaykr-aftermath: We have to take into account that `Sui` (along with `MoveStdlib` at "0x1"
AFAIK) is a special package which can be upgraded onchain 'in-place'
(unlike regular Move package upgrades on Sui that in effect create a new
package with a new address.)

The advantage of using `framework/testnet` is that the compiler will
automatically fetch the `Sui` package that's currently on testnet, hence
source verification will succeed.
```
$ sui client verify-source --verify-deps
UPDATING GIT DEPENDENCY https://github.com/MystenLabs/sui.git
INCLUDING DEPENDENCY Sui
INCLUDING DEPENDENCY MoveStdlib
BUILDING Wormhole
// ...
Source verification succeeded!
```

However, let's say we published the package in the past and set `Sui`
to `testnet-v1.17.1` in the manifest. Verification could fail in the
future to to upgrades to the `Sui` package.
```
$ sui client verify-source --verify-deps
FETCHING GIT DEPENDENCY https://github.com/MystenLabs/sui.git
INCLUDING DEPENDENCY Sui
INCLUDING DEPENDENCY MoveStdlib
BUILDING Wormhole
// ...
Multiple source verification errors found:

- Local dependency did not match its on-chain version at 0000000000000000000000000000000000000000000000000000000000000002::Sui::bls12381
- Local version of dependency 0000000000000000000000000000000000000000000000000000000000000002::group_ops was not found.
```

For all other packages, it definitely makes sense to pin them to a
commit/tag.

sui/wormhole/Move.toml
unmaykr-aftermath: That's interesting. Indeed, I changed `published-at` and `wormhole` to
`0xf47329f4344f3bf0f8e436e2f7b485466cff300f12a166563995d3888c296a94`
in `Move.toml` and source verification passed too. Begs the question of
why there are two different deployments with the exact same source code.

In fact, at Aftermath we've been using [`0xcc029e2810f17f9f43f52262f40026a71fbdca40ed3803ad2884994361910b7e`](https://suiexplorer.com/object/0xcc029e2810f17f9f43f52262f40026a71fbdca40ed3803ad2884994361910b7e?network=testnet)
to verify Wormhole messages (and there's more activity in the contract
than just ours looking at the txs.)

Would be good to clarify what's the official version.

djb15:
Closing as part of cleanup. The PR can be reopened if required.

PR #3736, opened 2024-01-24
node: Added RecoverChainId governance support

kev1n-peters:


sdk/vaa/payloads.go
```diff
@@ -257,6 +264,24 @@ func (r BodyTokenBridgeUpgradeContract) Serialize() []byte {
...
 
+func (r BodyRecoverChainId) Serialize() []byte {
+	// Module
+	buf := LeftPadBytes(r.Module, 32)
+	// Action
+	var action GovernanceAction
+	if r.Module == "Core" {
+		action = ActionCoreRecoverChainId
+	} else {
+		action = ActionTokenBridgeRecoverChainId
```
bruce-riley, line 275: What about NFT bridge? If they both use the same enum, should it be named something that indicates that? Or, if NFT bridge is not supported, check for anything other than "TokenBridge" and throw an error?

bruce-riley (approved):

sdk/vaa/payloads.go
kev1n-peters: `"NFTBridge"` is supported and it has the same action as the token bridge `3`. Fwiw the other payloads in this file (e.g. `BodyTokenBridgeUpgradeContract`) also work with the `"NFTBridge"` module since the token bridge and nft bridge handle the same governance message payloads

evan-gray (approved):

PR #3743, opened 2024-01-25
Node/CCQ: Solana library not honoring min context slot

bruce-riley:
The Solana library we are using does not honor the `MinConextSlot` option on the `GetMultipleAccountsWithOpts` like it does on the `GetAccountWithOpts` call.

An [issue](https://github.com/gagliardetto/solana-go/issues/170) has been opened against the library. In the meantime, the troublesome function has been copied into the watcher and tweaked to support the parameter.

### DIFF
--- a/node/pkg/watchers/solana/ccq.go
+++ b/node/pkg/watchers/solana/ccq.go
@@ -3,6 +3,7 @@ package solana
 import (
 	"context"
 	"encoding/hex"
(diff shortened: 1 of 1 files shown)


evan-gray (approved):
wonder if there's some way to add integration tests to cover these cases to avoid regressions

panoel (approved):

PR #3747, opened 2024-01-26
Node/CCQ: Solana min context slot support

bruce-riley:
This PR adds a fast retry mechanism to the Solana query handler when the RPC call fails due to the `MinContextSlot` not being met. In that case, we retry every 200 milliseconds, attempting to hit the desired slot.

This PR also changes the query audit interval to once per second from every ten seconds so that we will be more likely to retry a query every ten seconds.

### DIFF
--- a/node/pkg/query/query.go
+++ b/node/pkg/query/query.go
@@ -25,6 +25,9 @@ const (
 	// RetryInterval specifies how long we will wait between retry intervals. This is the interval of our ticker.
 	RetryInterval = 10 * time.Second
 
+	// AuditInterval specifies how often to audit the list of pending queries.
+	AuditInterval = time.Second
+
 	// SignedQueryRequestChannelSize is the buffer size of the incoming query request channel.
 	SignedQueryRequestChannelSize = 50
 
@@ -105,7 +108,7 @@ func (qh *QueryHandler) Start(ctx context.Context) error {
 
 // handleQueryRequests multiplexes observation requests to the appropriate chain
 func (qh *QueryHandler) handleQueryRequests(ctx context.Context) error {
-	return handleQueryRequestsImpl(ctx, qh.logger, qh.signedQueryReqC, qh.chainQueryReqC, qh.allowedRequestors, qh.queryResponseReadC, qh.queryResponseWriteC, qh.env, RequestTimeout, RetryInterval)
+	return handleQueryRequestsImpl(ctx, qh.logger, qh.signedQueryReqC, qh.chainQueryReqC, qh.allowedRequestors, qh.queryResponseReadC, qh.queryResponseWriteC, qh.env, RequestTimeout, RetryInterval, AuditInterval)
 }
 
 // handleQueryRequestsImpl allows instantiating the handler in the test environment with shorter timeout and retry parameters.
@@ -120,6 +123,7 @@ func handleQueryRequestsImpl(
 	env common.Environment,
 	requestTimeoutImpl time.Duration,
 	retryIntervalImpl time.Duration,
+	auditIntervalImpl time.Duration,
 ) error {
 	qLogger := logger.With(zap.String("component", "ccqhandler"))
 	qLogger.Info("cross chain queries are enabled", zap.Any("allowedRequestors", allowedRequestors), zap.String("env", string(env)))
@@ -165,7 +169,7 @@ func handleQueryRequestsImpl(
 		}
 	}
 
-	ticker := time.NewTicker(retryIntervalImpl)
+	ticker := time.NewTicker(auditIntervalImpl)
 	defer ticker.Stop()
 
 	for {
--- a/node/pkg/query/query.go
+++ b/node/pkg/query/query.go
@@ -22,7 +22,7 @@ const (
 	// RequestTimeout indicates how long before a request is considered to have timed out.
 	RequestTimeout = 1 * time.Minute
 
-	// RetryInterval specifies how long we will wait between retry intervals. This is the interval of our ticker.
+	// RetryInterval specifies how long we will wait between retry intervals.
 	RetryInterval = 10 * time.Second
 
 	// AuditInterval specifies how often to audit the list of pending queries.
--- a/node/pkg/query/query_test.go
+++ b/node/pkg/query/query_test.go
@@ -36,6 +36,7 @@ const (
 	// Speed things up for testing purposes.
 	requestTimeoutForTest = 100 * time.Millisecond
 	retryIntervalForTest  = 10 * time.Millisecond
+	auditIntervalForTest  = 10 * time.Millisecond
 	pollIntervalForTest   = 5 * time.Millisecond
 )
 
@@ -436,7 +437,7 @@ func createQueryHandlerForTestWithoutPublisher(t *testing.T, ctx context.Context
 
 	go func() {
 		err := handleQueryRequestsImpl(ctx, logger, md.signedQueryReqReadC, md.chainQueryReqC, ccqAllowedRequestersList,
-			md.queryResponseReadC, md.queryResponsePublicationWriteC, common.GoTest, requestTimeoutForTest, retryIntervalForTest)
+			md.queryResponseReadC, md.queryResponsePublicationWriteC, common.GoTest, requestTimeoutForTest, retryIntervalForTest, auditIntervalForTest)
 		assert.NoError(t, err)
 	}()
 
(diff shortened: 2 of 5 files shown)


node/pkg/query/query.go
```diff
@@ -25,6 +25,9 @@ const (
 	// RetryInterval specifies how long we will wait between retry intervals. This is the interval of our ticker.
```
panoel, line 25: But this is no longer the interval of the ticker (line 172)?

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +193,109 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+		w.ccqLogger.Error("failed to parse for min context slot error", zap.Error(err))
+		return false
+	}
+
+	if !isMinContext {
+		return false
+	}
+
+	// Estimate how far in the future the requested slot is, assuming a slot time of 400 ms.
+	msInTheFuture := (req.MinContextSlot - currentSlot) * 400
```
evan-gray, line 220: should this be `futureSlotEstimate := (req.MinContextSlot - currentSlot) * 400 * time.Millisecond` and maybe even have `400 * time.Millisecond` as a const somewhere like `ESTIMATED_SLOT_TIME`

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +193,109 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+
+	// If the requested slot is more than ten seconds in the future, use the regular retry mechanism.
+	if msInTheFuture > 10000 {
```
evan-gray, line 223: then this could be `futureSlotEstimate > RetryInterval`?

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +193,109 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+		)
+		return false
+	}
+
+	// Kick off the retry after a short delay.
+	go w.ccqSleepAndRetryAccountQuery(ctx, queryRequest, req, requestId, currentSlot, giveUpTime, log)
+	return true
+}
+
+const CCQ_FAST_RETRY_INTERVAL = 200
```
evan-gray, line 238: should this be `CCQ_FAST_RETRY_INTERVAL = 200 * time.Millisecond` here insetad of line 251 so the const is more obvious? or does that not log well

node/pkg/query/query_test.go
```diff
@@ -36,6 +36,7 @@ const (
 	// Speed things up for testing purposes.
 	requestTimeoutForTest = 100 * time.Millisecond
 	retryIntervalForTest  = 10 * time.Millisecond
+	auditIntervalForTest  = 10 * time.Millisecond
```
panoel, line 39: Did you try a test scenario where the audit and retry intervals are different just in case there is some weird interaction?

node/pkg/watchers/solana/ccq.go
```diff
@@ -37,7 +41,8 @@ func (w *SolanaWatcher) ccqHandleQuery(ctx context.Context, queryRequest *query.
 
 	switch req := queryRequest.Request.Query.(type) {
 	case *query.SolanaAccountQueryRequest:
-		w.ccqHandleSolanaAccountQueryRequest(ctx, queryRequest, req)
+		giveUpTime := start.Add(query.RetryInterval).Add(-250 * time.Millisecond)
```
evan-gray, line 44: thoughts on `start.Add(query.RetryInterval).Add(-CCQ_FAST_RETRY_INTERVAL)` or something? wondering what the intention with the `-250` is so maybe we could make that more clear :slightly_smiling_face:

panoel, line 44: Does the RetryInterval need to be > 250mS?

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +193,109 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+	// Kick off the retry after a short delay.
+	go w.ccqSleepAndRetryAccountQuery(ctx, queryRequest, req, requestId, currentSlot, giveUpTime, log)
+	return true
+}
+
+const CCQ_FAST_RETRY_INTERVAL = 200
+
+// ccqSleepAndRetryAccountQuery does a short sleep and then initiates a retry.
+func (w *SolanaWatcher) ccqSleepAndRetryAccountQuery(ctx context.Context, queryRequest *query.PerChainQueryInternal, req *query.SolanaAccountQueryRequest, requestId string, currentSlot uint64, giveUpTime time.Time, log bool) {
+	if log {
```
evan-gray, line 242: if I'm reading correctly, this doesn't log the first time around, but I'm not sure why it wouldn't log all the time - wouldn't it be the only log if the initial attempt was a min slot error?

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +193,109 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+			zap.String("requestId", requestId),
+			zap.Uint64("currentSlot", currentSlot),
+			zap.Uint64("minContextSlot", req.MinContextSlot),
+			zap.Int("retryInterval", CCQ_FAST_RETRY_INTERVAL),
+		)
+	}
+
+	time.Sleep(CCQ_FAST_RETRY_INTERVAL * time.Millisecond)
+
+	if log {
```
evan-gray, line 253: same, seems like this is worthwhile to log even on the first retry, but I could be missing something

evan-gray (commented):
overall I think this is an improvement and the structure is good, just a couple nits

node/pkg/watchers/solana/ccq.go
panoel: If RetryInterval < 250mS, then ccqCheckForMinSlotContext() will give up and return false and just not do a retry?  Sounds fair?

node/pkg/query/query_test.go
bruce-riley: No, I didn't, but this behavior is exactly the same as before.

node/pkg/watchers/solana/ccq.go
bruce-riley: No, it only logs the first time (!retry is passed in above).

node/pkg/watchers/solana/ccq.go
bruce-riley: Same response as above.

panoel (approved):

node/pkg/watchers/solana/ccq.go
bruce-riley: Added a unit test to ensure this.

bruce-riley: Created a const and added a comment.

panoel (dismissed):

sdk/js-query/src/query/solana.test.ts
```diff
@@ -27,6 +27,7 @@ const ENV = "DEVNET";
 const SERVER_URL = CI ? "http://query-server:" : "http://localhost:";
 const CCQ_SERVER_URL = SERVER_URL + "6069/v1";
 const QUERY_URL = CCQ_SERVER_URL + "/query";
+const SOLANA_NODE_URL = CI ? "http://solana-devnet" : "http://localhost:8899";
```
evan-gray, line 30: I think this needs to be `const SOLANA_NODE_URL = CI ? "http://solana-devnet:8899" : "http://localhost:8899";`

node/pkg/watchers/solana/ccq.go
```diff
@@ -182,6 +205,116 @@ func (w *SolanaWatcher) ccqHandleSolanaAccountQueryRequest(ctx context.Context,
...
+		w.ccqLogger.Error("failed to parse for min context slot error", zap.Error(err))
+		return false
+	}
+
+	if !isMinContext {
+		return false
+	}
+
+	// Estimate how far in the future the requested slot is, using our estimated slot time.
+	futureSlotEstimate := time.Duration(req.MinContextSlot-currentSlot) * CCQ_ESTIMATED_SLOT_TIME
```
evan-gray, line 241: the go time stuff always looks strange to me :laughing:

evan-gray (approved):

panoel (approved):

evan-gray (approved):

PR #3748, opened 2024-01-26
ethereum: PUSH0 test

djb15:
Checks that we don't inadvertently build contracts with compiler versions that support PUSH0 opcodes whilst support is spotty.

It's brittle but it works whilst this is a transient state of play.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
           node-version: "16"
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
-      - run: cd ethereum && make test
+      - run: cd ethereum && make test && make test-push0
 
   ethereum-upgrade:
     runs-on: ubuntu-latest
--- a/ethereum/Makefile
+++ b/ethereum/Makefile
@@ -85,5 +85,10 @@ test-identifiers: dependencies
 test-forge: dependencies
 	forge test --no-match-test .*_KEVM # ignore KEVM tests (see PROOFS.md)
 
+# Verify that the contracts do not include PUSH0 opcodes
+test-push0: dependencies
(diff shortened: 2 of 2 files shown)


ethereum/Makefile
```diff
@@ -85,5 +85,10 @@ test-identifiers: dependencies
 test-forge: dependencies
 	forge test --no-match-test .*_KEVM # ignore KEVM tests (see PROOFS.md)
 
+# Verify that the contracts do not include PUSH0 opcodes
+test-push0: dependencies
+	forge build --extra-output evm.bytecode.opcodes
+	@if grep -qr --include \*.json PUSH0 ./build-forge; then echo "Contract uses PUSH0 instruction"; exit 1; fi
```
kcsongor, line 91: would be nicer to echo to stderr here

kcsongor (dismissed):
very cool!

kcsongor (approved):

a5-pickle (approved):

.github/workflows/build.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           node-version: "16"
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
-      - run: cd ethereum && make test
+      - run: cd ethereum && make test && make test-push0
```
evan-gray, line 74: maybe this could have just been put under `test` as well?

evan-gray (approved):

.github/workflows/build.yml
djb15: Yeah this was deliberate actually to avoid coupling these tests together. In the future when we do want to support PUSH0 instructions in our deployments we can just edit this line back to what it was, but still keep the command around in the Makefile available for use

PR #3758, opened 2024-01-30
Node/CCQServer: Quorum not met

bruce-riley:
This PR changes the CCQ proxy server so that it will give up on a query when the number of matching responses plus the potential number of outstanding responses would not reach quorum. When that happens, it returns a "quorum not met" error to the client.


panoel (approved):

evan-gray (approved):
nice changes!

PR #3412, opened 2023-10-02
Update 0008_batch_messaging.md

hydrogenbond007:
added solana batch

PR #3759, opened 2024-01-30
Remove batchVAA remnants

evan-gray:
This PR removes the dead batchVAA feature related code. A summary of the rationale shared amongst contributors was

> During the implementation of the guardian component of batch VAAs we realized a design limitation that we previously missed.

> Due to safety features like governor and accounting, individual VAAs, that might be part of a batch, may be held back indefinitely making the whole batch fail. Batches were intended to be atomic but still allow relatively loose coupling. As a single component may now make a whole batch fail though, it breaks delivery / VAA production guarantees and would require a significant redesign (e.g. removal of the atomicity guarantees) additionaly a lot of complexity would be added to make governor/accounting batch-aware (massive unexpected explosion of complexity).

> Batch VAAs were mainly designed for 2 reasons: A: Cost reduction when sending multiple messages B: allowing interoperability through atomic batches of VAAs particularly to make generic relaying easier.

> A: can be addressed by message accumulators that can give similar levels of cost reduction with less code-complexity
B: Generic relayers have already been re-designed without batch VAAs achieving similar functionality. 

Note to reviewers:
The `node/pkg/proto` code is generated and verified by the action, but please check that the `.proto` removals themselves. I mostly just searched for `BatchVAA` and tried to ensure that all of the code that had been previously merged in related to this feature has since been removed.

Closes: #3412


barnjamin (approved):

whitepapers/0009_guardian_key.md
```diff
@@ -1,37 +1,39 @@
...
 
 Rationale
-* Gossip messages cannot be confused with other gossip messages because the message type prefix is prepended.
-* Gossip messages cannot be confused with observations because observations utilize a double-Keccak256 and the payload is enforced to be `>=34` bytes.
-* v2 VAAs cannot be confused as v1 VAAs because their payload when parsed as a v1 VAA is only 33 bytes, which does not constitute a valid observation.
-* v1 VAAs cannot be confused as v2 VAAs because observations are longer than 33 bytes and hence do not constitute a valid v2 VAA body.
 
+- Gossip messages cannot be confused with other gossip messages because the message type prefix is prepended.
+- Gossip messages cannot be confused with observations because observations utilize a double-Keccak256 and the payload is enforced to be `>=34` bytes.
+- v2 VAAs cannot be confused as v1 VAAs because their payload when parsed as a v1 VAA is only 33 bytes, which does not constitute a valid observation.
```
SEJeff, line 38:
```suggestion
```

After this PR, version 2 VAAs no longer exist.

SEJeff (approved):

whitepapers/0009_guardian_key.md
evan-gray: thanks, removed :slightly_smiling_face:

SEJeff (approved):

bruce-riley (approved):

barnjamin (approved):

PR #3760, opened 2024-01-30
Generalize accountant parsing 

RahulMaganti47:
Currently, the global accountant supports tracking of balances for the wrapped asset standard. While the logic for tracking balances and checking the relevant invariants is generalizable, the parsing logic is not (e.g relaying transfers that do not go through AR may suffer from a double emitter problem).


evan-gray:
As far as I understand, this PR is superseded by #3805

PR #3763, opened 2024-02-01
Constants: Add Dymension to cosmwasm chains

barnjamin:
Adds Dymension chain to constants, requested by @nik-suri


sdk/js/src/cosmwasm/address.ts
```diff
@@ -35,6 +36,7 @@ export const isNativeDenomNeutron = (denom: string) => denom === "untrn";
 export const isNativeDenomCelestia = (denom: string) => denom === "utia";
 export const isNativeDenomStargaze = (denom: string) => denom === "ustars";
 export const isNativeDenomSeda = (denom: string) => denom === "aseda";
+export const isNativeDenomDymension = (denom: string) => denom === "dym";
```
evan-gray, line 39: I think this is `udym`, according to the readme https://github.com/dymensionxyz/dymension?tab=readme-ov-file#initializing-dymd

evan-gray (changes requested):

sdk/js/src/cosmwasm/address.ts
barnjamin: good spot, poking around i found [here](https://github.com/dymensionxyz/dymension/blob/9fa05e7f1f0dfc9d6d83ecf3a309c24eda203bee/scripts/src/genesis_config_commands.sh#L82) is `adym` too, in code search on that repo `udym` is always in some `_test` file

nik-suri: Confirmed that this should be `adym` with the Dymension team.

barnjamin: updated this

evan-gray (approved):

evan-gray:
> LGTM, thanks @barnjamin !
> 
> Can we also add this chain ID to the rust SDK constants in `chain.rs`?

maybe save this for a separate PR so as to bother those code owners with that change 🙂

SEJeff (approved):

PR #3766, opened 2024-02-05
Node/CCQServer: Metrics publish exiting

bruce-riley:


evan-gray (approved):
agree, should be safe to simply try again at the timer and not error out.

PR #3768, opened 2024-02-06
CCQ/EVM: Testing library for building responses

bruce-riley:


ethereum/forge-test/query/QueryTest.sol
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: Apache 2
+
+// forge test --match-contract QueryTest
+
+pragma solidity ^0.8.4;
+
+// @dev QueryTest is a library to build Cross Chain Query (CCQ) responses for testing purposes.
+abstract contract QueryTest {
```
jaydenwindle, line 8: This might be more useful as a library rather than an abstract contract (since all functions are pure). That way you could use it within deployed contracts as well if needed.

bruce-riley: Converted it to a library.

bruce-riley: Converted it to a library.

ethereum/forge-test/query/QueryResponse.t.sol
```diff
@@ -74,15 +75,18 @@ contract TestQueryResponse is Test {
...
+            _queryRequestVersion,
+            _queryRequestNonce,
+            _numPerChainQueries,
+            _perChainQueries
+        );
+        return QueryTest.buildQueryResponseBytes(
             _version,
             _senderChainId,
             _signature,
             _queryRequestLen,
```
djb15, line 88: Feels weird passing in the length here. Can we instead fetch the length from `queryRequest` we've just built?

djb15 (commented):
What do you think about all the `buildXXXBytes` methods also returning the length of the bytes being returned?

There are a lot of lengths passed around that presumably a user has to manually fetch themselves with `BYTES.length`.

If we did this you can disregard my inline comment

djb15 (approved):
Changes look good to me! Byte array lengths aren't being checked but I think we'll be hard pressed to be using 4GB byte arrays

gator-boi (approved):

PR #3770, opened 2024-02-07
Fixed typo in docs

shanama1:
Life cycle is two separate words.


evan-gray:
It is not incorrect to spell it as one word. I don't believe anyone would be confused by this. e.g. https://aws.amazon.com/what-is/sdlc/ , https://www.oed.com/search/advanced/Quotations?textTermText0=lifecycle&textTermOpt0=QuotText

PR #3773, opened 2024-02-08
Update evm.ts

kolen44:
Simplify checking for the presence of a value: Instead of using the ?? operator to check the value of _rpc, use the || operator.


evan-gray:

PR #3775, opened 2024-02-08
Update injective.ts

kolen44:
This version of the code uses an if-else construct instead of a nested switch. This makes the code more readable and easy to extend.


evan-gray:

PR #3776, opened 2024-02-08
Update algorand.ts

kolen44:
This code creates an ALGORAND_HOST object by initializing the algodToken and algodPort properties based on the network value. These properties are then used when creating an Algodv2 object.


evan-gray:

PR #3777, opened 2024-02-08
Update injective.ts

kolen44:
The following changes have been made:
- Added variable declaration decodedAddress to clearly indicate that the address must be decoded before assigning a value to results[cname];
- Moved the address decoding line to a separate decodedAddress variable to improve readability and code support;


evan-gray:

PR #3778, opened 2024-02-08
Update main.go

kolen44:
In improved code:

- Instead of using cmd.PrintErrln and os.Exit, we use log.Fatal to improve readability and provide consistent error handling.
- We use hex.DecodeString to decode the message from hexadecimal format.

### DIFF
--- a/clients/eth/main.go
+++ b/clients/eth/main.go
@@ -52,24 +52,24 @@ func init() {
 
 func postMessage(cmd *cobra.Command, args []string) {
 	nonce, err := strconv.ParseUint(args[0], 10, 32)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


clients/eth/main.go
```diff
@@ -52,24 +52,24 @@ func init() {
...
+  	}
 
-	message := common.Hex2Bytes(args[2])
+  	message, err := hex.DecodeString(args[2])
+  	if err != nil {
+    	    log.Fatal("Could not decode message from hex:", err)
+  	}
 
-	ethC, err := getEthClient()
+  	ethClient, err := getEthClient()
```
evan-gray, line 69: wouldn't changing this variable break line 74?

evan-gray:

PR #3786, opened 2024-02-14
node: support overriding the p2p advertised ip

SEJeff:
When behind a private network such as in the case of a guardian running in Kubernetes, or behind a NAT, the default p2p setup doesn't work. It looks at all addresses that it is listening on and advertises them on p2p as the addresses for contacting the guardian. This patch is the first step towards allowing specifying a custom ip address to be advertised on the gossip p2p network.

For example, a guardian running inside kubernetes can post a reserved ip address of the incoming ingress/load balancer that sends the traffic in to their guardian.

To use this feature, just add `--gossipAdvertiseAddress $externally_reachable_ip_address` to the guardiand flags.

### DIFF
--- a/node/pkg/node/options.go
+++ b/node/pkg/node/options.go
@@ -39,7 +39,7 @@ type GuardianOption struct {
 
 // GuardianOptionP2P configures p2p networking.
 // Dependencies: Accountant, Governor
-func GuardianOptionP2P(p2pKey libp2p_crypto.PrivKey, networkId string, bootstrapPeers string, nodeName string, disableHeartbeatVerify bool, port uint, ccqBootstrapPeers string, ccqPort uint, ccqAllowedPeers string, ibcFeaturesFunc func() string) *GuardianOption {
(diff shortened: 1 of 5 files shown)


node/pkg/node/options.go
```diff
@@ -52,6 +52,11 @@ func GuardianOptionP2P(p2pKey libp2p_crypto.PrivKey, networkId string, bootstrap
 				components.SignedHeartbeatLogLevel = zapcore.InfoLevel
 			}
 
+			// Add the gossip advertisement address if it was specified
+			if gossipAdvertiseAddress != "" {
```
bruce-riley, line 56: Looks like you could always just do this since the field in the component defaults to the empty string anyway.

bruce-riley (approved):
Looks good to me. Just one minor comment.

node/pkg/node/options.go
SEJeff: Fixed in e306a743d5244fe03d59c65f39a923cc1bd376a9

evan-gray (approved):

bingyuyap (approved):

PR #3797, opened 2024-02-20
node: upgrade grpc to v1.57.1

bingyuyap:
- tested with tilt up
- ci tests passes 
- also read patch notes to check if upgrading other packages could lead to other unintended effects

@bruce-riley @evan-gray wondering if there are any more tests to do to make sure a safe upgrade


bruce-riley (approved):
Looks good to me, but I would suggest that @SEJeff review as well. Also probably want to push out as a pre-release to testnet first.

SEJeff (approved):

PR #3791, opened 2024-02-16
Node/Gov: Add target to db

bruce-riley:
This PR adds the target chain and address to the governor transfers stored in the DB. It also handles reloading transfers that do not have those fields and migrating them to the new format.

PR #3798, opened 2024-02-20
Implementation of 'Flow Cancel' modifications to Governor calculations

johnsaigle:
## EDIT 

The flow cancel logic has been rewritten following Bruce's feedback. Updated design can be found in the commit message here: https://github.com/wormhole-foundation/wormhole/pull/3798/commits/21c96c19c2a60b5a4a3ce219f8a87722f45754f4

## Overview

This pull requests contains modifications to the [Governor](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0007_governor.md).
In particular, it allows a certain allow-list of incoming tokens to "pay down" the Governor usage for the destination chain.

~~**Note**: This implementation is meant to be minimally invasive to highlight the flow cancel design and mechanism. As a result, it is not very efficient. Once the overall design is clear and agreed upon, I am more than happy to refactor the code to make it more performant.~~

## Background
The motivation and design document for this change is detailed in the GitHub Discussion ["Wormhole Governor - Flow-Canceling Design Proposal (Temp Check)"](https://github.com/wormhole-foundation/wormhole/discussions/3783).

~~Note that this feature depends on and includes commits from [Bruce's pull request](https://github.com/wormhole-foundation/wormhole/pull/3791), namely the modifications to `node/pkg/db/governor.go`.~~ (The other PR was merged and this one is rebased to include it)

### Allow list
The allow-list is in the same format as other hard-coded token implementations in Wormhole, such as the ['manual token list'](https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/governor/manual_tokens.go).

Currently, the allow list contains all implementations of UDSC, USDT, and DAI that are present in the ['generated mainnet token list'](https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/governor/generated_mainnet_tokens.go)

### Flow Cancel

A 'flow cancel' transfer occurs when all of the following are true:
- An `emitter` chain is governed by the Governor
- A transfer has a `TargetChain` equal to the `emitter`
- A transfer's `Timestamp` is within the usual 24 hour window used by the Governor
- The asset in the transfer has fields `OriginChain` and `OriginAddress` equal to an entry in the allow list

#### Example scenario
- A particular `USDC` implementation minted on Ethereum. It is allow-listed. 
- A transfer of $1,000 of this asset is sent from Solana to Sui.
- Before the transfer, Sui's usage is $10,000.
- Before the transfer, Solana's usage is $0.

After the transfer is processed, Sui's new usage will be $9,000 (flow cancel) and Solana's new usage will be $1,000 (as usual).

### ~~Efficiency~~ (resolved by rewrite)

There is a [triple for-loop](https://github.com/wormhole-foundation/wormhole/pull/3798/files#diff-eb54ff24f55c72fed426ce283242bef4d6f101186712b80b6703f192cdcb0e2bR712-R726) involved in the calculations. _(For every governed chain --> for every transfer --> for every allow-listed asset ...)_

Obviously this isn't optimal. However, I wanted to implement the logic using the existing database structure and Go structs where possible so that this change is not too intrusive.

In general, anything we can do to optimize the way `Transfer`s are filtered and accessed would be a big improvement. While there are 3 loops involved here, the token list and `gov.chains` are small, fixed numbers. Our real issue is processing the transfers.

One clear improvement would be to make a struct similar to `chainEntry` that indexes based on target chain, rather than emitter chain. This would allow for more efficient lookups when calculating the net flows into a chain via the `FlowCancellingTransfersForChain()`. 

Another possible approach is to cache the Governor usage calculations, perhaps in the `ChainGovernor`.

Some limitations:
- The `sum` values in governor.go as well as the Value of transfers are unsigned ints.
- The governor usage is dynamically calculated and not stored in the database
- `chainEntry`'s are indexed by source (emitter) chains, not destination (target).

Another area for improvement: `FlowCancelTokenList()` is a function, not a field of `ChainGovernor` but it probably could be (similarly to its fields `tokens` and `tokensByCoinGeckoId`).

### Contents of the allow-list
Are there any assets that should be removed? ~~I wanted to include all stablecoins from the generated mainnet tokens list and to avoid making editiorial decisions about specific tokens.~~ Felix suggested in a review to use only the assets originating from Ethereum and Solana chains for an initial rollout.

~~That said, there are some likely candidates for removal; the following appear to have de-pegged:~~

- ~~https://www.coingecko.com/en/coins/bridged-tether-orbit-bridge
`{chain: 13, addr: "000000000000000000000000cee8faf64bb97a73bb51e115aa89c17ffa8dd167", symbol: "oUSDT", coinGeckoId: "orbit-bridge-klaytn-usd-tether", decimals: 6, price: 0.490247},  // Addr: 0xcee8faf64bb97a73bb51e115aa89c17ffa8dd167, Notional: 1.5499198124759999Z}`~~

- ~~https://www.coingecko.com/en/coins/klaytn-dai
`{chain: 13, addr: "0000000000000000000000005c74070fdea071359b86082bd9f9b3deaafbe32b", symbol: "KDAI", coinGeckoId: "klaytn-dai", decimals: 18, price: 0.490008}, // Addr: 0x5c74070fdea071359b86082bd9f9b3deaafbe32b, Notional: 0.00980016`~~

#### Unit tests

I added unit tests to cover the new functionality. I'm happy to add more if others have suggestions.

#### Dynamic test coverage

I got as far as running Tilt on my machine but I'm not exactly clear on the best way to run smoke tests for this feature.

#### Benchmarking

As we refine these changes, it would be beneficial to benchmark the Flow Cancelling calculations with a large number of transfers as it is likely to introduce a performance hit.


node/pkg/governor/flow_cancel_tokens.go
```diff
@@ -0,0 +1,57 @@
+package governor
+
+// FlowCancelTokenList Returns a list of `tokenConfigEntry`s representing tokens that can 'Flow Cancel'. This means that incoming transfers
+// that use these tokens can reduce the 'daily limit' of the Governor configured for the destination chain.
```
bruce-riley, line 4: Nit: "daily usage" not "daily limit".

bruce-riley (commented):
Please see my comment above.

johnsaigle:
@bruce-riley I don't have objections to that method in general, and I looked into it following our intial discussion. The main caveat that most of the values in question (in the structs and database IIRC) are using unsigned types, so it's not trivial to modify the code to use negative values. This would also limit the effective upper bound of the values (since half of the data type would need to be reserved to represent negative values.)

It may be possible to change to e.g. `int64` instead of `uint64` but depending on how these values are used across the codebase, it may be that pulling on the thread actually ends up being a much more invasive change.

bruce-riley:
The value is in USD, so I don't think we need to worry about exceeding the max of an int64. Also, I'm not proposing changing the DB. That would need to stay what's in the actual transfer. The signed value could be stored in the transfers array in addition to the DB Transfer object (so the new payload in that array would be a struct containing the DB transfer and a signed value).

Anyway, I think it would be useful to have another face to face discussion to talk through the alternatives.

node/pkg/governor/governor.go
```diff
@@ -93,19 +94,85 @@ type (
...
+		return fmt.Errorf("Value for transfer.dbTransfer exceeds MaxInt64: %d", transfer.dbTransfer.Value)
+	}
+	// Type conversion is safe here because of the MaxInt64 bounds check above
+	if value != -int64(transfer.dbTransfer.Value) {
+		return fmt.Errorf("Transfer is invalid: transfer.value %d must equal the inverse of transfer.dbTransfer.Value %d", value, transfer.dbTransfer.Value)
+	}
+	if targetChain != ce.emitterChainId {
+		return fmt.Errorf("Flow cancel transfer TargetChain %s does not match this chainEntry %s", targetChain, ce.emitterChainId)
+	}
+	// TODO: Verify the asset?
```
djb15, line 150: What would you want to verify about the asset? Or can we get rid of this?

node/pkg/governor/governor.go
```diff
@@ -230,6 +305,19 @@ func (gov *ChainGovernor) initConfig() error {
...
+		addr, err := vaa.StringToAddress(flowCancelConfigEntry.addr)
+		if err != nil {
+			return err
+		}
+		key := tokenKey{chain: vaa.ChainID(flowCancelConfigEntry.chain), addr: addr}
+
+		// Only add flow cancelling for tokens that are already configured for rate-limiting.
+		if _, ok := gov.tokens[key]; ok {
+			gov.tokens[key].flowCancels = true
+		}
```
djb15, line 318: Should we log here if the token isn't already being tracked? Rather than just moving on silently?

Maybe in both cases we can log to debug

node/pkg/governor/governor.go
```diff
@@ -656,34 +796,83 @@ func computeValue(amount *big.Int, token *tokenEntry) (uint64, error) {
...
+	return sum, nil
+
 }
 
-func (gov *ChainGovernor) TrimAndSumValue(transfers []*db.Transfer, startTime time.Time) (uint64, []*db.Transfer, error) {
+// TrimAndSumValue iterates over a slice of transfer structs. It filters out transfers that have Timestamp values that
+// are earlier than the parameter `startTime`. The function then iterates over the remaining transfers, sums their Value,
+// and returns the sum and the filtered transfers.
+// The `transfers` slice must be sorted by Timestamp. We expect this to be the case as transfers are added to the
+// Governor in chronological order as they arrive.
```
djb15, line 843: Is this still true with the flow cancelling transfers since they come from a different chain with it's own timestamp that might drift from the timestamp of the destination chain (being flow cancelled with)?

node/pkg/governor/governor.go
```diff
@@ -701,3 +890,48 @@ func (gov *ChainGovernor) HashFromMsg(msg *common.MessagePublication) string {
 	digest := v.SigningDigest()
 	return hex.EncodeToString(digest.Bytes())
 }
+
+// Addition of two uint64 values with overflow checks
+func CheckedAddUint64(x uint64, y uint64) (uint64, error) {
+	if x == 0 {
+		return y, nil
+	}
+	if y == 0 {
```
djb15, line 899: else if?

node/pkg/governor/governor.go
johnsaigle: My thinking was that since this method adds a Flow Cancel token, maybe it could validate that the asset being added in the `transfer` has the `flowCancels` bool property set to true. Thought it might be overkill but I suppose there's no harm?

node/pkg/governor/governor.go
johnsaigle: Yeah that's a good call out. It would be helpful to know if there's a configuration issue

node/pkg/governor/governor.go
johnsaigle: Yep, the `TimeStamp` is generated by this file in the method `ProcessMsg` and `ProcessMsgForTime`; it isn't read from the transfer itself.
I'll modify the code comment here and in the db.Transfer struct to make this more clear.

node/pkg/governor/governor.go
johnsaigle: I pushed a new commit that removes this TODO and adds a comment that calling code has to do this verification. I also modified the unit tests to make sure that this verification is done.

node/pkg/governor/governor.go
```diff
@@ -606,12 +750,17 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
...
 							return nil, err
 						}
 
-						ce.transfers = append(ce.transfers, &xfer)
+						transfer, err := newTransferFromDbTransfer(&dbTransfer)
+						if err != nil {
+							return nil, err
+						}
+						ce.transfers = append(ce.transfers, transfer)
+						// TODO: Add inverse transfer to destination?
```
djb15, line 763: Yeah I think if this is a transfer that fits in the daily limit (i.e. counts towards transfers) then it should have the inverse on the destination to flow cancel as well...wdyt?

johnsaigle: @bruce-riley do you have any thoughts here? I'm not exactly sure how pending transfers should be handled

node/pkg/governor/flow_cancel_tokens.go
```diff
@@ -0,0 +1,60 @@
...
+
+// FlowCancelTokenList Returns a list of `tokenConfigEntry`s representing tokens that can 'Flow Cancel'. This means that incoming transfers
+// that use these tokens can reduce the 'daily usage' of the Governor configured for the destination chain.
+// The list of tokens was generated by grepping the file `generated_mainnet_tokens.go` for "USDC", "USDT", and "DAI".
+//
+// Only tokens that are configured in the mainnet token list should be able to flow cancel. That is, if a token is
+// present in this list but not in the mainnet token lists, it should not flow cancel.
+//
+// Note that the field `symbol` is unused. It is retained in this file only for convenience.
+func FlowCancelTokenList() []tokenConfigEntry {
```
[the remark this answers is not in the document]
johnsaigle: Sure, I have no problem with using a very small number especially just as we're getting started.

node/pkg/governor/governor_test.go
```diff
@@ -1936,3 +2186,84 @@ func TestPendingTransferWithBadPayloadGetsDroppedNotReleased(t *testing.T) {
 	_, exists = gov.msgsSeen[gov.HashFromMsg(&msg2)]
 	assert.False(t, exists)
 }
+
+func TestCheckedAddUint64HappyPath(t *testing.T) {
```
SEJeff, line 2190: We can come back later and replace these with [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) but these are nice as is.

SEJeff (approved):
Lots of nits, mostly style stuff since I know you're not a gopher natively, but this is good. I couldn't find anything obviously bad and this is definitely a big improvement.

claudijd (approved):

node/pkg/governor/governor_monitoring.go
```diff
@@ -242,20 +241,34 @@ func (gov *ChainGovernor) resetReleaseTimerForTime(vaaId string, now time.Time)
...
-			sum += t.Value
+		if t.dbTransfer.Timestamp.Before(startTime) {
+			continue
+		}
+		checkedSum, err := CheckedAddInt64(sum, t.value)
+		if err != nil {
+			// We have to stop and return an error here (rather than saturate, for example). The
+			// transfers are not sorted by value so we can't make any guarantee on the final value
+			// if we hit the upper or lower bound. We don't expect this to happen in any case
+			// because we don't expect this much value to flow between two chains in a 24h period.
```
panoel, line 261: Nit: is the word 'much' needed in this sentence?

panoel (approved):

node/pkg/governor/governor.go
bruce-riley: I think that makes sense.

node/pkg/governor/governor.go
```diff
@@ -230,6 +306,26 @@ func (gov *ChainGovernor) initConfig() error {
 		}
 	}
 
+	for _, flowCancelConfigEntry := range flowCancelTokens {
+		addr, err := vaa.StringToAddress(flowCancelConfigEntry.addr)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
bruce-riley: FWIW, I don't like ignoring the error return. If you add tests so this would only fail as the result of a program bug, then this could panic.

node/pkg/governor/governor.go
```diff
@@ -442,7 +550,35 @@ func (gov *ChainGovernor) ProcessMsgForTime(msg *common.MessagePublication, now
...
+	//	the summed Value of Sui to decrease.
+	emitterChainEntry.transfers = append(emitterChainEntry.transfers, transfer)
+
+	// Add inverse transfer to destination chain entry if this asset can cancel flows.
+	key := tokenKey{chain: msg.EmitterChain, addr: msg.EmitterAddress}
+	tokenEntry := gov.tokens[key]
+	if tokenEntry != nil {
+		// Mandatory check to ensure that the token should be able to reduce the Governor limit.
+		if tokenEntry.flowCancels {
+			destinationChainEntry := gov.chains[payload.TargetChain]
```
bruce-riley, line 574: Nil pointer check please.

node/pkg/governor/governor.go
```diff
@@ -656,34 +805,84 @@ func computeValue(amount *big.Int, token *tokenEntry) (uint64, error) {
...
 		} else {
-			sum += t.Value
+			checkedSum, err := CheckedAddInt64(sum, t.value)
+			if err != nil {
+				// We have to stop and return an error here (rather than saturate, for example). The
+				// transfers are not sorted by value so we can't make any guarantee on the final value
+				// if we hit the upper or lower bound. We don't expect this to happen in any case
+				// because we don't expect this much value to flow between two chains in a 24h period.
+				return 0, transfers, err
+			}
```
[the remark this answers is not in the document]
johnsaigle: This `err` should propagate to the calling code and be logged by `CheckPendingForTime`.
Did you want me to add an additional log here?

node/pkg/governor/governor.go
```diff
@@ -93,19 +94,86 @@ type (
...
+		transfers []transfer
 		pending   []*pendingEntry
 	}
 )
 
+// newTransferFromDbTransfer function    Perform a bounds check on dbTransfer.Value to ensure it can fit into int64.
+// This should always be the case for normal operation as dbTransfer.Value represents the USD value of a transfer.
+func newTransferFromDbTransfer(dbTransfer *db.Transfer) (tx transfer, err error) {
+	if dbTransfer.Value > math.MaxInt64 {
+		return tx, fmt.Errorf("Value for db.Transfer exceeds MaxInt64: %d", dbTransfer.Value)
```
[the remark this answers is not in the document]
johnsaigle: We should add this to a linting tool if possible

node/pkg/governor/governor.go
```diff
@@ -230,6 +306,26 @@ func (gov *ChainGovernor) initConfig() error {
...
+			gov.tokens[key].flowCancels = true
+		} else {
+			gov.logger.Debug("token present in flow cancel list but absent from main token list:",
+				zap.Stringer("chain", key.chain),
+				zap.Stringer("addr", key.addr),
+				zap.String("symbol", flowCancelConfigEntry.symbol),
+				zap.String("coinGeckoId", flowCancelConfigEntry.coinGeckoId),
+			)
+		}
+	}
```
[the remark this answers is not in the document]
djb15: Ah I suggested we add this error, but I actually agree with Jeff here that it would be better to assert this at CI time since that's what we do with the token list already. No harm in keeping this logging in, but having a test to assert validity would be great if it's not too painful

node/pkg/governor/governor_monitoring.go
```diff
@@ -96,14 +95,16 @@ import (
 )
 
 // Admin command to display status to the log.
-func (gov *ChainGovernor) Status() string {
+func (gov *ChainGovernor) Status() (resp string, err error) {
```
johnsaigle, line 98: I changed the function signature here because `sumValue` can theoretically overflow though in practice should never do so.

Two approaches here:
1. We can swallow the error and just return 0 and log
2. I can keep pulling at this thread and propagate the error through the codebase. It might make this PR needlessly complex for such an edge-case.

Let me know what seems right.

johnsaigle: Yeah I think my preference is to not return the error. Thanks I'll issue a commit soon to change this.

node/pkg/governor/governor_monitoring.go
```diff
@@ -96,14 +95,18 @@ import (
...
 	gov.mutex.Lock()
 	defer gov.mutex.Unlock()
 
 	startTime := time.Now().Add(-time.Minute * time.Duration(gov.dayLengthInMinutes))
-	var resp string
 	for _, ce := range gov.chains {
-		valueTrans := sumValue(ce.transfers, startTime)
+		valueTrans, err := sumValue(ce.transfers, startTime)
+		if err != nil {
+			// Print an error to the logs. We don't want to actually return an error or otherwise stop
```
bruce-riley, line 106: Sorry for a very minor nit pick, but the comment is wrong. You're not printing to the logs, you're returning an error message.

node/pkg/governor/governor.go
```diff
@@ -170,6 +238,7 @@ func (gov *ChainGovernor) initConfig() error {
 
 	gov.dayLengthInMinutes = 24 * 60
 	configTokens := tokenList()
+	flowCancelTokens := FlowCancelTokenList()
```
bruce-riley, line 241: Something to think about. We have different lists of monitored tokens for devnet and testnet. Should we have something similar for flow cancelled tokens?

node/pkg/governor/governor.go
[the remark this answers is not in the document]
bruce-riley: This will probably happen in devnet and testnet (although we don't usually have governor enabled in those cases) since we only have a single list of flow cancelled tokens, not per-environment.

bruce-riley (dismissed):

node/pkg/governor/governor.go
johnsaigle: Yeah that seems sensible. I'll look into this today

bruce-riley (approved):

SEJeff (approved):

node/pkg/governor/governor.go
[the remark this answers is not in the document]
SEJeff: No that should be fine as-is.

SEJeff (approved):

PR #3799, opened 2024-02-21
NTT Accountant Contract

evan-gray:
This PR adds a new Global Accountant contract for Native Token Transfers. Essentially, this is a copy/paste of the Token Bridge Global Accountant with modifications for tracking networks of NTT endpoints by their locking hub instead of the singular Token Bridge network of varying origin tokens. Interfacing with the contract is almost entirely the same except

- the observation prefix was changed to `ntt_acct_sub_obsfig_00000000000000|`
- the `chain_registration` query was changed to `relayer_chain_registration` for clarity
- `all_endpoint_hubs` query was added
- `all_endpoint_peers` query was added

The additional integration test was run locally but is disabled for now until it is provisioned an additional key to avoid race conditions. I expect this can be done in #3800 when the tests are expanded to cover observations as well, since it adds several keys already.


tampz85 (approved):

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -437,59 +465,181 @@ fn handle_accountant_governance_vaa(
...
+        // this code will only add peers for which it can assure their hubs match so one less key can be loaded on transfers
+        let message = EndpointRegister::deserialize(&payload)?;
+
+        let peer_hub_key =
+            ENDPOINT_TO_HUB.key((message.endpoint_chain_id, message.endpoint_address.into()));
+
+        let peer_hub = peer_hub_key
+            .may_load(deps.storage)
+            .context("failed to load peer hub")?
+            .ok_or(ContractError::MissingHubRegistration)?;
```
djb15, line 594: So this will fail early for any peer registration that isn't the chainid of the locking chain where the hub has been registered?

djb15 (commented):
Worth renaming "endpoints" to "transceivers" now?

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
evan-gray: the first registration for a spoke (submitted to the accountant contract), has to be with its hub. this is so that the contract can enforce that all spokes share the same hub. they do not get peered together otherwise

evan-gray:
GH bugged out on this PR, reopened as #3805

evan-gray:
attempting to reopen

evan-gray:
Due to GH bug, reopened as #3805

PR #3800, opened 2024-02-21
NTT/ Acct / Node: Guardian support

bruce-riley:


node/pkg/node/options.go
```diff
@@ -132,20 +139,23 @@ func GuardianOptionAccountant(contract string, websocket string, enforcing bool,
...
 			if enforcing {
-				logger.Info("acct: accountant is enabled and will be enforced", zap.String("component", "gacct"))
+				logger.Info("accountant is enabled and will be enforced", zap.String("component", "gacct"))
 			} else {
-				logger.Info("acct: accountant is enabled but will not be enforced", zap.String("component", "gacct"))
+				logger.Info("accountant is enabled but will not be enforced", zap.String("component", "gacct"))
+			}
+			if nttContract != "" {
+				logger.Info("NTT accountant is enabled", zap.String("component", "gacct"))
 			}
```
panoel, line 159: Did you want to add a check ```if nttWormchainConn == nil {...}```

node/pkg/accountant/accountant.go
```diff
@@ -41,26 +41,27 @@ type (
...
+	emitterKey struct {
 		emitterChainId vaa.ChainID
 		emitterAddr    vaa.Address
 	}
 
-	// tokenBridgeEntry is the payload of the map of the token bridges being monitored
-	tokenBridgeEntry struct {
-	}
+	// validEmitters is a set of supported emitter chain / address pairs. The payload is the enforcement flag.
+	validEmitters map[emitterKey]bool
```
panoel, line 51: I'm not expecting a name change here.  Just getting my bearings. Isn't this really *enforcedEmitters*?  Any emitter in a map would just assumed to be valid?

node/pkg/accountant/ntt.go
```diff
@@ -0,0 +1,193 @@
...
+	return nil
+}
+
+var WH_PREFIX = []byte{0x99, 0x45, 0xFF, 0x10}
+var NTT_PREFIX = []byte{0x99, 0x4E, 0x54, 0x54}
+
+// isNTT determines if the payload bytes are for a Native Token Transfer, according to the following implementation:
+// https://github.com/wormhole-foundation/example-native-token-transfers/blob/41ac7baae5bb0f60fff2ec87603970af39dced01/test/EndpointStructs.t.sol
+func nttIsPayloadNTT(payload []byte) bool {
+	if len(payload) < 116 {
```
panoel, line 66: What is 116?  Is that a standard NTT payload size?

bingyuyap: would be nice to have a constant defined for this!

node/pkg/accountant/ntt.go
```diff
@@ -0,0 +1,193 @@
...
+func nttParseArPayload(msgPayload []byte) (bool, [32]byte) {
+	var senderAddress [32]byte
+	reader := bytes.NewReader(msgPayload[:])
+
+	var deliveryInstruction uint8
+	if err := binary.Read(reader, binary.BigEndian, &deliveryInstruction); err != nil {
+		return false, senderAddress
+	}
+
+	if deliveryInstruction != 1 { // PAYLOAD_ID_DELIVERY_INSTRUCTION
```
bingyuyap, line 124: `PAYLOAD_ID_DELIVERY_INSTRUCTION` -> constant instead of comment?

wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
```diff
@@ -104,17 +138,35 @@ const host = ci
   : devnetConsts.chains[3104].tendermintUrlLocal;
 // TODO: have a mnemonic dedicated for this test
 const mnemonic =
-  devnetConsts.chains[3104].accounts.wormchainNodeOfGuardian0.mnemonic;
+  "quality vacuum heart guard buzz spike sight swarm shove special gym robust assume sudden deposit grid alcohol choice devote leader tilt noodle tide penalty";
```
panoel, line 141: Is this a different account than what is in the consts file?

panoel (approved):

node/pkg/accountant/accountant.go
bruce-riley: No, the map would contain all valid emitters, but some of them may not have enforcement enabled ("log only"). So this map really does contain all valid emitters.

node/pkg/accountant/ntt.go
bruce-riley: This is the offset into the message of the NTT prefix code that I need to check to see if this is a Native Token Transfer (as opposed to some other protocol message).

I added constants for these things.

wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
evan-gray: yes, only in that those accounts are on Terra while this one is on wormchain (added in this PR)

PR #3806, opened 2024-03-01
NTT Accountant Contract

evan-gray:
Replaces #3799 which GH killed and cleans up the commits.

This PR adds a new Global Accountant contract for Native Token Transfers. Essentially, this is a copy/paste of the Token Bridge Global Accountant with modifications for tracking networks of NTT endpoints by their locking hub instead of the singular Token Bridge network of varying origin tokens. Interfacing with the contract is almost entirely the same except

- the observation prefix was changed to `ntt_acct_sub_obsfig_00000000000000|`
- the `chain_registration` query was changed to `relayer_chain_registration` for clarity
- `all_endpoint_hubs` query was added
- `all_endpoint_peers` query was added

The additional integration test was run locally but is disabled for now until it is provisioned an additional key to avoid race conditions. I expect this can be done in #3800 when the tests are expanded to cover observations as well, since it adds several keys already.


cosmwasm/contracts/ntt-global-accountant/tests/query.rs
```diff
@@ -32,56 +30,56 @@ fn create_accounts(wh: &fake::WormholeKeeper, contract: &mut Contract, count: us
...
-                token_address: token_address.into(),
-                recipient_chain,
-            },
-        });
-    }
-
-    contract.submit_vaas(vaas).unwrap();
-    out
-}
+// fn create_transfers(
```
djb15, line 33: Any reason you're commenting out this test rather the editing or removing? Same with the test below?

djb15 (approved):
Spent a fair bit of time stepping through the peer registration logic this morning and it makes sense to me. The ordering seems to be well enforced and I couldn't figure out a way to get into an undefined state.

Just to double check, we don't want to support NTT deployments with multiple locking chains right? I'm not sure why an integrator would want to do that, but maybe worth being explicit in the README as I think then we could have multiple active hub keys?

cosmwasm/contracts/ntt-global-accountant/src/structs/relayer.rs
```diff
@@ -0,0 +1,273 @@
...
+            0x97, 0x1a, 0xcb, 0x22, 0x17, 0xb0, 0xdb, 0x81, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x7a, 0x0a, 0x53, 0x84, 0x77, 0x76, 0xf7, 0xe9,
+            0x4c, 0xc3, 0x57, 0x42, 0x97, 0x1a, 0xcb, 0x22, 0x17, 0xb0, 0xdb, 0x81, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xe4, 0x93, 0xcc, 0x4f,
+            0x06, 0x98, 0x21, 0x40, 0x4d, 0x27, 0x2b, 0x99, 0x4b, 0xb8, 0x0b, 0x1b, 0xa1, 0x63,
+            0x19, 0x14, 0x00,
+        ];
+        let delivery = DeliveryInstruction::deserialize(&payload).unwrap();
+
+        assert_eq!(delivery.target_chain, 10002);
```
a5-pickle, line 183: FYI: If `DeliveryInstruction` derives `PartialEq` and `Eq`, we could just assert that your `delivery` struct is equal to an expected struct. It's nicer to do it this way in case the test inadvertently misses checking a struct field.

evan-gray: great idea, thanks. done!

sdk/rust/core/src/ntt_accountant.rs
```diff
@@ -0,0 +1,424 @@
...
+}
+
+impl<'de> Deserialize<'de> for ModificationKind {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        <u8 as Deserialize>::deserialize(deserializer).map(Self::from)
+    }
+}
```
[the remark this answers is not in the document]
evan-gray: agreed and done!

cosmwasm/contracts/ntt-global-accountant/tests/query.rs
evan-gray: removed for now but left TODOs

evan-gray:
> Spent a fair bit of time stepping through the peer registration logic this morning and it makes sense to me. The ordering seems to be well enforced and I couldn't figure out a way to get into an undefined state (but we should spend more time in internal audit trying to break this).
> 
> Just to double check, we don't want to support NTT deployments with multiple locking chains right? I'm not sure why an integrator would want to do that, but maybe worth being explicit in the README as I think then we could have multiple active hub keys?

Updated the statement in the README to be more explicit. Thanks :slightly_smiling_face:

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+        ExecuteMsg::SubmitObservations {
+            observations,
+            guardian_set_index,
+            signature,
+        } => submit_observations(deps, info, observations, guardian_set_index, signature),
+
+        ExecuteMsg::SubmitVaas { vaas } => submit_vaas(deps, info, vaas),
+    }
+}
+
```
a5-pickle, line 97: Can remove this in favor of `TRIMMED_DECIMALS` from https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/solana/modules/ntt-messages/src/trimmed_amount.rs#L17.

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+fn normalize_transfer_amount(trimmed_amount: TrimmedAmount) -> Uint256 {
+    let to_decimals = 8;
```
a5-pickle, line 99: Can use `untrim` using `TRIMMED_DECIMALS` instead.

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+                delivery_instruction.sender_address.into(),
+                delivery_instruction.payload,
+            )
+        } else {
+            // otherwise, the sender and payload is the same as the VAA
+            (o.emitter_address.into(), o.payload.0)
+        };
+
+    let hub_key = TRANSCEIVER_TO_HUB.key((o.emitter_chain, sender));
+
```
a5-pickle, line 208:
```suggestion
    let hub = hub_key
        .load(deps.storage)
        .map_err(|_| ContractError::MissingHubRegistration)?;
```

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+        .may_load(deps.storage)
+        .context("failed to load hub")?
+        .ok_or(ContractError::MissingHubRegistration)?;
+
+    let message: TransceiverMessage<WormholeTransceiver, NativeTokenTransfer> =
+        TypePrefixedPayload::read_payload(&mut payload.as_slice())
+            .context("failed to parse observation payload")?;
+
+    let destination_chain = message.ntt_manager_payload.payload.to_chain.id;
+    let source_peer_key = TRANSCEIVER_PEER.key((o.emitter_chain, sender, destination_chain));
```
a5-pickle, line 219:
```suggestion
    let source_peer = source_peer_key
        .load(deps.storage)
        .map_err(|_| ContractError::MissingSourcePeerRegistration(destination_chain.into()))?;
```

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+    let source_peer = source_peer_key
+        .may_load(deps.storage)
+        .context("failed to load source peer")?
+        .ok_or_else(|| ContractError::MissingSourcePeerRegistration(destination_chain.into()))?;
+    let destination_peer_key =
+        TRANSCEIVER_PEER.key((destination_chain, source_peer, o.emitter_chain));
```
a5-pickle, line 225: `may_load` is used a lot when there is no case to handle the option in a special way (like with attempting to fetch the relayer emitter above). These all can be changed to `load` and mapping to a specific error.

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
a5-pickle, line 99: Can remove this in favor of `TRIMMED_DECIMALS` from https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/solana/modules/ntt-messages/src/trimmed_amount.rs#L17.

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+                delivery_instruction.payload,
+            )
+        } else {
+            // otherwise, the sender and payload is the same as the VAA
+            (o.emitter_address.into(), o.payload.0)
+        };
+
+    let hub_key = TRANSCEIVER_TO_HUB.key((o.emitter_chain, sender));
+
+    let hub = hub_key
```
a5-pickle, line 209:
```suggestion
    let hub = hub_key
        .load(deps.storage)
        .map_err(|_| ContractError::MissingHubRegistration)?;
```

cosmwasm/contracts/ntt-global-accountant/src/contract.rs
```diff
@@ -0,0 +1,962 @@
...
+        .ok_or(ContractError::MissingHubRegistration)?;
+
+    let message: TransceiverMessage<WormholeTransceiver, NativeTokenTransfer> =
+        TypePrefixedPayload::read_payload(&mut payload.as_slice())
+            .context("failed to parse observation payload")?;
+
+    let destination_chain = message.ntt_manager_payload.payload.to_chain.id;
+    let source_peer_key = TRANSCEIVER_PEER.key((o.emitter_chain, sender, destination_chain));
+    let source_peer = source_peer_key
+        .may_load(deps.storage)
```
a5-pickle, line 221: `may_load` is used a lot when there is no case to handle the option in a special way (like with attempting to fetch the relayer emitter above). These all can be changed to `load` and mapping to a specific error.

a5-pickle (approved):

kev1n-peters (approved):

PR #3807, opened 2024-03-01
NTT/ Acct / Node: Guardian support

evan-gray:
Reopening #3800 due to GH bug.


node/pkg/accountant/accountant.go
```diff
@@ -41,26 +41,27 @@ type (
...
 
 	// pendingEntry is the payload for each pending transfer
 	pendingEntry struct {
-		msg    *common.MessagePublication
-		msgId  string
-		digest string
+		msg         *common.MessagePublication
+		msgId       string
+		digest      string
+		isNTT       bool
```
mdulin2, line 58: Would it make sense to make this more expandable than just an NTT or token bridge? If this is made more generic then changes like this in the future for new products would probably be easier to expand.

node/pkg/accountant/ntt.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/certusone/wormhole/node/pkg/common"
+	"github.com/certusone/wormhole/node/pkg/supervisor"
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+	"go.uber.org/zap"
+)
+
+// nttEnabled returns true if NTT is enabled, false if not.
+func (acct *Accountant) nttEnabled() bool {
+	return acct.nttContract != ""
+}
```
RahulMaganti47, line 17: why not check against a single valid string instead of against the empty string? checking against empty can be prone to initialization errors (this will return true for any non-empty string)

bruce-riley:
Closing to create a new PR with me as owner.

wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
```diff
@@ -191,6 +189,32 @@ const fetchGlobalAccountantMetrics = async (): Promise<{
       return p;
     }, {} as any);
 
+async function waitForMetricsChange(
+  failurePredicate: (GuardianMetrics) => boolean,
+  retryTimeout: number = 1000,
+  retryAttempts: number = 30
```
RahulMaganti47, line 195: might be nice to encapsulate these in an object

wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
```diff
@@ -191,6 +189,32 @@ const fetchGlobalAccountantMetrics = async (): Promise<{
...
+  let passed = false;
+  let attempts = 0;
+  while (!passed) {
+    attempts++;
+    await new Promise((resolve) => setTimeout(resolve, retryTimeout));
+    let afterMetrics;
+    try {
+      afterMetrics = await fetchGlobalAccountantMetrics();
+    } catch (e) {
+      continue;
```
RahulMaganti47, line 206: can you log an error before continuing?

node/pkg/accountant/accountant.go
bruce-riley: Expanding the accountant to more bridges will require extensive changes, so this can be changed if / when that happens.

PR #3823, opened 2024-03-08
Node/EVM: Batch poller refactor

bruce-riley:
Simplify the batch poller to remove a go routine and a channel hop.

The main change here is to not start a go routine in `NewBatchPollConnector` but wait until `SubscribeForBlocks` like other connectors do. This allows us to publish directly to the sink passed in, rather than having to use an intermediate go-ethereum event feed.

Also exposing some get block methods to be used by a future PR.


node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -61,48 +58,50 @@ func NewBatchPollConnector(ctx context.Context, logger *zap.Logger, baseConnecto
...
+	errCount := 0
+
 	common.RunWithScissors(ctx, errC, "block_poll_subscribe_for_blocks", func(ctx context.Context) error {
+		timer := time.NewTimer(b.Delay)
+		defer timer.Stop()
 		for {
 			select {
 			case <-ctx.Done():
-				blockSub.Unsubscribe()
-				innerErrSub.Unsubscribe()
```
evan-gray, line 96: just to make sure I'm following, this can be removed because the caller has a `defer headerSubscription.Unsubscribe()`

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -61,48 +58,50 @@ func NewBatchPollConnector(ctx context.Context, logger *zap.Logger, baseConnecto
...
-			case v := <-innerErrSink:
-				b.logger.Info("batch_poller innerErrSink received an error, forwarding it on", zap.String("v", v))
-				sub.err <- fmt.Errorf(v)
+			case <-timer.C:
+				lastBlocks, err = b.pollBlocks(ctx, sink, lastBlocks)
+				if err != nil {
+					errCount++
+					b.logger.Error("batch polling encountered an error", zap.Int("errCount", errCount), zap.Error(err))
+					if errCount > 3 {
+						errC <- fmt.Errorf("polling encountered too many errors: %w", err)
```
evan-gray, line 98: should this return the error so that runWithScissors can handle the channel publish?

evan-gray (approved):

node/pkg/watchers/evm/connectors/batch_poller.go
bruce-riley: Mostly they can go because those are the two go-ethereum feeds that no longer exist.

node/pkg/watchers/evm/connectors/batch_poller.go
bruce-riley: I believe the pattern in the watchers is that the go routines should post to errC and return nil. Isn't that right @panoel?

node/pkg/watchers/evm/connectors/finalizer_poller.go
```diff
@@ -188,19 +188,19 @@ func (b *FinalizerPollConnector) pollBlock(ctx context.Context, logger *zap.Logg
 	return
 }
 
-func getLatestBlock(ctx context.Context, logger *zap.Logger, conn Connector) (*NewBlock, error) {
-	return getBlockByFinality(ctx, logger, conn, Latest)
+func GetLatestBlock(ctx context.Context, logger *zap.Logger, conn Connector) (*NewBlock, error) {
```
panoel, line 191: I didn't see where these were referenced outside this file and needed to be exported.

panoel (approved):

PR #3830, opened 2024-03-12
Deploy Mantle to Sepolia

bruce-riley:
Wormhole was deployed to Mantle Goerli testnet last year, but the chain was never enabled in Testnet. Since Mantle Goerli will be shut down next month in favor of Mantle Sepolia, this PR does the deployment to Mantle Sepolia Testnet.

The chain ID and basic guardian support already exists so those changes are not part of this PR.


evan-gray (approved):

node/cmd/guardiand/node.go
```diff
@@ -692,8 +692,6 @@ func runNode(cmd *cobra.Command, args []string) {
 		logger.Fatal("Both --scrollContract and --scrollRPC must be set together or both unset")
 	}
 
-	// Mantle should not be allowed in mainnet until its finality policy is understood and implemented in the watcher.
-	// Note that as of 11/9/2023 Mantle does not support querying for `finalized` or `safe`, just `latest`, so we will need to implement a finalizer.
```
SEJeff, line 696: Was this sorted?

SEJeff (approved):

node/cmd/guardiand/node.go
evan-gray: Yes, they updated to (OP) Bedrock.

bruce-riley: Yup, they upgraded to bedrock and added support for querying for `finalized`.

PR #3834, opened 2024-03-13
Node/CCQ: Track concurrent queries

bruce-riley:
This PR adds a metric to the proxy server to track the maximum number of concurrent requests by chain. This will be useful for scaling the amount of parallelism in the guardian watchers.

### DIFF
--- a/node/cmd/ccq/http.go
+++ b/node/cmd/ccq/http.go
@@ -112,7 +112,8 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 		Signature:    signature,
 	}
 
-	if status, err := validateRequest(s.logger, s.env, s.permissions, s.signerKey, apiKey, signedQueryRequest); err != nil {
+	status, queryReq, err := validateRequest(s.logger, s.env, s.permissions, s.signerKey, apiKey, signedQueryRequest)
+	if err != nil {
 		s.logger.Error("failed to validate request", zap.String("userId", permEntry.userName), zap.String("requestId", hex.EncodeToString(signedQueryRequest.Signature)), zap.Int("status", status), zap.Error(err))
 		http.Error(w, err.Error(), status)
 		// Error specific metric has already been pegged.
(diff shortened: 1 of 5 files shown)


node/cmd/ccq/metrics.go
```diff
@@ -108,4 +111,19 @@ var (
 			Name: "ccq_server_total_number_of_successful_reconnects",
 			Help: "Total number of successful reconnects to bootstrap peers",
 		})
+
+	maxConcurrentQueriesByChain = promauto.NewGaugeVec(
```
evan-gray, line 115: I wonder if it's also worth tracking `currentConcurrentQueriesByChain` to see how bursty the overlapping requests are.

evan-gray (approved):

node/cmd/ccq/http.go
```diff
@@ -112,7 +112,8 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 		Signature:    signature,
 	}
 
-	if status, err := validateRequest(s.logger, s.env, s.permissions, s.signerKey, apiKey, signedQueryRequest); err != nil {
+	status, queryReq, err := validateRequest(s.logger, s.env, s.permissions, s.signerKey, apiKey, signedQueryRequest)
+	if err != nil {
```
panoel, line 116: Is it necessary to check if queryReq is nil?

panoel (approved):

node/cmd/ccq/http.go
bruce-riley: The function always returns queryReq when the error is nil, so I don't think it's necessary to check the pointer for nil.

node/cmd/ccq/utils.go
```diff
@@ -89,14 +89,14 @@ func validateRequest(logger *zap.Logger, env common.Environment, perms *Permissi
 	if err != nil {
```
bingyuyap, line 89: does `queryRequest.Unmarshal()` return error if `len(qr.QueryRequest)` is 0? If not would that unmarshal it into an empty object?

bingyuyap: it seems like `UnmarshalFromReader` does that. All good!

bingyuyap (approved):
LGTM!

node/cmd/ccq/utils.go
bruce-riley: Unmarshal from an empty buffer returns this error:
`failed to read message version: EOF`

PR #3843, opened 2024-03-18
Node/NTT/Acct: Defense in depth

bruce-riley:
This PR adds a defense in depth check to the guardian NTT accountant to the code that parses an AR payload, to make sure that the entire payload is as expected, not just up through what we need.


mdulin2 (approved):
Exactly what I was looking for. Thanks for taking the check seriously! Could save us some day.

evan-gray (approved):

bingyuyap (approved):

PR #3844, opened 2024-03-18
Node/CCQ: Concurrent queries

bruce-riley:
This PR allows a configurable (but hard coded) number of concurrent queries per chain.

TODO: We need to come up with a reasonable number of workers per chain!

### DIFF
--- a/node/pkg/query/query.go
+++ b/node/pkg/query/query.go
@@ -87,7 +87,46 @@ type (
 		channel        chan *PerChainQueryInternal
 		lastUpdateTime time.Time
 	}
+
+	PerChainConfig struct {
+		TimestampCacheSupported bool
+		NumWorkers              int
+	}
 )
+
+var perChainConfig = map[vaa.ChainID]PerChainConfig{
+	vaa.ChainIDSolana:          {NumWorkers: 10, TimestampCacheSupported: false},
+	vaa.ChainIDEthereum:        {NumWorkers: 5, TimestampCacheSupported: true},
+	vaa.ChainIDBSC:             {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDPolygon:         {NumWorkers: 5, TimestampCacheSupported: true},
+	vaa.ChainIDAvalanche:       {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDOasis:           {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDAurora:          {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDFantom:          {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDKarura:          {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDAcala:           {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDKlaytn:          {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDCelo:            {NumWorkers: 1, TimestampCacheSupported: true},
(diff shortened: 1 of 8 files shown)


bingyuyap (approved):
LGTM, but lets wait for the discussion on workers per chain

node/pkg/query/query.go
```diff
@@ -87,8 +87,47 @@ type (
...
+	vaa.ChainIDBaseSepolia:     {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDOptimismSepolia: {NumWorkers: 1, TimestampCacheSupported: true},
+	vaa.ChainIDPolygonSepolia:  {NumWorkers: 1, TimestampCacheSupported: true},
+}
+
+func GetPerChainConfig(chainID vaa.ChainID) PerChainConfig {
+	if pcc, exists := perChainConfig[chainID]; exists {
+		return pcc
+	}
+	return PerChainConfig{}
```
panoel, line 128: An empty return value is not checked anywhere.  Should it be?  Or it's good enough to notice the watcher rolling?

node/pkg/watchers/evm/ccq.go
```diff
@@ -19,6 +20,30 @@ import (
...
+	for count := 0; count < w.ccqConfig.NumWorkers; count++ {
+		workerId := count
+		common.RunWithScissors(ctx, errC, fmt.Sprintf("evm_fetch_query_req_%d", workerId), func(ctx context.Context) error {
+			w.ccqLogger.Debug("CONCURRENT: starting worker", zap.Int("worker", workerId))
+			for {
+				select {
+				case <-ctx.Done():
+					return nil
+				case queryRequest := <-w.queryReqC:
+					w.ccqLogger.Debug("CONCURRENT: processing query request", zap.Int("worker", workerId))
```
panoel, line 38: Since the workerIds will repeat for all the chains, does the logger identify which watcher this is?  I wasn't sure about the ccqLogger.

panoel (dismissed):

node/pkg/watchers/evm/ccq.go
bingyuyap: I think when setting up the logger it does 
[```w.ccqLogger = logger.With(zap.String("component", "ccqsol"))```](https://github.com/wormhole-foundation/wormhole/blob/0248837bf374a762b57560699edde83ef92f1248/node/pkg/watchers/solana/ccq.go#L35)
[```w.ccqLogger = logger.With(zap.String("component", "ccqevm"))```](https://github.com/wormhole-foundation/wormhole/blob/0248837bf374a762b57560699edde83ef92f1248/node/pkg/watchers/evm/watcher.go#L203-L204)

bruce-riley: Yeah, all of the watcher loggers have the chain name in them. However, I have reconsidered having the worker number in the runnable name. It really adds no value and just pollutes the scissors metrics. Please see the rework commit and comment again if you have any other concerns.

node/pkg/query/query.go
bingyuyap: lgtm since if the chain config does not exist or workers are 0 this would evaluate to false and not start ccq for the watche

```go
// QueriesSupported can be used by the watcher to determine if queries are supported for the chain.
func (config PerChainConfig) QueriesSupported() bool {
	return config.NumWorkers > 0
}
```

bingyuyap (approved):

evan-gray (approved):
Looks good. The only concern that I see this introduces is that now `QueryHandler` needs to avoid any race / lock contention. It probably already did, but now it definitely does.

bruce-riley:
@evan-gray I went through all the code again and I don't see any issues with concurrency. The only state that is shared between the workers is the RPC (which is already being used in multiple go routines) and, in the case of EVM, the timestamp cache, which already has a mutex.

So I think the concurrency of the workers is okay.

PR #3846, opened 2024-03-19
Node/NTT/Acct: Add tests to verify emitter config

bruce-riley:


node/pkg/accountant/ntt_test.go
```diff
@@ -271,3 +271,24 @@ func TestNttParseArMsgUnknownArEmitter(t *testing.T) {
...
+	directEmitters, arEmitters, err := nttGetEmitters(common.MainNet)
+	require.NoError(t, err)
+	assert.Equal(t, 0, len(directEmitters)) // TODO: Change this when we add a mainnet emitter!
+	assert.NotEqual(t, 0, len(arEmitters))
+}
+
+func TestNttVerifyTestnetEmitters(t *testing.T) {
+	directEmitters, arEmitters, err := nttGetEmitters(common.TestNet)
+	require.NoError(t, err)
+	assert.NotEqual(t, 0, len(directEmitters))
```
panoel, line 285: We don't have any testnet emitters, yet?

panoel (approved):

node/pkg/accountant/ntt_test.go
bruce-riley: That is a `NotEqual`, so this is ensuring that we have at least one. I don't want to check for the actual value because that is a maintenance headache.

PR #3803, opened 2024-03-01
Upgrade sui packages and compiler to 1.19.1-mainnet

kcsongor:
This should not require a contract upgrade, as the packages are backwards compatible, but it allows integrators to more seamlessly build on wormhole, as the previous dependencies were fairly out of date.

Marking it as draft for now, as I don't have permissions to push to the `sui` ghcr package, so I pushed to `sui-tmp` for now in order to be unblocked on testing CI.

PR #3847, opened 2024-03-19
update rev of sui dependencies 

PutraLaksmana:
update the rev sui dependencies.

so that anyone can build and test it.


evan-gray:
First, this appears to be the testnet rev (though I don't see that the commit matches a tag), which should likely not be the one merged into main. Second, there is already PR #3803 open to update the packages and compiler which is in draft due to CI failures. Once that is resolved, updating the package dependencies in `sui/testnet` can be considered.

In the interim, please consider using the `sui/testnet` or `sui-upgrade-testnet` branches.

PR #3849, opened 2024-03-19
sdk/rust: prepare to publish crate

a5-pickle:
Desperately needs documentation


.github/workflows/build.yml
evan-gray, line 1: What is the motivation for / impact of removing `locked` here?

sdk/rust/serde_wormhole/Cargo.toml
```diff
@@ -1,17 +1,21 @@
...
 
 [dependencies]
-base64 = "0.13"
-itoa = "1.0.1"
-serde = { version = "1.0.103", default-features = false }
-serde_bytes = "0.11.5"
-thiserror = "1.0"
+base64.workspace = true
+itoa.workspace = true
+serde.workspace = true
```
evan-gray, line 14: Is this a downgrade of the serde version?

sdk/rust/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
evan-gray, line 1: As far as I recall, conventionally, lock files should be committed. Why is this an exception?

evan-gray (commented):
The split generally seems good. Unsure about any downstream impacts of other packages now using the workspace instead of path overrides like they were.

sdk/rust/.gitignore
a5-pickle: When crates are published, lock files are ignored. So because these things aren't built into any sort of independent binary, there is no use in maintaining a lock file. And as far as I can tell with the repo, there is nothing that requires that the rust SDK maintain its lock file.

.github/workflows/build.yml
a5-pickle: I can reintroduce it for the smart contract directories. I actually wanted to see if anything broke if I took out that flag for everything in that workflow.

Should I split the workflow so we run --locked for everything but the sdk/rust directory (see comment about removing the lock file)?

sdk/rust/serde_wormhole/Cargo.toml
a5-pickle: When publishing a crate, the highest version will be fetched to validate that the library can be built. So when this package is used for something in our repo like the smart contracts, it is just requiring that any version fetched is okay as long as it is major version 1.

The serde patches do not break compatibility with previous versions as far as I can tell... Those libraries have been on 1.0 for quite a while. But I can set the version to 1.0 to be extra safe. How do you feel about this?

sdk/rust/.gitignore
evan-gray: Looking at some popular crates, I see they follow this pattern 👍

sdk/rust/serde_wormhole/Cargo.toml
evan-gray: If that is the standard, I don't feel it's unreasonable to comply, and then I suppose it comes down to the integrator's `Cargo.lock` anyway.

.github/workflows/build.yml
evan-gray: Personally, I'd lean towards keeping locked for the things with lock files and therefore splitting the workflow as you described.

a5-pickle: Fixed with your help: https://github.com/wormhole-foundation/wormhole/pull/3849/commits/cb7271a2a44d05c359013a397b84f34e23e0cd73. Thank you

evan-gray (approved):
What a masterful change of the workflow yaml :clap: :clap: :clap:

kcsongor (approved):

PR #3850, opened 2024-03-20
update rev of sui dependencies 

PutraLaksmana:
#### When trying to build and test the wormhole package, it will fail, because the dependencies of sui are deprecated.


evan-gray:
Please do not open duplicate PRs, especially after a previous one was closed with a comment. I recommend starting with a GitHub issue if https://github.com/wormhole-foundation/wormhole/pull/3847#issuecomment-2007466187 was insufficient.

PR #3851, opened 2024-03-20
Node/EVM: Remove obsolete field in logs

bruce-riley:


node/pkg/watchers/evm/connectors/finalizer_poller.go
```diff
@@ -74,9 +76,8 @@ func (b *FinalizerPollConnector) SubscribeForBlocks(ctx context.Context, errC ch
 }
 
 func (b *FinalizerPollConnector) runFromSupervisor(ctx context.Context) error {
-	logger := supervisor.Logger(ctx).With(zap.String("eth_network", b.Connector.NetworkName()))
 	supervisor.Signal(ctx, supervisor.SignalHealthy)
-	return b.run(ctx, logger)
+	return b.run(ctx, b.logger)
 }
 
 func (b *FinalizerPollConnector) run(ctx context.Context, logger *zap.Logger) error {
```
bingyuyap, line 83: if `FinalizerPollConnector` has the field `logger`, is there a reason why we are passing the logger into the method instead of calling `b.logger` from the receiver?

```go
func (b *FinalizerPollConnector) methodA() {
  b.logger.Info("logging...")
 }
 
// instead of 
func (b *FinalizerPollConnector) methodA(logger *zap.Logger) {
  logger.Info("logging...")
 }
```

bingyuyap (dismissed):
LGTM. Left a comment about the usage of logger using receiver vs parameters

evan-gray (dismissed):
Looks good, though I concur with @bingyuyap 's comment - might be worthwhile to clean up the logger parameter on the methods if they are no longer needed due to it being on the struct.

node/pkg/watchers/evm/connectors/finalizer_poller.go
evan-gray: seconded, would be good to clean up if it isn't needed anymore.

node/pkg/watchers/evm/connectors/logpoller.go
```diff
@@ -62,7 +65,7 @@ func (l *LogPollConnector) run(ctx context.Context) error {
 		case err := <-errC:
 			return err
 		case block := <-blockChan:
-			if err := l.processBlock(ctx, logger, block); err != nil {
+			if err := l.processBlock(ctx, l.logger, block); err != nil {
```
bingyuyap, line 68: there's still one more over here! 😅

node/pkg/watchers/evm/connectors/finalizer_poller.go
```diff
@@ -101,10 +102,10 @@ func (b *FinalizerPollConnector) run(ctx context.Context, logger *zap.Logger) er
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-timer.C:
-			prevLatest, prevFinalized, err = b.pollBlock(ctx, logger, prevLatest, prevFinalized)
+			prevLatest, prevFinalized, err = b.pollBlock(ctx, b.logger, prevLatest, prevFinalized)
```
bingyuyap, line 105: this should be using the logger from the receiver

bingyuyap (commented):
I think we missed 2 methods!

bingyuyap (approved):
Thank you for the changes!

Issue #2888, opened 2023-05-08
Wormhole Spy should validate signed VAAs before further processing


kev1n-peters:
## Description and context
The Wormhole Spy should fetch (periodically?) the latest guardian set and use it to validate inbound signed VAAs. If the signed VAA is not valid, then it shouldn't process it any further (e.g. in `PublishSignedVAA` or `HandleGossipVAA`).

Here is an example of fetching the latest guardian set: https://github.com/wormhole-foundation/wormhole-dashboard/blob/0a07acab591de0dbd071d00e064bd88bc67daf40/fly/cmd/fly/main.go#L177

ljb630:
```
package main

import (
	"fmt"
	"time"
)

func main() {
	// Run the Wormhole Spy to periodically fetch and validate VAAs
	RunWormholeSpy()
}

func RunWormholeSpy() {
	// Set the desired frequency for fetching the latest guardian set
	fetchFrequency := time.Hour // Fetch the latest guardian set every hour

	// Start a ticker to trigger periodic fetches
	ticker := time.NewTicker(fetchFrequency)
	defer ticker.Stop()

	// Initialize the guardian set
	guardianSet := FetchGuardianSet()

	// Loop to fetch and validate VAAs periodically
	for {
		select {
		case <-ticker.C:
			// Fetch the latest guardian set
			latestGuardianSet := FetchGuardianSet()

			// Update the current guardian set if a new one is available
			if latestGuardianSet != nil {
				guardianSet = latestGuardianSet
			}

			// Get inbound signed VAAs
			inboundVAAs := GetInboundVAAs()

			// Validate and process the VAAs
			for _, signedVAA := range inboundVAAs {
				if IsValidVAA(signedVAA, guardianSet) {
					PublishSignedVAA(signedVAA)
					HandleGossipVAA(signedVAA)
				}
			}
		}
	}
}

// FetchGuardianSet simulates fetching the latest guardian set
func FetchGuardianSet() *GuardianSet {
	// Implement the logic to fetch the latest guardian set
	// Return the fetched guardian set or nil if an error occurs
	fmt.Println("Fetching the latest guardian set...")
	// Simulating fetching the guardian set
	guardianSet := &GuardianSet{
		Index: 1,
		Keys:  []string{"key1", "key2", "key3"},
	}
	return guardianSet
}

// GetInboundVAAs simulates retrieving inbound signed VAAs
func GetInboundVAAs() []SignedVAA {
[...]

PR #3854, opened 2024-03-24
Node/spy: Adds optional VAA signature verification

bruce-riley:
This PR enhances the spy to support optional verification of VAA signatures. Any VAA that has an invalid signature is dropped.

To use this feature, add the following config parameters.

These values should work for mainnet:

```
--ethRPC https://eth.drpc.org
--ethContract 0x98f3c9e6E3fAce36bAAd05FE09d375Ef1464288B
```

These values should work for Sepolia testnet:

```
--ethRPC https://sepolia.drpc.org/
--ethContract 0x4a8bc80Ed5a4067f1CCf107057b8270E0cC11A78
```

This PR fixes issue #2888.


node/cmd/spy/spy.go
```diff
@@ -304,6 +311,17 @@ func runSpy(cmd *cobra.Command, args []string) {
...
 	gst := common.NewGuardianSetState(nil)
 
+	// VAA verifier (optional)
+	if *ethRPC != "" {
+		if *ethContract == "" {
+			logger.Fatal(`If "--ethRPC" is specified, "--ethContract" must also be specified`)
+		}
+		vaaVerifier = NewVaaVerifier(logger, *ethRPC, *ethContract)
+		if err := vaaVerifier.GetInitialGuardianSet(); err != nil {
+			logger.Fatal(`Failed to read initial guardian set for VAA verification`, zap.Error(err))
```
bingyuyap, line 321: since we don't need to log double quotes `"` in the string here should we standardize this to double quotes instead of backticks?

bruce-riley: Is there a downside to using back ticks? Because I'm actually thinking I should start always using back ticks so I can double quote things in the message without thinking about it. . .

node/cmd/spy/vaa_verifier.go
```diff
@@ -0,0 +1,136 @@
...
+func (v *VaaVerifier) fetchGuardianSet(gsIndex uint32) (*common.GuardianSet, error) {
+	timeout, cancel := context.WithTimeout(context.Background(), RpcTimeout)
+	defer cancel()
+
+	rawClient, err := ethRpc.DialContext(timeout, v.rpcUrl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to ethereum: %w", err)
+	}
+
+	client := ethClient.NewClient(rawClient)
```
bingyuyap, line 121: If this is used very frequently we might want to consider reusing the same client instead of creating a new one on each invocation. But I would assume it to be otherwise since we are only invoking this method when we don't have the guardian set in the map.

bingyuyap (approved):
Good work!

node/cmd/spy/spy.go
bingyuyap: I don't think so! In fact it is more readable when you have double quotes in the string. Was just nitpicking because I thought double quote was the standard, since most of the code is using double quotes.

node/cmd/spy/vaa_verifier.go
```diff
@@ -0,0 +1,136 @@
...
+		v.logger.Error("failed to unmarshal VAA", zap.Error(err), zap.Any("vaa", vaaBytes))
+		return false, err
+	}
+
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	gs, exists := v.guardianSets[vv.GuardianSetIndex]
+	if !exists {
+		gs, err = v.fetchGuardianSet(vv.GuardianSetIndex)
```
evan-gray, line 95: I like this just-in-time guardian set fetching!

node/cmd/spy/vaa_verifier.go
evan-gray: I would expect this to only ever be called when the guardian set changes while the spy is running. In the history of Wormhole v2, that has only happened 3 times, so I'm not too worried at this stage. Given the infrequency, seems better than keeping the client around.

evan-gray (approved):

node/cmd/spy/vaa_verifier.go
bruce-riley: It will be used *very* infrequently, only when the guardian set index changes (once a year??).

evan-gray (approved):

PR #3862, opened 2024-03-28
Ethereum: deploy with forge

bruce-riley:
This PR adds the ability to deploy the core and token bridge contracts using forge rather than truffle.

This PR lifts the scripts in `ethereum/forge-scripts` from PR https://github.com/wormhole-foundation/wormhole/pull/3395 and modifies only the following:

- deployCoreBridge.sh
- DeployCore.s.sol
- deployTokenBridge.sh
- DeployTokenBridge.s.sol

The standard `foundry.toml` file uses a newer version of solc, and the relayer code relies on that. Since we don't want to risk upgrading the compiler used by the WH contracts at this time, this PR also adds the `forge-deploy-build.sh` script which modifies the toml file, deletes the relayer contracts and build the contracts needed for a standard deployment.


evan-gray (dismissed):

evan-gray (approved):

evan-gray:
Compared the `deployedBytecode` between `build` and `build-forge` and they appear to be the same aside from what I understand to be the hashes.

ethereum/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build:core": "truffle compile",
-    "build:forge": "if [ \"$DEV\" = \"True\" ]; then forge build; else forge build --via-ir; fi",
+    "build:forge": "forge build",
```
kcsongor, line 21: This (together with the foundry.toml file) changes the default behaviour from building `via-ir` to building without it. Also, previously to disable this behaviour, one would have to run
```sh
export DEV=True
```
but now, to *enabled* this behaviour, one has to run
```sh
export FOUNDRY_PROFILE=production
```

This is fine, but might be worth documenting in the README?

kcsongor (approved):
Overall looks great!

ethereum/package.json
bruce-riley: The `--via-ir` flag as really only intended for the relayer contracts. Since we don't currently build the base contracts with that flag, I removed it from the command. If we decide to "upgrade" to building with IR in the future, it can be re-added.

PR #3864, opened 2024-03-29
CI: Change chain id for evm devnet node

barnjamin:
The Ethereum Mainnet (eip155) chain id is `1`, our Ethereum node for devnet also sets its config to report chain id of `1`

This conflicting chain id to identify the network breaks some assumptions.

This PR is meant to identify what else might break if the devnet chain id is changed to something else


barnjamin:
First attempt with only the yaml file update passed all tests except Tilt which was still running but in the logs:

image

updated the `.env.test` file which is copied to `.env` in the eth Makefile

evan-gray (approved):

kcsongor (approved):

PR #3866, opened 2024-04-01
Ethereum: Move relayer

bruce-riley:
This PR moves the relayer contracts out of the `ethereum` folder into `relayer/ethereum`.

This change is necessary as part of the migration from deploying the core and token bridge contracts with truffle to using forge, which is necessary because truffle is deprecated.

The original plan for this PR was to change the relayer to use the `wormhole-solidity-sdk` for the `IWormhole` interface so that it is not directly dependent on the old folder. However, this changed the byte code for the contracts, which changed the devnet deploy address for the proxy contract (because it is deployed with `create2`), breaking various tests. The decision was made to "kick that can down the road" and not make the move to use the sdk. Instead `IWormhole` has been copied into the relayer folder. The same goes for the `BytesLib` library, which was also copied into the relayer folder.


ethereum-relayer/Makefile
```diff
@@ -0,0 +1,51 @@
...
+forge_dependencies: lib/forge-std lib/openzeppelin-contracts lib/wormhole-solidity-sdk
+
+lib/forge-std:
+	forge install foundry-rs/forge-std@v1.5.5 --no-git --no-commit
+
+lib/openzeppelin-contracts:
+	forge install openzeppelin/openzeppelin-contracts@0457042d93d9dfd760dbaa06a4d2f1216fdbe297 --no-git --no-commit
+
+lib/wormhole-solidity-sdk:
+	forge install wormhole-foundation/wormhole-solidity-sdk --no-git --no-commit
```
bruce-riley, line 29: Should there be a tag for this?

evan-gray: Probably the latest commit as of right now?

.github/CODEOWNERS
```diff
@@ -1,6 +1,6 @@
 # Default
 
-* @evan-gray
+- @evan-gray
```
evan-gray, line 3: don't think you meant to change this.

ethereum/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@wormhole-foundation/contracts-ethereum",
+  "name": "@wormhole-foundation/contracts-ethereum-relayer",
```
evan-gray, line 2: wouldn't this one stay `contracts-ethereum`?

evan-gray (dismissed):
Confirmed the diffs of the copied files are sane
```
diff ethereum/.gitignore relayer/ethereum/.gitignore
diff ethereum/Makefile relayer/ethereum/Makefile 
diff ethereum/contracts/interfaces/IWormhole.sol relayer/ethereum/contracts/interfaces/IWormhole.sol
diff ethereum/contracts/libraries/external/BytesLib.sol relayer/ethereum/contracts/libraries/external/BytesLib.sol
diff ethereum/foundry.toml relayer/ethereum/foundry.toml
diff ethereum/package-lock.json relayer/ethereum/package-lock.json
diff ethereum/package.json relayer/ethereum/package.json
```

gator-boi:
I think the new relayer directory could use a readme with instructions for building and testing :)

relayer/ethereum/contracts/interfaces/IWormhole.sol
```diff
@@ -0,0 +1,142 @@
+// contracts/Messages.sol
```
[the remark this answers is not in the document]
bruce-riley: The real solution is to make the relayer contracts use the `wormhole-solidity-sdk`, but that results in the byte code changing, which I was not willing to do as part of this PR. That is a task that should be taken on by the owners of the relayer contracts at the appropriate time.

In the meantime, we did not want to introduce a hard dependency on the `ethereum` folder, so the decision was made to copy the file for now. (If we linked to the one in the ethereum folder, and someone changed that, the byte code of the relayer contracts would change.)

bruce-riley:
@gator-boi I added a basic readme file. More details would need to be provided by whoever owns the relayer contracts.

evan-gray (approved):

PR #3867, opened 2024-04-01
node: Throttle CoinGecko queries

djb15:
We're consistently hitting the CoinGecko API rate limit now on the free plan which means we're not always updating prices and the query check fails when trying to update the governor token list. Rather than paying for the premium API plans we could instead just throttle our requests, since these price updates are not time critical.

Someone better then me at Go can tell me if this is rubbish!

### DIFF
--- a/node/pkg/governor/governor_prices.go
+++ b/node/pkg/governor/governor_prices.go
@@ -135,6 +135,9 @@ func (gov *ChainGovernor) queryCoinGecko(ctx context.Context) error {
 	params := url.Values{}
 	params.Add("bust", strconv.Itoa(int(time.Now().Unix()))+strconv.Itoa(rand.Int())) // #nosec G404
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	// Throttle the queries to the CoinGecko API. We query for 200 tokens at a time, so this throttling would
 	// allow us to query up to 18,000 tokens in a 15 minute window (the query interval). Currently there are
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):
Thanks for doing this!

node/pkg/governor/governor_prices.go
```diff
@@ -128,14 +128,35 @@ func (gov *ChainGovernor) PriceQuery(ctx context.Context) error {
...
 	params := url.Values{}
 	params.Add("bust", strconv.Itoa(int(time.Now().Unix()))+strconv.Itoa(rand.Int())) // #nosec G404
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Throttle the queries to the CoinGecko API. We query for 200 tokens at a time, so this throttling would
+	// allow us to query up to 12,000 tokens in a 15 minute window (the query interval). Currently there are
+	// between 1000 and 2000 tokens.
+	throttle := make(chan int, 1)
```
SEJeff, line 144: As a followup, for your governor update script, I almost wonder if it makes sense to add a check for `len(governed_tokens) > 12000`? If that fails, there will be something in the test to tell you to come update this. Not for this PR, but makes sense as a followup I think to keep everyone honest.

SEJeff (approved):

node/pkg/governor/governor_prices.go
djb15: Great suggestion! Have made a note to open a PR for this

PR #3869, opened 2024-04-04
ci: switch from ganache to anvil

evan-gray:
This PR switches the devnet / ci EVM node from `ganache` to `anvil`. Notably, this had a couple impacts:
- `anvil` is set to automine at 1s with `block-time=1` so `mine.js` is no longer needed
- `anvil` simulates 2 epoch finality (including `safe` and `finalized` tags), so waiting for `finalized` blocks is now required as needed. This is done with `await ethProvider.send("anvil_mine", ["0x40"]);`
  - Essentially this required mining 64 blocks before attempting to fetch and finalized "Eth" or "BSC" VAA
image

While touching the CI tests for `sdk/js`, `sdk/js-query`, and `accountant`, I also took the opportunity to
- switch to `JsonRpcProvider` to avoid worrying about calling `.destroy()`
- suppress `secp256k1` and `bigint` log messages
- unify the `testTimeout` in the jest config instead of using `jest.setTimeout` everywhere
- use the standard `jest.config.json` filename
- clean up some frivolous logging
- suppress guardian warning `rollback detected in timestamp cache` in `unsafeDevMode` as it gets triggered thousands of times when `anvil_mine` is used
- remove `--guardiand_loglevel=info` from the workflow as it's no longer needed for debugging the noted issue

Bonus: All that cleanup means the Tilt CI logs are now ~8300 lines instead of > 10,000


bruce-riley (approved):

panoel (approved):
I only inspected the files I am code owner on.

sdk/js/jest.config.json
```diff
@@ -5,5 +5,5 @@
   "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$",
   "testPathIgnorePatterns": ["__tests__/utils"],
   "moduleFileExtensions": ["ts", "tsx", "js", "jsx", "json", "node"],
-  "testTimeout": 60000
+  "testTimeout": 300000
```
barnjamin, line 8: the manual jest timeout settings i saw were 120k, this change takes timeout from 60k->300k

just picked a big number or experimentally found to be long enough?

barnjamin (approved):

kcsongor (approved):
very nice 👍

sdk/js/jest.config.json
evan-gray: Heh, basically. Some of the tests with conflicting resources caused the timeouts to grow slightly higher (which triggered the last change to 120) and this switch to Anvil seemed to take some just over that limit. I would hope this could be resolved with some better / hard-coded resource allocation, but that will be quite a slog to get right. `300k` just felt slightly better than doubling it to `240k` :shrug:

PR #3871, opened 2024-04-09
Deploy XLayer, Linea and Berachain to testnet

bruce-riley:
This PR adds support for X Layer, Linea and Berachain, but limits it to testnet.

There is a known shortcoming in the watcher for Linea, in that they do not support polling for finalized, so we will need to read the roll up contract on Ethereum. That will be handled as a follow on PR.

These deployments were done using Forge, not Truffle.


evan-gray (approved):

SEJeff (approved):
go bits look good ✅

PR #3872, opened 2024-04-10
Node/EVM: Linea poller

bruce-riley:
This PR modifies the watcher for Linea to track the publication of finalized blocks from the roll up contract on Ethereum.

There are still questions about whether or not this implementation is adequate, but it is at least a step in the right direction.

This PR also updates the `getFinality` function in the watcher to make the finality assumptions of every EVM chain explicit.

Note to @panoel : Don't be frightened by the size of this PR. Most of it is in the generated ABI file.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -182,8 +182,10 @@ var (
 	xlayerRPC      *string
 	xlayerContract *string
 
-	lineaRPC      *string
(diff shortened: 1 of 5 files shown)


panoel (commented):
I did not review the ABI.

panoel (approved):

bingyuyap (approved):

PR #3877, opened 2024-04-12
Tilt : Deploy with forge instead of truffle

bruce-riley:
This PR changes the tilt deployment to use forge rather than truffle. It borrows from PR #3395.


evan-gray (dismissed):

gator-boi:
Forge does not do any validation of the input values for the deployment scripts. It might be nice to verify that addresses and bytes32 types are not the zero address (unless that's the intended value). Also, I don't think the shell scripts belong in the forge-scripts directory. Other than that, everything looks good.

gator-boi (approved):
I reviewed the forge deployment scripts. In a future PR it would be nice to remove the dummy contract deploys and fix the tests using the new ethereum contract addresses.

PR #3884, opened 2024-04-16
node: add emitters to ntt allowlist

nik-suri:
Please double check that these are addresses that point to valid and verified NTT contracts!


bruce-riley (approved):

bruce-riley (approved):

evan-gray (approved):
confirmed that the EVM addresses are verified `WormholeTransceiver`s and that the Solana address has emitted NTT VAAs

PR #3888, opened 2024-04-18
Node: Logging cleanup

bruce-riley:
This PR does two things.

First, it adds the standard VAA / Message ID (emitterChain/emitterAddress/seqNo) to messages in the processor to make it easier to follow a message from end-to-end in the logs.

Second, it adds a log level check around a lot of the debug messages so that we don't waste time formatting complex log messages that get dropped anyway.

### DIFF
--- a/node/pkg/processor/processor.go
+++ b/node/pkg/processor/processor.go
@@ -3,6 +3,7 @@ package processor
 import (
 	"context"
 	"crypto/ecdsa"
+	"encoding/hex"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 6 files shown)


node/pkg/processor/processor.go
```diff
@@ -82,6 +83,14 @@ type (
 	}
 )
 
+func (s *state) MessageID() string {
+	if s.ourObservation != nil {
+		return s.ourObservation.MessageID()
+	}
+
+	return hex.EncodeToString(s.txHash)
```
evan-gray, line 91: Is this safe to sustitute, generally speaking? Should there be a comment on this method that this should not be used to uniquely identify `state` structs?

evan-gray (approved):
approving, though slight concern about the applicability / safety of txHash for MessageID

panoel (approved):

node/pkg/processor/processor.go
bruce-riley: Changed it to `LoggingID` and added a comment.

panoel (approved):

PR #3889, opened 2024-04-18
Deploy SeiEVM Read Only to Testnet

bruce-riley:


ethereum/env/.env.seievm
```diff
@@ -0,0 +1,13 @@
...
+
+RPC_URL=""
+FORGE_ARGS="--legacy --with-gas-price 3000000000" # 3 gwei
+
+# Wormhole Core
+INIT_SIGNERS=["0x58CC3AE5C097b213cE3c81979e1B9f9570746AA5"]
+INIT_CHAIN_ID=40
+INIT_GOV_CHAIN_ID=0x1
+INIT_GOV_CONTRACT=0x0000000000000000000000000000000000000000000000000000000000000004
+INIT_EVM_CHAIN_ID=713715
```
evan-gray, line 13: this is the chain ID for Sei Devnet - https://chainlist.org/chain/713715
should this file be `.env.seievm.testnet`?

sdk/js/src/utils/consts.ts
```diff
@@ -339,7 +339,7 @@ const MAINNET = {
     nft_bridge: undefined,
   },
   seievm: {
-    core: undefined,
+    core: "0x07782FCe991dAb4DE7a3124032E534A0D059B4d8",
```
evan-gray, line 342: This explorer also seems like it's supposed to only work for arctic-1 (public devnet) where it shows this contract  https://seitrace.com/address/0x07782FCe991dAb4DE7a3124032E534A0D059B4d8

ethereum/env/.env.seievm
```diff
@@ -0,0 +1,13 @@
+# Sei EVM mainnet read only env. Use to deploy the core contract with forge.
+
+# ethereum$ ln -s env/.env.seievm .env
+
+RPC_URL=""
```
evan-gray, line 5: I think the RPC for arctic-1 is `https://evm-rpc-arctic-1.sei-apis.com/` [[source](https://v2.docs.sei.io/setting-up-a-wallet#switching-networks)]

evan-gray (changes requested):
Judging by the Chain ID and Sei docs, I think this should be for testnet, no?

sdk/js/src/utils/consts.ts
bruce-riley: Moved address to TESTNET.

evan-gray (approved):

barnjamin (approved):

PR #3896, opened 2024-04-22
Client/js: Add new chains

bruce-riley:
This PR upgrades the version of the SDK used by the worm client to pick up the recently added chains.

### DIFF
--- a/clients/js/src/consts/networks.ts
+++ b/clients/js/src/consts/networks.ts
@@ -171,6 +171,41 @@ const MAINNET = {
     key: getEnvVar("ETH_KEY"),
     chain_id: 30,
   },
+  scroll: {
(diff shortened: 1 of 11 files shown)


clients/js/src/consts/networks.ts
```diff
@@ -171,47 +171,72 @@ const MAINNET = {
     key: getEnvVar("ETH_KEY"),
     chain_id: 30,
   },
-  sepolia: {
```
bruce-riley, line 174: Sorry this looks so ugly. All I really did was pull Scroll and Mantle up to put them in the right place based on chain ID.

panoel (approved):

evan-gray (approved):

PR #3906, opened 2024-04-25
node: shorten hostname for load testing

evan-gray:


bruce-riley (approved):

node/cmd/guardiand/node.go
```diff
@@ -1003,10 +1003,16 @@ func runNode(cmd *cobra.Command, args []string) {
...
 				_, err := net.LookupIP("guardian-0.guardian")
 				if err == nil {
 					break
 				}
+				// load tests use this hostname format
+				_, err = net.LookupIP("guardian-0")
+				if err == nil {
+					break
+				}
 				logger.Info("Error resolving guardian-0.guardian. Trying again...")
```
panoel, line 1016: I don't know if this needs to be changed?  I could go either way.

panoel (approved):

node/cmd/guardiand/node.go
evan-gray: I don't think so, it's only in the devnet branch and the point is pretty much the same.

PR #3913, opened 2024-04-26
Node: Default devnet evm chains

bruce-riley:


SEJeff (approved):
This is a nice change.

panoel (approved):

SEJeff (approved):

node/cmd/guardiand/node.go
```diff
@@ -1818,10 +1596,49 @@ func shouldStart(rpc *string) bool {
...
+			if contractAddr != "" {
+				logger.Fatal(fmt.Sprintf("If --%sRPC is not set, --%sContract must not be set", chainLabel, chainLabel))
+			}
+		} else {
+			if contractAddr == "" {
+				contractAddr = devnet.GanacheWormholeContractAddress.Hex()
+			}
+		}
+	}
+	if contractAddr != "" && !mainnetSupported && mainnetMode() {
```
evan-gray, line 1619: should this use `env` instead? might have to be passed as a parameter, but it seems safer / saner to start using `env` more places.

node/cmd/guardiand/node.go
```diff
@@ -1818,10 +1596,49 @@ func shouldStart(rpc *string) bool {
...
+		logger.Fatal(fmt.Sprintf("Chain %s not supported in mainnet", chainLabel))
+	}
+	return contractAddr
+}
+
+func mainnetMode() bool {
```
evan-gray, line 1625: if we use `env` above, we can drop this function

evan-gray (approved):
Really like the cleanup :heart: Just one small suggestion to keep it even cleaner :pray:

node/cmd/guardiand/node.go
bruce-riley: Created an issue to address this: https://github.com/wormhole-foundation/wormhole/issues/3920

node/cmd/guardiand/node.go
bruce-riley: Created an issue to address this: https://github.com/wormhole-foundation/wormhole/issues/3920

PR #3937, opened 2024-05-09
Replace Terra2 RPCs with new endpoint

terencelimzhengwei:
https://phoenix-lcd.terra.dev is being deprecated. The new LCD for Terra2 is https://lcd-terra.tfl.foundation

### DIFF
--- a/clients/js/src/consts/networks.ts
+++ b/clients/js/src/consts/networks.ts
@@ -138,7 +138,7 @@ const MAINNET = {
     chain_id: undefined,
   },
   terra2: {
-    rpc: "https://phoenix-lcd.terra.dev",
+    rpc: "https://lcd-terra.tfl.foundation",
     chain_id: "phoenix-1",
     key: getEnvVar("TERRA_MNEMONIC"),
   },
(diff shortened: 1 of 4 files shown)


bruce-riley (approved):

evan-gray (changes requested):
If it is the case this will be deprecated (I could not immediately find supporting documentation). I would sooner recommend switching to `https://terra-lcd.publicnode.com/`

terencelimzhengwei:
@evan-gray - i'm from Terraform Labs. We have not put out comms regarding this yet, but we have swapped out the phoenix LCD endpoint to the lcd-terra endpoint in all our repos. An example here (https://github.com/terra-money/station-assets/blob/main/chains/mainnet/terra.js). 

Swapping to publicnode.com works as well if that is what the wormhole team prefers. But just wanted to notify that we've updated the URL of our public infra. Comms and docs will be updated over the next few days.

evan-gray:
As far as I can tell, [the docs](https://docs.terra.money/develop/endpoints/) have not been updated. Closing for the time being.

PR #3950, opened 2024-05-23
script: uses only first digest as identifier for governance prototxt

scnale:
This fixes the governance message document template when using multi mode.

When generating more than one governance message you could observe this being generated as the shell command to generate the prototxt:
```
cat << EOF > governance-206690B
259B804.prototxt
current_set_index:  4
[...]
```

Which breaks the format of the governance message.


kcsongor (approved):

evan-gray (approved):
Thanks!

PR #3951, opened 2024-05-24
clients/js: add chain and rpc support for worm verify-vaa command

HamphreyR26:
## Changes
* Added `--rpc`, `--chain` and `--contract-address` args to `worm verify-vaa` command
* Updated README.md


evan-gray (changes requested):
This fails some checks. Please fix them. Also could you provide the motivation for verifying a VAA against the other EVM core bridges? The result should be identical as long as they are on the same guardian set.

djb15:
Closing as part of cleanup. The PR can be reopened if required.

PR #3964, opened 2024-05-31
ci: add required-review

evan-gray:
The motivation for this PR is to allow for required reviewers which do not have write access to the repo. In order to achieve this, it leverages [action-required-review](https://github.com/Automattic/action-required-review/).

In the short term this could co-exist with the full set of CODEOWNERS and eventually be the sole enforcer for much of the codebase, with the likely exception of the `.github` folder itself, or at least this action, which should probably remain protected by CODEOWNERS. The fallback could also remain with some users with write access if explicit approval from someone with write on every PR was desired.

Reviewers - I have attempted to replicate the existing CODEOWNERS rules in `.github/required-review.yml`, but please confirm they match.

The name field is optional, but I was leaning towards providing it for clarity.

Note that this config will match multiple entries, unlike CODEOWNERS which matches the last one. Therefore the ordering has been switched and some entries set to `consume: true`. See the [Requirements Format](https://github.com/Automattic/action-required-review?tab=readme-ov-file#requirements-format) for more details.

### DIFF
--- a/.github/required-review.yml
+++ b/.github/required-review.yml
@@ -1,0 +1,349 @@
+# see https://github.com/Automattic/action-required-review for more details
+
+# Chains
+
+- name: Chains - Algorand
+  paths:
+    - "algorand/**"
+  teams:
+    - "@evan-gray"
+    - "@kcsongor"
+
+- name: Chains - Aptos
+  paths:
+    - "aptos/**"
+  teams:
+    - "@kcsongor"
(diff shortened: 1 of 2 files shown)


evan-gray:
After further review, there are several shortcomings with a GHA actions approach, including
- Users without write do not properly have their approvals dismissed
- There are different status check results based on trigger making resolution difficult / clunky

PR #3967, opened 2024-06-05
Node metric and performance tweaks

bruce-riley:
This PR does a few things to help with guardian performance.

First, it adds some metrics to the processor to better track the flow of observations.

Secondly, it enhances the `common.GuardianSet` object to cache the quorum value once so that we do not have to repeatedly calculate quorum while processing observations. It also adds a map of guardian address to guardian index to speed up the look up of guardian index.

Finally, it modifies `handleObservation` to bail out early when we have already submitted the VAA.

### DIFF
--- a/node/pkg/common/guardianset.go
+++ b/node/pkg/common/guardianset.go
@@ -8,6 +8,7 @@ import (
 	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -49,8 +50,30 @@ const MaxStateAge = 1 * time.Minute
 type GuardianSet struct {
 	// Guardian's public key hashes truncated by the ETH standard hashing mechanism (20 bytes).
 	Keys []common.Address
+
 	// On-chain set index
 	Index uint32
+
+	// Quorum value for this set of keys
+	Quorum int
+
+	// A map from address to index. Testing showed that, on average, a map is almost three times faster than a sequential search of the key slice.
+	// Testing also showed that the map was twice as fast as using a sorted slice and `slices.BinarySearchFunc`. That being said, on a 4GHz CPU,
+	// the sequential search takes an average of 800 nanos and the map look up takes about 260 nanos. Is this worth doing?
+	keyMap map[common.Address]int
+}
+
+func NewGuardianSet(keys []common.Address, index uint32) *GuardianSet {
+	keyMap := map[common.Address]int{}
+	for idx, key := range keys {
+		keyMap[key] = idx
+	}
+	return &GuardianSet{
+		Keys:   keys,
+		Index:  index,
+		Quorum: vaa.CalculateQuorum(len(keys)),
+		keyMap: keyMap,
+	}
 }
 
 func (g *GuardianSet) KeysAsHexStrings() []string {
@@ -66,9 +89,15 @@ func (g *GuardianSet) KeysAsHexStrings() []string {
 // KeyIndex returns a given address index from the guardian set. Returns (-1, false)
 // if the address wasn't found and (addr, true) otherwise.
 func (g *GuardianSet) KeyIndex(addr common.Address) (int, bool) {
-	for n, k := range g.Keys {
-		if k == addr {
-			return n, true
+	if g.keyMap != nil {
+		if idx, found := g.keyMap[addr]; found {
+			return idx, true
+		}
+	} else {
+		for n, k := range g.Keys {
+			if k == addr {
+				return n, true
+			}
 		}
 	}
 
--- a/node/pkg/common/guardianset_test.go
+++ b/node/pkg/common/guardianset_test.go
@@ -1,11 +1,41 @@
 package common
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/assert"
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
+
+func TestNewGuardianSet(t *testing.T) {
+	keys := []common.Address{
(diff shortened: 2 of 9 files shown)


evan-gray (approved):
Just one nit about always having tracing on.

panoel (approved):

node/pkg/common/guardianset.go
```diff
@@ -66,9 +89,15 @@ func (g *GuardianSet) KeysAsHexStrings() []string {
...
-			return n, true
+	if g.keyMap != nil {
+		if idx, found := g.keyMap[addr]; found {
+			return idx, true
+		}
+	} else {
+		for n, k := range g.Keys {
+			if k == addr {
+				return n, true
+			}
```
SEJeff, line 100: Your code is totally reasonable here, but I always try to do as few indents as feasible to make code a bit easier to read. If you agree, something like this would work:

```suggestion
	if g.keyMap != nil {
		if idx, found := g.keyMap[addr]; found {
			return idx, true
		}
		return -1, false
	}
	for n, k := range g.Keys {
		if k == addr {
			return n, true
		}
```

node/pkg/common/guardianset_test.go
```diff
@@ -1,12 +1,42 @@
...
+		common.HexToAddress("0x2edb0D8530E31A218E72B9480202AcBaeB06178d"),
+		common.HexToAddress("0xa78858e5e5c4705CdD4B668FFe3Be5bae4867c9D"),
+		common.HexToAddress("0x5Efe3A05Efc62D60e1D19fAeB56A80223CDd3472"),
+		common.HexToAddress("0xD791b7D32C05aBB1cc00b6381FA0c4928f0c56fC"),
+		common.HexToAddress("0x14Bc029B8809069093D712A3fd4DfAb31963597e"),
+		common.HexToAddress("0x246Ab29FC6EBeDf2D392a51ab2Dc5C59d0902A03"),
+		common.HexToAddress("0x132A84dFD920b35a3D0BA5f7A0635dF298F9033e"),
+	}
+	gs := NewGuardianSet(keys, 1)
+	assert.True(t, reflect.DeepEqual(keys, gs.Keys))
```
SEJeff, line 35: Sucks that you can't use `assert.Equal()` for this, but I guess that won't work due to them not being byte arrays.

node/pkg/p2p/p2p.go
```diff
@@ -71,6 +72,16 @@ var (
...
+	p2pDrop = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Name: "wormhole_p2p_drops",
+			Help: "Total number of messages that were dropped by libp2p",
+		})
+	p2pReject = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Name: "wormhole_p2p_rejects",
+			Help: "Total number of messages rejected by libp2p",
+		})
```
SEJeff, line 84: A more prometheus idiomatic approach would be more like:

```
wormhole_p2p_messages{type="dropped"} 4
wormhole_p2p_messages{type="rejected"} 5
wormhole_p2p_messages{type="successful"} 500
```

Labels are how you differentiate metrics in prometheus generally.

node/pkg/p2p/p2p.go
```diff
@@ -342,9 +368,13 @@ func Run(
 		}
 
 		logger.Info("Subscribing pubsub topic", zap.String("topic", topic))
+		ourTracer := &traceHandler{}
 		ps, err := pubsub.NewGossipSub(ctx, h,
 			pubsub.WithValidateQueueSize(P2P_VALIDATE_QUEUE_SIZE),
 			pubsub.WithGossipSubParams(components.GossipParams),
+			pubsub.WithEventTracer(ourTracer),
```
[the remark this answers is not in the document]
SEJeff: I don't disagree with your suggestion, but are you sure that it adds much overhead in the current form? The most expensive part I can see (from trace.go in go-libp2p-pubsub) in probably the unique msgId generation. This code is built to use opencensus / opentelemetry which is generally designed to be run all the time in production in sampling or non-sampling mode. A CLI flag to enable it without changing the log debug level might also be more preferable to node operators.

My understanding of past profiling of p2p was the slow bits weren't the p2p per-se but the dht bits layered ontop. [This](https://github.com/libp2p/go-libp2p-kad-dht) was the library that caused so many issues if my memory is correct here.

node/pkg/processor/observation.go
```diff
@@ -245,18 +250,18 @@ func (p *Processor) handleObservation(ctx context.Context, obs *node_common.MsgW
...
-				zap.Int("required_sigs", quorum),
+				zap.Int("required_sigs", gs.Quorum),
 				zap.Int("have_sigs", len(sigsVaaFormat)),
-				zap.Bool("quorum", len(sigsVaaFormat) >= quorum),
+				zap.Bool("quorum", len(sigsVaaFormat) >= gs.Quorum),
 			)
 		}
 
-		if len(sigsVaaFormat) >= quorum && !s.submitted {
+		if len(sigsVaaFormat) >= gs.Quorum {
```
SEJeff, line 259: Nice

SEJeff (approved):
Feel free to toss out my nits if you disagree. This seems like a very sensible change. The only "style difference" thing I might have done is not made `GuardianSet.Quorum` a public struct member as it seems wrong to be able to arbitrarily change it. I'd have made it private and used a getter on it.

node/pkg/p2p/p2p.go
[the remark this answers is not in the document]
evan-gray: I agree it shouldn’t be much overhead comparably, just being paranoid from doing performance analysis and then seeing a couple `if p.tracer =! nil` checks when briefly clicking through. I concur we can leave it as is and revisit later if it turns out to be a concern. Will resolve this for now.

node/pkg/processor/observation.go
```diff
@@ -213,17 +213,22 @@ func (p *Processor) handleObservation(ctx context.Context, obs *node_common.MsgW
...
 	s.signatures[their_addr] = m.Signature
 
-	if s.ourObservation != nil {
+	if s.submitted {
+		if p.logger.Level().Enabled(zapcore.DebugLevel) {
+			p.logger.Debug("already submitted, doing nothing",
+				zap.String("messageId", m.MessageId),
+				zap.String("digest", hash),
+			)
+		}
```
ali-bahjati, line 222: this code path seems unreachable because there is a condition above returning if it's submitted and is not null. if it's null the state is constructed without setting this :?

PR #3968, opened 2024-06-06
node: Allow tokens to change their symbol in generator script

djb15:
When trying to update the governor token list, the script was failing on this check:

```
// Sanity check to make sure the script is doing what we think it is
    if (existingTokenKeys.length + addedTokens.length - removedTokens.length != newTokensCount) {
      console.error(`Num existing tokens (${existingTokenKeys.length}) + Added tokens (${addedTokens.length}) - Removed tokens (${removedTokens.length}) != Num new tokens (${newTokensCount})`);
      process.exit(1);
    }
```

It turned out that a token was being classified as "removed" because its symbol had changed. We should allow tokens to change their symbol (like we do on the new token side) so this PR updates the script to only use the chain + address key for removed token lookups; that's a unique identifier after all.

Even if the token symbol included a "-" character this change is safe since we only concat the first 2 elements of the array, and the symbol is the last element in the key strings.

### DIFF
--- a/node/hack/governor/src/index.ts
+++ b/node/hack/governor/src/index.ts
@@ -215,7 +215,7 @@ axios
               notional +
               "\n";
 
-            newTokenKeys[chain + "-" + wormholeAddr + "-" + data.Symbol] = true;
+            newTokenKeys[chain + "-" + wormholeAddr] = true;
             newTokensCount += 1;
           }
         }
@@ -224,7 +224,9 @@ axios
 
     for (var token of existingTokenKeys) {
       // A token has been removed from the token list 
-      if (!newTokenKeys[token]) {
+      // We cut the symbol off the end of the key as it's possible for a token to change its symbol
+      var tokenParts = token.split("-");
+      if (!newTokenKeys[tokenParts[0] + "-" + tokenParts[1]]) {
         removedTokens.push(token);
       }
     }


bruce-riley (approved):

node/hack/governor/src/index.ts
```diff
@@ -224,7 +224,9 @@ axios
 
     for (var token of existingTokenKeys) {
       // A token has been removed from the token list 
-      if (!newTokenKeys[token]) {
+      // We cut the symbol off the end of the key as it's possible for a token to change its symbol
+      var tokenParts = token.split("-");
+      if (!newTokenKeys[tokenParts[0] + "-" + tokenParts[1]]) {
```
johnsaigle, line 229: Any risk here that `tokenParts` will be empty or that its length can be < 2? If so you might get a NaN or null or crash or "oopsie-doopsie" or however JS/TS handles these things

djb15: Every `token` is read from the previously generated token list, so no. If the token list were malformed then it would have broken on the previous guardian update anyway and we have tests to check the validity of the generated token list

evan-gray: lol, fwiw, you get `undefined` which would then get coerced so you would then get `undefined-undefined` on an empty list.

johnsaigle (approved):
The change conceptually makes sense and I agree that it is still a unique identifier.
I ran the script locally and didn't encounter any issues.

I tried following up with `go run check_query.go` but I think I am rate-limited so that didn't work for me. Probably just something on my side though.

PR #3977, opened 2024-06-10
Node: Observation batching

bruce-riley:
This PR introduces the batching of observations by adding the `SignedObservationBatch` gossip message. It modifies the guardian to wait up to one second in order to batch as many observations as possible (up to 4000) into a single gossip message. This not only reduces gossip traffic but also the load on the guardians and applications that listen to observations.

This PR is meant to go hand-in-hand with PR #4000, which further improves gossip traffic by splitting it across different topics, allowing the guardians and gossip listeners (like spies and flies) to only subscribe to the topics they need.

NOTE: Once the gossip network cutover is complete, support for `SignedObservation` (individual observation) messages will be dropped, but in the meantime, applications that listen to observations need to support both message streams.


node/pkg/p2p/p2p.go
```diff
@@ -74,11 +80,6 @@ var (
 			Name: "wormhole_p2p_drops",
 			Help: "Total number of messages that were dropped by libp2p",
 		})
```
panoel, line 82: Do you want to add a gauge for observations in a batch? :shrug:

node/pkg/p2p/run_params.go
```diff
@@ -95,14 +98,22 @@ func NewRunParams(
 	return p, nil
 }
 
-// WithSignedObservationListener is used to set the channel to receive `SignedObservation“ messages.
+// WithSignedObservationListener is used to set the channel to receive `SignedObservation` messages.
```
panoel, line 101: Since you corrected this one, did you notice all the others only have a beginning tick and not an ending tick?

node/pkg/processor/cleanup.go
```diff
@@ -228,7 +228,10 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 					if err := common.PostObservationRequest(p.obsvReqSendC, req); err != nil {
 						p.logger.Warn("failed to broadcast re-observation request", zap.String("message_id", s.LoggingID()), zap.Error(err))
 					}
-					p.gossipSendC <- s.ourMsg
+					p.postObservationToBatch(s.ourObs)
+					if s.ourMsg != nil {
+						p.gossipSendC <- s.ourMsg // TODO: Get rid of this
```
panoel, line 233: Are you getting rid of this in a future PR?

panoel (dismissed):

node/pkg/processor/cleanup.go
bruce-riley: Yes, after the cutover is complete.

node/pkg/p2p/p2p.go
bruce-riley: Meh.

node/pkg/processor/broadcast.go
```diff
@@ -72,21 +85,14 @@ func (p *Processor) broadcastSignature(
...
-	// We can't block here because the same process would be responsible for reading from obsvC.
-	om := node_common.CreateMsgWithTimestamp[gossipv1.SignedObservation](&obsv)
-	select {
-	case p.obsvC <- om:
-	default:
-		go func() { p.obsvC <- om }()
-	}
-
+	// Post the signature to ourselves.
+	p.handleSingleObservation(addr, ourObs)
```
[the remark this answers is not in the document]
evan-gray: My initial guess is that, since the same process is responsible for this anyway, it could save resources handling this directly, but at the risk of expedited treatment for our own observations. That may not necessarily be a bad thing though, since publishing consensus currently requires our own observation.

evan-gray (approved):

node/pkg/processor/broadcast.go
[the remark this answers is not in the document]
evan-gray: It seems reasonable to me to give this a shot with the aim to further optimize it, particularly with #3988 on the horizon, lessening the immediate impact of this call. I always thought it was odd that the injection was handled in `broadcastSignature` anyway when `HandleQuorum` handles its own state setting and only uses `broadcastSignedVAA` to perform the broadcast.

bruce-riley:
A concern has been raised regarding possible latency impacts of this change. Therefore I am converting this PR back to draft until those concerns can be addressed.

aadam-10:
The availability of Pyth's market data with as little latency as possible is a very strong usecase to warrant not being part of observation batching, and so they should be excluded from it.

bruce-riley:
This PR is being superseded by PR #4066, so this one is being closed.

PR #3988, opened 2024-06-17
Node: Processor performance improvements

bruce-riley:
This PR contains a number of small performance improvements to the processor.

- Writes to the badger DB on a separate go routine. 
- Adds the ability to write multiple VAAs using the BadgerDB batch API.
- Speeds up processing of local observations.
- Adds some measurements to monitor the time to handle observations and quorum.
- Removes some unnecessary map look ups.
- Removes unnecessary debugging code.

### DIFF
--- a/node/pkg/db/db.go
+++ b/node/pkg/db/db.go
@@ -149,12 +149,11 @@ func (d *Database) StoreSignedVAABatch(vaaBatch []*vaa.VAA) error {
 		if err != nil {
 			return err
 		}
-
-		storedVaaTotal.Inc()
 	}
 
 	// Wait for the batch to finish.
 	err := batchTx.Flush()
+	storedVaaTotal.Add(float64(len(vaaBatch)))
 	return err
 }
 
--- a/node/pkg/db/db.go
+++ b/node/pkg/db/db.go
@@ -128,6 +128,36 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
 	return nil
 }
 
+// StoreSignedVAABatch writes multiple VAAs to the database using the BadgerDB batch API.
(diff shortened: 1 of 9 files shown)


node/pkg/processor/processor.go
```diff
@@ -139,6 +140,14 @@ type Processor struct {
 	acctReadC      <-chan *common.MessagePublication
 	pythnetVaas    map[string]PythNetVaaEntry
 	gatewayRelayer *gwrelayer.GatewayRelayer
+	updateVAALock  sync.Mutex
+	updatedVAAs    map[string]*updateVaaEntry
```
panoel, line 144: Would it be useful to anyone to have a prometheus metric showing how big this map gets?

panoel (approved):

node/pkg/processor/processor.go
bruce-riley: I kinda don't think so. I think there are a lot of metrics that no one ever looks at, so I don't really want to add to that. . .

evan-gray (approved):

node/pkg/db/db.go
```diff
@@ -128,6 +128,35 @@ func (d *Database) StoreSignedVAA(v *vaa.VAA) error {
...
+
+		err = batchTx.Set(VaaIDFromVAA(v).Bytes(), b)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Wait for the batch to finish.
+	err := batchTx.Flush()
+	storedVaaTotal.Add(float64(len(vaaBatch)))
```
evan-gray, line 156: Should this only `.Add` when `err != nil`? or follow the pattern / flow from `StoreSignedVAA` above?

evan-gray (approved):

panoel (approved):

PR #3996, opened 2024-06-25
Node: p2p.Run interface change

bruce-riley:
This PR changes `p2p.Run` to take a parameter structure, rather than over two dozen individual parameters. The idea is that most of those parameters only have meaning to the guardian, so we should not require applications like the spy, fly and WormholeScan to setup numerous parameters that they don't need, just to suit the API. Additionally, this will allow us to add extra parameters that are of use only to the guardian without having to modify those other applications.

Once this PR is approved and merged, this new functionality with simplify two upcoming guardian features:

- Observation Batching
- Gossip Topic Split

Note that this PR follows a typical go approach for passing in optional parameters using `With` functions. For instance, this same approach is used in `libp2p`, which served as a pattern for this PR.


node/pkg/node/options.go
```diff
@@ -55,33 +55,38 @@ func GuardianOptionP2P(p2pKey libp2p_crypto.PrivKey, networkId, bootstrapPeers,
...
+			params, err := p2p.NewRunParams(
+				bootstrapPeers,
+				networkId,
 				p2pKey,
-				g.gk,
 				g.gst,
-				networkId,
-				bootstrapPeers,
-				nodeName,
-				disableHeartbeatVerify,
```
evan-gray, line 70: I think this one got lost. Not sure if that was intentional.

evan-gray (approved):

node/pkg/node/options.go
bruce-riley: Good catch. I added it to `WithGuardianOptions`.

evan-gray (approved):

ftocal (approved):

walker-16 (approved):

node/pkg/p2p/run_params.go
```diff
@@ -0,0 +1,220 @@
...
+		p.disableHeartbeatVerify = disableHeartbeatVerify
+		return nil
+	}
+}
+
+// WithGuardianOptions is used to set options that are only meaningful to the guardian.
+func WithGuardianOptions(
+	nodeName string,
+	gk *ecdsa.PrivateKey,
+	obsvC chan<- *common.MsgWithTimeStamp[gossipv1.SignedObservation],
```
panoel, line 150: Isn't this one set with WithSignedObservationListener?

panoel: Is this so that you don't have to call all the functions and just pass all the values in?

bruce-riley: Yes, but that is meant to be used by clients. I wanted `WithGuardianOptions` to take everything that the guardian requires.

node/pkg/p2p/run_params.go
```diff
@@ -0,0 +1,220 @@
...
+	ibcFeaturesFunc func() string,
+	gatewayRelayerEnabled bool,
+	ccqEnabled bool,
+	signedQueryReqC chan<- *gossipv1.SignedQueryRequest,
+	queryResponseReadC <-chan *query.QueryResponsePublication,
+	ccqBootstrapPeers string,
+	ccqPort uint,
+	ccqAllowedPeers string,
+) RunOpt {
+	return func(p *RunParams) error {
```
panoel, line 168: Do you have to return an error type for this?  It's always nil.

panoel (approved):

node/pkg/p2p/run_params.go
bruce-riley: That is the signature of `RunOpt`. Although none of those return an error at this point, some might some day, so I would like to leave it. Also, that is the pattern that I copied from libp2p...

PR #4000, opened 2024-06-27
Node: Gossip Topic Split

bruce-riley:
### Overview
This PR splits the guardian gossip traffic across multiple p2p topics:

- `control` - used for heartbeats, observation requests and governor messages
- `attestation` - used for observations
-  `broadcast` - used for VAAs

The `broadcast` topic is the old topic. It was maintained rather than using `vaa` to maintain backwards compatibility during the cutover interval.

This PR provides the logic to perform the cutover from the old topic to the new ones at a predetermine time. If the time has not passed yet (or is not set), we publish everything on and listen for everything on the old topic. After the cutover time, we will publish on the above topics. After the cutover, we will still continue to listen for everything on the old topic, but there should be nothing on that topic except for VAAs (since it becomes the VAA topic). 

### Roll Out
Although this PR fully implements the gossip split and cutover logic, the cutover times are not set, so this PR will have no impact on publishing. Future PRs will be used to set the cutover time for first testnet, then mainnet. At some point after the cutover is fully complete, a new PR will be taken out to remove publishing to and listening on the old topic.


node/pkg/p2p/p2p.go
```diff
@@ -355,24 +366,84 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+				}
+			}()
+
+			if params.obsvReqC != nil || params.signedGovCfg != nil || params.signedGovSt != nil {
+				logger.Info("subscribing to the control topic", zap.String("topic", controlTopic))
+				controlSubscription, err = controlPubsubTopic.Subscribe(pubsub.WithBufferSize(P2P_SUBSCRIPTION_BUFFER_SIZE))
+				if err != nil {
+					return fmt.Errorf("failed to subscribe to the control topic: %w", err)
+				}
+				defer controlSubscription.Cancel()
```
panoel, line 394: This is different than the other 2 channels?  This one doesn't have a defer func to close?

node/pkg/p2p/p2p.go
```diff
@@ -426,332 +497,618 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+								logger.Error("failed to publish message from control queue to old topic", zap.Error(err))
 							}
 						}
-						if params.gatewayRelayerEnabled {
-							features = append(features, "gwrelayer")
-						}
-						if params.ccqEnabled {
-							features = append(features, "ccq")
+					} else {
+						logger.Error("received a message on the control queue when we do not have a control topic")
```
panoel, line 630: If we are after the cutover time and we don't have a control channel, shouldn't we do something more drastic than log?

node/pkg/p2p/p2p.go
```diff
@@ -426,332 +497,618 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+							}
 						}
+					case *gossipv1.GossipMessage_SignedVaaWithQuorum:
+						if params.signedInC != nil {
+							select {
+							case params.signedInC <- m.SignedVaaWithQuorum:
+								p2pMessagesReceived.WithLabelValues("signed_vaa_with_quorum").Inc()
+							default:
+								if params.components.WarnChannelOverflow {
+									// TODO do not log this in production
```
panoel, line 1043: Don't log what?  The warning?

panoel (approved):

node/pkg/p2p/p2p.go
bruce-riley: That's because `pubsub.Subscription.Cancel` doesn't return a value, so the defer just calls it directly. That's the way the subscribe for the old topic was too.

node/pkg/p2p/p2p.go
bruce-riley: This is an old tbjump thing. I removed the todo and added a comment saying that parameter should not be enabled in prod.

node/pkg/p2p/p2p.go
bruce-riley: That would actually be a program bug, so I changed it (and similar cases for the other topics) to a panic.

node/pkg/p2p/p2p.go
```diff
@@ -355,24 +367,84 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+				return fmt.Errorf("failed to join the control topic: %w", err)
+			}
+
+			defer func() {
+				if err := controlPubsubTopic.Close(); err != nil && !errors.Is(err, context.Canceled) {
+					logger.Error("Error closing the control topic", zap.Error(err))
+				}
+			}()
+
+			if params.obsvReqC != nil || params.signedGovCfg != nil || params.signedGovSt != nil {
```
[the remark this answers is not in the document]
evan-gray: Also, it doesn't seem like there is a way from someone to get heartbeats but not one of the other three types or have nodeName set. Not sure if there are any consumers like that.

node/pkg/p2p/p2p.go
evan-gray: that panic makes me nervous, but agree it would be a programming error. wonder if there's any safer way to set this up.

evan-gray (dismissed):

node/pkg/p2p/p2p.go
bruce-riley: After some offline discussion, it was agreed that the panic is reasonable since this function always creates the control topic if the control channel is not nil, so this would definitely be a program bug.

node/pkg/p2p/p2p.go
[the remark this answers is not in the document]
bruce-riley: I'm renaming the receive channels.

Regarding receiving heartbeats, I don't think there has ever been a way to subscribe to heartbeats. We can add that someday if someone wants it.

evan-gray (approved):
thanks for the renames, much easier to read imo

panoel (dismissed):

evan-gray (approved):

panoel (approved):

PR #4001, opened 2024-06-27
node: Fix tokenEntry when checking flow cancel for pending transfers

johnsaigle:
## Bug

Related: #3990. The same issue exists when generating the `tokenEntry` for messages that are queued/pending. The result is that when "small transfers" (i.e. those under the "big tx limit") were moved from the queue into `transfers` slice, the corresponding flow-cancelling transfer was not added to the `TargetChain`.

Note that this shouldn't meaningfully affect the Governor usage that we can currently observe. The only chain that is near or at capacity is Sui and thus it is the only one with queued small transfers. That means that transfers coming _from_ Sui were not reducing the governor usage of the destination chain after they were dequeued. 

The only result will be a reduction in Governor usage for non-Sui chains that Sui users are interacting with. 

## Enhancements

This PR adds a few logs in places where the `sumValue` function in the governor stats can return an error. This should help Guardians to get some insight if these error conditions are hit.

I also changed the name of a `value` variable used in a monitoring function. `value` was used first to indicate the "governor usage" and then reused to indicate the remaining available notional value for a chain. This refactor tries to make it clear that these are different concepts.

A new unit test is added to ensure that flow-cancelling works when a pending transfer is released from the queue.

## Commit breakdown

Commit 1:

Similar to a previous issue in the function `ProcessMsgForTime`, the
tokenEntry was not being generated properly.
This should result in queued "small transfers" being able to flow cancel
when they are released from the queue.
Also adds a comment on the CheckedInt64 function to indicate what its
error states mean and when they occur.

Commit 2:

- Adds governor logging to error cases
- Add function comments to `governor_monitoring` explain what they do and what their error
  states mean
- Change variable names in publishStatus function. `value` was used
  first to indicate the "governor usage" and then reused to indicate the
  remaining available notional value for a chain. This refactor tries to
  make it clear that these are different concepts
  
 Commit 3:

- Add a unit test to ensure that, when a pending transfer is released,
  it also does flow-cancelling on the TargetChain (previously we had a
  bug here)
- Add documentation for CheckPendingForTime to clarify that it has
  side-effects

### DIFF
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -794,8 +794,10 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
 							// Mandatory check to ensure that the token should be able to reduce the Governor limit.
 							if tokenEntry.flowCancels {
 								if destinationChainEntry, ok := gov.chains[payload.TargetChain]; ok {
+									// Defense in depth check. A transfer with this value should have been
+									// added to the pending list above instead of reaching this branch.
 									if ce.isBigTransfer(value) {
-										return nil, err
+										return nil, fmt.Errorf("refusing to add flow cancel transfer with value %d above the bigTransactionSize %d", transfer.dbTransfer.Value, ce.bigTransactionSize)
 									}
 
 									if err := destinationChainEntry.addFlowCancelTransferFromDbTransfer(&dbTransfer); err != nil {
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -149,10 +149,6 @@ func (ce *chainEntry) addFlowCancelTransfer(transfer transfer) error {
 	if targetChain != ce.emitterChainId {
 		return fmt.Errorf("flow cancel transfer TargetChain %s does not match this chainEntry %s", targetChain, ce.emitterChainId)
 	}
-	// Big transfers should not flow cancel under any circumstances
-	if ce.isBigTransfer(transfer.dbTransfer.Value) {
-		return fmt.Errorf("refusing to add flow cancel transfer with value %d above the bigTransactionSize %d", transfer.dbTransfer.Value, ce.bigTransactionSize)
-	}
 
 	ce.transfers = append(ce.transfers, transfer)
 	return nil
@@ -794,11 +790,6 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
 							// Mandatory check to ensure that the token should be able to reduce the Governor limit.
 							if tokenEntry.flowCancels {
 								if destinationChainEntry, ok := gov.chains[payload.TargetChain]; ok {
-									// Defense in depth check. A transfer with this value should have been
-									// added to the pending list above instead of reaching this branch.
-									if ce.isBigTransfer(value) {
-										return nil, fmt.Errorf("refusing to add flow cancel transfer with value %d above the bigTransactionSize %d", transfer.dbTransfer.Value, ce.bigTransactionSize)
-									}
 
 									if err := destinationChainEntry.addFlowCancelTransferFromDbTransfer(&dbTransfer); err != nil {
 										return nil, err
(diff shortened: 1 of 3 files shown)


bruce-riley (approved):

node/pkg/governor/governor.go
```diff
@@ -774,7 +774,7 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
 						ce.transfers = append(ce.transfers, transfer)
 
 						// Add inverse transfer to destination chain entry if this asset can cancel flows.
-						key := tokenKey{chain: dbTransfer.EmitterChain, addr: dbTransfer.EmitterAddress}
+						key := tokenKey{chain: pe.token.token.chain, addr: pe.token.token.addr}
```
djb15, line 777: Is there a test case we could/should add here to check flow cancelling when loading from pending entries?

johnsaigle: Yeah that's probably a good idea. In `governor_test.go` there's a test `TestPendingTransferBeingReleased`. I can modify this to also include a flow cancel transfer to make sure that it's included in this flow.

johnsaigle: @djb15 Added a new unit test in the latest commit

djb15 (dismissed):

node/pkg/governor/governor.go
djb15: Nice! Even cooler that you do it without advancing time and just using the flow cancelling to build the headroom for the queued transfer

node/pkg/governor/governor.go
```diff
@@ -774,12 +788,16 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
...
 						tokenEntry := gov.tokens[key]
 						if tokenEntry != nil {
 							// Mandatory check to ensure that the token should be able to reduce the Governor limit.
 							if tokenEntry.flowCancels {
 								if destinationChainEntry, ok := gov.chains[payload.TargetChain]; ok {
+									if ce.isBigTransfer(value) {
+										return nil, err
```
djb15, line 798: Isn't `err` going to be nil here?

claudijd (approved):

node/pkg/governor/governor.go
johnsaigle: Yes! Thanks I replaced this with an error string

djb15 (dismissed):

mdulin2 (approved):

mdulin2 (dismissed):

node/pkg/governor/governor.go
```diff
@@ -148,6 +149,10 @@ func (ce *chainEntry) addFlowCancelTransfer(transfer transfer) error {
 	if targetChain != ce.emitterChainId {
 		return fmt.Errorf("flow cancel transfer TargetChain %s does not match this chainEntry %s", targetChain, ce.emitterChainId)
 	}
+	// Big transfers should not flow cancel under any circumstances
```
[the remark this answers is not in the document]
johnsaigle: Thanks I removed it. The other errors returned here are cases where processing the transfer is impossible. This new case would just be when it's undesirable at a high-level. So it doesn't fit here.

node/pkg/governor/governor.go
```diff
@@ -774,12 +788,18 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
...
-						key := tokenKey{chain: dbTransfer.EmitterChain, addr: dbTransfer.EmitterAddress}
+						key := tokenKey{chain: pe.token.token.chain, addr: pe.token.token.addr}
 						tokenEntry := gov.tokens[key]
 						if tokenEntry != nil {
 							// Mandatory check to ensure that the token should be able to reduce the Governor limit.
 							if tokenEntry.flowCancels {
 								if destinationChainEntry, ok := gov.chains[payload.TargetChain]; ok {
+									// Defense in depth check. A transfer with this value should have been
+									// added to the pending list above instead of reaching this branch.
+									if ce.isBigTransfer(value) {
```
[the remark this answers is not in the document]
johnsaigle: Thanks I deleted this

johnsaigle:
I've converted this to draft as the Flow Cancel changes have been rolled-back. I will rework this PR once we decide on a strategy

johnsaigle:
These changes have been incorporated into https://github.com/wormhole-foundation/wormhole/pull/4016.

PR #4002, opened 2024-07-02
node: Fix issue where transfers that were loaded from the DB did not add a flow-cancel transfer on the TargetChain

johnsaigle:
Flow-canceling is done in the `ProcessMsgForTime` loop when a new message occurs. However, this was not done when a node restarted and reloaded transfers from the past 24 hours. As a result it was possible for the node to calculate a result that showed that the outgoing transfers for an emitter chain exceeded the daily limit. In effect this is true but only with the condition that there was incoming flow to allow this to happen. This appeared to violate an invariant and so the node did not start properly.

### DIFF
--- a/node/pkg/governor/governor_db.go
+++ b/node/pkg/governor/governor_db.go
@@ -243,11 +243,6 @@ func (gov *ChainGovernor) reloadTransfer(xfer *db.Transfer) error {
 	}
 	ce.transfers = append(ce.transfers, transfer)
 
-	// If the transfer does not flow cancel, we're done now. Transfers about the bigTransactionSize never flow cancel.
-	if ce.isBigTransfer(xfer.Value) {
-		return nil
-	}
-
 	// Reload flow-cancel transfers for the TargetChain. This is important when node restarts so that a corresponding,
 	// inverse transfer is added to the TargetChain. This is already done during the `ProcessMsgForTime` loop but
 	// that function does not capture flow-cancelling when the node is restarted.
@@ -257,7 +252,7 @@ func (gov *ChainGovernor) reloadTransfer(xfer *db.Transfer) error {
 		if tokenEntry.flowCancels {
 			if destinationChainEntry, ok := gov.chains[xfer.TargetChain]; ok {
 				if err := destinationChainEntry.addFlowCancelTransferFromDbTransfer(xfer); err != nil {
-					return  err
+					return err
 				}
 			} else {
 				gov.logger.Warn("tried to cancel flow but chain entry for target chain does not exist",
--- a/node/pkg/governor/governor_db.go
+++ b/node/pkg/governor/governor_db.go
@@ -18,6 +18,9 @@ func (gov *ChainGovernor) loadFromDB() error {
 	return gov.loadFromDBAlreadyLocked()
 }
 
+// Loads transfers and pending data from the database and modifies the corresponding fields in the ChainGovernor.
+// These fields are slices transfers or pendingTransfers and will be sorted by their Timestamp property.
+// Modifies the state of the database as a side-effect: 'transfers' that are older than 24 hours are deleted.
 func (gov *ChainGovernor) loadFromDBAlreadyLocked() error {
 	xfers, pending, err := gov.db.GetChainGovernorData(gov.logger)
 	if err != nil {
@@ -154,10 +157,16 @@ func (gov *ChainGovernor) reloadPendingTransfer(pending *db.PendingTransfer) {
 		zap.String("Hash", hash),
 	)
 
+	// Note: no flow cancel added here. We only want to add an inverse, flow-cancel transfer when the transfer is
+	// released from the pending queue, not when it's added.
 	ce.pending = append(ce.pending, &pendingEntry{token: token, amount: payload.Amount, hash: hash, dbData: *pending})
 	gov.msgsSeen[hash] = transferEnqueued
 }
 
+// Processes a db.Transfer and validates that it should be loaded into `gov`.
+// Modifies `gov` as a side-effect: when valid transfer is loaded, the properties 'transfers' and 'msgsSeen' are
+// updated with information about the loaded transfer. In the case of a loading a transfer of a flow-canceling asset,
+// both chain entries (emitter and target) will be updated.
 func (gov *ChainGovernor) reloadTransfer(xfer *db.Transfer) error {
 	ce, exists := gov.chains[xfer.EmitterChain]
 	if !exists {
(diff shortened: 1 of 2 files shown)


node/pkg/governor/governor_db.go
```diff
@@ -18,6 +18,9 @@ func (gov *ChainGovernor) loadFromDB() error {
 	return gov.loadFromDBAlreadyLocked()
 }
 
+// Loads transfers and pending data from the database and modifies the corresponding fields in the ChainGovernor.
+// These fields are slices transfers or pendingTransfers and will be sorted by their Timestamp property.
```
bruce-riley, line 22: Nit "slices of transfers".

node/pkg/governor/governor_db.go
```diff
@@ -18,6 +18,9 @@ func (gov *ChainGovernor) loadFromDB() error {
 	return gov.loadFromDBAlreadyLocked()
 }
 
+// Loads transfers and pending data from the database and modifies the corresponding fields in the ChainGovernor.
```
bruce-riley, line 21: I think the doc generator expects the function prologues to start with the function name.

node/pkg/governor/governor_db.go
```diff
@@ -154,10 +157,16 @@ func (gov *ChainGovernor) reloadPendingTransfer(pending *db.PendingTransfer) {
...
 	)
 
+	// Note: no flow cancel added here. We only want to add an inverse, flow-cancel transfer when the transfer is
+	// released from the pending queue, not when it's added.
 	ce.pending = append(ce.pending, &pendingEntry{token: token, amount: payload.Amount, hash: hash, dbData: *pending})
 	gov.msgsSeen[hash] = transferEnqueued
 }
 
+// Processes a db.Transfer and validates that it should be loaded into `gov`.
+// Modifies `gov` as a side-effect: when valid transfer is loaded, the properties 'transfers' and 'msgsSeen' are
```
bruce-riley, line 167: Nit: "when a valid transfer"

node/pkg/governor/governor_test.go
```diff
@@ -2337,6 +2337,206 @@ func TestDontReloadDuplicates(t *testing.T) {
...
+	// Sum of absolute value of all transfers, including inverse flow cancel transfers:
+	// 2 * (10_000) = 20_000
+	_, valueTransferred, _, _ := gov.getStatsForAllChains()
+	assert.Equal(t, uint64(20000), valueTransferred)
+
+	governorUsageEth, err := gov.TrimAndSumValueForChain(chainEntryEth, time.Unix(int64(transferTime.Unix()-1000), 0))
+	assert.Equal(t, uint64(10000), governorUsageEth)
+	assert.Zero(t, governorUsageEth-chainEntryEth.dailyLimit) // Make sure we used the whole capacity
+	require.NoError(t, err)
+	governorUsageSui, err := gov.TrimAndSumValueForChain(chainEntrySui, time.Unix(int64(transferTime.Unix()-1000), 0))
```
bruce-riley, line 2484: Looks like you want `require.NoError(t, err)` after 2484, 2507 and 2529.

node/pkg/governor/governor_test.go
```diff
@@ -2337,6 +2337,205 @@ func TestDontReloadDuplicates(t *testing.T) {
...
+	tokenBridgeAddrStrEthereum := "0x0290fb167208af455bb137780163b7b7a9a10c16" //nolint:gosec
+	tokenBridgeAddrEthereum, err := vaa.StringToAddress(tokenBridgeAddrStrEthereum)
+	require.NoError(t, err)
+	// recipientEthereum := "0x707f9118e33a9b8998bea41dd0d46f38bb963fc8" //nolint:gosec
+
+	// Data for Sui
+	tokenBridgeAddrStrSui := "0xc57508ee0d4595e5a8728974a4a93a787d38f339757230d441e895422c07aba9" //nolint:gosec
+	tokenBridgeAddrSui, err := vaa.StringToAddress(tokenBridgeAddrStrSui)
+	require.NoError(t, err)
+	// recipientSui := "0x84a5f374d29fc77e370014dce4fd6a55b58ad608de8074b0be5571701724da31"
```
pleasew8t, line 2378: Do these comments on 2372, 2378 and 2381 need to stay? Just asking in case you were meaning to remove them.

node/pkg/governor/governor_test.go
```diff
@@ -2337,6 +2337,205 @@ func TestDontReloadDuplicates(t *testing.T) {
...
+	// Setup
+	ctx := context.Background()
+	gov, err := newChainGovernorForTest(ctx)
+
+	require.NoError(t, err)
+	assert.NotNil(t, gov)
+
+	// Set-up time
+	gov.setDayLengthInMinutes(24 * 60)
+	transferTime := time.Unix(int64(1654543099), 0)
```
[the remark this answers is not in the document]
johnsaigle: Done in latest commit

node/pkg/governor/governor_test.go
johnsaigle: I removed the commented `recipient...` fields in the latest commit.

bruce-riley (approved):

pleasew8t (approved):

djb15 (approved):

node/pkg/governor/governor_db.go
```diff
@@ -233,5 +242,27 @@ func (gov *ChainGovernor) reloadTransfer(xfer *db.Transfer) error {
 		return err
 	}
 	ce.transfers = append(ce.transfers, transfer)
+
+	// Reload flow-cancel transfers for the TargetChain. This is important when node restarts so that a corresponding,
+	// inverse transfer is added to the TargetChain. This is already done during the `ProcessMsgForTime` loop but
+	// that function does not capture flow-cancelling when the node is restarted.
+	tokenEntry := gov.tokens[tk]
+	if tokenEntry != nil {
```
mdulin2, line 250: To me, I'd prefer that the *negative* part of the flow were stored in the database. However, the dBTransfer object is unsigned, meaning that's not possible without modifying the DB entries. So, adding in the flow canceling upon reload makes sense.

One issue this may have is a race condition on token entries on reload vs. what was originally set at the time of usage. We are *assuming* that all of the token entries in the flow cancel list are the same now as prior to the reload, which may not be the case.

So, a token that was NOT flow canceled before will be on reload and a token that wasn't originally flow canceled WILL be on reload. I don't see how to avoid this in the current architecture though... Hmmm...

mdulin2: After discussions in Slack, we decided that the complexity of fixing this issue is too high in a nice way. It would require modifying the database storage of transfers to include an unsigned value, which would be difficult to get right in all of the places. 

If the second case (token removed) is triggered in production, then the governor just panics and doesn't allow anything through. This can also happen through governor limit decreases midcycle and reload. 

A better solution to this is changing the error handling on the reload to allow for the sum to be larger than dailylimit on reload. Although there may be a delay on the chain with the removed flow cancel entries, it's very unlikely to happen. 

In the backfilled flow cancels (token added case), it's fine. We've decided to add the token in so it's not much of a threat.

PR #4008, opened 2024-07-03
node: remove websockets from SUI watcher

panoel:


node/cmd/guardiand/node.go
evan-gray, line 1: can we remove the unused `suiWS` parameter?

node/pkg/watchers/sui/watcher.go
```diff
@@ -45,7 +39,25 @@ type (
 		obsvReqC      <-chan *gossipv1.ObservationRequest
 		readinessSync readiness.Component
 
-		subId int64
+		subId                     int64
```
evan-gray, line 42: `subId` is no longer needed.

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,36 +303,31 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-				if err != nil {
-					logger.Error(fmt.Sprintf("ReadMessage: '%s'", err.Error()))
-					p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDSui, 1)
-					suiConnectionErrors.WithLabelValues("channel_read_error").Inc()
-					return err
-				}
-
-				var res SuiEventMsg
-				err = json.Unmarshal(msg, &res)
+				fmt.Println("Getting events...", time.Now().Format(time.RFC3339))
```
evan-gray, line 306: Was this intended to stay here? If so, should it be `logger.Debug`?

node/pkg/watchers/sui/watcher.go
```diff
@@ -155,30 +147,32 @@ var (
...
+		suiMoveEventType:          suiMoveEventType,
+		unsafeDevMode:             unsafeDevMode,
+		msgChan:                   messageEvents,
+		obsvReqC:                  obsvReqC,
+		readinessSync:             common.MustConvertChainIdToReadinessSyncing(vaa.ChainIDSui),
+		subId:                     0,
+		latestProcessedCheckpoint: 0,
+		maximumBatchSize:          10,
+		descendingOrder:           true, // Retrieve newest events first
+		loopDelayInSecs:           3,    // SUI produces a checkpoint every ~3 seconds
```
evan-gray, line 166: Perhaps this could go a little faster, like every second, so we don't introduce an up-to 3 second delay?

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,36 +303,31 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-					p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDSui, 1)
-					suiConnectionErrors.WithLabelValues("channel_read_error").Inc()
-					return err
-				}
-
-				var res SuiEventMsg
-				err = json.Unmarshal(msg, &res)
+				fmt.Println("Getting events...", time.Now().Format(time.RFC3339))
+				// This will return an array of events in the correct range and order
+				event, err := e.getEvents()
```
evan-gray, line 308: nit: `event` is actually the response wrapper, containing `Data` which is the list of events, right?

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,36 +303,31 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-				logger.Debug("SUI result message", zap.String("message", string(msg)), zap.Any("event", res))
-				if res.ID != nil {
-					logger.Error("Found an unexpected res.ID")
+					logger.Error(fmt.Sprintf("sui_data_pump Error: %s", err.Error()))
 					continue
 				}
-
-				if res.Params != nil && (*res.Params).Result != nil {
-					err := e.inspectBody(logger, *(*res.Params).Result, false)
+				for _, datum := range event.Data {
```
evan-gray, line 313: nit: which makes `datum` here an individual event? might be clearer to call the above `response` and this one `event`

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +450,154 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		})
+	}
+	return height, nil
+}
+
+func (e *Watcher) getCheckpointForEvent(event SuiResult) (int64, error) {
+	retVal := int64(0)
+	reader := strings.NewReader(fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "sui_getTransactionBlock", "params": [ "%s" ]}`,
+		*event.ID.TxDigest))
+	resp, err := http.Post(e.suiRPC, "application/json", reader) //nolint:noctx // TODO FIXME we should propagate context with Deadline here.
```
evan-gray, line 568: should this be fixed?

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +450,154 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		if firstTime {
+			reader = strings.NewReader(
+				fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "suix_queryEvents", "params": [{ "MoveEventType": "%s" }, null, %d, %t]}`,
+					w.suiMoveEventType, w.maximumBatchSize, w.descendingOrder))
+		} else {
+			reader = strings.NewReader(
+				fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "suix_queryEvents", "params": [{ "MoveEventType": "%s" }, { "txDigest": "%s", "eventSeq": "%s" }, %d, %t]}`,
+					w.suiMoveEventType, nextCursor.TxDigest, nextCursor.EventSeq, w.maximumBatchSize, w.descendingOrder))
+		}
+		resp, err := http.Post(w.suiRPC, "application/json", reader) //nolint:noctx // TODO FIXME we should propagate context with Deadline here.
```
evan-gray, line 479: should this be fixed?

node/pkg/watchers/sui/watcher.go
evan-gray: Same comment for the other occurrences.

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +450,154 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+
+		var res SuiEventResponse
+		err = json.Unmarshal(body, &res)
+		if err != nil {
+			return retVal, fmt.Errorf("suix_queryEvents failed to unmarshal body: %s, error: %w", string(body), err)
+		}
+		fmt.Println("Number of events:", len(res.Result.Data))
+		for _, datum := range res.Result.Data {
+			// Check if the event is newer than the last processed height
+			height, hErr := w.getCheckpointForEvent(datum)
```
evan-gray, line 497: In the dashboard I only fetched the last event's checkpoint instead of each one, which will save you some RPC calls. You can then, later use `sui_multiGetTransactionBlocks` to get all of the checkpoints instead of making an RPC call for each one later on. i.e. just calculate `done` and append the results first, then after having fetched at least to the desired checkpoint (after this for loop), use that batch call to get all of the blocks at once.

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +450,154 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		} else {
+			done = true
+		}
+	}
+
+	retVal.Data = reverseArray(retVal.Data)
+	return retVal, nil
+}
+
+func (e *Watcher) getLatestCheckpointSN(logger *zap.Logger) (int64, error) {
```
evan-gray, line 525: Can this function be used by `sui_block_height` as well?

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,36 +303,31 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+				for _, datum := range event.Data {
+					err = e.inspectBody(logger, datum, false)
 					if err != nil {
-						logger.Error(fmt.Sprintf("inspectBody: %s", err.Error()))
+						logger.Error(fmt.Sprintf("inspectBody Error: %s", err.Error()))
+						continue
+					}
+					// Get the checkpoint for the event
+					fmt.Println("Getting checkpoint for returned event")
+					lph, err := e.getCheckpointForEvent(datum)
```
evan-gray, line 321: `getEvents` already fetched all the checkpoints for each of these transactions, can it also return them so they don't have to be fetched again?

evan-gray (changes requested):
Seems to work, mostly clean up / efficiency comments.

node/pkg/watchers/sui/watcher.go
panoel: Yes. Changed.

PR #4009, opened 2024-07-05
Remove sui ws

panoel:


node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +443,146 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-	if u == nil || u.Scheme == "" || (u.Scheme != "" && u.Host == "") {
-		logger.Warn(fmt.Sprintf("DEPRECATED: Prefix --suiWS address with the url scheme e.g.: ws://%s or wss://%s", e.suiWS, e.suiWS))
-		u = &url.URL{Scheme: "ws", Host: e.suiWS}
-	} else if u.Scheme != "ws" && u.Scheme != "wss" {
-		return fmt.Errorf("invalid url scheme specified for --suiWS, try ws:// or wss://: %s", e.suiWS)
+func (e *Watcher) getCheckpointForEvent(event SuiResult) (int64, error) {
+	retVal := int64(0)
+	reader := strings.NewReader(fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "sui_getTransactionBlock", "params": [ "%s" ]}`,
+		*event.ID.TxDigest))
+	resp, err := http.Post(e.suiRPC, "application/json", reader) //nolint:noctx // TODO FIXME we should propagate context with Deadline here.
```
panoel, line 553: @evan-gray  Should this be fixed?

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +443,146 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		} else {
+			done = true
+		}
+	}
+
+	retVal.Data = reverseArray(retVal.Data)
+	return retVal, nil
+}
+
+func (e *Watcher) getLatestCheckpointSN(logger *zap.Logger) (int64, error) {
```
panoel, line 512: @evan-gray Can this function be used by sui_block_height as well?

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,35 +300,26 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-					err := e.inspectBody(logger, *(*res.Params).Result, false)
+				for _, event := range dataWithEvents.Data {
+					err = e.inspectBody(logger, event, false)
 					if err != nil {
-						logger.Error(fmt.Sprintf("inspectBody: %s", err.Error()))
+						logger.Error(fmt.Sprintf("inspectBody Error: %s", err.Error()))
+						continue
 					}
-					continue
+					lph, err := e.getCheckpointForEvent(event)
```
panoel, line 314: @evan-gray getEvents already fetched all the checkpoints for each of these transactions, can it also return them so they don't have to be fetched again?

node/pkg/watchers/sui/watcher.go
panoel: Yes.  Changed.

node/pkg/watchers/sui/watcher.go
panoel: This was how Toby left things.  Fix here or in a different PR?

node/pkg/watchers/sui/watcher.go
```diff
@@ -36,16 +31,31 @@ type (
...
-		msgChan       chan<- *common.MessagePublication
-		obsvReqC      <-chan *gossipv1.ObservationRequest
-		readinessSync readiness.Component
+		msgChan                   chan<- *common.MessagePublication
+		obsvReqC                  <-chan *gossipv1.ObservationRequest
+		readinessSync             readiness.Component
+		latestProcessedCheckpoint int64
+		maximumBatchSize          int
+		descendingOrder           bool
+		loopDelayInSecs           int
```
bruce-riley, line 44: Make this a `time.Duration` then initialize it below with:
`loopDelayInSecs: time.Second`
Then you don't need to do that conversion every time you use it.

node/pkg/watchers/sui/watcher.go
evan-gray: I realized that too after my comment. I guess it can be fixed in the future, though I don't know if there's an increased impact with the increased number of rpc calls.

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,35 +300,26 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-					logger.Error("Failed to unmarshal SuiEventMsg", zap.String("body", string(msg)), zap.Error(err))
-					return fmt.Errorf("Failed to unmarshal SuiEventMsg, body: %s, error: %w", string(msg), err)
-				}
-				if res.Error != nil {
-					return fmt.Errorf("Bad SuiEventMsg, body: %s, error: %w", string(msg), err)
-				}
-				logger.Debug("SUI result message", zap.String("message", string(msg)), zap.Any("event", res))
-				if res.ID != nil {
-					logger.Error("Found an unexpected res.ID")
+					logger.Error(fmt.Sprintf("sui_data_pump Error: %s", err.Error()))
```
bruce-riley, line 305: Can you change this (and any other instances) to the usual `zap.Error(err)` rather than using `fmt.Sprintf`?

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +420,141 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		TxDigest string
+		EventSeq string
+	}
+	firstTime := true
+	done := false
+	for !done {
+		var reader io.Reader
+		if firstTime {
+			reader = strings.NewReader(
+				fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "suix_queryEvents", "params": [{ "MoveEventType": "%s" }, null, %d, %t]}`,
```
bruce-riley, line 442: Since this string doesn't change, could you create it at start up and store it in w, so you don't have to do it every time?

node/pkg/watchers/sui/watcher.go
panoel: Changed.

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,35 +300,26 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-				logger.Debug("SUI result message", zap.String("message", string(msg)), zap.Any("event", res))
-				if res.ID != nil {
-					logger.Error("Found an unexpected res.ID")
+					logger.Error(fmt.Sprintf("sui_data_pump Error: %s", err.Error()))
 					continue
 				}
-
-				if res.Params != nil && (*res.Params).Result != nil {
-					err := e.inspectBody(logger, *(*res.Params).Result, false)
+				for _, event := range dataWithEvents.Data {
```
bruce-riley, line 308: Rather than using `reverseArray` in `getEvents`, couldn't you just do the following here:
`
if len(dataWithEvents.Data) > 0 {
   for idx := len(dataWithEvents.Data) - 1; (idx >= 0); idx-- {
      event := dataEvents.Data[idx]
`

Sorry the formatting is so ugly!

node/pkg/watchers/sui/watcher.go
panoel: Changed.

node/pkg/watchers/sui/watcher.go
panoel: Changed

node/pkg/watchers/sui/watcher.go
panoel: Changed

node/pkg/watchers/sui/watcher.go
panoel: Changed

node/pkg/watchers/sui/watcher.go
panoel: If I do a `context.WithTimeout()`, what should I set the timeout value to?
@evan-gray ☝️

bruce-riley: I'm not evan-gray, but I play one on TV. 😁 

Maybe five or ten seconds?

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +443,146 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		resp.Body.Close()
+
+		var res SuiEventResponse
+		err = json.Unmarshal(body, &res)
+		if err != nil {
+			return retVal, fmt.Errorf("suix_queryEvents failed to unmarshal body: %s, error: %w", string(body), err)
+		}
+		for _, datum := range res.Result.Data {
+			// Check if the event is newer than the last processed height
+			height, hErr := w.getCheckpointForEvent(datum)
```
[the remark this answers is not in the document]
panoel: Done is based on checkpoint.  Without actually getting the checkpoint for the event in question, how would I know if I am done?  Also, we are polling 3 times faster than blocks are being produced.  I think it is unlikely to get more than a few events of interest.  I'm happy to make this more efficient.  But I don't get what you are saying.

node/pkg/watchers/sui/watcher.go
```diff
@@ -351,36 +309,26 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+							logger.Error("inspectBody Error", zap.Error(err))
+							continue
+						}
+						if event.checkpoint > e.latestProcessedCheckpoint {
+							e.latestProcessedCheckpoint = event.checkpoint
+						}
 					}
-					continue
 				}
+				time.Sleep(time.Duration(e.loopDelay))
```
bruce-riley, line 331: `time.Sleep(e.loopDelay)`

bruce-riley (approved):

node/pkg/watchers/sui/watcher.go
panoel: Fixed.

bruce-riley (dismissed):

node/pkg/watchers/sui/watcher.go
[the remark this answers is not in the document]
evan-gray: On this point, I meant that instead of getting every event’s checkpoint in the first loop, you only need to get the _last_ event’s checkpoint since you are querying in descending order and that will be the oldest. That alone should be enough to calculate done, and gather all the applicable events. Then, after having gathered all the potentially applicable events, you can use the batch call to get all the checkpoints in one RPC call, then filter out the ones that are older than necessary.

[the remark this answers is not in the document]
panoel: Changed

node/pkg/watchers/sui/watcher.go
panoel: I made it a variable.

bruce-riley (approved):

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +440,208 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+func (w *Watcher) getEvents() ([]SuiResultInfo, error) {
+	// Only get events newer than the last processed height
+	var retVal []SuiResultInfo
+	var results []SuiResult
+	var txs []string
+	var nextCursor struct {
+		TxDigest string
+		EventSeq string
+	}
+	firstTime := true
```
johnsaigle, line 458: It looks like `firstTime` is never set to `false` so I'm not sure the else branch gets executed here.

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +440,208 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+		}
+		// Get and check the checkpoint for the last event against the lastProcessedHeight to see if we are done.
+		height, hErr := w.getCheckpointForDigest(txs[len(txs)-1])
+		if hErr != nil {
+			return retVal, hErr
+		}
+		if height <= w.latestProcessedCheckpoint {
+			break
+		}
+		if (!done) && res.Result.HasNextPage {
```
johnsaigle, line 491: I think `done` is always `false` here because this is a `for <bool>` loop.
With that in mind, I think this section could be simplified to:

```go
    ...
    if !res.Result.Has.NextPage {
        break
    }
    nextCursor.TxDigest = res.Result.NextCursor.TxDigest
    nextCursor.EventSeq = res.Result.NextCursor.EventSeq
}
```
In that case you could also change the loop to `for { ... }` instead of `for !done {...}`. 

IMO this style makes reasoning about control flow a little easier. (Compared with tracking a mix of `break`s, `return`s, and checks against `done`.)

Anyway it's a style thing at the end of the day so feel free to ignore.

johnsaigle (dismissed):

node/pkg/watchers/sui/watcher.go
panoel: Good catch.  The else branch needs to be used for subsequent calls in order to index to the next page of results.

node/pkg/watchers/sui/watcher.go
panoel: Sounds reasonable.  Changing.

node/pkg/watchers/sui/watcher.go
```diff
@@ -393,35 +362,12 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-				var res SuiCheckpointSN
-				err = json.Unmarshal(body, &res)
-				if err != nil {
-					logger.Error("unmarshal failed into uint64", zap.String("body", string(body)), zap.Error(err))
-					p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDSui, 1)
-					return fmt.Errorf("sui_getLatestCheckpointSequenceNumber failed to unmarshal body: %s, error: %w", string(body), err)
-				}
-
-				height, pErr := strconv.ParseInt(res.Result, 0, 64)
+				height, pErr := e.getLatestCheckpointSN(logger)
```
evan-gray, line 365: `pErr` can probably just be `err` now since there's only one :slightly_smiling_face:

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +440,204 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+	if (len(mbRes) == 0) || (len(mbRes) != len(txs)) {
+		return retVal, errors.New("getEvents error getting multiple blocks")
+	}
+	for idx, block := range mbRes {
+		cp, err := strconv.ParseInt(block.Checkpoint, 10, 64)
+		if err != nil {
+			return retVal, fmt.Errorf("getEvents failed to ParseInt: %w", err)
+		}
+		if cp > w.latestProcessedCheckpoint {
+			// Can we double check the digests here?
```
evan-gray, line 510: I feel like we can, since it's doing it here

evan-gray (changes requested):
couple nits but looks good.

node/pkg/watchers/sui/watcher.go
panoel: Removed

node/pkg/watchers/sui/watcher.go
panoel: Changed

node/pkg/watchers/sui/watcher.go
```diff
@@ -503,25 +440,204 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+}
+
+func (w *Watcher) createAndExecReq(payload string) ([]byte, error) {
+	var retVal []byte
+	ctx, cancel := context.WithTimeout(context.Background(), w.postTimeout)
+	defer cancel()
+	// Create a new request with the context
+	req, err := http.NewRequestWithContext(ctx, "POST", w.suiRPC, strings.NewReader(payload))
+	if err != nil {
+		return retVal, fmt.Errorf("getMultipleBlocks failed to create request: %w", err)
```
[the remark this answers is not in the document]
panoel: Changed

evan-gray (approved):

bruce-riley (approved):

PR #4016, opened 2024-07-09
node: Flow cancel enhancements and bug fixes

johnsaigle:
It may be helpful to read commits individually while reviewing.

### Bug fixes

- Flow cancel did not trigger when pending transfers were released be3696a2b4354457e779b495e9e7a0b3269afa30
- Flow cancel did not trigger when loading pending transfers from the database 0524d8b6e3b61f9004a59cecb395c65ef9a0960a
- Changed error handling to prevent "crash looping" when Governor usage exceeds daily limit 013d79acb47ed1e561aa0ab72ce3ad8f2e2bc68a

### Additional features

- **Toggle Flow Cancel**: Added a command-line flag to enable or disable flow-canceling when starting the node. This should allow Guardians to disable flow canceling in the case of future bugs or during a security incident. This should prevent the need to rollback to earlier Guardian versions. (@mdulin2 ) 9120d17760c89d215ef43a0279d45c1ae3ae8825
- **Deterministic iteration**: Use deterministic ordering when iterating over chain entries and modifying their state. 38c79eef5f0f543dfbf3de33209472729bb8c042
- **Addional metrics**: Modify the monitoring code and protobuf files to make the status of the Governor more legible when flow-canceling is enabled. This can be consumed by Wormhole Dashboard to better reflect the effects of flow cancelling. a29158ce72b0ea19ee9e6844d65b3cab6f5cd2ff https://github.com/wormhole-foundation/wormhole/commit/eb13ff9db9029886bf8d138932d094ee9ee8054d


node/pkg/governor/governor_test.go
```diff
@@ -1061,6 +1792,69 @@ func TestTestnetConfigIsValid(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestNumDaysForReleaseTimerReset(t *testing.T) {
```
johnsaigle, line 1795: Note to self: delete this. It's from Jason's PR

node/pkg/governor/governor_test.go
```diff
@@ -1785,76 +2109,13 @@ func TestMainnetConfigIsValid(t *testing.T) {
...
 	var db db.MockGovernorDB
-	gov := NewChainGovernor(logger, &db, common.GoTest)
+	gov := NewChainGovernor(logger, &db, common.GoTest, true)
 
 	gov.env = common.TestNet
 	err := gov.initConfig()
 	require.NoError(t, err)
 }
 
-func TestNumDaysForReleaseTimerReset(t *testing.T) {
```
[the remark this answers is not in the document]
johnsaigle: Ah this came up during rebasing. I changed the branch target a few times. I'll restore this section

node/pkg/governor/governor_monitoring.go
```diff
@@ -244,68 +245,112 @@ func (gov *ChainGovernor) resetReleaseTimerForTime(vaaId string, now time.Time,
...
+func sumValue(transfers []transfer, startTime time.Time) (netNotional int64, smallTxOutgoingNotional uint64, flowCancelNotional uint64, err error) {
 	if len(transfers) == 0 {
-		return 0, nil
+		return 0, 0, 0, nil
 	}
 
-	var sum int64
+	// Sum of all outgoing small tranasfers minus incoming flow cancel transfers. Big transfers are excluded
+	netNotional = int64(0)
+	smallTxOutgoingNotional = uint64(0)
```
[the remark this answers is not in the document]
johnsaigle: I prefer the verbosity here. It took me a long time to sort through the different flow cancel edge-cases and I want it to be really clear going forward. Even if the variable name is long I think the concept it's describing represents a net reduction in mental effort for understanding what the code does

node/pkg/governor/governor_test.go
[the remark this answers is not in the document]
johnsaigle: Let me know if this looks better from your side now

pleasew8t (approved):

node/pkg/governor/governor.go
```diff
@@ -132,6 +135,7 @@ func newTransferFromDbTransfer(dbTransfer *db.Transfer) (tx transfer, err error)
 // ensure that the Governor usage cannot be lowered due to malicious or invalid transfers.
 // - the Value must be negative (in order to represent an incoming value)
 // - the TargetChain must match the chain ID of the Chain Entry
+// - the flow cancel value must always be less than the big transfer limit
```
bruce-riley, line 138: What if we put out a release that reduces the big transfer size and there is something in the 24 hour window that exceeds the new BTS? Would that violate this constraint? I tried looking through the code, but I don't see where that's handled. I could've missed something though.

node/pkg/governor/governor.go
```diff
@@ -179,12 +183,15 @@ func (ce *chainEntry) isBigTransfer(value uint64) bool {
...
-	chains                map[vaa.ChainID]*chainEntry  // protected by `mutex`
+	db                  db.GovernorDB // protected by `mutex`
+	logger              *zap.Logger
+	mutex               sync.Mutex
+	tokens              map[tokenKey]*tokenEntry    // protected by `mutex`
+	tokensByCoinGeckoId map[string][]*tokenEntry    // protected by `mutex`
+	chains              map[vaa.ChainID]*chainEntry // protected by `mutex`
+	// We maintain a sorted slice of governed chainIds so we can iterate over maps in a deterministic way
+	// This slice should be sorted in ascending order by (Wormhole) Chain ID.
+	chainIds              []vaa.ChainID
```
bruce-riley, line 194: If you make this a slice of `*chainEntry` you would not have to do map look ups everywhere you iterate through it. Of course they you should probably rename it to something like `chainsById`.

node/pkg/governor/governor.go
```diff
@@ -651,10 +680,23 @@ func (gov *ChainGovernor) parseMsgAlreadyLocked(
...
+// moving a `dbTransfer` element from `pending` to `transfers`. Returns a slice of Messages that will be published.
+// A transfer is ready to be released when one of the following conditions holds:
+//   - The 'release time' duration has passed since `now` (i.e. the transfer has been queued for 24 hours, regardless of
+//     the Governor's current capacity)
+//   - Within the release time duration, other transfers have been processed and have freed up outbound Governor capacity.
+//     This happens either because other transfers get released after 24 hours or because incoming transfers of
+//     flow-cancelling assets have freed up outbound capacity.
+//
+// WARNING: When this function returns an error, it propagates to the `processor` which in turn interprets this as a
+// signal to RESTART THE NODE. Therefore, errors returned by this function effectively act as panics.
```
bruce-riley, line 699: Doesn't restart the whole node, just the processor (which is still very bad).

node/pkg/governor/governor.go
```diff
@@ -651,10 +680,23 @@ func (gov *ChainGovernor) parseMsgAlreadyLocked(
 	return true, ce, token, payload, nil
 }
 
+// CheckPending is a wrapper method for CheckPendingForTime. It is called by the processor with the purpose of releasing
+// queued transfers.
 func (gov *ChainGovernor) CheckPending() ([]*common.MessagePublication, error) {
 	return gov.CheckPendingForTime(time.Now())
 }
 
+// CheckPendingForTime If a pending message is ready to be released, modifies the chain entry's `pending` and `transfers` slices by
```
bruce-riley, line 689: Nit: "CheckPendingForTime checks if a pending message is ready to be released, and if so, modifies..."

node/pkg/governor/governor.go
```diff
@@ -669,15 +711,26 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
 		gov.msgsToPublish = nil
 	}
 
-	for _, ce := range gov.chains {
+	// Iterate deterministically by accessing keys from this slice instead of the chainEntry map directly
+	for _, chainId := range gov.chainIds {
+		ce, ok := gov.chains[chainId]
+		if !ok {
```
bruce-riley, line 717: Since this happens on the processor routine, I think it would definitely be better to avoid these map look ups and replace `gov.chainIds` with a slice of pointers.

node/pkg/governor/governor.go
```diff
@@ -762,26 +818,48 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
...
-							gov.msgsToPublish = msgsToPublish
+						transfer, err := newTransferFromDbTransfer(&dbTransfer)
+						if err != nil {
+							// Should never occur unless dbTransfer.Value overflows MaxInt64
+							gov.logger.Error("could not convert dbTransfer to transfer",
+								zap.String("msgID", dbTransfer.MsgID),
+								zap.String("hash", pe.hash),
+								zap.Error(err),
+							)
+							// This causes the process to die. We don't want to process transfers that
```
bruce-riley, line 829: Nit: "processor" not "process".

Same comment below.

node/pkg/governor/governor.go
```diff
@@ -836,46 +913,46 @@ func computeValue(amount *big.Int, token *tokenEntry) (uint64, error) {
...
 // SECURITY Invariant: The `sum` return value should never be less than 0
-// SECURITY Invariant: The `sum` return value should never exceed the "Governor limit" for the chain
-func (gov *ChainGovernor) TrimAndSumValueForChain(emitter *chainEntry, startTime time.Time) (sum uint64, err error) {
-	// Sum the value of all outgoing transfers
-	var sumOutgoing int64
-	sumOutgoing, emitter.transfers, err = gov.TrimAndSumValue(emitter.transfers, startTime)
+func (gov *ChainGovernor) TrimAndSumValueForChain(chainEntry *chainEntry, startTime time.Time) (sum uint64, err error) {
+	if chainEntry == nil {
+		// We don't expect this to happen but this prevents a nil pointer deference
+		return 0, errors.New("TrimAndSumValeForChain parameter chainEntry must not be nil")
```
bruce-riley, line 932: Seems like a program bug, so maybe panic is better?

node/pkg/governor/governor_test.go
```diff
@@ -185,6 +185,21 @@ func TestTrimEmptyTransfers(t *testing.T) {
 	assert.Equal(t, 0, len(updatedTransfers))
 }
 
+// Make sure that the code doesn't panic if called with a nil chainEntry
+func TestTrimAndSumValueForChainReturnsErrorForNilChainEntry(t *testing.T) {
```
bruce-riley, line 189: I think it should panic because this would be a program bug, right?

node/pkg/governor/governor_db.go
```diff
@@ -243,15 +243,19 @@ func (gov *ChainGovernor) reloadTransfer(xfer *db.Transfer) error {
...
+	// Reload flow-cancel transfers for the TargetChain. This is important when the node restarts so that a corresponding,
+	// inverse transfer is added to the TargetChain. This is already done during the `ProcessMsgForTime` and
+	// `CheckPending` loops but those functions do not capture flow-cancelling when the node is restarted.
 	tokenEntry := gov.tokens[tk]
 	if tokenEntry != nil {
 		// Mandatory check to ensure that the token should be able to reduce the Governor limit.
 		if tokenEntry.flowCancels {
 			if destinationChainEntry, ok := gov.chains[xfer.TargetChain]; ok {
 				if err := destinationChainEntry.addFlowCancelTransferFromDbTransfer(xfer); err != nil {
+					gov.logger.Warn("could not add flow canceling transfer to destination chain",
```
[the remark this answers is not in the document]
johnsaigle: Sure I'm not opposed to upgrading the severity to Error.

node/pkg/governor/governor_prices.go
```diff
@@ -309,7 +309,7 @@ func CheckQuery(logger *zap.Logger) error {
 	logger.Info("Instantiating governor.")
 	ctx := context.Background()
 	var db db.MockGovernorDB
-	gov := NewChainGovernor(logger, &db, common.MainNet)
+	gov := NewChainGovernor(logger, &db, common.MainNet, false)
```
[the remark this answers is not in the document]
johnsaigle: Yes, thanks!

proto/gossip/v1/gossip.proto
```diff
@@ -156,6 +156,7 @@ message ChainGovernorConfig {
   int64 timestamp = 3;
   repeated Chain chains = 4;
   repeated Token tokens = 5;
+  bool flow_cancel_enabled = 6;
```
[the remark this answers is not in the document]
johnsaigle: Sure I'll take a look at this and get back to you

node/pkg/governor/governor.go
johnsaigle: Do you know the effects this would have on the processor? I want to avoid a situation where a bug in the Governor logic causes it or related processes to restart. I see your point that this is a serious problem we don't want to encounter but I'm hesitant to add panics or otherwise affect liveness

node/pkg/governor/governor.go
johnsaigle: Do you mean changing `chains` from a map to a slice and removing `chainIds`?

node/pkg/governor/governor.go
johnsaigle: Thanks I'll update the comment!

node/pkg/governor/governor.go
bruce-riley: I guess I wasn't as concerned about "governor wonkiness" like the limits being a bit messed up until it clears. I just wanted to make sure we wouldn't panic or kill the processor in that case. If you feel that's covered, okay. (Or maybe have a test to verify that. :grin:).

node/pkg/governor/governor.go
bruce-riley: No, keep `chains` as is so we can do a map look up by chainID. Just replace `chainIds` with `chainsById` where the payload would be `*chainEntry` so you could use it to directly access the chain entry without doing that extra map lookup.

node/pkg/governor/governor.go
bruce-riley: Returning an error would cause the processor to restart, which is a silent failure unless someone is watching the logs.

A panic would kill the application, which should be much more visible.

My point was is that this should only happen in the case of a program bug, as in someone broke the code, which should be detected during testing, and never make it to production.

node/pkg/governor/governor.go
johnsaigle: Ah got it. The only remaining case where the processor should restart is here: https://github.com/wormhole-foundation/wormhole/pull/4016/files#diff-f68b867da294e995f0449fff8116515f60d0e718da580251c6fab60f5c4ce4abR277

This represents the sum of the transfers overflowing the bounds of int64

node/pkg/governor/governor.go
johnsaigle: The code that calls should `TrimAndSumValueForChain` log and swallows the error in all cases rather than propagating it up to the processor. (This is the intention anyway, let me know if you see otherwise.)

The idea here wasn't too fail silently or loudly in the sense of causing the processor to restart. Instead the error is logged.

My understanding is that the only path to restart the processor is via [returning an error from `CheckPending`](https://github.com/wormhole-foundation/wormhole/blob/80c26eb8729a4aab5ff338c1ac01a526531ead50/node/pkg/processor/processor.go#L264-L267). Errors from `ProcessMsg` do not cause a restart; [instead there's a continue.](https://github.com/wormhole-foundation/wormhole/blob/80c26eb8729a4aab5ff338c1ac01a526531ead50/node/pkg/processor/processor.go#L231)

node/pkg/governor/governor.go
johnsaigle: Could we break this out into a separate issue and not have it block the PR?
I see how it might be more optimal to iterate over pointers directly but I think having both data structures also means that we need to keep the pointers in sync.

I tried making this change and it broke a lot of things in the test suite. I think the reason is that there are helper functions there to update the `chainEntry` pointer in the map, which is just fine. But it means every time we do that we also need to update the `chainsById` slice. In the actual governor code we don't update the `chainEntry` pointer in the map but it feels like an accident waiting to happen. 

My preference would be to keep it simple and iterate over a static list of ~30 ints rather than worry about keeping these two things in sync.

proto/gossip/v1/gossip.proto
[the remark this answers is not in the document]
johnsaigle: Added here: https://github.com/wormhole-foundation/wormhole/pull/4016/commits/eb13ff9db9029886bf8d138932d094ee9ee8054d

node/pkg/governor/governor.go
bruce-riley: I do feel that it would be better to eliminate a bunch of unnecessary map look ups, especially in any code called from the main processor routine, but I'm okay with deferring that to a future PR.

johnsaigle: Thanks I appreciate the flexibility. I'll write up a GitHub Issue summarizing the above

node/pkg/governor/governor.go
johnsaigle: I've added https://github.com/wormhole-foundation/wormhole/issues/4029 so we can discuss future performance enhancements

johnsaigle: #4029 recorded as an Issue here

node/pkg/governor/governor_prices.go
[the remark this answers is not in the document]
johnsaigle: Just taking a look at this more closely. I don't think the check query test function interacts with flow cancelling at all. The flow cancel token list doesn't store price information, only the OriginChain, OriginAddress, and Symbol.

node/pkg/governor/governor_test.go
johnsaigle: Reposting [a related comment](https://github.com/wormhole-foundation/wormhole/pull/4016#discussion_r1683201546) for easier reviewing:

> The code that calls should TrimAndSumValueForChain log and swallows the error in all cases rather than propagating it up to the processor. (This is the intention anyway, let me know if you see otherwise.)
The idea here wasn't too fail silently or loudly in the sense of causing the processor to restart. Instead the error is logged.
My understanding is that the only path to restart the processor is via [returning an error from CheckPending](https://github.com/wormhole-foundation/wormhole/blob/80c26eb8729a4aab5ff338c1ac01a526531ead50/node/pkg/processor/processor.go#L264-L267). Errors from ProcessMsg do not cause a restart; [instead there's a continue.](https://github.com/wormhole-foundation/wormhole/blob/80c26eb8729a4aab5ff338c1ac01a526531ead50/node/pkg/processor/processor.go#L231)

node/pkg/governor/governor_prices.go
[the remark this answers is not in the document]
bruce-riley: What I meant is that if the flow cancel token list was bad (it returns an error if `vaa.StringToAddress` fails), passing in true above would cause `check_query` to fail, allowing the problem to be caught earlier.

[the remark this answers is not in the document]
johnsaigle: Ah I see. Sure, no harm adding it

pleasew8t (approved):

node/pkg/governor/governor.go
```diff
@@ -762,26 +821,48 @@ func (gov *ChainGovernor) CheckPendingForTime(now time.Time) ([]*common.MessageP
...
+							// have USD value in excess of MaxInt64 under any circumstances.
+							// This check should occur before the call to the database so
+							// that we don't store a problematic transfer.
 							return nil, err
 						}
 
-						transfer, err := newTransferFromDbTransfer(&dbTransfer)
-						if err != nil {
+						if err := gov.db.StoreTransfer(&dbTransfer); err != nil {
+							gov.msgsToPublish = msgsToPublish
```
djb15, line 840: What's the purpose of the assignment here? Especially if the processor is going to die?

node/pkg/governor/governor_monitoring.go
```diff
@@ -244,62 +245,107 @@ func (gov *ChainGovernor) resetReleaseTimerForTime(vaaId string, now time.Time,
...
-		} else {
-			value = ce.dailyLimit - value
+
+		remaining := gov.availableNotionalValue(chainId, netUsage)
+
+		if !gov.flowCancelEnabled {
+			// When flow cancel is disabled, we expect that both the netUsage and remaining notional should be
+			// within the range of [0, dailyLimit]. Flow cancel allows flexibility here. netUsage may be
+			// negative if there is a lot of incoming flow; conversely, it may exceed dailyLimit if incoming
+			// flow added space, allowed additional transfers through, and then expired after 24h.
```
djb15, line 326: I don't think this is true if flow cancel was on but is then switched off, right? I think keeping the warnings for this is fine (since this toggling of flow cancel from on to off is an edge case), but probably worth clarifying in the comment?

bruce-riley (dismissed):

node/pkg/governor/governor.go
johnsaigle: Fair point. It's not very clear from the diff but I think I just moved (old) lines from 765-767 down here. Agreed it seems unnesscary

djb15 (approved):

djb15 (dismissed):

djb15 (approved):

bruce-riley (approved):

proto/gossip/v1/gossip.proto
```diff
@@ -182,6 +183,9 @@ message ChainGovernorStatus {
     string emitter_address = 1; // human-readable hex-encoded (leading 0x)
     uint64 total_enqueued_vaas = 2;
     repeated EnqueuedVAA enqueued_vaas = 3; // Only the first 20 will be included.
+    int64 small_tx_net_notional_value = 4;
```
evan-gray, line 186: Just confirming, should this be here instead of under `Chain` alongside `remaining_available_notional`? Does each emitter have it's own set of notional values in a flow cancel world?

evan-gray (approved):
Reviewed proto. Approved pending confirmation of proto change.

proto/gossip/v1/gossip.proto
johnsaigle: Yeah maybe under `Chain` makes more sense especially since that's where the existing `notional` measure is. I'll add a new commit.

evan-gray: Okay, changing my review to request changes in the interim

evan-gray (changes requested):
Holding for proto change.

evan-gray (approved):
reviewed proto

bruce-riley (approved):

PR #4022, opened 2024-07-17
Coin Gecko API Key Support

mdulin2:
In case the Guardians want to use API keys for Coin Gecko, support for the API keys has been added. This changes code in the Guardian and adds a command line flag.

Testing this is difficult because we don't want to include a key in here. Prior to this, I don't think we had tests for the query success from CoinGecko anyway. So, I added tests that make requests to coin gecko using the standard and pro APIs. If the API key isn't specified in the test file, the test is just skipped. If there's a better way to do this or relying on external APIs for builds to succeed is bad practice, then please say something and I'm happy to make changes accordingly.


mdulin2:
Included too many commits. I need to fix the history of this code before making the PR.

PR #4025, opened 2024-07-18
Add coin gecko pro API usage for the governor

mdulin2:
In case the Guardians want to use API keys for Coin Gecko, support for the API keys has been added. This changes code in the Guardian and adds a command line flag.

Testing this is difficult because we don't want to include a key in here. Prior to this, I don't think we had tests for the query success from CoinGecko anyway. So, I added tests that make requests to coin gecko using the standard and pro APIs. If the API key isn't specified in the test file, the test is just skipped. If there's a better way to do this or relying on external APIs for builds to succeed is bad practice, then please say something and I'm happy to make changes accordingly.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -245,6 +245,7 @@ var (
 
 	chainGovernorEnabled      *bool
 	governorFlowCancelEnabled *bool
+	coinGeckoApiKey           *string
 
 	ccqEnabled           *bool
 	ccqAllowedRequesters *string
@@ -462,6 +463,7 @@ func init() {
 
 	chainGovernorEnabled = NodeCmd.Flags().Bool("chainGovernorEnabled", false, "Run the chain governor")
 	governorFlowCancelEnabled = NodeCmd.Flags().Bool("governorFlowCancelEnabled", false, "Enable flow cancel on the governor")
+	coinGeckoApiKey = NodeCmd.Flags().String("coinGeckoApiKey", "", "Coin gecko pro API key. Defaults to off via no API key")
 
 	ccqEnabled = NodeCmd.Flags().Bool("ccqEnabled", false, "Enable cross chain query support")
 	ccqAllowedRequesters = NodeCmd.Flags().String("ccqAllowedRequesters", "", "Comma separated list of signers allowed to submit cross chain queries")
@@ -866,6 +868,10 @@ func runNode(cmd *cobra.Command, args []string) {
 
 	if !argsConsistent([]string{*gatewayContract, *gatewayWS, *gatewayLCD}) {
 		logger.Fatal("Either --gatewayContract, --gatewayWS and --gatewayLCD must all be set or all unset")
+	}
+
+	if !*chainGovernorEnabled && *coinGeckoApiKey != "" {
+		logger.Fatal("If coinGeckoApiKey is set, then chainGovernorEnabled must be set")
 	}
 
 	var publicRpcLogDetail common.GrpcLogDetail
@@ -1673,7 +1679,7 @@ func runNode(cmd *cobra.Command, args []string) {
 		node.GuardianOptionDatabase(db),
 		node.GuardianOptionWatchers(watcherConfigs, ibcWatcherConfig),
 		node.GuardianOptionAccountant(*accountantWS, *accountantContract, *accountantCheckEnabled, accountantWormchainConn, *accountantNttContract, accountantNttWormchainConn),
-		node.GuardianOptionGovernor(*chainGovernorEnabled, *governorFlowCancelEnabled),
+		node.GuardianOptionGovernor(*chainGovernorEnabled, *governorFlowCancelEnabled, *coinGeckoApiKey),
 		node.GuardianOptionGatewayRelayer(*gatewayRelayerContract, gatewayRelayerWormchainConn),
 		node.GuardianOptionQueryHandler(*ccqEnabled, *ccqAllowedRequesters),
 		node.GuardianOptionAdminService(*adminSocketPath, ethRPC, ethContract, rpcMap),
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -463,7 +463,7 @@ func init() {
 
 	chainGovernorEnabled = NodeCmd.Flags().Bool("chainGovernorEnabled", false, "Run the chain governor")
 	governorFlowCancelEnabled = NodeCmd.Flags().Bool("governorFlowCancelEnabled", false, "Enable flow cancel on the governor")
-	coinGeckoApiKey = NodeCmd.Flags().String("coinGeckoApiKey", "", "Coin gecko pro API key. Defaults to off via no API key")
+	coinGeckoApiKey = NodeCmd.Flags().String("coinGeckoApiKey", "", "CoinGecko Pro API key. If no API key is provided, CoinGecko requests may be throttled or blocked.")
 
 	ccqEnabled = NodeCmd.Flags().Bool("ccqEnabled", false, "Enable cross chain query support")
 	ccqAllowedRequesters = NodeCmd.Flags().String("ccqAllowedRequesters", "", "Comma separated list of signers allowed to submit cross chain queries")
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -211,7 +211,6 @@ func NewChainGovernor(
 	flowCancelEnabled bool,
 	coinGeckoApiKey string,
 ) *ChainGovernor {
-
 	return &ChainGovernor{
 		db:                  db,
 		logger:              logger.With(zap.String("component", "cgov")),
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -211,6 +211,7 @@ func NewChainGovernor(
 	flowCancelEnabled bool,
 	coinGeckoApiKey string,
 ) *ChainGovernor {
+
 	return &ChainGovernor{
 		db:                  db,
 		logger:              logger.With(zap.String("component", "cgov")),
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -201,6 +201,7 @@ type ChainGovernor struct {
 	statusPublishCounter  int64
 	configPublishCounter  int64
 	flowCancelEnabled     bool
+	coinGeckoApiKey       string
 }
 
 func NewChainGovernor(
@@ -208,7 +209,9 @@ func NewChainGovernor(
 	db db.GovernorDB,
 	env common.Environment,
 	flowCancelEnabled bool,
+	coinGeckoApiKey string,
 ) *ChainGovernor {
+
 	return &ChainGovernor{
 		db:                  db,
 		logger:              logger.With(zap.String("component", "cgov")),
@@ -218,6 +221,7 @@ func NewChainGovernor(
 		msgsSeen:            make(map[string]bool),
 		env:                 env,
 		flowCancelEnabled:   flowCancelEnabled,
+		coinGeckoApiKey:     coinGeckoApiKey,
 	}
 }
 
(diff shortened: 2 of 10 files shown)


mdulin2:
I'm currently trying to fix up the tests that are failing. The Guardian doesn't like to build on Macbooks.

mdulin2:
Once John's changes for cancel flow are approved at https://github.com/wormhole-foundation/wormhole/pull/4016 I'll have to fix a few locations since it touches similar code.

node/cmd/guardiand/node.go
```diff
@@ -431,6 +432,7 @@ func init() {
 	promRemoteURL = NodeCmd.Flags().String("promRemoteURL", "", "Prometheus remote write URL (Grafana)")
 
 	chainGovernorEnabled = NodeCmd.Flags().Bool("chainGovernorEnabled", false, "Run the chain governor")
+	coinGeckoApiKey = NodeCmd.Flags().String("coinGeckoApiKey", "", "Coin gecko pro API key. Defaults to off via no API key")
```
djb15, line 435: Nit pick on the string here, wdyt of the suggestion below?

"CoinGecko Pro API key. If no API key is provided, CoinGecko requests may be throttled or blocked"

node/pkg/governor/governor.go
```diff
@@ -194,13 +194,16 @@ type ChainGovernor struct {
...
+	coinGeckoApiKey       string
 }
 
 func NewChainGovernor(
 	logger *zap.Logger,
 	db db.GovernorDB,
 	env common.Environment,
+	coinGeckoApiKey string,
 ) *ChainGovernor {
+
```
djb15, line 206: Nit, unnecessary new line

djb15 (commented):
Not sure if this is the best way to do it, but for the Pro API test we could load a string from an environment variable? That way we could add a key as a secret in GitHub and load it in the Action

node/pkg/governor/governor_prices.go
```diff
@@ -88,19 +88,30 @@ func createCoinGeckoQueries(idList []string, tokensPerQuery int) []string {
...
 // createCoinGeckoQuery creates a CoinGecko query for the specified set of IDs.
-func createCoinGeckoQuery(ids string) string {
+func createCoinGeckoQuery(ids string, coinGeckoApiKey string) string {
 	params := url.Values{}
 	params.Add("ids", ids)
 	params.Add("vs_currencies", "usd")
 
-	query := "https://api.coingecko.com/api/v3/simple/price?" + params.Encode()
+	// If modifying this code, ensure that the test 'TestCoinGeckoPriceChecks' passes when adding a pro API key to it.
+	// Since the code is requires an API key (which we don't want to publish to git), this
```
bruce-riley, line 104: Nit: Wording ("is requires").

node/pkg/governor/governor_test.go
```diff
@@ -2808,6 +2808,100 @@ func TestCoinGeckoQueries(t *testing.T) {
...
+// Test the URL of CoinGecko queries to be correct
+func TestCoinGeckoQueryFormat(t *testing.T) {
+	id_amount := 10
+	ids := make([]string, id_amount)
+	for idx := 0; idx < id_amount; idx++ {
+		ids[idx] = fmt.Sprintf("id%d", idx)
+	}
+
+	// Create and parse the query
+	queries := createCoinGeckoQueries(ids, 100, "") // Not API key
```
bruce-riley, line 2820: Nit: "No API key"?

node/pkg/governor/governor_test.go
```diff
@@ -2808,6 +2808,100 @@ func TestCoinGeckoQueries(t *testing.T) {
...
+	require.Equal(t, params.Has("x_cg_pro_api_key"), true)
+	require.Equal(t, params["x_cg_pro_api_key"][0], "FAKE_KEY")
+	require.Equal(t, params.Has("vs_currencies"), true)
+	require.Equal(t, params["vs_currencies"][0], "usd")
+	require.Equal(t, params.Has("ids"), true)
+
+}
+
+// Test that the prices are gathering from the CoinGecko API
+func TestCoinGeckoPriceChecks(t *testing.T) {
```
bruce-riley, line 2857: When I originally wrote the stuff to query Coin Gecko, I suggested writing tests to do queries, and some people (@SEJeff  as I recall) were adamant that go tests should not require access to the internet. That's why I went the `check_query.go` route instead. Not sure if that feeling has changed.

bruce-riley (approved):

node/pkg/governor/governor_test.go
johnsaigle: 👍🏻 I think connecting to the internet gets into integration test territory. Maybe Tilt tests would be more appropriate?

mdulin2: Not testing this functionality at all in the codebase seems like a mistake waiting to happen tbh. What's wrong with connecting to the Internet on this test?

mdulin2:
> Not sure if this is the best way to do it, but for the Pro API test we could load a string from an environment variable? That way we could add a key as a secret in GitHub and load it in the Action

Good thought! I wasn't sure if testing this small feature was worth the pain of doing something like that. What do you think? 

Also, it depends on whether we can even connect to the Internet during our tests.

node/pkg/governor/governor_test.go
johnsaigle: It's kind of a philosophical thing but a lot of people have the view that unit tests should strictly test your own functions/things strictly under your control/only your software and not interactions with other software.

So in this case that interaction is with the network, but similarly an author might avoid interacting with a database or the file system. 

In these circumstances the unit tests aren't very unit-y. Here we would be testing 1) parsing JSON into Go structs but also 2) our connectivity to CoinGecko. If CoinGecko is down, should this make our build fail? If a developer is running tests against an unrelated feature, how will they know whether they broke logic or whether this test failed due to internet connectivity issues? What if CoingGecko changes their API interface or the format of their responses? We might not want changes in their software to break our software development pipeline. 

We might also want to run the unit tests in a restricted environment that doesn't have general outgoing network access. 

I think if you want to test this you could try to mock it: in-line JSON of the right format here and make sure that the data structures get parsed correctly.
You could also try breaking the functions into smaller pieces. You isolate the network request to a function that's a single line long, and then unit test every thing else around it so you can be sure that part is working fine.

djb15: > What if CoingGecko changes their API interface or the format of their responses?

This is exactly what I think would be great to test actually. Would mean we catch any breaking changes early. If there's a way to differentiate using the HTTP error codes between CoinGecko being down and any other issues, that could alleviate some of the downside of having an internet connected test

mdulin2: I see the point. It may make the testing unstable. If the API bans the IP testing or is having issues resolving for whatever reason then it would cause the test to fail for no reason. Hmmmm. 

If the API interface breaks for CoinGecko, we would WANT to know that in our tests I would think; I would agree with Dirk on that. Additionally, I want people who are touching this code to make sure that the endpoint is correct and configured properly when the code is modified. 

> I think if you want to test this you could try to mock it: in-line JSON of the right format here and make sure that the data structures get parsed correctly.

I feel this would test SOME of the code paths but not all of them. This is why I wrote the tests this may.

mdulin2: @bruce-riley Here are the options that I see:

- Leave the test as is with the network requests
- Change the test to *mock* the response in some way. This would miss large parts I wanted to test though.
- Comment out the test but keep it for people who want to use it. 
- Remove the test altogether. 

What are your thoughts?

SEJeff: > Not testing this functionality at all in the codebase seems like a mistake waiting to happen tbh. What's wrong with connecting to the Internet on this test?

Unit tests and for testing controlled environments. Relying on an external web service is NOT a unit test. This is the wrong way to do software engineering. You can mockup an example coingecko server with fixture json returned with the [httptest](https://pkg.go.dev/net/http/httptest) package. @bruce-riley is correct here.

mdulin2: From discussions on another channel, I think Jeff and Bruce are right. I removed this test but plan on making a Github Actions PR cron job instead. I'll. be making that in a separate PR though. Thanks all!

node/pkg/governor/governor_prices.go
```diff
@@ -88,19 +88,30 @@ func createCoinGeckoQueries(idList []string, tokensPerQuery int) []string {
...
+	// If modifying this code, ensure that the test 'TestCoinGeckoPriceChecks' passes when adding a pro API key to it.
+	// Since the code requires an API key (which we don't want to publish to git), this
+	// part of the test is normally skipped but mods to sensitive places should still be checked
+	query := ""
+	if coinGeckoApiKey == "" {
+		query = "https://api.coingecko.com/api/v3/simple/price?" + params.Encode()
+	} else { // Pro version API key path
+		query = "https://pro-api.coingecko.com/api/v3/simple/price?"
+		params.Add("x_cg_pro_api_key", coinGeckoApiKey)
+		query += "&" + params.Encode()
```
djb15, line 112: You don't need the "&" here. The params encoding handles that for us.

I would add the param in the first line of the `else` block and then do the assignment to query once (like we did before and how it's done when no api key is provided)

node/pkg/node/options.go
```diff
@@ -195,14 +195,18 @@ func GuardianOptionAccountant(
...
 // GuardianOptionGovernor enables or disables the governor.
 // Dependencies: db
-func GuardianOptionGovernor(governorEnabled bool) *GuardianOption {
+func GuardianOptionGovernor(governorEnabled bool, coinGeckoApiKey string) *GuardianOption {
 	return &GuardianOption{
 		name:         "governor",
 		dependencies: []string{"db"},
 		f: func(ctx context.Context, logger *zap.Logger, g *G) error {
 			if governorEnabled {
+
```
djb15, line 204: Nit, unnecessary new line

evan-gray:
> Not sure if this is the best way to do it, but for the Pro API test we could load a string from an environment variable? That way we could add a key as a secret in GitHub and load it in the Action

@djb15 - I saw this comment and wanted to note that I don’t believe secrets for this repo would be available to PRs from forks, so I would not rely on that approach. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

It may be better to simply give reviewers a “hack” script or the like to prove the updated fetching logic does indeed work with and without an API key and have the unit tests ensure that URL is formatted appropriately.

djb15:
> 
> @djb15 - I saw this comment and wanted to note that I don’t believe secrets for this repo would be available to PRs from forks, so I would not rely on that approach. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow
> 
> It may be better to simply give reviewers a “hack” script or the like to prove the updated fetching logic does indeed work with and without an API key and have the unit tests ensure that URL is formatted appropriately.

Ah thanks, yeah that's not an option then. Anyhow I think the approach we're going for with this now is to avoid adding an internet dependency here in these unit tests: https://github.com/wormhole-foundation/wormhole/pull/4025#discussion_r1689830680

evan-gray:
converting to draft while this has conflicts

mdulin2:
I accidentally added a test CoinGecko API key in the last commit. I deleted the key from the PR and revoked the key from my coin gecko account.

node/pkg/governor/governor_prices.go
```diff
@@ -88,19 +88,30 @@ func createCoinGeckoQueries(idList []string, tokensPerQuery int) []string {
...
+	// If modifying this code, ensure that the test 'TestCoinGeckoPriceChecks' passes when adding a pro API key to it.
+	// Since the code requires an API key (which we don't want to publish to git), this
+	// part of the test is normally skipped but mods to sensitive places should still be checked
+	query := ""
+	if coinGeckoApiKey == "" {
+		query = "https://api.coingecko.com/api/v3/simple/price?" + params.Encode()
+	} else { // Pro version API key path
+		query = "https://pro-api.coingecko.com/api/v3/simple/price?"
+		params.Add("x_cg_pro_api_key", coinGeckoApiKey)
+		query += params.Encode()
```
djb15, line 112: Looks like this changed slightly with the rebase but it's fine as is, just a nit to do the assignment on one line

djb15 (dismissed):

node/pkg/governor/governor_prices.go
mdulin2: Fixed this. Good catch on the regression.

djb15 (approved):

SEJeff (approved):
Nice

PR #4026, opened 2024-07-18
Input Validation on Generated Token List

mdulin2:
There's a script that adds a list of tokens into a file called `generated_mainnet_tokens.go` once every two weeks. This adds some input validation on the CoinGecko data before being inserted into the Go code. 

This is meant to be a temporary fix while a more robust mechanism, such as a JSON config file or the usage of a real Go code generation library, is completed. 

I ran this on the production endpoint and this did not violate any tokens in the current list either.

### DIFF
--- a/node/hack/governor/src/index.ts
+++ b/node/hack/governor/src/index.ts
@@ -163,9 +163,9 @@ axios
                   continue;
                 }
               }
-
+              
               // If the character list is violated, then skip the coin. The error is logged in the function if something happens to have some sort of check on it.
-              if(!(await safetyCheck(chain, wormholeAddr, data.Symbol, data.CoinGeckoId, data.TokenDecimals, data.TokenPrice, data.Address, notional))){
+              if(!(safetyCheck(chain, wormholeAddr, data.Symbol, data.CoinGeckoId, data.TokenDecimals, data.TokenPrice, data.Address, notional))){
                 failedInputValidationTokens.push(chain + "-" + wormholeAddr + "-" + data.symbol)
                 continue; 
               }
@@ -312,24 +312,24 @@ axios
 
   Example data: 30 000000000000000000000000b5c457ddb4ce3312a6c5a2b056a1652bd542a208 O404 omni404 18 1128.69 0xb5c457ddb4ce3312a6c5a2b056a1652bd542a208 7.4832146999999996
 */
-async function safetyCheck(chain, wormholeAddr, symbol, coinGeckoId, tokenDecimals, tokenPrice, address, notional)  : Promise<boolean>{
+function safetyCheck(chain, wormholeAddr, symbol, coinGeckoId, tokenDecimals, tokenPrice, address, notional)  : boolean{
   
   if(isNaN(chain)){
     console.log("Invalid chain ID ", chain, " provided")
     return false; 
   }
 
-  if(await inputHasInvalidChars(wormholeAddr)){
+  if(inputHasInvalidChars(wormholeAddr)){
     console.log("Invalid wormhole address ", wormholeAddr, " provided")
     return false; 
   }
 
-  if(await inputHasInvalidChars(symbol)){
+  if(inputHasInvalidChars(symbol)){
     console.log("Invalid token symbol ", symbol, " provided")
     return false; 
   }
 
-  if(await inputHasInvalidChars(coinGeckoId)){
+  if(inputHasInvalidChars(coinGeckoId)){
     console.log("Invalid coin gecko id ", coinGeckoId, " provided")
     return false; 
   }
@@ -344,7 +344,7 @@ async function safetyCheck(chain, wormholeAddr, symbol, coinGeckoId, tokenDecima
     return false; 
   }
 
-  if(await inputHasInvalidChars(address)){
+  if(inputHasInvalidChars(address)){
     console.log("Invalid address ", address, " provided")
     return false; 
   }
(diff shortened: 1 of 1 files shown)


node/hack/governor/src/index.ts
```diff
@@ -290,5 +295,71 @@ axios
...
+    return false; 
+  }
+
+  return true; 
+}
+
+// Validates that a list of characters are not in the provided string.
+// If a character is found then return true. Otherwise, return false. 
+async function inputHasInvalidChars(input) : Promise<boolean>{
+  var deny_list = ["\"", "%", "\n","\r", "\\"]
```
johnsaigle, line 357: I think it's worth checking the JSON spec (e.g. here) for valid escape sequences and adding them to the list.

In addition I think we'd be safe to include:
* `/`
* `\`
* `'`
* `"`
* `{`
* `}`
* `[`
* `]`
* `(`
* `)`

node/hack/governor/src/index.ts
```diff
@@ -290,5 +295,71 @@ axios
...
+  }
+  if(isNaN(notional)){
+    console.log("Invalid notional", notional, " provided")
+    return false; 
+  }
+
+  return true; 
+}
+
+// Validates that a list of characters are not in the provided string.
```
johnsaigle, line 354: I think it actually does the opposite:
```suggestion
// Checks whether an illegal character is present in the provided string
```

node/hack/governor/src/index.ts
```diff
@@ -136,8 +136,8 @@ axios
                   await (async () => {
                     const provider = new JsonRpcProvider(
                       new Connection({
-                        // fullnode: "https://fullnode.mainnet.sui.io",
-                        fullnode: "https://sui-mainnet.g.allthatnode.com/full/json_rpc",
+                         fullnode: "https://fullnode.mainnet.sui.io",
```
johnsaigle, line 139: Why is the URL being changed here?

johnsaigle (dismissed):

node/hack/governor/src/index.ts
```diff
@@ -290,5 +295,71 @@ axios
...
+    console.log("Invalid notional", notional, " provided")
+    return false; 
+  }
+
+  return true; 
+}
+
+// Validates that a list of characters are not in the provided string.
+// If a character is found then return true. Otherwise, return false. 
+async function inputHasInvalidChars(input) : Promise<boolean>{
```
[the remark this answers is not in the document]
mdulin2: This is the code to check for the strings being valid. For the numeric fields, I do an `isNaN` check to validate that they are numbers. Is this acceptable for what you mean?

node/hack/governor/src/index.ts
mdulin2: The script wasn't working with the other rpc. So, I tried this one in the comments and it worked. I can change it back though.

johnsaigle: What was the issue?

mdulin2: The request was simply failing - like a bad rpc.

node/hack/governor/src/index.ts
mdulin2: I think that the `[` and `(` may reasonably used in a token symbol. Since they have no meaning within a string, I'd prefer to keep those out of it. I can add `{}`, `\` and `'` to it though!

node/hack/governor/src/index.ts
djb15: That RPC has been spotty for a while, I've also had more luck with the fullnode endpoint so happy to switch that to the default again

node/hack/governor/src/index.ts
```diff
@@ -162,6 +162,11 @@ axios
                   continue;
                 }
               }
+
+              // If the character list is violated, then skip the coin. The error is logged in the function if something happens to have some sort of check on it.
+              if(!(await safetyCheck(chain, wormholeAddr, data.Symbol, data.CoinGeckoId, data.TokenDecimals, data.TokenPrice, data.Address, notional))){
```
djb15, line 167: Besides logging to the console can we also add the tokens that violate the safety check to the txt summary output? Since this runs in CI most of the time nobody is going to read the console output unless we flag that tokens have failed the safety check

node/hack/governor/src/index.ts
johnsaigle: When I test it locally, we have no governed tokens that use brackets or parentheses. Since this is the case, we might as well add them to the blocklist imo and we can enable them only when we need to later.

The other additions are a nice improvement 👍🏻

mdulin2: I know there's nothing there but I could see it happening, was the point. Is there any security issues in Go with not including the brace and paren?

johnsaigle: I'm not enough of a Go expert to know how to "jailbreak" from the string using them, but since `{}` are used to represent code blocks in Go it might be nice to disable them. Parentheses might be fine?

mdulin2: Alright, I just added all of them.

node/hack/governor/src/index.ts
mdulin2: Added this as well. Good catch!

node/hack/governor/src/index.ts
```diff
@@ -290,5 +300,71 @@ axios
...
+
+/*
+  Perform type checks on the incoming values
+  Check for a denylist set of characters
+
+  If either of these fail, we reject adding the token.
+
+  Example data: 30 000000000000000000000000b5c457ddb4ce3312a6c5a2b056a1652bd542a208 O404 omni404 18 1128.69 0xb5c457ddb4ce3312a6c5a2b056a1652bd542a208 7.4832146999999996
+*/
+async function safetyCheck(chain, wormholeAddr, symbol, coinGeckoId, tokenDecimals, tokenPrice, address, notional)  : Promise<boolean>{
```
djb15, line 315: Dumb question but what's the reasoning for making these functions asynchronous?

mdulin2: Good question!

Honestly, I've had so many issues over the years trying to write JavaScript with promises that I just have async and awaits on most things to avoid dealing with async issues. It's super hard to debug and I hate it.

I just ran a test without all of the async stuff and it works fine. I can remove it to simplify things.

bruce-riley (approved):

PR #4027, opened 2024-07-19
Fixes

OKEAMAH:


johnsaigle:
Closing this PR because it contains undocumented changes to important dependencies. We consider dependency management to be a sensitive operation that requires careful discussion.

Issue #4031, opened 2024-07-22
build guardiand on an ARM mac M1 machine


kakucodes:
## Description
`guardiand` and `wormholed` are both not able to be built on arm64 macbooks. 
while this is not an urgent issue i wanted to create discussion around it as it does increase dx friction.

in the meantime max gave a workaround [here](https://github.com/strangelove-ventures/wormhole/issues/58#issuecomment-2239435485) which can be used in a pinch for the guardiand binary

evan-gray:
The guardian software is only supported on Linux. You may find developing with https://code.visualstudio.com/docs/devcontainers/containers to be useful on other operating systems. Some contributors use WSL on Windows, but as mentioned elsewhere there are several other ecosystem features not supported on ARM, such as building production cosmwasm contracts (which must be built with amd64). While you can run amd64 code/containers on M-series processors, it is significantly slower.

johnsaigle:
FWIW I use the following flags for running tests locally on an arm64 machine:

```
go test -ldflags '-extldflags "-Wl,-ld_classic " ' ./...
```

Contrast with the flags used for amd64 builds:

https://github.com/wormhole-foundation/wormhole/blob/main/node/Makefile#L1C1-L2C76

evan-gray:
> FWIW I use the following flags for running tests locally on an arm64 machine:
> 
> ```
> go test -ldflags '-extldflags "-Wl,-ld_classic " ' ./...
> ```
> 
> Contrast with the flags used for amd64 builds:
> 
> https://github.com/wormhole-foundation/wormhole/blob/main/node/Makefile#L1C1-L2C76

@johnsaigle maybe PR this up to the makefile for like `make test-arm` or something?

johnsaigle:
I believe this is no longer an issue following https://github.com/wormhole-foundation/wormhole/pull/4210

I'm able to run `make node` on main without any issues and then run `guardiand`. @kakucodes would you mind testing and closing the issue if this works for you?

evan-gray:
Can confirm I explicitly tested this when collaborating on that PR! 😁 Thanks for noticing, forgot to close this issue.

PR #4032, opened 2024-07-23
node: devcontainer support

evan-gray:
This PR adds a [devcontainer.json](https://code.visualstudio.com/docs/devcontainers/containers) to aid in the development of node / guardian code. While development is currently possible in Linux (including WSL on Windows), providing this container adds an option for non-WSL Windows and ARM (M-series) Mac users which only requires Docker and VS Code. It also shares the same dockerfile used in CI for added assurance of compatibility. Relevant to #4031. Please note, this dockerfile _explicitly_ specifies amd64, which means it will run via qemu or rosetta (depending on your Docker settings) - as far as I understand, the cosmwasm dependency (and possibly others) requires this and there will be a slight performance hit, but it is useable.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+  "name": "Guardian Development",
+  "build": {
+    "context": "..",
+    "dockerfile": "../node/Dockerfile",
+    "target": "dev"
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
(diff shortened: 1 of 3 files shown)


SEJeff (approved):
Multistage builds ftw

bruce-riley (approved):

kcsongor (approved):
neat, I downloaded vs code and and opened the project using the devcontainer, and it seems to have worked 👍

PR #4034, opened 2024-07-24
feat: Wormchain v0.47

joelsmith-2019:


wormchain/app/app.go
```diff
@@ -343,95 +349,166 @@ func New(
...
 	)
 	// If evidence needs to be handled for the app, set routes in router here and seal
 	app.EvidenceKeeper = *evidenceKeeper
 
-	govRouter.AddRoute(wormholemoduletypes.RouterKey, wormholemodule.NewWormholeGovernanceProposalHandler(app.WormholeKeeper))
+	// register the proposal types
+	govRouter := govv1beta.NewRouter().
+		AddRoute(govtypes.RouterKey, govv1beta.ProposalHandler).
+		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(app.UpgradeKeeper)).
+		AddRoute(ibcclienttypes.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper)).
```
Reecepbcups, line 498: ref: above to remove?

wormchain/app/apptesting/test_helpers.go
```diff
@@ -213,6 +213,7 @@ func genesisStateWithValSet(t *testing.T,
 		balances,
 		totalSupply,
 		[]banktypes.Metadata{},
+		[]banktypes.SendEnabled{}, // TODO: JOEL - MAY NEED TO ADD A DENOM TO THE SEND ENABLED
```
Reecepbcups, line 216: not needed

wormchain/cmd/wormchaind/root.go
```diff
@@ -0,0 +1,294 @@
...
+	initClientCtx := client.Context{}.
+		WithCodec(encodingConfig.Marshaler).
+		WithInterfaceRegistry(encodingConfig.InterfaceRegistry).
+		WithTxConfig(encodingConfig.TxConfig).
+		WithLegacyAmino(encodingConfig.Amino).
+		WithInput(os.Stdin).
+		WithAccountRetriever(authtypes.AccountRetriever{}).
+		WithBroadcastMode(flags.BroadcastSync).
+		WithHomeDir(app.DefaultNodeHome).
+		WithViper("")
```
Reecepbcups, line 52: WORMCHAIND

wormchain/cmd/wormchaind/root.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Allows you to add extra params to your client.toml
+	// gas, gas-price, gas-adjustment, fees, note, etc.
+	setCustomEnvVariablesFromClientToml(initClientCtx)
+
+	rootCmd := &cobra.Command{
+		Use:   version.AppName,
+		Short: "Neutron",
```
Reecepbcups, line 60: name

wormchain/ts-sdk/src/modules/cosmos.tx.v1beta1/types/tendermint/abci/types.ts
```diff
@@ -528,7 +528,7 @@ export interface Evidence {
   /**
    * Total voting power of the validator set in case the ABCI application does
    * not store historical validators.
-   * https://github.com/tendermint/tendermint/issues/4581
+   * https://github.com/cometbft/cometbft/issues/4581
```
Reecepbcups, line 531: regen? proto & then ts

wormchain/x/wormhole/governance.go
```diff
@@ -17,8 +17,8 @@ import (
 // NewWormholeGovernanceProposalHandler creates a governance handler to manage new proposal types.
 // It enables GuardianSetProposal to update the guardian set and GenericWormholeMessageProposal to emit a generic wormhole
 // message from the governance emitter.
-func NewWormholeGovernanceProposalHandler(k keeper.Keeper) govtypes.Handler {
-	return func(ctx sdk.Context, content govtypes.Content) error {
+func NewWormholeGovernanceProposalHandler(k keeper.Keeper) govv1beta.Handler {
```
Reecepbcups, line 20: rm me?

wormchain/Makefile
```diff
@@ -78,3 +93,50 @@ bootstrap:
 clean:
 	rm -rf build/wormchaind build/wormchaind-* build/**/*.db build/**/*.wal vue
 	echo "{\"height\":\"0\",\"round\":0,\"step\":0}" > build/data/priv_validator_state.json
+
+get-heighliner:
+	git clone https://github.com/strangelove-ventures/heighliner.git
+	cd heighliner && go install
```
pharr117, line 99: We may want to make this:

```
get-heighliner:
        go install github.com/strangelove-ventures/heighliner@latest
```

to avoid the git clone

wormchain/Dockerfile
```diff
@@ -1,9 +1,11 @@
...
 
 #used for a readiness probe
-RUN apt-get update
-RUN apt install -y netcat
-RUN apt install -y jq
+RUN apt update && \
+    apt-get install -y \
+        build-essential \
+        ca-certificates \
+        curl
```
pharr117, line 8: The comment on the original file and the downloads of netcat and jq seem to indicate some sort of "readiness probe" requirement for netcat, did you replace these with curl elsewhere?

joelsmith-2019: Good catch, I'll add those back. I must've been under the impression build-essential included those packages as well.

wormchain/Makefile
joelsmith-2019: Very good point! I was actually supposed to remove heighliner completely from the Make file. Looks like I only took out 1 of 2 targets. I'll remove this completely now.

wormchain/go.mod
```diff
@@ -1,146 +1,206 @@
...
+	pgregory.net/rapid v1.1.0 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
 replace (
-	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
-	github.com/CosmWasm/wasmd v0.30.0 => github.com/wormhole-foundation/wasmd v0.30.0-wormchain-2
-	github.com/cosmos/cosmos-sdk => github.com/wormhole-foundation/cosmos-sdk v0.45.9-wormhole-2
+	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
+	github.com/cosmos/cosmos-sdk => github.com/strangelove-ventures/wh-cosmos-sdk v0.0.0-20240903174520-1c952abb4033
```
pharr117, line 201: Looks like we still need to make a release in the [wormhole-foundation/cosmos-sdk](https://github.com/wormhole-foundation/cosmos-sdk) repo with the changes at this SL fork version and then update the replace statement.

wormchain/Makefile
```diff
@@ -78,3 +93,44 @@ bootstrap:
...
+	docker build -t wormchain:latest -f Dockerfile.ict ..
+
+# Individual Tests ($$ is interpreted as $)
+rm-testcache:
+	go clean -testcache
+
+ictest-cancel-upgrade: rm-testcache
+	cd interchaintest && go test -race -v -run ^TestCancelUpgrade$$ ./...
+
+ictest-chain-start: rm-testcache
```
pharr117, line 112: When running this test, it fails with:

```
wormchain:local: pull image wormchain:local: Error response from daemon: pull access denied for wormchain
```

It then only passed if I tagged the local docker image built for wormchain ICT as "local" instead of "latest", we may need to update the make command on line 102 above to build the image with the tag local instead of latest, like:

```
local-image: build/wormchaind
	docker build -t wormchain:local -f Dockerfile.ict ..
```

We may want to consider a way to ensure the local docker image is built with the latest version of the current branch's wormchaind software before running the tests.

pharr117: Good points there, I agree that rebuilding the image every test would be an unnecessary annoyance and make test runs take longer, especially if chain code stays the same. I will spend some time thinking about it, but this should work just fine for now.

node/go.mod
```diff
@@ -1,113 +1,137 @@
...
 	github.com/gorilla/websocket v1.5.1 // indirect
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
-	github.com/grpc-ecosystem/grpc-gateway/v2 v2.16.2
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.21.0
 	github.com/improbable-eng/grpc-web v0.15.0
 	github.com/ipfs/go-log/v2 v2.5.1
 	github.com/libp2p/go-libp2p v0.33.1
-	github.com/libp2p/go-libp2p-kad-dht v0.25.2
+	github.com/libp2p/go-libp2p-kad-dht v0.25.0
```
evan-gray, line 22: Why was there a downgrade here?

joelsmith-2019: Good catch, I'll update that now. It might've been me running `go get` on one of the newer dependencies which ultimately downgraded it.

jtieri:
Could someone remove me as a reviewer on this? Please and thank you!

PR #4038, opened 2024-07-26
node/cleanup: Add some documentation and an ignore-list to mainnet_chains_test.go

johnsaigle:
If not true, then there exists some chain that Wormhole governs yet has no governed tokens. That means assets native to the chain are not actually governed and can be transferred across all other supported chains without increasing their Governor limit. We don't want this.

This unit test should help when deploying a new chain to the Governor by helping developers to remember to add an initial Governor token list.

### DIFF
--- a/node/pkg/governor/mainnet_chains.go
+++ b/node/pkg/governor/mainnet_chains.go
@@ -8,7 +8,7 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
-func chainList() []chainConfigEntry {
+func ChainList() []chainConfigEntry {
 	return []chainConfigEntry{
 		{emitterChainID: vaa.ChainIDSolana, dailyLimit: 25_000_000, bigTransactionSize: 2_500_000},
 		{emitterChainID: vaa.ChainIDEthereum, dailyLimit: 50_000_000, bigTransactionSize: 5_000_000},
--- a/node/pkg/governor/mainnet_chains.go
+++ b/node/pkg/governor/mainnet_chains.go
@@ -8,7 +8,7 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
-func ChainList() []chainConfigEntry {
+func chainList() []chainConfigEntry {
 	return []chainConfigEntry{
 		{emitterChainID: vaa.ChainIDSolana, dailyLimit: 25_000_000, bigTransactionSize: 2_500_000},
 		{emitterChainID: vaa.ChainIDEthereum, dailyLimit: 50_000_000, bigTransactionSize: 5_000_000},
--- a/node/pkg/governor/mainnet_tokens_test.go
+++ b/node/pkg/governor/mainnet_tokens_test.go
@@ -2,12 +2,10 @@ package governor
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/wormhole-foundation/wormhole/sdk"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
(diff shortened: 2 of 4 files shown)


node/pkg/governor/mainnet_tokens_test.go
```diff
@@ -88,3 +90,87 @@ func TestTokenListEmptyCoinGeckoId(t *testing.T) {
 		assert.Greater(t, len(tokenConfigEntry.coinGeckoId), 0)
 	}
 }
+
+// If not true, then there exists some chain that Wormhole governs yet has no governed tokens. That means assets
```
evan-gray, line 94: Tokens which are governed affect all chains they are transferred to/from. Why should it be necessary that a chains native assets are governed?

evan-gray (changes requested):

node/pkg/governor/mainnet_tokens_test.go
```diff
@@ -88,3 +90,87 @@ func TestTokenListEmptyCoinGeckoId(t *testing.T) {
...
+// native to the chain are not actually governed and can be transferred across all other supported chains without
+// increasing their Governor limit. We don't want this.
+func TestAllGovernedChainsHaveGovernedAssets(t *testing.T) {
+
+	// Build a list of chains that are not governed we will filter these out. Compare with `mainnet_chains.go`
+	// When a new ChainID is added, it must either be added to this list or else have entries populated in
+	// one of the list of tokens accessed by `tokenList()`.
+	ignoredChains := map[vaa.ChainID]bool{
```
evan-gray, line 102: Can node/pkg/governor/mainnet_chains.go be used instead? Seems like this would require maintaining any time a new chain is added, while this is the source of Governor truth.

node/pkg/governor/mainnet_tokens_test.go
bruce-riley: There was a time when we had governed chains that did not have governed tokens. See lines 51 of this file. I guess that line is obsolete now, but there was a reason that line was added.

node/pkg/governor/mainnet_tokens_test.go
```diff
@@ -88,3 +90,87 @@ func TestTokenListEmptyCoinGeckoId(t *testing.T) {
...
+		vaa.ChainIDNeutron:    true,
+		vaa.ChainIDCelestia:   true,
+		vaa.ChainIDStargaze:   true,
+		vaa.ChainIDSeda:       true,
+		vaa.ChainIDDymension:  true,
+		vaa.ChainIDProvenance: true,
+		// ID reserved but not yet Governed
+		vaa.ChainIDBerachain: true,
+		vaa.ChainIDSnaxchain: true,
+		// Testnets
```
bruce-riley, line 122: This seems like a maintenance headache. When a new chain goes to testnet, I would need to add it here. Later, when the chain goes to mainnet, someone needs to remember to come back here and take it out.

node/pkg/governor/mainnet_tokens_test.go
johnsaigle: Ah thanks Bruce, I missed that somehow. I think because I stated with an approach that iterated over all the vaa ChainIDs so at first it looked like there _were_ some that were missing.
I'll take some time to review my test and see if it adds anything new. If not I'll close the PR, otherwise I might rework the existing test rather than add a mostly-redundant one.

bruce-riley: Actually, how is this test different from `TestTokenListChainTokensPresent`?

node/pkg/governor/mainnet_tokens_test.go
johnsaigle: Yep! Evan raised the same point and I agree this isn't a great implementation. I'm going to read from https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/governor/mainnet_chains.go#L11 chainList instead and see if that feels better. That way we only care about things that 1) have to do with the Governor and 2) are on Governor's mainnet.

node/pkg/governor/mainnet_tokens_test.go
johnsaigle: It's not - I didn't see the other test 😅 
I've changed the PR so that it updates the existing test to be a little more verbose and commented. It's fine as-is but since I already wrote the comments, might as well add them here.
I deleted my new test and have (hopefully) combined the best of both worlds.

node/pkg/governor/mainnet_chains.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
-func chainList() []chainConfigEntry {
+func ChainList() []chainConfigEntry {
```
evan-gray, line 11: Does this need to be made public? It is only referenced inside of this package.

evan-gray (changes requested):

node/pkg/governor/mainnet_chains.go
johnsaigle: You're right, thanks. I've reverted the change.

bruce-riley (approved):

evan-gray (approved):

PR #4046, opened 2024-07-31
CI: query-sdk-ci fails intermittently

bruce-riley:
This PR addresses several more issues seen with CI tests:

- The `query-ci` test were failing quite frequently, I think due to p2p connectivity issues. Since this test is redundant now that we have the `query-sdk-ci` tests, we decided to get rid of it.
- The Solana `query-sdk-ci` tests were failing intermittently because they were not all setting the min context slot, so the guardians were returning different blocks.
- The Solana concurrency test was failing to reach quorum so we reduced the number of concurrent queries to match the number of solana workers.
- The Solana `query-sdk-ci` tests were failing sometimes while converting BigInt to Number.
- The accountant on guardian-0 had a sequence number error while submitting an observation to the contract, so I created new, unique keys for the accountants on both guardians.


evan-gray (approved):

evan-gray:
Example failure of querying Solana without `minContextSlot` [here](https://github.com/wormhole-foundation/wormhole/actions/runs/10187748135/job/28182484122#step:6:9056)

panoel (approved):

PR #4049, opened 2024-07-31
ci: bump foundry to 2024-08-04 nightly

evan-gray:
We have seen [some instances](https://github.com/wormholelabs-xyz/wormhole/actions/runs/10162658975/job/28103791466#step:6:9835) in CI where anvil returns a block number which doesn't match the one queried. Perhaps this issue has been fixed since the [nightly currently being used (2024-05-26)](https://github.com/foundry-rs/foundry/commit/ea2eff95b5c17edd3ffbdfc6daab5ce5cc80afc0). Confirmed that the forge output is still the same by exec-ing into the container and running the following before and after the change then diffing the result.
```bash
cd ~/app/ethereum/build-forge
find . -type f -exec sha256sum {} \;
cd ~/app/relayer/ethereum/build-forge
find . -type f -exec sha256sum {} \;
```

### DIFF
--- a/ethereum/Dockerfile
+++ b/ethereum/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker.io/docker/dockerfile:1.3@sha256:42399d4635eddd7a9b8a24be879d2f9a930d0ed040a61324cfdf59ef1357b3b2
 FROM const-gen AS const-export
-FROM ghcr.io/foundry-rs/foundry:nightly-ea2eff95b5c17edd3ffbdfc6daab5ce5cc80afc0@sha256:2a774f86765258a0d176366fc46f92bc14f5040faae7a3c3ba59b1c24c5fa7cb as foundry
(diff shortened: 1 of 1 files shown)


kcsongor (dismissed):
thanks for checking the build output, other than that this should be a straightforward change

bruce-riley (dismissed):

evan-gray:
Just hit [this error](https://github.com/wormhole-foundation/wormhole/actions/runs/10188715890/job/28185870774?pr=4049#step:6:5440) so perhaps the bump does not resolve it, but may be worth considering nonetheless.

evan-gray:
Isolated the bug and [reported](https://github.com/foundry-rs/foundry/issues/8590)

evan-gray:
updated to a version which includes [the fix](https://github.com/foundry-rs/foundry/commit/910850011c09437b57e99f8b94b25f03610f9e8a) and checked the build output again

kcsongor (approved):

bruce-riley (approved):

PR #4053, opened 2024-08-01
node: Fix missing support for chains in generated token list due to outdated SDK

johnsaigle:
- The script used to generate the updated token list was using an outdated version of the TS SDK. This caused newer chains to return a result of 'undefined' when they were converted to the custom ChainId type as they were not encluded in the enum of valid chain IDs
- Update the local SDK version used by the script via `npm update @certusone/wormhole-sdk`
- Adds additional logging to the script to help troubleshoot similar errors in the future

To test:
- Before this commit: run `node/hack/governor/src/index.ts`, observe error output relating to Scroll. Note that Scroll has no governed tokens in the generated tokens list
- On this commit: do the same. No error and Scroll will be updated to be in the generated token list.

### DIFF
--- a/node/hack/governor/package-lock.json
+++ b/node/hack/governor/package-lock.json
@@ -690,11 +690,12 @@
       "dev": true
     },
     "node_modules/@certusone/wormhole-sdk": {
-      "version": "0.10.5",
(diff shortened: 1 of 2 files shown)


evan-gray (approved):
Might have done `npm i @certusone/wormhole-sdk@latest` just to make the package.json more obvious. 🤷

bruce-riley (approved):

djb15 (approved):

PR #4066, opened 2024-08-07
Node: Observation batching with override

bruce-riley:
This PR supersedes #3977, which is being closed due to merge / rebase issues, as well has new requirements.

This PR introduces the batching of observations by adding the `SignedObservationBatch` gossip message. It modifies the guardian to wait up to one second in order to batch as many observations as possible (up to 4000) into a single gossip message. This not only reduces gossip traffic but also the load on the guardians and applications that listen to observations.

This feature allows for certain types of observations to be published immediately, as specified in the `shouldPublishImmediately` function in `node/pkg/processor/batch_obs.go`. If that function returns true, then the observation will be published without delay. Note that it will still be published in a `SignedObservationBatch` message, but the batch size will always be one.

Any spies or flies that listen to observations will need to be updated to also pass ` WithSignedObservationBatchListener` into `p2p.NewRunParams`, in addition to `WithSignedObservationListener`.

This feature is currently disabled, and will be enabled at a later date by setting the cutover time in `node/pkg/processor/cutover.go`.

NOTE: Once the cutover is complete, support for `SignedObservation` (individual observation) messages will be dropped, but in the meantime, applications that listen to observations need to support both message streams.


node/pkg/node/node.go
```diff
@@ -33,6 +33,11 @@ const (
 	// One observation takes roughly 0.1ms to process on one core, so the whole queue could be processed in 1s
 	inboundObservationBufferSize = 10000
 
+	// inboundBatchObservationBufferSize configures the size of the batchObsvC channel that contains batches of observations from other Guardians.
+	// Since a batch contains many observations, the guardians should not be publishing too many of these, so we can keep the channel small. With
+	// 19 guardians, we would expect 19 messages per second during normal operations. This gives us plenty of extra room.
```
evan-gray, line 38: Does this math change with the new batching exception? I would expect this channel may need to be a bit larger to accomodate for several message in a second being included in a batch of their own.

node/pkg/p2p/p2p.go
```diff
@@ -397,7 +397,7 @@ func Run(params *RunParams) func(ctx context.Context) error {
 		}
 
 		// Set up the attestation channel. ////////////////////////////////////////////////////////////////////
-		if params.gossipAttestationSendC != nil || params.obsvRecvC != nil {
+		if params.gossipAttestationSendC != nil || params.obsvRecvC != nil || params.batchObsvRecvC != nil {
```
evan-gray, line 400: Looks like the `if params.obsvRecvC != nil {` statement below should be updated as well.

node/pkg/processor/batch_obs_test.go
```diff
@@ -0,0 +1,85 @@
...
+	assert.Equal(t, 205, len(obsBuf))
+
+	batch := gossipv1.SignedObservationBatch{
+		Addr:         crypto.PubkeyToAddress(gk.PublicKey).Bytes(),
+		Observations: observations,
+	}
+
+	buf, err := proto.Marshal((&batch))
+	require.NoError(t, err)
+	assert.Greater(t, pubsub.DefaultMaxMessageSize, len(buf))
```
evan-gray, line 71: This test is great!

node/pkg/processor/broadcast.go
```diff
@@ -31,26 +31,46 @@ func (p *Processor) broadcastSignature(
...
 		MessageId: messageID,
 	}
 
-	w := gossipv1.GossipMessage{Message: &gossipv1.GossipMessage_SignedObservation{SignedObservation: &obsv}}
+	if batchCutoverComplete() {
+		if shouldPublishImmediately {
+			msg = p.publishImmediately(ourObs)
+			observationsBroadcast.Inc()
+		} else {
+			p.postObservationToBatch(ourObs)
```
evan-gray, line 49: Is there a corresponding prometheus metric being pegged for observation batches?

node/pkg/processor/cleanup.go
```diff
@@ -228,8 +228,11 @@ func (p *Processor) handleCleanup(ctx context.Context) {
 					if err := common.PostObservationRequest(p.obsvReqSendC, req); err != nil {
 						p.logger.Warn("failed to broadcast re-observation request", zap.String("message_id", s.LoggingID()), zap.Error(err))
 					}
-
-					p.gossipAttestationSendC <- s.ourMsg
+					if s.ourMsg != nil {
+						p.gossipAttestationSendC <- s.ourMsg
```
evan-gray, line 232: Seems like this will potentially re-send un-batched observations after the cutover - I think that's probably okay as long as folks are aware of this possibility (that both will need to be supported until after `retryLimitOurs` which is currently 30 hours) - but just wanted to double-check.

evan-gray (changes requested):
Overall looks good, mostly had questions and a couple tiny changes.

node/pkg/node/node.go
bruce-riley: Increased it to 1000 and updated the comment.

node/pkg/processor/cleanup.go
bruce-riley: That is correct, after the cutover any retries on old messages will still use the old message format. As long as we don't immediately rip out the old code, that should be fine.

Additionally, this case also covers the immediately published messages, so this code will have to remain when we rip out the cutover stuff. I added a comment about that.

node/pkg/processor/broadcast.go
bruce-riley: Added one.

evan-gray (approved):

evan-gray (approved):

bruce-riley:
Tested the following scenarios in local tilt:

- No batching, no gossip split
- Batching, no gossip split
- No batching, gossip split
- Batching, gossip split
- Pyth messages get published immediately

node/pkg/processor/batch_obs_test.go
```diff
@@ -0,0 +1,85 @@
...
+		ConsistencyLevel: uint8(32),
+		EmitterChain:     vaa.ChainIDSolana,
+		EmitterAddress:   governanceEmitter,
+		Payload:          payload,
+	}
+
+	return vaa
+}
+
+func TestMarshalSignedObservationBatch(t *testing.T) {
```
panoel, line 40: Do you need a test with 1 in the batch and then a test with >1 in the batch?  Or it doesn't matter?

panoel (approved):

node/pkg/processor/batch_obs_test.go
bruce-riley: Not sure the single observation test would be any different.

PR #4068, opened 2024-08-08
Node/CCQ: Solana should retry block read

bruce-riley:
We are seeing quite a few cases where the Solana watcher is able to read the desired slot, but then fails to look up the block associated with that slot (to get the block time). This is most likely due to load balancing. This PR adds a retry mechanism for that case.

### DIFF
--- a/node/pkg/watchers/solana/ccq.go
+++ b/node/pkg/watchers/solana/ccq.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


evan-gray (dismissed):

evan-gray (approved):

bruce-riley:
I tested this using my local tilt pointed at a public mainnet RPC, with a modified query test that did 100 queries for one slot in the future. Of the 100 attempts, 26 of them required a single retry before succeeding. One still failed after three attempts. Calling that success.

bingyuyap (approved):

PR #4073, opened 2024-08-13
doc: use markdown rather than HTML in README

DemoYeti:


evan-gray:
This does not seem important to change, the code is equivalent.

PR #4095, opened 2024-08-25
Wormhole

OKEAMAH:


evan-gray:
[As stated previously](https://github.com/wormhole-foundation/wormhole/pull/3619#issuecomment-1997481456), upgrading dependencies should be carefully considered and there is no justification provided as to why these packages should be bumped. Additionally, such changes should be done in an isolated manner, instead of bumping a bunch of independent dependencies. If you feel there is a strong motivation for updating these dependencies, consider raising it in an issue with the necessary context.

PR #4099, opened 2024-08-27
Node: Solana watcher cleanup

bruce-riley:
This PR does some simple clean up / performance improvements in the Solana watcher.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"nhooyr.io/websocket"
 )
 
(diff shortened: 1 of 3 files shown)


panoel (commented):
Don't know if there is only 1 account key matching the contract or if you want the last one?  :shrug:

bruce-riley:
> Don't know if there is only 1 account key matching the contract or if you want the last one? 🤷

That is a question that I asked @evan-gray offline. If we are uncomfortable with this change, I can take it out.

bingyuyap (approved):
lgtm assuming you want the 1st account key you found

bruce-riley:
I removed the break that was causing some concern.

bingyuyap (approved):

node/pkg/watchers/solana/client.go
```diff
@@ -215,21 +222,27 @@ func NewSolanaWatcher(
 	queryReqC <-chan *query.PerChainQueryInternal,
 	queryResponseC chan<- *query.PerChainQueryResponseInternal,
 ) *SolanaWatcher {
+	msgObservedLogLevel := zapcore.DebugLevel
+	if chainID != vaa.ChainIDPythNet {
```
evan-gray, line 226: this did confuse me the first time I read it. might be easier to read to have the default be info and then if it *is* pythnet, set it to debug.

panoel (approved):

evan-gray (approved):

PR #3366, opened 2023-09-08
Ethereum: add optimized verification endpoint to the core bridge

gator-boi:
The purpose of this PR is to add an optimized `parseAndVerify` endpoint to `Messages.sol` along with various other optimizations. See the following table for the estimated gas savings:

call | before | after | savings
-----|--------|-------|--------
parseAndVerify | 129k | 83k | 46k
parseVM | 26.5k | 16.2k | 10.3k

PR #4100, opened 2024-08-27
Solidity core contract differential fuzzing

djb15:
The original purpose of this fuzzing suite was to give more confidence around the changes in https://github.com/wormhole-foundation/wormhole/pull/3366. This fuzz test suite uses both the currently deployed core contract, and deploys the current local implementation, before performing an arbitrary sequence of calls across both forks.

Given the timeline and consensus around that PR is unclear I've decided to pull these fuzz tests into this PR.

Fuzz tests can be executed with:

```
echidna fuzz-upgrade/DiffFuzzUpgrades.sol --contract DiffFuzzUpgrades --config fuzz-upgrade/CryticConfig.yaml
```

### DIFF
--- a/ethereum/.gitignore
+++ b/ethereum/.gitignore
@@ -5,3 +5,6 @@ flattened
 ethers-contracts
 build-forge/
 broadcast/
+corpus/
+crytic-cache/
+crytic-export/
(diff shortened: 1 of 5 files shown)


evan-gray (commented):
If this test is valuable, should it be run in CI alongside the others?

djb15:
> If this test is valuable, should it be run in CI alongside the others?

I was concerned about tying CI to an RPC given we're forking mainnet for this fuzz test, and from what I've seen it looks like we hit the endpoint pretty hard

djb15:
Closing as part of cleanup. The branch can be retrieved [here](https://github.com/wormhole-foundation/wormhole-branch-clone/tree/differential-fuzzing) and the PR reopened from a fork if required.

PR #4106, opened 2024-08-30
Node/Processor: Peg observation metric for us

bruce-riley:
The recent refactoring of the processor included a change to not do all of the `handleObservation` processing on our own observations. This avoids unnecessary work, but the change also caused us to stop pegging `wormhole_observations_signed_by_guardian_total`, which at least one guardian does not like. This PR make it peg that metric for our own observations.

### DIFF
--- a/node/pkg/processor/message.go
+++ b/node/pkg/processor/message.go
@@ -95,6 +95,10 @@ func (p *Processor) handleMessage(k *common.MessagePublication) {
 	// Broadcast the signature.
 	ourObs, msg := p.broadcastSignature(v.MessageID(), k.TxHash.Bytes(), digest, signature, shouldPublishImmediately)
 
+	// Indicate that we observed this one.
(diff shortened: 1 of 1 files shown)


node/pkg/processor/message.go
```diff
@@ -95,6 +95,9 @@ func (p *Processor) handleMessage(k *common.MessagePublication) {
 	// Broadcast the signature.
 	ourObs, msg := p.broadcastSignature(v.MessageID(), k.TxHash.Bytes(), digest, signature, shouldPublishImmediately)
 
+	// Indicate that we observed this one.
+	observationsReceivedByGuardianAddressTotal.WithLabelValues(p.ourAddr.Hex()).Inc()
```
evan-gray, line 99: should `observationsReceivedTotal` be `Inc`'ed as well?

panoel (dismissed):

bingyuyap (approved):

PR #4120, opened 2024-09-19
Replace Guardian Key with abstracted Guardian Signer

pleasew8t:
This pull request introduces the `GuardianSigner` interface, defined in the `guardiansigner` package, which is meant to replace directly using the guardian key (private key) for data signing. In doing so, it becomes easier to introduce alternative signing mechanisms without too much modification of code outside of the new `guardiansigner` package. Additional signing mechanisms include HSMs or KMS's (such as AWS or GCP).

The changes made to the repository are summarised as follows:
* Introduction of the new `guardiansigner` package, which includes a `FileSigner` implementation that works with the current guardian key.
* Addition of the `--guardianSignerUri` commandline argument. Node operators can still make use of `guardianKeyPath`, as support is provided to translate the path to the appropriate `guardianSignerUri`. As additional signer implementations are introduced, more URI schemes will be added. But for now only `file://` is supported, loading a private key from disk.
* Replace data signing that used the guardian key with signing using a `GuardianSigner`.

## Notable Code Change

One change in the PR that we are not yet certain about is the use of `AddSignature` defined in `sdk/vaa/structs.go`. The changes in `node/pkg/adminrpc/adminserver.go` removed the use of `AddSignature` and adds the signature manually, to make use of a `GuardianSigner`. There were some alternative ideas surrounding this:

An alternative idea was to make `AddSignature` accept a `GuardianSigner` instead of a private key. This did not seem right, as it would create a dependency on the `node` packages that seems unneccessary, and also modify what is defined as an SDK, potentially breaking functionality for other projects that might make use of the SDK.

Additional comment by @johnsaigle that is also worth considering: https://github.com/wormhole-foundation/wormhole/pull/4120#issuecomment-2361061102


node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,66 @@
...
+type SignerType int
+
+const (
+	InvalidSignerType SignerType = iota
+	FileSignerType
+)
+
+// GuardianSigner interface
+type GuardianSigner interface {
+	// TODO: document that the signer implementations assume they are recieving a keccack256 hash
```
banescusebi, line 19: Not sure if you want to leave this one in there or not.

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,66 @@
...
+func NewGuardianSignerFromUri(signerUri string, unsafeDevMode bool) (GuardianSigner, error) {
+
+	// Get the signer type
+	signerType, signerKeyConfig := ParseSignerUri(signerUri)
+
+	switch signerType {
+	case FileSignerType:
+		return NewFileSigner(unsafeDevMode, signerKeyConfig)
+	case InvalidSignerType:
+		return nil, fmt.Errorf("unsupported guardian signer type")
```
banescusebi, line 34: This case seems identical to the `default` one. Is it explicitly needed?

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,66 @@
...
+	default:
+		return nil, fmt.Errorf("unsupported guardian signer type")
+	}
+}
+
+func ParseSignerUri(signerUri string) (signerType SignerType, signerKeyConfig string) {
+	signerUriSplit := strings.Split(signerUri, "://")
+
+	if len(signerUriSplit) < 2 {
```
banescusebi, line 43: Seems pretty easy to understand but you might want to add a comment as to why the value `2` is used here to improve maintainability

node/cmd/guardiand/node.go
```diff
@@ -626,7 +629,14 @@ func runNode(cmd *cobra.Command, args []string) {
...
 	if *guardianKeyPath == "" {
-		logger.Fatal("Please specify --guardianKey")
+		if *guardianSignerUri == "" {
+			logger.Fatal("Please specify --guardianKey or --guardianSignerUri")
+		}
+	} else {
+		// If guardianKeyPath is set, set guardianSignerUri to the file signer URI, pointing to guardianKeyPath.
+		// This ensures that the signer-abstracted guardian has backwards compatibility with guardians that would
+		// just like to ignore the new guardianSignerUri altogether.
+		*guardianSignerUri = fmt.Sprintf("file://%s", *guardianKeyPath)
```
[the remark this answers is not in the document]
johnsaigle: My preference would be to use https://pkg.go.dev/net/url#URL.JoinPath instead of Sprintf so that the program can help with repeated characters, relative filepaths and so on.

node/hack/accountant/send_obs.go
```diff
@@ -26,7 +26,8 @@ func main() {
 	wormchainURL := string("localhost:9090")
 	wormchainKeyPath := string("./dev.wormchain.key")
 	contract := "wormhole14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9srrg465"
-	guardianKeyPath := string("./dev.guardian.key")
+	// guardianKeyPath := string("./dev.guardian.key")
```
johnsaigle, line 29: delete?

node/pkg/adminrpc/adminserver.go
```diff
@@ -1162,7 +1162,18 @@ func (s *nodePrivilegedService) SignExistingVAA(ctx context.Context, req *nodev1
 	}
 
 	// Add local signature
-	newVAA.AddSignature(s.gk, uint8(localGuardianIndex))
+	sig, err := s.guardianSigner.Sign(v.SigningDigest().Bytes())
+	if err != nil {
+		panic(err)
```
johnsaigle, line 1167: Any chance we can do better than `panic` here? I see that the old method with `AddSignature` also panicked if the signature failed but maybe there's another way to handle errors here.

node/pkg/adminrpc/adminserver.go
```diff
@@ -1162,7 +1162,18 @@ func (s *nodePrivilegedService) SignExistingVAA(ctx context.Context, req *nodev1
...
+	}
+
+	sigData := [65]byte{}
```
johnsaigle, line 1170: I like how these lines are captured by `AddSignature` in the existing implementation. Any thoughts on putting this code into a method on the VAA rather than modifying it directly?

node/pkg/guardiansigner/filesigner.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"google.golang.org/protobuf/proto"
+
+	nodev1 "github.com/certusone/wormhole/node/pkg/proto/node/v1"
+	"golang.org/x/crypto/openpgp/armor" // nolint
```
johnsaigle, line 15: What's the lint issue here?

node/pkg/guardiansigner/filesigner.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign wormchain address: %w", err)
+	}
+
+	return sig, nil
+}
+
+func (fs *FileSigner) PublicKey() ecdsa.PublicKey {
+	publicKey := fs.privateKey.PublicKey
```
johnsaigle, line 83: I think you can just return here rather than assign to `publicKey` first

node/pkg/guardiansigner/generatedsigner.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign wormchain address: %w", err)
+	}
+
+	return sig, nil
+}
+
+func (gs *GeneratedSigner) PublicKey() (pubKey ecdsa.PublicKey) {
+	publicKey := gs.pk.PublicKey
+	return publicKey
```
johnsaigle, line 39: Same here, I think you can just return instead of assign to a new variable.

node/pkg/guardiansigner/generatedsigner.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (gs *GeneratedSigner) Verify(sig []byte, hash []byte) (valid bool, err error) {
+	recoveredPubKey, err := ethcrypto.SigToPub(hash, sig)
+
+	if err != nil {
+		return false, err
+	}
+
+	fsPubkey := gs.pk.PublicKey
```
johnsaigle, line 49: This is very minor but in other crypto libraries I've seen the naming convention of `pk == public key` and `sk == private key (secret key)`. I think this naming is more clear if you want to use concise variable names.

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,66 @@
...
+		return NewFileSigner(unsafeDevMode, signerKeyConfig)
+	case InvalidSignerType:
+		return nil, fmt.Errorf("unsupported guardian signer type")
+	default:
+		return nil, fmt.Errorf("unsupported guardian signer type")
+	}
+}
+
+func ParseSignerUri(signerUri string) (signerType SignerType, signerKeyConfig string) {
+	signerUriSplit := strings.Split(signerUri, "://")
```
johnsaigle, line 41: I think this is another place where using the Uri package is appropriate rather than operating on strings

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,66 @@
...
+
+	switch typeStr {
+	case "file":
+		return FileSignerType, keyConfig
+	default:
+		return InvalidSignerType, ""
+	}
+}
+
+// WARNING: DO NOT USE THIS SIGNER OUTSIDE OF TESTS
```
johnsaigle, line 58: Maybe it would be helpful to rename it to something like `GenerateSignerWithPrivatekeyUnsafe` or else move it into a testing library if that's possible.

node/pkg/guardiansigner/guardiansigner_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseSignerUri(t *testing.T) {
+	tests := []struct {
```
johnsaigle, line 12: ❤️ the use of table-driven tests

node/pkg/guardiansigner/guardiansigner_test.go
```diff
@@ -0,0 +1,88 @@
...
+		label        string
+		path         string
+		expectedType SignerType
+	}{
+		{label: "RandomText", path: "RandomText", expectedType: InvalidSignerType},
+		{label: "ArbitraryUri", path: "arb://data", expectedType: InvalidSignerType},
+		// File
+		{label: "FileURI", path: "file://whatever", expectedType: FileSignerType},
+		{label: "FileUriNoDotDotSlash", path: "filewhatever", expectedType: InvalidSignerType},
+		{label: "FileUriMultipleDotDotSlash", path: "file://testing://this://", expectedType: FileSignerType},
```
johnsaigle, line 22: There are no `DotDotSlash` sequences in either of these. Maybe you meant scheme separators `://`?

johnsaigle (commented):
Left some minor comments related to naming and clarity. Overall the implementation looks good

johnsaigle:
`AddSignature` in `/sdk/vaa/structs.go` is now unused outside of tests. Should we [mark it as deprecated](https://go.dev/wiki/Deprecated)? Are there any other related functions that should be deprecated now?

node/cmd/guardiand/node.go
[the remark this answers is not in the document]
pleasew8t: @johnsaigle  I'm going to leave the `Sprintf` as is, since I feel like an incorrect path should just lead to failure and we aren't really doing that now with the `guardianKeyPath` anyway.

@banescusebi My intention is that the `guardianKeyPath` gets preference, so leaving this as is for now.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: I was meaning to document the interface. Will do that. Thanks for pointing this out.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: Yeah I'm dropping the `InvalidSignerType` case. Wanted it in there to make use of it, but I really think it's a waste of code.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: Added some more comments to this function in general.

node/pkg/adminrpc/adminserver.go
pleasew8t: Going to resolve this comment and make a note in the PR description about this decision in particular. I'm keen to get WH feedback on this.

node/pkg/guardiansigner/filesigner.go
pleasew8t: This is because "golang.org/x/crypto/openpgp/armor" is deprecated. We do the same thing in `armoredKey.go:LoadGuardianKey()`.

node/pkg/guardiansigner/generatedsigner.go
pleasew8t: I like this suggestion. Just opted for `privateKey`.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: Renamed the function. Will move it to a different file if there are more requests like this.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: For now I like the simplicity. I'll use the URI package if there's demand for it.

node/pkg/guardiansigner/guardiansigner_test.go
pleasew8t: Updated the naming.

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func ParseSignerUri(signerUri string) (signerType SignerType, signerKeyConfig string) {
+	// Split the URI using the standard "://" scheme separator
+	signerUriSplit := strings.Split(signerUri, "://")
+
+	// Length smaller than 2 implies that there is no path following the scheme separator,
+	// which is invalid.
+	if len(signerUriSplit) < 2 {
```
djb15, line 49: If there was no path following the scheme separator won't we still get 2 elements, but the second is just an empty string?

node/pkg/guardiansigner/guardiansigner_test.go
djb15, line 1: Can you add a test for an empty string following the `://` separator? i.e. `file://`

node/pkg/p2p/run_params.go
```diff
@@ -243,13 +243,13 @@ func (p *RunParams) verify() error {
 		return errors.New("rootCtxCancel may not be nil")
 	}
 	if p.nodeName != "" { // Heartbeating is enabled.
-		if p.gk == nil {
-			return errors.New("if heart beating is enabled, gk may not be nil")
+		if p.guardianSigner == nil {
+			return errors.New("if heart beating is enabled, vs may not be nil")
```
djb15, line 247: Supposed to be "gs" instead of "vs"?

djb15 (commented):
Very cool!

node/pkg/guardiansigner/filesigner.go
```diff
@@ -0,0 +1,99 @@
...
+	return fileSigner, nil
+}
+
+func (fs *FileSigner) Sign(hash []byte) ([]byte, error) {
+
+	// Sign the hash
+	sig, err := crypto.Sign(hash, fs.privateKey)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign wormchain address: %w", err)
```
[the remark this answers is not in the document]
johnsaigle: It's kinda nice IMO to use this strategy or else use [errors.Join](https://pkg.go.dev/errors#Join) to provide some context on where the errors are coming from. Otherwise, in Go, it's easy for an error to pop-up but have no idea where it was called because errors don't come with line numbers or stack traces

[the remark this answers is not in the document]
djb15: Yeah I like logging like this in general, but from what I've seen the caller of `Sign` gets the error and then wraps it in another string, both of which seem to be the same error string, so it ends up nested without any extra useful information

node/cmd/guardiand/node.go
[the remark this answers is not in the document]
pleasew8t: Yes I'm reverting back to nested if.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: will update comment that we are checking for a separator, and we don't care about the rest.

node/pkg/guardiansigner/filesigner.go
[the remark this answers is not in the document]
djb15: For clarity I didn't like the fact that in other places we also formatted the error with string "failed to sign wormchain address", so we'd end up with an error logged that looked like:

```
failed to sign wormchain address: failed to sign wormchain address: [ERROR]
```

Anyhow we chatted in a call and the decision was to tweak the error here to something like "failed to sign hash"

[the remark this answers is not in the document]
pleasew8t: I'm going to leave this as is for now, except for updating the error message in `FileSigner.Sign()`, because I overlooked that. In some instances the guardian panics if the signing failed, and other instances it wraps the error and continues.

node/cmd/guardiand/node.go
```diff
@@ -622,7 +625,17 @@ func runNode(cmd *cobra.Command, args []string) {
...
 	if *guardianKeyPath == "" {
-		logger.Fatal("Please specify --guardianKey")
+		// This if-statement is nested, since checking if both are empty at once will always result in the else-branch
+		// being executed if at least one is specified. For example, in the case where the singer URI is specified and
+		// the guardianKeyPath not, then the else-statement will create an empty `file://` URI.
+		if *guardianSignerUri == "" {
+			logger.Fatal("Please specify --guardianKey or --guardianSignerUri")
+		}
+	} else {
+		// If guardianKeyPath is set, set guardianSignerUri to the file signer URI, pointing to guardianKeyPath.
```
bruce-riley, line 635: Shouldn't you throw if both are set?

node/pkg/guardiansigner/generatedsigner.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+)
+
+type GeneratedSigner struct {
+	privateKey *ecdsa.PrivateKey
+}
+
```
bruce-riley, line 15: I know you have a comment in `guardianSigner.go` about this being for test only, but could you please put something in here as well?

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,73 @@
...
+
+	switch signerType {
+	case FileSignerType:
+		return NewFileSigner(unsafeDevMode, signerKeyConfig)
+	default:
+		return nil, fmt.Errorf("unsupported guardian signer type")
+	}
+}
+
+func ParseSignerUri(signerUri string) (signerType SignerType, signerKeyConfig string) {
```
bruce-riley, line 43: Should this return an error so line 39 could print more details?

bruce-riley (approved):

node/cmd/guardiand/node.go
pleasew8t: So I decided that I would default to using `guardianKeyPath`, even if the signer URI is set. I don't have any strong feelings about this decision. Do you think it should rather throw if both are set?

node/pkg/guardiansigner/generatedsigner.go
pleasew8t: I added a comment to the struct. I also got paranoid, and moved the `GenerateSignerWithPrivatekeyUnsafe` function to `generatedsigner.go`.

node/pkg/guardiansigner/guardiansigner.go
pleasew8t: Updated this to return an error.

node/cmd/guardiand/node.go
bruce-riley: It seems like someone is confused if both are set, so I would suggest forcing them to fix their config. The release notes could / should clearly state that they don't *have* to change anything, but if they do set the new parameter, they should remove the old one.

node/pkg/guardiansigner/guardiansigner.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func ParseSignerUri(signerUri string) (signerType SignerType, signerKeyConfig string, err error) {
+	// Split the URI using the standard "://" scheme separator
+	signerUriSplit := strings.Split(signerUri, "://")
+
+	// This check is purely for ensuring that there is actually a path separator.
+	if len(signerUriSplit) < 2 {
+		return InvalidSignerType, "", fmt.Errorf("No path separator in guardian signer URI")
```
bruce-riley, line 53: Sorry I didn't catch this the first time around, but @SEJeff taught me to use `errors.New` if the error string is constant. 😁

bruce-riley (approved):

djb15 (dismissed):

node/pkg/adminrpc/adminserver.go
SEJeff, line 1170: As @bruce-riley can likely attest, I do hate [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)). I personally know that an ecdsa signature is 64 bytes, but these magic numbers are best (in the future, or a followup PR) as a constant. There are often bugs when magic numbers aren't updated in the future that can be mitigated by using constants when sensible.

node/pkg/adminrpc/adminserver_test.go
```diff
@@ -120,8 +120,20 @@ func generateMockVAA(gsIndex uint32, gsKeys []*ecdsa.PrivateKey) []byte {
 		EmitterAddress:   vaa.Address{},
 		Payload:          []byte("test"),
 	}
-	for i, key := range gsKeys {
-		v.AddSignature(key, uint8(i))
+	for i, signer := range signers {
+		sig, err := signer.Sign(v.SigningDigest().Bytes())
+		if err != nil {
+			panic(err)
+		}
```
SEJeff, line 127: panicing like this from within a unit test is poor form, please don't do this! Just use testify with [require.NoError(err)](https://pkg.go.dev/github.com/stretchr/testify/assert#NoError). Also, pass in a `*testing.T` to this function and in the very first line, do something like: [t.Helper()](https://pkg.go.dev/testing#T.Helper)

node/pkg/adminrpc/adminserver_test.go
```diff
@@ -120,8 +120,20 @@ func generateMockVAA(gsIndex uint32, gsKeys []*ecdsa.PrivateKey) []byte {
...
+
+		sigData := [65]byte{}
```
SEJeff, line 129: Please use a constant for these sir.

node/pkg/guardiansigner/guardiansigner_test.go
```diff
@@ -0,0 +1,97 @@
...
+	//	That the signer returned is not nil
+	//	No error is returned
+	//	The public key returned by PublicKey(), converted to an eth address,
+	//		matches the expected address.
+
+	// Attempt to generate signer using top-level generator
+	fileSigner1, err := NewGuardianSignerFromUri(fileUri, true)
+	assert.NoError(t, err)
+	assert.NotNil(t, fileSigner1)
+	assert.Equal(t, ethcrypto.PubkeyToAddress(fileSigner1.PublicKey()).Hex(), expectedEthAddress)
```
SEJeff, line 71: If these top level assertions fail, I assume the rest of the stuff below will also fail. For these setup stuff that "needs to always return correctly or the entire rest of the state is wrong", consider not using assert, but instead doing [require.NoError(t, err)](https://pkg.go.dev/github.com/stretchr/testify/require#NoError).

The [require](https://pkg.go.dev/github.com/stretchr/testify/require) testify package is for ending execution of the rest of a given test if the required assertion fails. That way you have 1 failure in the console output instead of 11 if the `NewGuardianSignerFromUri()` unexpectedly fails.

SEJeff (approved):
All of my suggestions here are in general nits or things that make future maintenance a bit nicer. This is a nice clean implementation from all I see here. Nicely done sir! :partying_face:

node/pkg/guardiansigner/generatedsigner.go
```diff
@@ -0,0 +1,63 @@
...
+func (gs *GeneratedSigner) Verify(sig []byte, hash []byte) (valid bool, err error) {
+	recoveredPubKey, err := ethcrypto.SigToPub(hash, sig)
+
+	if err != nil {
+		return false, err
+	}
+
+	// Need to use gs.privateKey.Public() instead of PublicKey to ensure
+	// the returned public key has the right interface for Equal() to work.
+	fsPubkey := gs.privateKey.Public()
```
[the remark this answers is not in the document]
pleasew8t: I'm going to leave this as is for now. I remember fooling around with this when I was writing it. Casting the way you suggested fails, because the compiler complains that `gs.PublicKey() is not an interface`.

What I don't understand is why passing in `gs.PublicKey()` (type `ecdsa.PublicKey`) doesn't work, but `gs.privateKey.Public()` does (type `crypto.PublicKey`), despite the `crypto.PublicKey` being casted to `ecdsa.PublicKey` in the `Equal()` function [here](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/crypto/ecdsa/ecdsa.go;l=77). I might be missing something obvious.

node/pkg/guardiansigner/guardiansigner_test.go
```diff
@@ -0,0 +1,97 @@
...
+		label        string
+		path         string
+		expectedType SignerType
+	}{
+		{label: "RandomText", path: "RandomText", expectedType: InvalidSignerType},
+		{label: "ArbitraryUriScheme", path: "arb://data", expectedType: InvalidSignerType},
+		// File
+		{label: "FileURI", path: "file://whatever", expectedType: FileSignerType},
+		{label: "FileUriNoSchemeSeparator", path: "filewhatever", expectedType: InvalidSignerType},
+		{label: "FileUriMultipleSchemeSeparators", path: "file://testing://this://", expectedType: FileSignerType},
```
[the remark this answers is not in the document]
pleasew8t: I wanted to keep it flexible for any instances where a specific URI signer's path needs to contain a URI scheme operator. One use-case could be a signer implementation that has sub-implementations. Leaving this as is for now, but happy to change in my follow-up PR if the argument is a bit weak. I am in favor of a more opinionated approach, where the parsing is more strict. But I'm also trying to keep options open for the unknown.

node/pkg/guardiansigner/filesigner.go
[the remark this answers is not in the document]
pleasew8t: I am going to do this in the follow-up PR to avoid the need for review right now.

djb15 (approved):

node/pkg/adminrpc/adminserver.go
bruce-riley: I think for now this panic is okay. I think the original thinking was that signing is just a calculation, so if it fails,  something is really wrong. I think when we start adding signers that rely on external resources, we will need to rethink this. Having an external failure cause the guardian to roll in panics doesn't sound like a good plan.

bruce-riley (approved):

PR #4139, opened 2024-10-18
spy: respect --port if set

pires:
`guardiand spy --port` wouldn't be respected and the default port was always used.


pires:
IIUC the CI failure is not related to the code changes.

bruce-riley (approved):
Nice catch. Thanks!

panoel (approved):

PR #4143, opened 2024-10-23
docs: Audit reports

djb15:


johnsaigle (dismissed):

pleasew8t (commented):
The only thing that gets me is the naming convention. I think we should consider something like `<date>_<audit-firm>_<project-name>.pdf`.

johnsaigle (approved):

pleasew8t (approved):

PR #4148, opened 2024-10-29
node: add amazon kms and benchmark signers (old)

pleasew8t:
## Amazon KMS Guardian Signer

This PR adds an Amazon AWS KMS Guardian signer, allowing observations to be signed using KMS! The new signer can be used by specifying the ARN of the KMS key to use, through the `--guardianSignerUri` commandline argument, as follows:

* `--guardianSignerUri=amazonkms://<ARN>`

_NOTE_ For the best possible performance, it is recommended that the Guardian be run from an EC2 instance that is in the same region as the KMS key. 

The KMS key's spec should be `ECC_SECG_P256K1`, and should be enabled for signing. In order for the Guardian to authenticate against the KMS service, one of two options are available:

* Create new API keys in the AWS console that are permissioned to use the KMS key for signing, and add the keys to the EC2 instance's `~/.aws/credentials` file. ([example here](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html)).
* Create a role that is permissioned to use the KMS key and attach that role to the Guardian EC2 instance.


node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,167 @@
...
+	// try recovery id 0
+	ecSigWithRecid := append(signature, []byte{0}...)
+	pubkey, err := ethcrypto.SigToPub(hash[:], ecSigWithRecid)
+
+	if bytes.Equal(ethcrypto.CompressPubkey(pubkey), ethcrypto.CompressPubkey(&expectedPublicKey)) {
+		return ecSigWithRecid, nil
+	}
+
+	ecSigWithRecid = append(signature, []byte{1}...)
+	pubkey, err = ethcrypto.SigToPub(hash[:], ecSigWithRecid)
```
pires, line 101: Same comment as above.

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, errors.New("failed to load default config")
+	}
+
+	amazonKmsSigner.svc = kms.NewFromConfig(cfg)
+
+	return &amazonKmsSigner, nil
+}
+
+func (a *AmazonKms) Sign(hash []byte) (signature []byte, err error) {
```
pires, line 60: I'm sorry if I failed to flag this before, but i think it would be ideal if both `Sign` and `Verify` were to have a `context.Context` parameter.

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,167 @@
...
+//  5. the r value as a big-endian integer
+//  6. 0x02: header byte indicating an integer
+//  7. one byte to encode the length of the following s value
+//  8. the s value as a big-endian integer
+func derSignatureToRS(signature []byte) (rBytes []byte, sBytes []byte) {
+	var sigAsn1 asn1EcSig
+	_, err := asn1.Unmarshal(signature, &sigAsn1)
+
+	if err != nil {
+		panic(err)
```
pires, line 145: Is panicking part of the final design?

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	return sigAsn1.R.Bytes, sigAsn1.S.Bytes
+	// return rBytes, sBytes
+}
+
+func adjustBufferSize(b []byte) []byte {
```
pires, line 152: I suggest adding a comment explaining what this function is expected to do.

pires: And a unit table test to prove it works as expected, ie supports buffer with sizes equal, gt and lt 32, and the contents match the expected.

node/pkg/guardiansigner/benchmarksigner.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (b *BenchmarkSigner) PublicKey() ecdsa.PublicKey {
+
+	start := time.Now()
+
+	pubKey := b.innerSigner.PublicKey()
+
+	duration := time.Since(start)
+	fmt.Printf("Public key retrieval time: %v\n", duration)
```
pires, line 44: This could be exposed through [a Prometheus histogram](https://prometheus.io/docs/practices/histograms/).

node/pkg/guardiansigner/benchmarksigner.go
```diff
@@ -0,0 +1,59 @@
+package guardiansigner
+
+import (
+	"crypto/ecdsa"
+	"fmt"
+	"time"
+)
+
+type BenchmarkSigner struct {
```
pires, line 9: Please, document this.

pires: IIUC this means we can benchmark AWS KMS signer by passing its configuration to a signer of type `benchmark`, correct?

pires (commented):
Excited about trying this one 👏🏻

node/pkg/guardiansigner/amazonkms.go
pleasew8t: No, I don't think it should panic here. I removed it.

node/pkg/guardiansigner/amazonkms.go
pleasew8t: added comments and tests

node/pkg/guardiansigner/amazonkms.go
pleasew8t: Do you think I can just pass in the contexts that are available in the calling scope, and create a new one in scopes that don't have an existing context available?

pires: It's better than having no Context so I'm good with it.

node/pkg/guardiansigner/benchmarksigner.go
pires: Here's an example of the definition of an histogram https://github.com/wormhole-foundation/wormhole/blob/02f468f776f6c8b8d20c1cb5d192818c4fc20c75/node/pkg/processor/processor.go#L169 and how to observe latency https://github.com/wormhole-foundation/wormhole/blob/02f468f776f6c8b8d20c1cb5d192818c4fc20c75/node/pkg/processor/processor.go#L323

pires (commented):
For the record, as discussed privately, I think metrics should be produced all the time and not limited to when testing/benchmarking.

node/cmd/guardiand/adminclient.go
```diff
@@ -236,13 +236,15 @@ func getPublicRPCServiceClient(ctx context.Context, addr string) (*grpc.ClientCo
...
+	ctx := context.Background()
+
 	guardianSignerUri := args[0]
 	wormchainAddress := args[1]
 	if !strings.HasPrefix(wormchainAddress, "wormhole") || strings.HasPrefix(wormchainAddress, "wormholeval") {
 		return errors.New("must provide a bech32 address that has 'wormhole' prefix")
 	}
 
-	guardianSigner, err := guardiansigner.NewGuardianSignerFromUri(guardianSignerUri, *unsafeDevnetMode)
+	guardianSigner, err := guardiansigner.NewGuardianSignerFromUri(ctx, guardianSignerUri, *unsafeDevnetMode)
```
pires, line 247: Context shouldn't be passed to constructors but to calls, such as `Sign` and `Verify` by the callers.
- https://go.dev/blog/context-and-structs#prefer-contexts-passed-as-arguments
- https://go.dev/blog/context-and-structs#storing-context-in-structs-leads-to-confusion

That said, I think we can keep this as is and simply add comments explaining why this is a special case, i.e. you need a context for calling AWS KMS during signer initialization.

pires: Alternatively, the initialization may not ask for a parent Context and simply manage its own, e.g. a `Context.WithTimeout(timeout)` should suffice.

node/pkg/governor/governor_monitoring.go
```diff
@@ -600,7 +601,7 @@ func (gov *ChainGovernor) publishConfig(hb *gossipv1.Heartbeat, sendC chan<- []b
 
 	digest := ethCrypto.Keccak256Hash(append(governorMessagePrefixConfig, b...))
 
-	sig, err := guardianSigner.Sign(digest.Bytes())
+	sig, err := guardianSigner.Sign(context.Background(), digest.Bytes())
```
pires, line 604: There's isn't a Context being passed by the `cmd/guardiand/node.go`?

node/pkg/guardiansigner/benchmarksigner.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+
+	return &BenchmarkSigner{
+		innerSigner: innerSigner,
+	}
+}
+
+func (b *BenchmarkSigner) Sign(ctx context.Context, hash []byte) ([]byte, error) {
+
```
pires, line 51: Some methods have an empty newline after method signature, some don't. My OCD is going nuts.

pires (commented):
Other reviewers may have a different perspective on how this code relies on `context.Context` but aside from that I think this is ready for broader review.

node/pkg/governor/governor_monitoring.go
pleasew8t: The run function in `node/pkg/p2p/p2p.go` that gets produced does accept a context. I've modified `CollectMetrics` in governor monitoring to accept the context, and pass it to `publishConfig`, so it gets passed to `Sign`.

PR #4149, opened 2024-10-30
Node: Remove cutover checks

bruce-riley:
This PR removes the cutover checking now that the cutover time has passed and all guardians are publishing using the new formats.

Note that there is no particular urgency to release this PR. It can go out whenever.


node/pkg/processor/processor.go
```diff
@@ -474,8 +467,5 @@ func (p *Processor) vaaWriter(ctx context.Context) error {
 
 // GetFeatures returns the processor feature string that can be published in heartbeat messages.
 func GetFeatures() string {
-	if batchCutoverComplete() {
-		return "processor:batching"
```
panoel, line 478: Is there a way that we can get rid of "p2p:new_gossip" from the guardian heartbeat messages?  Should that be an issue?

panoel (approved):

node/pkg/processor/processor.go
bruce-riley: I'm removing lines 531-533 in `p2p.go`. That will remove it.

bingyuyap (approved):

PR #4152, opened 2024-11-04
Node: p2p.Run not always subscribing to heartbeat channel when it should

bruce-riley:
The `p2p.Run` method is not subscribing to the channel topic if a client (such as the historical uptime monitor) subscribes to heartbeats via the guardian set state but not any other of the other events that use the control topic. This results in never seeing heartbeats.

I tested this by hacking the spy to subscribe to heartbeats using the guardian set state.

### DIFF
--- a/node/pkg/common/guardianset.go
+++ b/node/pkg/common/guardianset.go
@@ -214,3 +214,8 @@ func (st *GuardianSetState) Cleanup() {
 		}
 	}
 }
+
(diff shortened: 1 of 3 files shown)


bingyuyap (approved):
lgtm!

panoel (approved):

SEJeff (approved):

PR #4164, opened 2024-11-20
Temporarily increasing governor limits for a few chains

banescusebi:
There is a discussion #4163 recommending that the daily limit for Solana and Ethereum be increased.


SEJeff (approved):
📈 🧔‍♂️

PR #4167, opened 2024-11-21
Update 0012_ibc_generic_messaging.md with permissionless channel creation information

mdulin2:
The `wormhole-ibc-receiver` contract contains an allowlist of Wormhole chain ID to channel mappings (other side of Cosmos chain). The allowlist is not used when receiving a packet because we don't want to handle the case of a token transfer failure on the other side. Instead, we just emit the message and ignore it in the IBC watcher. I am just updating the documentation to reflect this.


evan-gray (approved):
Per my understanding, this description appears to be accurate.

kcsongor (approved):

PR #4168, opened 2024-11-22
node: add amazon kms and benchmark signers

pleasew8t:
## Amazon KMS Guardian Signer

This PR adds an Amazon AWS KMS Guardian signer, allowing observations to be signed using KMS! The new signer can be used by specifying the ARN of the KMS key to use, through the `--guardianSignerUri` commandline argument, as follows:

* `--guardianSignerUri=amazonkms://<ARN>`

_NOTE_ For the best possible performance, it is recommended that the Guardian be run from an EC2 instance that is in the same region as the KMS key. 

The KMS key's spec should be `ECC_SECG_P256K1`, and should be enabled for signing. In order for the Guardian to authenticate against the KMS service, one of two options are available:

* Create new API keys in the AWS console that are permissioned to use the KMS key for signing, and add the keys to the EC2 instance's `~/.aws/credentials` file. ([example here](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html)).
* Create a role that is permissioned to use the KMS key and attach that role to the Guardian EC2 instance.

## Benchmark Signer

The PR also includes a benchmark signer, which wraps any other signer, logging signing and verification latency to prometheus histograms. External signing services might at times introduce unwanted latency, and if an event occurs where observation processing is particularly slow, the histograms would provide insight into whether or not the signing service is to blame.

_NOTE_
This is a redo of a previous pull request, which Pires spent time looking at. Below are key points following that review that informs the current state of the code:

 * `Context`s should be supplied to signing, verification and public key retrieval, as these functions potentially interact with 3rd party services that could timeout or block indefinitely.
* The `GuardianSigner` constructor (`NewGuardianSignerFromUri`) accepts a `Context`, as the new signer might need to interact with the 3rd party service to validate configurations (such as the `AmazonKms` signer). A different approach could be to have the constructor define its own context, to avoid the necessity of passing a context to it.
*


node/pkg/guardiansigner/benchmarksigner.go
```diff
@@ -0,0 +1,78 @@
...
+	if innerSigner == nil {
+		return nil
+	}
+
+	return &BenchmarkSigner{
+		innerSigner: innerSigner,
+	}
+}
+
+func (b *BenchmarkSigner) Sign(ctx context.Context, hash []byte) ([]byte, error) {
```
bruce-riley, line 52: Will errors skew the data? Should you add a metric for errors and bucket the latency separately or exclude them from the latency numbers?

bruce-riley (dismissed):

node/pkg/guardiansigner/benchmarksigner.go
pleasew8t: Thanks for this! I've added error counters.

pires (approved):
I've been involved in reviewing this before this PR was open so I want to signal I think this is in a good shape to be seriously considered for merge by other more seasoned maintainers.
That said, @andreclaro and xLabs Nodes team will be soaking this in testnet and report back 🙏🏻

node/pkg/guardiansigner/benchmarksigner.go
djb15: Following on from Bruce's comment, you're now tracking error counts but the latency figures aren't grouped by success/fail cases. Should we only observe the latency for a successful operation?

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,255 @@
...
+		keyId:  keyPath,
+		region: region,
+	}
+
+	// Create a configuration object to create a new KMS client from. The region passed to
+	// `config.WithDefaultRegion()` must match the region in the actual ARN, otherwise the SDK throws
+	// an error. This is why the region is first extracted from the keyPath.
+	cfg, err := config.LoadDefaultConfig(timeoutCtx, config.WithDefaultRegion(amazonKmsSigner.region))
+	if err != nil {
+		return nil, errors.New("Failed to load default config")
```
djb15, line 99: Should this error be more descriptive re being related to KMS? i.e. should every error string in this signer include the string "KMS"?

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,255 @@
...
+	var asn1Pubkey asn1EcPublicKey
+	_, err = asn1.Unmarshal(pubKeyOutput.PublicKey, &asn1Pubkey)
+
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unmarshal KMS public key: %w", err)
+	}
+
+	ecdsaPubkey := ecdsa.PublicKey{
+		X: new(big.Int).SetBytes(asn1Pubkey.PublicKey.Bytes[1 : 1+32]),
+		Y: new(big.Int).SetBytes(asn1Pubkey.PublicKey.Bytes[1+32:]),
```
djb15, line 123: Comment to explain the byte array indexes?

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,255 @@
...
+	// happen, unless there is something seriously wrong with the KMS service.
+	return nil, fmt.Errorf("Failed to generate valid signature")
+}
+
+func (a *AmazonKms) PublicKey(ctx context.Context) ecdsa.PublicKey {
+	return a.publicKey
+}
+
+func (a *AmazonKms) Verify(ctx context.Context, sig []byte, hash []byte) (bool, error) {
+	timeoutCtx, cancel := context.WithTimeout(ctx, time.Second*15)
```
djb15, line 195: Use `KMS_TIMEOUT` constant here

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return sigAsn1.R.Bytes, sigAsn1.S.Bytes, nil
+}
+
+// adjustBufferSize takes an input buffer and
+// a) trims it down to 32 bytes, if the input length is greater than 32, or
```
djb15, line 237: When I read "trim it down to 32 bytes" I would guess the least significant bytes are trimmed, but this trims from the front of the array. Maybe worth making this clearer in the comment?

node/cmd/guardiand/adminclient.go
```diff
@@ -236,13 +236,15 @@ func getPublicRPCServiceClient(ctx context.Context, addr string) (*grpc.ClientCo
 }
 
 func runSignWormchainValidatorAddress(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
```
djb15, line 239: Here and elsewhere in some other source files (e.g. `message.go`) you create new contexts without managing their lifecycle...what's the reasoning for this? Are we concerned about resource leaks?

node/pkg/guardiansigner/benchmarksigner.go
pires: >Should we only observe the latency for a successful operation?

We definitely should.

pleasew8t: Changed this. Thanks for the suggestion.

node/cmd/guardiand/adminclient.go
pleasew8t: Thanks for pointing this out. Didn't realise I missed a few. Where possible, I reuse the context from the caller if there is one available.

pleasew8t: FYI - I fixed this at multiple places, but the specific example you pointed out is actually fine imo. `runSignWormchainValidatorAddress` is a once-off command that is run from the cli.

node/pkg/guardiansigner/amazonkms.go
```diff
@@ -118,6 +118,12 @@ func NewAmazonKmsSigner(ctx context.Context, unsafeDevMode bool, keyPath string)
 		return nil, fmt.Errorf("Failed to unmarshal KMS public key: %w", err)
 	}
 
+	// It is possible to use `ethcrypto.UnmarshalPubkey(asn1Pubkey.PublicKey.Bytes)`` to get the public key,
+	// but `UnmarshalPubkey()` uses elliptic.Unmarshal() internally, which has been marked as deprecated.
+	// The following code implements similar logic, with the indexes meaning the following:
+	// 0: The first byte is the prefix byte, which is 0x04 for uncompressed keys.
+	// 1-32: The next 32 bytes are the X coordinate.
+	// 33-64: The next 32 bytes are the Y coordinate.
```
djb15, line 126: Thanks for the extra docstring. Is it worth doing an explicit length check on the byte array before slicing from it? Or is the length guaranteed in package code?

pleasew8t: I wouldn't feel comfortable making that assumption after checking the code. Added a minimum length check.

djb15 (approved):

bruce-riley (approved):

PR #4169, opened 2024-11-22
node: Add Transfer Verifier mechanism

johnsaigle:
# Transfer Verifier

_This PR is a joint effort between myself, @pleasew8t, and @mdulin2._ 

## Overview

This PR introduces the **Transfer Verifier**, a mechanism to improve Token
Bridge security by validating transaction receipts that correspond to Publish
Message events emitted from the Core Bridge.

When the core bridge emits a Publish Message event, the Transfer Verifier does
the following:
* Fetches the transaction artifacts corresponding to the Publish Message event
(e.g. an Ethereum receipt transaction)
* Checks whether the publish message corresponds to a token transfer and is
emitted by the Token Bridge contract
* Cross-references the data in the publish message against actual deposits or
transfers into the token bridge
* If there is publish message with no valid corresponding deposit or transfers
into the Token Bridge, the program emits an error to signal to the node that
there is a security incident.

This provides protection in the event that an attacker devises a way to emit
arbitrary publish message events from the core contract or else somehow tricks
the Token Bridge into making token transfer calls without any backing tokens.
While existing Integrity Checks (Governor and Accountant) would help mitigate
this attack, Transfer Verifier acts as another layer of protection.

## What this PR contains

This PR contains two initial implementations for Ethereum and Sui. 

Both implementations are bundled into a Transfer Verifier package. This package
is used by a new `transfer-verifier` binary that exists as a subcommand within
`guardiand`. This binary can be configured to monitor the mainnet token bridge and core
bridge contracts for this chain. 

## Testing

Both implementations contain a large number of unit tests. They can be invoked
as a normal part of testing the node via `make test`.

The Ethereum implementation has integration tests built into the Tilt set-up:
- After the eth-devnet container is set-up, the transfer-verifier binary is
invoked to monitor the Ethereum devnet
- A shell script invokes a malicious call to the core contract via `cast` (part
of the Foundry toolset)
- The transfer-verifier binary logs an error when detecting the malicious call
- A separate monitoring container looks for the presence of this error log and
returns a success when it appears

In order to impersonate a call from the Token Bridge to the Core Bridge, the
`--auto-impersonate` flag was added to the `anvil` invocation in eth-devnet.

## Future Work

The Transfer Verifier package is used here as a standalone binary. Its
fundamental output are logs that can be monitored and manually acted upon.

The next iteration of this project will be to integrate the package directly in
the Ethereum and Sui watchers

### Integration into the Ethereum watcher

Ethereum has an advantage over the Sui implementation due to the robust logging
and tooling present in the EVM and its ecosystem. We can validate the relevant
information for Transfer Verification in a highly robust way and detect
violations of its invariants with certainty.

As a result, we can use the logic in the Transfer Verifier to block
transactions at the level of the Ethereum watcher so that malicious message
publications will not be processed by the network.

### Integration into the Sui watcher

The nature of Sui makes it relatively difficult to provide guarantees about the
state of the core bridge relative to the token bridge. The Sui implementation
cannot provide the same guarantees as the Ethereum implementation due to limits
inherent in the chain.

It is still possible to integrate this protection into the Sui watcher but it
would be better to delay messages for e.g. 24h in this case rather than reject
them outright.

### Additional implementations for other chains 
This is an abstract algorithm that can in theory be applied to any chain supported by the Token Bridge.
Designing a Solana implementation of this script would be a fruitful choice
given that it is one of the chains that makes the most use of the Token Bridge.


node/cmd/transfer-verifier/transfer-verifier-sui.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	logger := ipfslog.Logger("wormhole-transfer-verifier-sui").Desugar()
+
+	ipfslog.SetAllLoggers(lvl)
+	logger.Info("Starting Sui transfer verifier")
+	logger.Debug("Sui rpc connection", zap.String("url", *suiRPC))
+	logger.Debug("Sui core contract", zap.String("address", *suiCoreContract))
+	logger.Debug("Sui token bridge contract", zap.String("address", *suiTokenBridgeContract))
+	logger.Debug("token bridge event emitter", zap.String("object id", *suiTokenBridgeEmitter))
```
panoel, line 60: Since these are all debugs, do you want to add `suiProcessInitialEvents`?

node/pkg/transfer-verifier/transfer-verifier-sui-structs.go
```diff
@@ -0,0 +1,436 @@
...
+}
+
+// Validate the type information of the object change. The following checks are performed:
+//   - pass the object through a regex that extractts the package ID, coin type, and asset type
+//   - ensure that the asset type is wrapped or native
+//   - ensure that the package IDs match the expected package ID
+//   - ensure that the coin types match
+func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success bool) {
+	// AI generated regex
+	re := regexp.MustCompile(`^0x2::dynamic_field::Field<([^:]+)::token_registry::Key<([^>]+)>, ([^:]+)::([^<]+)<([^>]+)>>$`)
```
panoel, line 98: I think this panics on error.  That seems a bit extreme?  Should we consider something less extreme?

node/pkg/transfer-verifier/transfer-verifier-sui-structs.go
```diff
@@ -0,0 +1,436 @@
...
+
+type ObjectChange struct {
+	ObjectType      string `json:"objectType"`
+	ObjectId        string `json:"objectId"`
+	Version         string `json:"version"`
+	PreviousVersion string `json:"previousVersion"`
+}
+
+// Validate the type information of the object change. The following checks are performed:
+//   - pass the object through a regex that extractts the package ID, coin type, and asset type
```
panoel, line 92: nit:

```suggestion
//   - pass the object through a regex that extracts the package ID, coin type, and asset type
```

node/pkg/transfer-verifier/transfer-verifier-evm-structs.go
```diff
@@ -0,0 +1,989 @@
...
+	// https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
+	WETH_ADDRESS     = common.HexToAddress("c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2")
+	ZERO_ADDRESS     = common.BytesToAddress([]byte{0x00})
+	ZERO_ADDRESS_VAA = VAAAddrFrom(ZERO_ADDRESS)
+)
+
+// EVM chain constants
+const (
+	// The Wormhole Chain ID for the chain being monitored
+	NATIVE_CHAIN_ID = 2
```
bruce-riley, line 62: This makes this Ethereum specific, rather than general EVM. Is that your intent? Maybe you should take the chain ID as an argument? Also, it won't even work for Sepolia / testnet.

node/cmd/transfer-verifier/transfer-verifier-evm.go
```diff
@@ -0,0 +1,120 @@
...
+
+var TransferVerifierCmdEvm = &cobra.Command{
+	Use:   "evm",
+	Short: "Transfer Verifier for EVM-based chains",
+	Run:   runTransferVerifierEvm,
+}
+
+// Configuration variables for EVM interactions.
+var (
+	// RPC endpoint URL for interacting with an EVM node.
```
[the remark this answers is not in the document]
mdulin2: Eventually, I think that's the goal of it. 

Right now, the package is standalone but should be inside of the watchers. Once there, it makes sense to add telemetry. That's the next step of the project.

node/pkg/transfer-verifier/transfer-verifier-evm-structs.go
mdulin2: Good call out! We can make this dynamic.

node/cmd/transfer-verifier/transfer-verifier-evm.go
```diff
@@ -0,0 +1,120 @@
...
+	// MAIN LOOP:
+	// - watch for LogMessagePublished events coming from the connector attached to the core bridge.
+	// - process the events through the transfer verifier.
+	for {
+		select {
+		case subErr := <-sub.Errors():
+			logger.Warn("error on subscription", zap.Error(subErr))
+
+		// Process observed LogMessagePublished events
+		case vLog := <-sub.Events():
```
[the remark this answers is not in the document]
mdulin2: Why what is here? I'm slightly confused at what you're calling out here.

node/pkg/transfer-verifier/transfer-verifier-evm-structs.go
pleasew8t: Good suggestion. Made a few changes to accommodate this. Instead of passing it as an argument to the constructor, we can obtain the chain ID from the connector.

node/pkg/transfer-verifier/transfer-verifier-sui.go
```diff
@@ -0,0 +1,301 @@
...
+		suiEventType:           fmt.Sprintf("%s::%s::%s", suiCoreContract, suiModule, suiEventName),
+	}
+}
+
+// func (s *SuiTransferVerifier) GetSuiEventType() string {
+// 	return s.suiEventType
+// }
+
+// Filter to be used for querying events
+// The `MoveEventType` filter doesn't seem to be available in the documentation. However, there is an example
```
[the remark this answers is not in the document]
pleasew8t: Yeah the docs don't offer this event filter: https://docs.sui.io/guides/developer/sui-101/using-events

node/pkg/transfer-verifier/transfer-verifier-sui-structs.go
pleasew8t: I was thinking about this, and the thing is that the actual regex remains static. Is there a reason it would compile during tests, but not during a live production run?

node/cmd/transfer-verifier/transfer-verifier-sui.go
```diff
@@ -0,0 +1,168 @@
...
+var TransferVerifierCmdSui = &cobra.Command{
+	Use:   "sui",
+	Short: "Transfer Verifier for Sui",
+	Run:   runTransferVerifierSui,
+}
+
+// CLI parameters
+func init() {
+	suiRPC = TransferVerifierCmdSui.Flags().String("suiRPC", "", "Sui RPC url")
+	suiCoreContract = TransferVerifierCmdSui.Flags().String("suiCoreContract", "", "Event to listen to in Sui")
```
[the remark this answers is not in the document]
pleasew8t: Redoing some of the cli arg text for evm and sui, thanks for bringing this to my attention.

node/cmd/transfer-verifier/transfer-verifier-evm.go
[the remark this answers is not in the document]
johnsaigle: I added some basic parameters to establish a connection to Loki after speaking with @panoel about it today.

node/cmd/transfer-verifier/transfer-verifier-evm.go
```diff
@@ -0,0 +1,120 @@
...
+			zap.Error(connectErr))
+	}
+
+	// Create main configuration for Transfer Verification
+	transferVerifier := txverifier.NewTransferVerifier(
+		ethConnector,
+		&txverifier.TVAddresses{
+			CoreBridgeAddr:  common.HexToAddress(*evmCoreContract),
+			TokenBridgeAddr: common.HexToAddress(*evmTokenBridgeContract),
+			// TODO: should be a CLI parameter so that we could support other EVM chains
```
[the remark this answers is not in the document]
bruce-riley: Well, this means this will not only work on ETH, but it will only work on ETH mainnet. Users might have the impression that they can set the RPC and the core and token bridge address for either Sepolia or any other EVM, but things will not work correctly because the WETH address would be wrong, right?

node/pkg/transfer-verifier/transfer-verifier-evm.go
```diff
@@ -0,0 +1,643 @@
...
+	if cmp(vLog.Sender, tv.Addresses.TokenBridgeAddr) != 0 {
+		tv.logger.Debug("skip: sender is not token bridge",
+			zap.String("txHash", vLog.Raw.TxHash.String()),
+			zap.String("sender", vLog.Sender.String()),
+			zap.String("tokenBridge", tv.Addresses.TokenBridgeAddr.String()))
+		return true
+	}
+
+	// Get the full transaction receipt for this log.
+	receipt, txReceiptErr := tv.evmConnector.TransactionReceipt(ctx, vLog.Raw.TxHash)
```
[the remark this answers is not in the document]
johnsaigle: Thanks, I've modified this in the latest commit

johnsaigle:
After discussing with Bruce, I've added a mapping from EVM --> Wormhole chain IDs where appropriate.

johnsaigle:
@bruce-riley @panoel When you have time, would you mind looking through the resolutions and updating your reviews?

bruce-riley (dismissed):

panoel (approved):

johnsaigle:
@bruce-riley @panoel sorry to tag you again, but I'll need you to review since I had to force-push after fixing a conflict.

bruce-riley (approved):

pleasew8t (dismissed):

panoel (approved):

bruce-riley (dismissed):

SEJeff (dismissed):

mdulin2 (approved):
I reviewed when we initially put it into Johns branch. So, I'm happy with the state of it.

node/pkg/transfer-verifier/transfer-verifier-evm-structs.go
```diff
@@ -0,0 +1,1056 @@
...
+	}
+	if len(result) < EVM_WORD_LENGTH {
+		tv.logger.Warn("isWrappedAsset() result length is too small", zap.String("result", fmt.Sprintf("%x", result)))
+		return false, err
+	}
+	tv.logger.Debug("isWrappedAsset result", zap.String("result", fmt.Sprintf("%x", result)))
+
+	// The boolean result will be returned as a byte string with length
+	// equal to EVM_WORD_LENGTH. Grab the last byte.
+	// TODO is 1 == true? Does this work?
```
djb15, line 800: Looks correct to me, think you can kill this todo

node/pkg/transfer-verifier/transfer-verifier-evm-structs.go
```diff
@@ -0,0 +1,1056 @@
...
+		tv.logger.Warn("failed to get decimals for token: result has insufficient length",
+			zap.String("tokenAddress", tokenAddress.String()),
+			zap.ByteString("result", result))
+		return 0, err
+	}
+
+	// An ERC20 token's decimals should fit in a single byte. A call to `decimals()`
+	// returns a uint8 value encoded in string with 32-bytes. To get the decimals,
+	// we grab the last byte, expecting all the preceding bytes to be equal to 0.
+	// TODO: find out if there is some official documentation for why this uint8 is in the last index of the 32byte return.
```
djb15, line 1002: Yellowpaper specifies big endianness

djb15 (approved):
Some nits to deal with in the next phase with the follow up PRs. Had a look through the EVM processing logic and the parsing + processing looks good to me

johnsaigle:
@djb15 @mdulin2 thanks for the reviews, mind doing it again? 😅 I decided to patch the endianness concerns but ended up doing a force-push or three in the process

djb15 (dismissed):

SEJeff (dismissed):

devnet/tx-verifier-monitor/transfer-verifier-test.sh
```diff
@@ -0,0 +1,117 @@
...
+TOKEN_BRIDGE_CONTRACT=0x0290FB167208Af455bB137780163b7B7a9a10C16
+
+MNEMONIC=0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
+
+ERC20_ADDR="0x47bdB2D7d6528C760b6f228b3B8F9F650169a10f" # Test token A
+
+VALUE="1000" # Wei value sent as msg.value
+TRANSFER_AMOUNT="10"
+
+ANVIL_USER0="0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1" # Account0 reported by anvil when run using $MNEMONIC
```
evan-gray, line 21: account 0 is used in the token bridge integration test [here](https://github.com/wormhole-foundation/wormhole/blob/a46bccaa9b743d3d23b1f956158b4e7bb2753497/sdk/js/src/token_bridge/__tests__/utils/consts.ts#L8). not sure if this could result in race conditions when used with `anvil_impersonateAccount` though. just flagging should it come up.

devnet/tx-verifier-monitor/Dockerfile.cast
```diff
@@ -0,0 +1,13 @@
+# These versions are pinned to match the Dockerfile in the `ethereum/`
+# directory. Otherwise, there is nothing special about them and they can be
+# updated alongside the other Dockerfile.
+FROM ghcr.io/foundry-rs/foundry:nightly-55bf41564f605cae3ca4c95ac5d468b1f14447f9@sha256:8c15d322da81a6deaf827222e173f3f81c653136a3518d5eeb41250a0f2e17ea as foundry
```
evan-gray, line 4: This does not work on mac without specifying `--platform=linux/amd64`. Please update (see `ethereum/Dockerfile`).

evan-gray: Alternatively, I built a multi-platform one over at https://github.com/orgs/wormholelabs-xyz/packages/container/package/foundry, but I haven't had time to debug #4186 yet.

Tiltfile
```diff
@@ -463,6 +463,7 @@ k8s_resource(
     trigger_mode = trigger_mode,
 )
```
evan-gray, line 465: new line?

Tiltfile
```diff
@@ -676,6 +697,20 @@ if ci_tests:
         trigger_mode = trigger_mode,
         resource_deps = [], # testing/querysdk.sh handles waiting for query-server, not having deps gets the build earlier
     )
+    # launches tx-verifier binary and sets up monitoring script
+    k8s_resource(
+        "tx-verifier-with-monitor",
+        resource_deps = ["eth-devnet"],
+        labels = ["evm", "tx-verifier"],
```
evan-gray, line 704: if you label this with two labels, it shows up twice in tilt. I'd probably only label these with `tx-verifier`.

node/cmd/root.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/certusone/wormhole/node/cmd/ccq"
 	"github.com/certusone/wormhole/node/cmd/debug"
 	"github.com/certusone/wormhole/node/cmd/spy"
+	transferverifier "github.com/certusone/wormhole/node/cmd/transfer-verifier"
```
evan-gray, line 10: It is [go convention](https://go.dev/blog/package-names) not to use dashes or underscores, e.g. guardiansigner. Please match your folders to your package name.

node/cmd/transfer-verifier/transfer-verifier-evm.go
```diff
@@ -0,0 +1,157 @@
+package transferverifier
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/certusone/wormhole/node/pkg/telemetry"
+	txverifier "github.com/certusone/wormhole/node/pkg/transfer-verifier"
```
evan-gray, line 9: It is [go convention](https://go.dev/blog/package-names) not to use dashes or underscores, e.g. guardiansigner. Please match your folders to your package name.

devnet/tx-verifier-monitor.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: apps/v1
```
evan-gray, line 1: is this file used anywhere?

ethereum/forge-scripts/DeployTestToken.s.sol
```diff
@@ -95,11 +98,26 @@ contract DeployTestToken is Script {
...
 
+        // Deploy a test token for Transfer Verification
+        ERC20PresetMinterPauser deployedA = new ERC20PresetMinterPauser(
+            "TransferVerifier Test Token A",
+            "TVA"
+        );
+
+        console.log("Test token A deployed at: ", address(deployedA));
+
+        for(uint16 i=0; i<11; i++) {
```
evan-gray, line 109: do you need to give 10 accounts these tokens? it looks like you only use 1

evan-gray (changes requested):
reviewed tilt and devnet changes

kcsongor (approved):
looked at the `ethereum` folder only

johnsaigle:
Thanks for the review @evan-gray, agreed with the feedback and I'll issue fixes soon.

djb15 (approved):

Tiltfile
```diff
@@ -676,6 +696,20 @@ if ci_tests:
...
+        "tx-verifier-with-monitor",
+        resource_deps = ["eth-devnet"],
+        labels = ["tx-verifier"],
+        trigger_mode = trigger_mode,
+    )
+    # triggers the integration tests that will be detected by the monitor
+    k8s_resource(
+        "tx-verifier-test",
+        resource_deps = ["eth-devnet", "tx-verifier-with-monitor"],
+        labels = ["evm", "tx-verifier"],
```
evan-gray, line 710: missed removing this `evm` label

ethereum/forge-scripts/DeployTestToken.s.sol
```diff
@@ -95,11 +98,31 @@ contract DeployTestToken is Script {
...
+        // Deploy a test token for Transfer Verification
+        ERC20PresetMinterPauser deployedA = new ERC20PresetMinterPauser(
+            "TransferVerifier Test Token A",
+            "TVA"
+        );
+
+        console.log("Test token A deployed at: ", address(deployedA));
+
+        // Mint test tokens to Anvil's default account at index 1.
+        // The account at index 0 is used for other tests in the devnet, so
```
evan-gray, line 110: account 1 is also used by the nft integration test [here](https://github.com/wormhole-foundation/wormhole/blob/883d94a92c2e6654d7414237f5c8bae7da57e038/sdk/js/src/nft_bridge/__tests__/utils/consts.ts#L11), as are many of the keys, which is why the anvil [accounts flag](https://github.com/wormhole-foundation/wormhole/blob/883d94a92c2e6654d7414237f5c8bae7da57e038/devnet/eth-devnet.yaml#L42) is set to 13. try searching the private keys (listed by anvil on startup) to ensure they are not used or bump the accounts flag to 14 and use the new key.

devnet/tx-verifier-monitor.yaml
evan-gray, line 1: is _this_ whole file used anywhere?

evan-gray (changes requested):

devnet/tx-verifier-monitor.yaml
johnsaigle: Nope, it was part of an earlier phase of building the integration tests. Thanks for the catch.

ethereum/forge-scripts/DeployTestToken.s.sol
johnsaigle: Got it, I've updated these tests to use account[13] and added some comments

Tiltfile
johnsaigle: Thanks, should be gone now.

evan-gray (approved):
requested changes resolved, thanks

kcsongor (approved):

PR #4174, opened 2024-11-29
aptos: upgrade to 4.5.0 (toolchain + libs)

ali-bahjati:
This PR is modeled on [the last change that previously bumped to 3.1.0](https://github.com/wormhole-foundation/wormhole/pull/3915)

This change bumps libraries and toolchain of Aptos to the Aptos CLI [v4.5.0](https://github.com/aptos-labs/aptos-core/releases/tag/aptos-cli-v4.5.0) version. The tests are slightly changed to accommodate the recent changes in the coin library that disallows coin operations on unitialized coins ([this PR](https://github.com/aptos-labs/aptos-core/pull/13228)).

The PR is in draft state because `aptos/Dockerfile` requires a docker hash that I am unable to push myself so assuming this PR looks OK, and someone with permission can do that push, I can fix the PR as well and mark ready.


aptos/Dockerfile.base
```diff
@@ -8,7 +8,7 @@ RUN git clone https://github.com/aptos-labs/aptos-core.git
 WORKDIR /tmp/aptos-core
 
 # Build aptos 2.0.3
```
kcsongor, line 10: maybe we should just remove this line?

evan-gray:
made the extra commits fixups and bumped the cli version of the `aptos` package to make this work, as it was handing on `client.simulateTransaction`

evan-gray:
"fixed" the cli complaints, but the real fix would be removing the reliance on the deprecated sdk

evan-gray:
This PR is failing due to two issues.
1. certain transactions during deploy and the NFT tests are failing on `client.simulateTransaction`. I added commits to bypass this, though I am unsure as to the root cause.
2. `aptosClient.getEventsByEventHandle` is not returning the expected data, leading to errors like the following

```
    TypeError: Cannot read properties of undefined (reading 'data')
    const depositTokenId = event.data.id;
```

I am not familiar enough with the Aptos changes to immediately identify what needs to be changed here.

PR #4184, opened 2024-12-10
tilt: more multi-platform support

evan-gray:
This PR removes the m1 flag in favor of the multi-platform `solana-test-validator` image [[src](https://github.com/wormholelabs-xyz/solana-test-validator), [image](https://github.com/orgs/wormholelabs-xyz/packages/container/package/solana-test-validator)]. It also uses the newly rebuilt multi-image aptos and sui images and makes the guardian image compatible with an arm64 or amd64 build.

With this PR, Near should now be the only container and test that fails on M-series Macs, which I tested via `tilt up -- --ci --num=2 --near=false`.

It is important to note that, when available, images should pin the *index* digest, which identifies the registry entry, not the manifext digest, which identifies a single platform image. See https://docs.docker.com/build/building/multi-platform/.

Reviewers: I wasn't able to build Sui from source with rust:1.62 but it appears to work with 1.79. Not sure if that would cause any other issues but it appears to work.

### DIFF
--- a/sdk/js-query/src/query/solana.test.ts
+++ b/sdk/js-query/src/query/solana.test.ts
@@ -204,7 +204,9 @@ describe("solana", () => {
     expect(sar.results[0].lamports.toString()).toEqual(
       BigInt(1461600).toString()
     );
-    expect(sar.results[0].rentEpoch.toString()).toEqual(BigInt(0).toString());
+    expect(sar.results[0].rentEpoch.toString()).toEqual(
+      BigInt("18446744073709551615").toString()
+    );
     expect(sar.results[0].executable).toEqual(false);
     expect(base58.encode(Buffer.from(sar.results[0].owner))).toEqual(
       "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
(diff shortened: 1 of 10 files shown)


sdk/js-query/src/query/solana.test.ts
```diff
@@ -204,7 +204,9 @@ describe("solana", () => {
     expect(sar.results[0].lamports.toString()).toEqual(
       BigInt(1461600).toString()
     );
-    expect(sar.results[0].rentEpoch.toString()).toEqual(BigInt(0).toString());
+    expect(sar.results[0].rentEpoch.toString()).toEqual(
+      BigInt("18446744073709551615").toString()
```
kcsongor, line 208: 🤔

kcsongor (approved):

sdk/js-query/src/query/solana.test.ts
evan-gray: *sigh*... the older version of `solana-test-validator` just returned a rentEpoch of `0` and the new one returns `18446744073709551615`. It is nice to have a value now.

bruce-riley: Is it worth defining a constant with a meaningful name rather than using the magic value everywhere? Make it easier if they change the value.

bruce-riley (approved):

PR #4185, opened 2024-12-11
node: add keyprint command

pleasew8t:
This PR adds a new `keyprint` command, which unmarshals a guardian-generated key file and prints out the private key as a hex string. With KMS incoming (along with other potential signing mechanisms) it might be desirable to reuse/import an existing key into the key management solution.

Usage: `guardiand keyprint <path-to-keyfile>`


bruce-riley:
Does this open a new security hole where anyone with access to a guardian box can print the private key?

andreclaro (approved):

djb15:
Stale

PR #4186, opened 2024-12-12
tilt: even more multi-platform support

evan-gray:
This PR adds more multi-platform support. With this PR, `tilt up -- --ci` should come up green!
- Near now works on Mac 🙌 
- Switched to a foundry registry that has the amd64 _and_ arm64 images

As noted in the markdown file, I ran into an issue doing the multi-platform build from a Mac, likely due to lack of AVX support - not an issue, just something for folks to be aware of should that image ever need to be updated. I simply built from a native arm64 machine and then a native amd64 machine, pushed them both, and made an index from those.

As of this writing, Foundry does not provide arm64 docker images on [ghcr](https://github.com/foundry-rs/foundry/pkgs/container/foundry), just amd64 😿 So I just built the multi-platform image from the matching commit and pushed it [here](https://github.com/orgs/wormholelabs-xyz/packages/container/package/foundry).

Leaving this in draft as I plan to make a couple of fixes for the various `node` images as well as fixing some of the tilt build triggers.


johnsaigle:
TL;DR I can't get a local build to work but I think it's because my machine doesn't have the resources to run the example commands.

If I run just `tilt up -- --wormchain` though, it works.

---

Things tried:
- confirmed local commit 8f1f3aae4
- increased Docker settings to 8 CPU/24GB
- disabled VPN in case of weird networking issues
- pruned all Docker images and deleted the kubectl config
- set the default context to wormhole as described in DEVELOP.md

wormchain itself actually shows up green for the tests, but generally I can't do `tilt up -- --ci --num=2 --maxWorkers=2`

My suspicion is that my computer is too resource-constrained and the various errors are basically bogus but I'm curious about your thoughts.

eth-devnet and Wormchain fail with 'disk pressure' problems:
```
Tracking new pod rollout (eth-devnet-0):
     ┃ Not Scheduled   - (Unschedulable): 0/1 nodes are available: 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure: }. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
     ┊ Initialized     - (…) Pending
     ┊ Ready           - (…) Pending
[event: pod wormhole/eth-devnet-0] 0/1 nodes are available: 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure: }. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
```

Wormchain
```
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] The node had condition: [DiskPressure].
[event: pod wormhole/wormchain-1] 0/1 nodes are available: 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure: }. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
```

Algorand
```
Detected container restart. Pod: algorand-0. Container: algorand-postgres.
[event: pod wormhole/algorand-0] 0/1 nodes are available: 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure: }. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
Reconnecting... Error port-forwarding algorand (8980 -> 8980): error upgrading connection: pod algorand-0 does not have a host assigned
```
NEAR
```
Build Failed: ImageBuild: const-gen:tilt-c24104124adf27cc: failed to resolve source metadata for docker.io/library/const-gen:tilt-c24104124adf27cc: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```
Terra2
```
Build Failed: ImageBuild: cosmwasm_artifacts:tilt-847dde7a4b3cb2c0: failed to resolve source metadata for docker.io/library/cosmwasm_artifacts:tilt-847dde7a4b3cb2c0: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

evan-gray:
@johnsaigle how much disk space do you have available to docker (in settings > resources)? May need 200 GB to be safe. I run with 400 GB set. I don’t think it should require nearly that much, but it’s definitely more than 20. The disk pressure errors make me think you’re OOSing the vm. Also just to double check, if you’re on a Mac, that you’re using the Apple Virtualization framework (settings > general). I can detail settings that work for me if it’s none of these.

PR #4189, opened 2024-12-12
chore: add cw wormhole interchain tests

joelsmith-2019:
Adds an interchain test which runs tests against the core contract on Wormchain. It tests the query and execution endpoints within the contract (both happy and error paths). Also, these tests will run in GitHub workflows.

**Side Note**
Looks like the whitepaper and the actual implementation of the TransferFees VAA do not align. In the whitepaper, it states that the payload should be structured as so: [whitepaper](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0004_message_publishing.md?plain=1#L103-L115). However, the implementation within the rust contract swaps the amount and recipient's order within the byte array: [contract](https://github.com/wormhole-foundation/wormhole/blob/main/cosmwasm/contracts/wormhole/src/state.rs#L386-L397).


wormchain/interchaintest/cw_wormhole_test.go
```diff
@@ -0,0 +1,482 @@
...
+			NumValidators: &numWormchainVals,
+			NumFullNodes:  &numFullNodes,
+		},
+	})
+
+	// Get chains from the chain factory
+	chains, err := cf.Chains(t.Name())
+	require.NoError(t, err)
+
+	return chains
```
kakucodes, line 46: i think it makes sense to return a single chain here instead of []ibc.Chain, to make it easier to consume in the tests

wormchain/interchaintest/helpers/cw_wormhole/helpers.go
```diff
@@ -0,0 +1,408 @@
...
+// VerifyFee verifies the fee amount in the contract state
+func VerifyFee(
+	t *testing.T,
+	ctx context.Context,
+	wormchain *cosmos.CosmosChain,
+	contractAddr string,
+	expectedAmount string,
+) {
+	var stateResp GetStateQueryResponse
+	err := wormchain.QueryContract(ctx, contractAddr, QueryMsg{
```
kakucodes, line 270: i think querying the fee is good, but it would be better if we also attempt to submit a tx with a lower fee and ensure that it fails and using the new fee succeeds.

kakucodes (changes requested):
Looks good to me with just a couple changes.

kakucodes (approved):
LGTM

nik-suri (approved):
LGTM. Verified that this change only affects non-functional testing code.

evan-gray (approved):
approving workflow, test only changes.

PR #4196, opened 2024-12-18
linea-mainnet: env file + jsclient constants

hernandp:


aadam-10:
included in https://github.com/wormhole-foundation/wormhole/pull/4464

PR #4203, opened 2024-12-19
feat: mainnet validator hot swaps

joelsmith-2019:
This PR allows mainnet validator hot swapping. Previously, only single validator clusters allowed hot swapping for testnet purposes. This PR removes that check and introduces an integration test which validates the process of swapping an existing guardian's validator to a new validator and then back again.


wormchain/x/wormhole/keeper/msg_server_register_account_as_guardian.go
```diff
@@ -50,16 +51,6 @@ func (k msgServer) RegisterAccountAsGuardian(goCtx context.Context, msg *types.M
 		return nil, types.ErrGuardianSetNotFound
 	}
 
-	consensusGuardianSetIndex, consensusIndexFound := k.GetConsensusGuardianSetIndex(ctx)
```
nik-suri, line 53: will removing this change the amount of gas that's used for the transaction? That could lead to consensus errors. If so we could gate this functionality behind a block height.

joelsmith-2019: I'm unsure of that. We could still keep the check in to verify an existing consensus guardian set index exists.

mdulin2: I believe that reads from storage require gas. So, this would be consensus-breaking.

This leads to three options: 
- Leave the lookup in there but don't use it. 
- Add a block-height check for the change that gives validators time to update. Personally, I don't want to add to many of these, as it makes the code base super messy.
- Do a hard-fork with a timed update. This would require a little downtime.


I'm partial to 1.

wormchain/x/wormhole/keeper/msg_server_register_account_as_guardian.go
```diff
@@ -25,6 +25,7 @@ func (k msgServer) RegisterAccountAsGuardian(goCtx context.Context, msg *types.M
 	if err != nil {
 		return nil, err
 	}
+
```
mdulin2, line 28: Can you update the comments above the function to be accurate? With this change, the "special case" is no longer true.

mdulin2 (commented):
We need to make this non-consensus breaking.

wormchain/x/wormhole/keeper/msg_server_register_account_as_guardian.go
joelsmith-2019: Yes, I'll update those comments now

pires:
Is this truly affecting only mainnet. Please, be reminded that in testnet the guardian set size *is* one (1) and we should keep the ability to rotate the key. cc @andreclaro

mdulin2:
This is *expanding* the ability to rotate keys. Right now, there is a restriction that the hot swap is only allowed when there is a single Guardian (aka testnet). This allows for all Guardians to do the hot swap at any point.

wormchain/x/wormhole/keeper/msg_server_register_account_as_guardian.go
nik-suri: Option 1 sounds good to me. Can you just add a comment for why we're leaving this seemingly useless lookup in there so someone doesn't accidentally remove it in the future?

wormchain/x/wormhole/keeper/msg_server_register_account_as_guardian.go
```diff
@@ -49,6 +49,14 @@ func (k msgServer) RegisterAccountAsGuardian(goCtx context.Context, msg *types.M
 		return nil, types.ErrGuardianSetNotFound
 	}
 
+	// With the change to allow hot swapping on validator sets > 1 validator, this check is no
+	// longer useful. However, it is necessary. This check ensures the gas usage of the transaction
+	// matches the previous implementation as this is included in a non-consensus breaking change.
+	_, consensusIndexFound := k.GetConsensusGuardianSetIndex(ctx)
```
mdulin2, line 55: Cool! Seems like a good fix to me.

mdulin2 (approved):

nik-suri (approved):

nik-suri (approved):

mdulin2 (approved):

evan-gray (approved):
reviewed `.github` changes only.

PR #4206, opened 2024-12-30
fix: make devcontainer valid JSON

glitched-w0rld:
Currently, this file is not a valid JSON file. This PR fixes that by removing the trailing comma.


johnsaigle:
I believe the devcontainer runs as-is. Can you share error logs if you're experiencing an issue?

PR #4210, opened 2025-01-07
Node: Update Celo watcher

bruce-riley:
When Celo support was originally introduced to the watcher, they required a separate ABI, which was implemented as a separate connector. They are now migrating to an op-stack, which means they now support finalized and safe blocks. This also means we don't need the special connector. These changes are currently available in testnet. In mainnet, they currently support finalized but not safe. Once mainnet cuts over, Celo can be treated just like any other op-stack chain.


node/Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY test
 test: 
 # Use this command on amd64 systems
-	go test -v -ldflags '-extldflags "-Wl,--allow-multiple-definition" ' ./...
+	go test -v -ldflags ./...
```
evan-gray, line 4: this should have `-ldflags` removed.

evan-gray: also the first line should be `.PHONY: node`, the colon is missing :)

evan-gray (changes requested):
so excited to be rid of `--allow-multiple-definition`! just a couple nits

evan-gray (approved):

evan-gray (approved):

panoel (approved):

PR #4215, opened 2025-01-11
feat(wormchain): optimize guardian set key lookup using binary search

SashaMalysehko:
- Replace linear search with binary search in ContainsKey method
- Add validation to ensure guardian keys are stored in sorted order
- Introduce SortKeys utility method for maintaining sorted order
- Improve code documentation and error messages

This change improves the time complexity of key lookups from O(n) to O(log n),
which will provide better performance as the number of guardians grows.
Resolves TODO comment regarding binary search optimization.

### DIFF
--- a/wormchain/x/wormhole/types/guardian_set.go
+++ b/wormchain/x/wormhole/types/guardian_set.go
@@ -3,6 +3,7 @@ package types
 import (
 	bytes "bytes"
 	"fmt"
+	"sort"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


johnsaigle:
Closing:
- The PR doesn't prove that the input is sorted. I think the TODO suggest binary search is better _if_ the keys are sorted, but that hasn't been shown
- Very stale

Feel free to open another PR if this work is important to you, but the work should include a test and/or link to documentation to show this is actually an imporvement.

PR #4219, opened 2025-01-14
Node: Change TxHash to TxID in MessagePublication

bruce-riley:
This PR updates the `MessagePublication` object used to publish message observations internal to the guardian.

The `TxHash` field (a go-ethereum `Hash`) is being replaced by the `TxID` field (a byte array) so that we can use something other than the 32 byte go-ethereum `Hash` as our transaction identifier. For instance, on Solana we want to use the 64 byte signature in some cases.

The `tx_hash` in the protobuf is already a byte array, so we don't need to mess with that (other than testing to verify nothing, such as Wormholescan, breaks if it is more than 32 bytes long.

This change also requires updating the governor and accountant DB code to migrate from the old format to the new one.

Testing:
- Node tests pass
- CI passes
- Ran load test and verified other guardians and the fly heartbeat monitor see observations and VAAs
- Ran load test while forcing the EVM `TxID` to be more than 32 characters. Verfied fly monitor saw correct value
- Verified the accountant and NTT accountant contracts properly handle TxHash of more than 32 bytes
- Used go tests to verify migration of database entries for governor and accountant
- Verified that the accountant can handle a full batch all of 64 byte TxHash values
- Verified that a `SignedObservationBatch` full of 64 byte TxHash values does not exceed the default p2p limit
- Verified with @M-Picco that these changes should not break Wormholescan

### DIFF
--- a/node/pkg/accountant/accountant_test.go
+++ b/node/pkg/accountant/accountant_test.go
@@ -138,12 +138,12 @@ func newAccountantForTest(
 }
 
 // Converts a string into a go-ethereum Hash object used as test input.
-func hashFromString(str string) ethCommon.Hash { //nolint:unparam
+func hashToTxID(str string) []byte {
 	if (len(str) > 2) && (str[0] == '0') && (str[1] == 'x') {
 		str = str[2:]
 	}
 
-	return ethCommon.HexToHash(str)
+	return ethCommon.HexToHash(str).Bytes()
 }
 
 // Note this method assumes 18 decimals for the amount.
@@ -188,7 +188,7 @@ func TestVaaFromUninterestingEmitter(t *testing.T) {
 	var payload = []byte{1, 97, 97, 97, 97, 97}
 
 	msg := common.MessagePublication{
-		TxHash:           hashFromString("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
+		TxID:             hashToTxID("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4064"),
 		Timestamp:        time.Unix(int64(1654543099), 0),
 		Nonce:            uint32(1),
 		Sequence:         uint64(1),
@@ -216,7 +216,7 @@ func TestVaaForUninterestingPayloadType(t *testing.T) {
 	var payload = []byte{2, 97, 97, 97, 97, 97}
 
 	msg := common.MessagePublication{
-		TxHash:           hashFromString("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
+		TxID:             hashToTxID("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
 		Timestamp:        time.Unix(int64(1654543099), 0),
 		Nonce:            uint32(1),
 		Sequence:         uint64(1),
@@ -251,7 +251,7 @@ func TestInterestingTransferShouldNotBeBlockedWhenNotEnforcingAccountant(t *test
 	)
 
 	msg := common.MessagePublication{
-		TxHash:           hashFromString("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
+		TxID:             hashToTxID("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
 		Timestamp:        time.Unix(int64(1654543099), 0),
 		Nonce:            uint32(1),
 		Sequence:         uint64(1),
@@ -295,7 +295,7 @@ func TestInterestingTransferShouldBeBlockedWhenEnforcingAccountant(t *testing.T)
 	)
 
 	msg := common.MessagePublication{
-		TxHash:           hashFromString("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
+		TxID:             hashToTxID("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
 		Timestamp:        time.Unix(int64(1654543099), 0),
 		Nonce:            uint32(1),
 		Sequence:         uint64(1),
@@ -347,7 +347,7 @@ func TestForDeadlock(t *testing.T) {
 	)
 
 	msg := common.MessagePublication{
-		TxHash:           hashFromString("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
+		TxID:             hashToTxID("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063"),
 		Timestamp:        time.Unix(int64(1654543099), 0),
 		Nonce:            uint32(1),
 		Sequence:         uint64(1683136244),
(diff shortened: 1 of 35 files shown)


node/pkg/processor/message.go
```diff
@@ -80,8 +78,7 @@ func (p *Processor) handleMessage(ctx context.Context, k *common.MessagePublicat
 	if p.logger.Core().Enabled(zapcore.DebugLevel) {
 		p.logger.Debug("observed and signed confirmed message publication",
 			zap.String("message_id", k.MessageIDString()),
-			zap.Stringer("txhash", k.TxHash),
-			zap.String("txhash_b58", base58.Encode(k.TxHash.Bytes())),
```
evan-gray, line 84: might be worth keeping this. IIRC it was useful for debugging Solana (since their account keys and signatures are b58 encoded).

node/pkg/processor/observation.go
```diff
@@ -113,7 +112,6 @@ func (p *Processor) handleSingleObservation(addr []byte, m *gossipv1.Observation
 			zap.String("signature", hex.EncodeToString(m.Signature)),
 			zap.String("addr", hex.EncodeToString(addr)),
 			zap.String("txhash", hex.EncodeToString(m.TxHash)),
-			zap.String("txhash_b58", base58.Encode(m.TxHash)),
```
evan-gray, line 116: ditto

evan-gray (approved):
agree with proposing the minimally invasive change in the watchers, though it did make me cringe a little to convert from bytes to hash back to bytes :grimacing:

node/pkg/db/governor.go
```diff
@@ -290,10 +290,10 @@ const transferLen = len(transfer)
 // The first time we run this new release, any existing entries with the "GOV:PENDING2" tag will get converted
```
panoel, line 290: Do you need to change this comment block for PENDING3 and PENDING4?

panoel (approved):

panoel (dismissed):

evan-gray (dismissed):

node/pkg/common/chainlock.go
```diff
@@ -48,7 +51,12 @@ const minMsgLength = 88 // Marshalled length with empty payload
 func (msg *MessagePublication) Marshal() ([]byte, error) {
 	buf := new(bytes.Buffer)
 
-	buf.Write(msg.TxHash[:])
+	if len(msg.TxID) > math.MaxUint32 {
```
djb15, line 54: Can we make this much more aggressive? The maximum permissible value here would mean a 4GB byte array. We could either cap at `MaxUint8` or at `64` explicitly?

node/pkg/accountant/accountant_test.go
```diff
@@ -138,12 +138,12 @@ func newAccountantForTest(
 }
 
 // Converts a string into a go-ethereum Hash object used as test input.
```
johnsaigle, line 140: This comment should be updated, but not worth blocking the PR

node/pkg/common/chainlock.go
```diff
@@ -35,6 +34,10 @@ type MessagePublication struct {
 	Unreliable bool
 }
 
+func (msg *MessagePublication) TxIDString() string {
+	return "0x" + hex.EncodeToString(msg.TxID)
```
johnsaigle, line 38: Does this make sense given that the goal of this TxID change is to grow to encompass Solana-based IDs as well?

Maybe this method should be renamed to `EVMString()`?

node/pkg/common/chainlock.go
```diff
@@ -5,21 +5,20 @@ import (
...
-
-	"github.com/ethereum/go-ethereum/common"
 )
 
 const HashLength = 32
 const AddressLength = 32
 
 type MessagePublication struct {
-	TxHash    common.Hash // TODO: rename to identifier? on Solana, this isn't actually the tx hash
+	TxID      []byte
```
johnsaigle, line 21: I'm wondering whether `TxID` should be a type of its own with some common interfaces implemented, e.g. `.String()`. It might better encapsulate some of the validation going on in this PR:

* bound checks (length of 64 or 32)
* string-to-hash conversions via something like `func (TxID *id) FromHashString(string s) []byte`, avoiding the need to use `eth_common.HexToHash("0x...").Bytes()`


I understand that might be a bit of an annoying change to make at this point but curious about your thoughts about the merits of this approach longer-term.

node/pkg/common/chainlock.go
johnsaigle: Related, it might be nice to have a `SolanaString()` function that does base58 encoding, if appropriate.

node/pkg/common/chainlock.go
```diff
@@ -35,6 +34,10 @@ type MessagePublication struct {
 	Unreliable bool
 }
 
+func (msg *MessagePublication) TxIDString() string {
+	return "0x" + hex.EncodeToString(msg.TxID)
+}
+
 func (msg *MessagePublication) MessageID() []byte {
 	return []byte(msg.MessageIDString())
```
johnsaigle, line 42: Could this return a copy of `msg.TxID` instead? I'm unclear on why this would go through an encoding process and back.

node/pkg/common/chainlock.go
```diff
@@ -61,13 +69,11 @@ func (msg *MessagePublication) Marshal() ([]byte, error) {
...
-
-// UnmarshalOldMessagePublicationBeforeIsReobservation deserializes a MessagePublication from prior to the addition of IsReobservation.
+// UnmarshalOldMessagePublicationWithTxHash deserializes a MessagePublication from when the TxHash was a fixed size ethCommon.Hash.
 // This function can be deleted once all guardians have been upgraded. That's why the code is just duplicated.
-func UnmarshalOldMessagePublicationBeforeIsReobservation(data []byte) (*MessagePublication, error) {
-	if len(data) < oldMinMsgLength {
-		return nil, errors.New("message is too short")
+func UnmarshalOldMessagePublicationWithTxHash(data []byte) (*MessagePublication, error) {
+	if len(data) < minMsgLength {
+		return nil, fmt.Errorf("message is too short")
```
johnsaigle, line 76: IMO `errors.New()` is preferable here given that there's no formatting in this message. Either is fine though.

node/pkg/common/chainlock.go
```diff
@@ -128,11 +138,15 @@ func UnmarshalMessagePublication(data []byte) (*MessagePublication, error) {
 
 	reader := bytes.NewReader(data[:])
 
-	txHash := common.Hash{}
-	if n, err := reader.Read(txHash[:]); err != nil || n != HashLength {
-		return nil, fmt.Errorf("failed to read TxHash [%d]: %w", n, err)
+	txIdLen := uint32(0)
```
johnsaigle, line 141: Should this be `uint8`, since the txID should only ever be 64? Maybe there's some compatibility concerns I'm not aware of.

johnsaigle (commented):
Suggested some changes, mostly along the lines of encapsulating the new functionality in a custom type.

node/pkg/common/chainlock.go
bruce-riley: This returns the message ID (chain/emitter/sequence) not the transaction ID. (A) I don't want to change that behavior and (B) there could be multiple wormhole messages in a single transaction so TxID may not uniquely identify a message.

node/pkg/common/chainlock.go
bruce-riley: I definitely don't want to have `EVMString` and `SolanaString` because we have many other chains. Right now, those all mangle their identifiers into `ethCommon.Hash` for historical reasons. I don't want to risk breaking other chains with this PR, so I'm leaving them as is. However, going forward, they could be changed to use something more relevant to each protocol.

Since the field is a byte array, logging it as hex seems like the safest approach. I add the "0x" on the front to keep it consistent with what it previously was.

Does all this make sense?

node/pkg/common/chainlock.go
bruce-riley: I'm not sure I see the value of that and I don't think I want to tackle that right now anyway.

johnsaigle: OK no problem

node/pkg/common/chainlock.go
johnsaigle: Sure, makes sense

johnsaigle (approved):

djb15 (approved):

PR #4224, opened 2025-01-17
ci: Bump golang linter version

djb15:
What started as a simple version bump of `golangci-lint` from `v1.60.1` to `v1.63.4` turned out to be much more involved. Interestingly, the version bump caused a huge number of new `gosec G115: integer overflow conversion` lints to appear.

Having these lints is valuable as it forces us to think about any explicit integer conversions. This PR therefore includes a mixture of adding explicit checks, and ignoring lints in areas where it's clearly safe (e.g. timestamp conversion to uint64).

Not the easiest review, so I would focus on checking that the ignores are safe, and that any places that I've added new checks error in the correct manner (some return errors, others only log, others panic).

### DIFF
--- a/node/cmd/ccq/http.go
+++ b/node/cmd/ccq/http.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"sort"
 	"strings"
@@ -201,9 +202,16 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 			return res.Signatures[i].Index < res.Signatures[j].Index
 		})
 		signatures := make([]string, 0, len(res.Signatures))
-		for _, s := range res.Signatures {
+		for _, sig := range res.Signatures {
+			if sig.Index > math.MaxUint8 {
+				s.logger.Error("Signature index out of bounds", zap.Int("sig.Index", sig.Index))
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				invalidQueryRequestReceived.WithLabelValues("failed_to_marshal_response").Inc()
+				failedQueriesByUser.WithLabelValues(permEntry.userName).Inc()
+				break
+			}
 			// ECDSA signature + a byte for the index of the guardian in the guardian set
-			signature := fmt.Sprintf("%s%02x", s.Signature, uint8(s.Index))
+			signature := fmt.Sprintf("%s%02x", sig.Signature, uint8(sig.Index)) // #nosec G115 -- This is validated above
 			signatures = append(signatures, signature)
 		}
 		w.Header().Add("Content-Type", "application/json")
--- a/node/cmd/ccq/http.go
+++ b/node/cmd/ccq/http.go
@@ -174,6 +174,7 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Wait for the response or timeout
+outer:
 	select {
 	case <-time.After(query.RequestTimeout + 5*time.Second):
 		maxMatchingResponses, outstandingResponses, quorum := pendingResponse.getStats()
@@ -208,7 +209,7 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				invalidQueryRequestReceived.WithLabelValues("failed_to_marshal_response").Inc()
 				failedQueriesByUser.WithLabelValues(permEntry.userName).Inc()
-				break
+				break outer
 			}
 			// ECDSA signature + a byte for the index of the guardian in the guardian set
 			signature := fmt.Sprintf("%s%02x", sig.Signature, uint8(sig.Index)) // #nosec G115 -- This is validated above
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"os"
 	"strconv"
 	"strings"
@@ -311,6 +312,10 @@ func runFindMissingMessages(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatalf("invalid chain ID: %v", err)
 	}
+	// Although chains ids are constrained to be 16 bit, we only constrain to 32 bit here given the protobuf usage later
+	if chainID > math.MaxUint32 {
+		log.Fatalf("chain ID is not a valid 32 bit unsigned integer: %v", err)
+	}
 	emitterAddress := args[1]
 
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
@@ -352,6 +357,10 @@ func runDumpVAAByMessageID(cmd *cobra.Command, args []string) {
 	chainID, err := strconv.ParseUint(parts[0], 10, 32)
 	if err != nil {
 		log.Fatalf("invalid chain ID: %v", err)
+	}
+	// We only constrain this to int32 now, not uint16, given the ChainID protobuf definition
+	if chainID > math.MaxInt32 {
+		log.Fatalf("chain id must not exceed the max int32: %v", chainID)
 	}
 	emitterAddress := parts[1]
 	seq, err := strconv.ParseUint(parts[2], 10, 64)
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -312,10 +312,7 @@ func runFindMissingMessages(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatalf("invalid chain ID: %v", err)
 	}
-	// Although chains ids are constrained to be 16 bit, we only constrain to 32 bit here given the protobuf usage later
-	if chainID > math.MaxUint32 {
-		log.Fatalf("chain ID is not a valid 32 bit unsigned integer: %v", err)
-	}
+
 	emitterAddress := args[1]
 
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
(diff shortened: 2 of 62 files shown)


node/cmd/guardiand/adminclient.go
```diff
@@ -312,6 +314,11 @@ func runFindMissingMessages(cmd *cobra.Command, args []string) {
 	}
 	defer conn.Close()
 
+	// Although chains ids are constrained to be 16 bit, we only constrain to 32 bit here given the protobuf usage later
+	if chainID > math.MaxUint32 {
```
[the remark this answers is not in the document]
djb15: Done

node/cmd/guardiand/adminclient.go
```diff
@@ -343,6 +350,10 @@ func runDumpVAAByMessageID(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatalf("invalid chain ID: %v", err)
 	}
+	// We only constrain this to int32 now, not uint16, given the ChainID protobuf definition
+	if chainID > math.MaxInt32 {
```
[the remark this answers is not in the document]
djb15: Done

node/pkg/watchers/algorand/watcher.go
```diff
@@ -298,9 +299,15 @@ func (e *Watcher) Run(ctx context.Context) error {
 				}
 			}
 
+			if status.LastRound > math.MaxInt64 {
```
mdulin2, line 302: Your error has uint64 vs. the check being on MaxInt64. Which one do you mean here?

mdulin2 (approved):
It's amazing that all of the tests still work! Thanks for going through and making all of these little annoying changes.

node/pkg/watchers/algorand/watcher.go
djb15: Good spot, updated the error string

node/pkg/adminrpc/adminserver.go
```diff
@@ -1267,7 +1276,12 @@ func (s *nodePrivilegedService) GetAndObserveMissingVAAs(ctx context.Context, re
 		splits := strings.Split(missingVAA.VaaKey, "/")
 		chainID, err := strconv.Atoi(splits[0])
 		if err != nil {
-			errMsgs += fmt.Sprintf("\nerror converting chainID [%s] to int", missingVAA.VaaKey)
+			errMsgs += fmt.Sprintf("\nerror converting chainID [%d] to int", chainID)
```
bruce-riley, line 1279: If `strconv.Atoi` fails, won't the value of `chainID` be nondeterministic? Why did you change this?

node/pkg/watchers/algorand/watcher.go
```diff
@@ -237,7 +238,7 @@ func (e *Watcher) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case r := <-e.obsvReqC:
-			if vaa.ChainID(r.ChainId) != vaa.ChainIDAlgorand {
+			if r.ChainId > math.MaxUint16 || vaa.ChainID(r.ChainId) != vaa.ChainIDAlgorand {
```
bruce-riley, line 241: When I saw this, my first thought was that someone could exploit this to cause a panic. However, an event is only posted on this channel in `reobserve.go` if `r.ChainId` is for this chain so that can't happen. Just wonder if that's worth a comment here (and in the other watchers) to keep others from wondering the same thing?

scripts/lint.sh
```diff
@@ -89,7 +89,7 @@ while getopts 'cwdlgh' opt; do
         SELF_ARGS_WITHOUT_DOCKER+="-l "
         ;;
     g)
-        GOLANGCI_LINT_ARGS+="--out-format=github-actions "
+        GOLANGCI_LINT_ARGS+=""
```
bruce-riley, line 92: Why are we doing this?

bruce-riley (commented):
Thanks for doing this! Just a couple of minor comments.

node/pkg/watchers/algorand/watcher.go
johnsaigle: Just FYI this type of thing motivated this PR: https://github.com/wormhole-foundation/wormhole/pull/4271

scripts/lint.sh
johnsaigle: New versions of golanci-lint raise a warning that this flag is deprecated. That said it does look weird to add an empty string here, so probably the correct change is to delete the whole line.

djb15: Removed that line as suggested

node/pkg/watchers/algorand/watcher.go
djb15: Added comments to the watchers

node/pkg/adminrpc/adminserver.go
djb15: Good catch, you're right. I think I changed that to conform to the new check I added below. Have reverted that back to what it was before

node/hack/repair_eth/repair_eth.go
```diff
@@ -397,10 +398,16 @@ func main() {
...
 		var from, to string
 		if lastHeight == 0 {
+			if currentHeight-step > math.MaxInt {
+				log.Fatalf("from block overflowed: %v", currentHeight-step)
+			}
 			from = strconv.Itoa(int(currentHeight - step))
 			to = "latest"
 			lastHeight = currentHeight
 		} else {
+			if lastHeight-step > math.MaxInt {
```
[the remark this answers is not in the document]
djb15: Thanks for the nudge on this! No I think it's fine to leave this as is. The default `step` is 10,000 and the block height is unlikely to be lower than that, that would only be true for very old VAAs. If you did try to fetch from a negative `from` then the `getLogs` call in a few lines would fail with `log.Fatalf`. I don't think it gives us more safety and there isn't a lint complaining about it so I'm inclined to leave this

johnsaigle (approved):

node/hack/repair_eth/repair_eth.go
```diff
@@ -403,12 +404,18 @@ func main() {
 
 		var from, to string
 		if lastHeight == 0 {
-			from = strconv.Itoa(int(currentHeight - step))
+			if currentHeight-step > math.MaxInt {
```
pleasew8t, line 407: I'm not sure this check is sufficient. Should we not check for an underflow here as well (`currentHeight - step > currentHeight`)? The current check only covers the case where `currentHeight` is essentially `> math.MaxInt64` and `step` is really small.

pleasew8t (approved):

johnsaigle (dismissed):

bruce-riley (approved):

bruce-riley (approved):

johnsaigle (dismissed):

pleasew8t (approved):

node/cmd/ccq/http.go
```diff
@@ -201,9 +202,16 @@ func (s *httpServer) handleQuery(w http.ResponseWriter, r *http.Request) {
...
 		})
 		signatures := make([]string, 0, len(res.Signatures))
-		for _, s := range res.Signatures {
+		for _, sig := range res.Signatures {
+			if sig.Index > math.MaxUint8 {
+				s.logger.Error("Signature index out of bounds", zap.Int("sig.Index", sig.Index))
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				invalidQueryRequestReceived.WithLabelValues("failed_to_marshal_response").Inc()
+				failedQueriesByUser.WithLabelValues(permEntry.userName).Inc()
+				break
```
evan-gray, line 211: this needs to break out of the `case` statement but it's in a for loop. won't it only break out of the loop and not the case?

e.g.
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	select {
	case <-time.After(time.Second):
		for i := 0; i < 3; i++ {
			if i == 2 {
				break
			}
			fmt.Printf("i: %d\n", i)
		}
		fmt.Printf("still here")
	}
}
```
will print "still here"

node/pkg/processor/observation.go
```diff
@@ -49,6 +50,11 @@ var (
 func signaturesToVaaFormat(signatures map[common.Address][]byte, gsKeys []common.Address) []*vaa.Signature {
 	// Aggregate all valid signatures into a list of vaa.Signature and construct signed VAA.
 	var sigs []*vaa.Signature
+
+	if len(gsKeys) > math.MaxUint8 {
+		panic(fmt.Sprintf("guardian set too large: %d", len(gsKeys)))
```
evan-gray, line 55: an explicit `panic` in the guardian innards doesn't make me feel good, but I suppose `gsKeys` being > 255 is a reasonable panic

node/pkg/watchers/evm/watcher.go
```diff
@@ -288,6 +288,12 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			case <-ctx.Done():
 				return nil
 			case r := <-w.obsvReqC:
+				if r.ChainId > math.MaxUint16 {
+					logger.Error("chain id for observation request is not a valid uint16",
```
evan-gray, line 292: does this need a `continue` or something afterwards? I think this just logs an error and then proceeds to use the value anyway :thinking:

node/pkg/watchers/solana/client.go
```diff
@@ -404,6 +406,12 @@ func (s *SolanaWatcher) Run(ctx context.Context) error {
 					return err
 				}
 			case m := <-s.obsvReqC:
+				if m.ChainId > math.MaxUint16 {
+					logger.Error("chain id for observation request is not a valid uint16",
```
evan-gray, line 410: does this need a continue or something afterwards? I think this just logs an error and then proceeds to use the value anyway 🤔

evan-gray (changes requested):
just want to confirm a few things :grimacing:

node/pkg/processor/observation.go
johnsaigle: Yeah agreed, maybe we can issue a follow-up PR to resolve this? This function feels like it could be incorporated into the vaa SDK with an error return value anyways.

evan-gray: I mean, it shouldn't happen and would be a big deal if it did, so I'm fine with this change as is.

node/cmd/ccq/http.go
djb15: Thanks, fixed

node/pkg/watchers/solana/client.go
djb15: Updated, agreed `continue` makes sense here

evan-gray (approved):

johnsaigle (approved):

node/cmd/ccq/http.go
djb15: Yeah it looks kinda weird but apparently that's the Go way of doing it. Appreciate the review

PR #4231, opened 2025-01-21
Node/Solana: Beef up commitment check

bruce-riley:


node/pkg/watchers/solana/client.go
```diff
@@ -923,19 +930,16 @@ func (s *SolanaWatcher) processMessageAccount(logger *zap.Logger, data []byte, a
...
 		return
 	}
-	if commitment != s.commitment {
-		if isReobservation && s.commitment == rpc.CommitmentFinalized {
-			// There is only a single reobservation request channel for each chain, which is assigned to the finalized watcher.
-			// If someone requests reobservation of a confirmed message, we should allow the observation to go through.
-			logger.Info("allowing reobservation although the commitment level does not match the watcher",
-				zap.Stringer("account", acc), zap.String("message commitment", string(commitment)), zap.String("watcher commitment", string(s.commitment)),
+
+	if !s.checkCommitment(commitment, isReobservation) {
```
panoel, line 934: You lose the log.info message above about mismatched commitment levels on a reobs.  Do you care?

node/pkg/watchers/solana/client_test.go
```diff
@@ -6,10 +6,53 @@ import (
...
+	type test struct {
+		commitment      string
+		watcher         string
+		isReobservation bool
+		result          bool
+	}
+	tests := []test{
+		// New observation success cases
+		{commitment: "finalized", watcher: "finalized", isReobservation: false, result: true},
+		{commitment: "confirmed", watcher: "confirmed", isReobservation: false, result: true},
```
panoel, line 28: Do you want to add a test case where commitments are the same and isReobs is true?

panoel (approved):

node/pkg/watchers/solana/client.go
bruce-riley: No, I think that was an unnecessary message anyway. This is a normal scenario (when we do a reobservation on a confirmed message) so doesn't warrant a log info.

evan-gray (approved):
this feels cleaner and the tests make me feel better. thanks! 😅

PR #4233, opened 2025-01-22
node: Integrate Transfer Verifier into the Ethereum watcher

johnsaigle:
# New Features
* Create a TxVerifier instance in the Run function of the EVM watcher
* Wrap all instances where a message would be published to the broadcasting msg channel with a new function,  `verifyAndPublish`
* Messages are published with a new status: `VerificationState`. 
* Modified the Transfer Verifier pkg API so that we can call the method with either an existing Receipt or a txHash.
* Added unit test in the watcher for new functionality

## If Transfer Verifier is enabled

- Messages that are token transfers will undergo Transfer Verification
- Message will be published with a status of Valid or Rejected depending on the result
- The calling code can then decide what to do based on this status

## If Transfer Verifier is not enabled
- Existing behaviour will be preserved, but messages will be published with a status of NotVerified. No further actions are taken when a Message Publication has this status

# Design Considerations

## Modifying MessagePublication

This PR modifies `MessagePublication` to add a new status based on whether the Message Publication is verified. This decision was made to handle Transfer Verification cases across many chains. For example, the EVM logs are reliable enough that we can confidently rule a message as Valid or Rejected. Other ecosystems (i.e. Sui, but perhaps also Solana, etc.) are not so clear cut. In this case, we may want to mark a transaction as Anomalous rather than outright rejecting it. 

Using this new enum allow us to do this.

Other potential benefits:
* Decouples the Verification of a Message from whether or not this can be published.
* Avoids scope-creep for the Watchers: they only watch messages, but do not need to reject them. (Instead this could be handled by the processor or some other security mechanism akin to the Governor or Accountant.)
* Allows configuring targeted action on a per-chain and per-status basis. For example, we may want to delay Anomalous messages but drop Rejected ones.
* Preserves a NotApplicable state that can be used as a fallback mechanism if the Transfer Verifier is disabled outright or on a particular chain
* This status could be used in other cases beyond Transfer Verification, but should not interfere with existing message handling.

## Scope of this PR

The idea with this PR is to test the modifications to the Ethereum watcher _without enabling the transfer verifier yet_. We can release this in testnet or mainnet to be sure that the changes here are stable. 

Once we're confident that the underlying mechanism is working well, we can add some logic in the Guardian to actually react when a MessagePublication has a status like Rejected or Anomalous. For now, changing the reaction of the watcher to a "bad" message is out of scope. 

# Questions

- Is it necessary to update the Protobuf files somewhere in order to capture the changes to Message Publication?

# Related Work
- [x] Note: #4277 should be merged before this PR.
- [x] #4280 extends the integration test coverage of the underlying package

### DIFF
--- a/devnet/node.yaml
+++ b/devnet/node.yaml
@@ -146,6 +146,7 @@ spec:
             - --ccqEnabled=true
             - --ccqAllowedRequesters=beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe,25021A4FCAf61F2EADC8202D3833Df48B2Fa0D54
             - --ccqAllowedPeers=12D3KooWSnju8zhywCYVi2JwTqky1sySPnmtYLsHHzc4WerMnDQH,12D3KooWM6WqedfR6ehtTd1y6rJu3ZUrEkTjcJJnJZYesjd89zj8
+            - --transferVerifierEnabledChains=2
             # - --logLevel=debug
           securityContext:
             capabilities:
--- a/devnet/node.yaml
+++ b/devnet/node.yaml
@@ -146,7 +146,7 @@ spec:
             - --ccqEnabled=true
             - --ccqAllowedRequesters=beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe,25021A4FCAf61F2EADC8202D3833Df48B2Fa0D54
             - --ccqAllowedPeers=12D3KooWSnju8zhywCYVi2JwTqky1sySPnmtYLsHHzc4WerMnDQH,12D3KooWM6WqedfR6ehtTd1y6rJu3ZUrEkTjcJJnJZYesjd89zj8
-            - --transferVerifierEnabledChains=2
+            # - --transferVerifierEnabledChains=1
             # - --logLevel=debug
           securityContext:
             capabilities:
--- a/devnet/node.yaml
+++ b/devnet/node.yaml
@@ -146,7 +146,7 @@ spec:
             - --ccqEnabled=true
             - --ccqAllowedRequesters=beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe,25021A4FCAf61F2EADC8202D3833Df48B2Fa0D54
             - --ccqAllowedPeers=12D3KooWSnju8zhywCYVi2JwTqky1sySPnmtYLsHHzc4WerMnDQH,12D3KooWM6WqedfR6ehtTd1y6rJu3ZUrEkTjcJJnJZYesjd89zj8
-            # - --transferVerifierEnabledChains=1
+            # - --transferVerifierEnabledChainIDs=2
             # - --logLevel=debug
           securityContext:
             capabilities:
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -286,7 +286,10 @@ var (
 
 	subscribeToVAAs *bool
 
+	// CLI parameter value
 	transferVerifierEnabledChains *string
+	// CLI parameter value
+	txVerifierChains []vaa.ChainID
 )
 
 func init() {
@@ -938,10 +941,12 @@ func runNode(cmd *cobra.Command, args []string) {
 	}
 
 	if cmd.Flags().Changed("transferVerifierEnabledChains") {
-		txVerifierChains, err := parseTxVerifierChains(*transferVerifierEnabledChains)
-		logger.Info("parsed txVerifierChains", zap.Any("chains", txVerifierChains))
-		if err != nil {
-			logger.Fatal("Could not parse transferVerifierEnabledChains", zap.Error(err))
+		var parseErr error
+		// NOTE: avoid shadowing txVerifierChains here. It should refer to the global variable.
+		txVerifierChains, parseErr = parseTxVerifierChains(*transferVerifierEnabledChains)
+		logger.Debug("parsed txVerifierChains", zap.Any("chains", txVerifierChains))
+		if parseErr != nil {
+			logger.Fatal("Could not parse transferVerifierEnabledChains", zap.Error(parseErr))
 		}
 	}
 
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -937,10 +937,12 @@ func runNode(cmd *cobra.Command, args []string) {
 		logger.Fatal("If coinGeckoApiKey is set, then chainGovernorEnabled must be set")
 	}
 
-	txVerifierChains, err := parseTxVerifierChains(*transferVerifierEnabledChains)
-	logger.Info("parsed txVerifierChains", zap.Any("chains", txVerifierChains))
-	if err != nil {
-		logger.Fatal("Could not parse transferVerifierEnabledChains", zap.Error(err))
+	if cmd.Flags().Changed("transferVerifierEnabledChains") {
+		txVerifierChains, err := parseTxVerifierChains(*transferVerifierEnabledChains)
+		logger.Info("parsed txVerifierChains", zap.Any("chains", txVerifierChains))
+		if err != nil {
+			logger.Fatal("Could not parse transferVerifierEnabledChains", zap.Error(err))
+		}
 	}
 
 	var publicRpcLogDetail common.GrpcLogDetail
@@ -1938,7 +1940,7 @@ func argsConsistent(args []string) bool {
 	return true
 }
 
-// Parse a list of chainIds from a comma separated string and validate that they have a Transfer Verifier implementation
+// Parse a list of chainIds from a comma separated string and validate that they have a Transfer Verifier implementation.
 func parseTxVerifierChains(
 	// A comma-separated list of Wormhole ChainIDs.
 	input string,
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1925,21 +1925,18 @@ func parseTxVerifierChains(
 	if len(input) == 0 {
 		return nil, errors.New("could not parse input to parseTxVerifierChains: input empty")
 	}
-	// These are the only supported chains. The argument to this function must only include elements
-	// that are members of this slice.
-	// TODO should this be a utility function in the Transfer Verifier package?
 	supported := txverifier.SupportedChains()
 	parsed := strings.Split(input, ",")
 	enabled := make([]vaa.ChainID, 0)
-	for _, c := range parsed {
-		cid, err := vaa.ChainIDFromString(c)
+	for _, chain := range parsed {
+		chainId, err := vaa.ChainIDFromString(chain)
 		if err != nil {
-			return nil, fmt.Errorf("%s is not a valid chainId. input: %s", c, input)
+			return nil, fmt.Errorf("%s is not a valid chainId: %w. input: %s", chain, err, input)
 		}
-		if !slices.Contains(supported, cid) {
-			return nil, fmt.Errorf("chainId %d is not supported by Transfer Verifier", cid)
+		if !slices.Contains(supported, chainId) {
+			return nil, fmt.Errorf("chainId %d is not supported by Transfer Verifier", chainId)
 		}
-		enabled = append(enabled, cid)
+		enabled = append(enabled, chainId)
 	}
 
 	return enabled, nil
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -285,9 +285,9 @@ var (
 	subscribeToVAAs *bool
 
 	// A list of chain IDs that should enable the Transfer Verifier. If empty, Transfer Verifier will not be enabled.
-	transferVerifierEnabledChains *[]uint
+	transferVerifierEnabledChainIDs *[]uint
 	// Global variable used to store enabled Chain IDs for Transfer Verification. Contents are parsed from
-	// transferVerifierEnabledChains.
+	// transferVerifierEnabledChainIDs.
 	txVerifierChains []vaa.ChainID
 )
 
@@ -519,7 +519,7 @@ func init() {
 
 	subscribeToVAAs = NodeCmd.Flags().Bool("subscribeToVAAs", false, "Guardiand should subscribe to incoming signed VAAs, set to true if running a public RPC node")
 
-	transferVerifierEnabledChains = NodeCmd.Flags().UintSlice("transferVerifierEnabledChains", make([]uint, 0), "Transfer Verifier will be enabled for these chain IDs (comma-separated)")
+	transferVerifierEnabledChainIDs = NodeCmd.Flags().UintSlice("transferVerifierEnabledChainIDs", make([]uint, 0), "Transfer Verifier will be enabled for these chain IDs (comma-separated)")
 }
 
 var (
@@ -940,14 +940,14 @@ func runNode(cmd *cobra.Command, args []string) {
 	}
 
 	// NOTE: If this flag isn't set, or the list is empty, Transfer Verifier should not be enabled.
-	if cmd.Flags().Changed("transferVerifierEnabledChains") {
+	if len(*transferVerifierEnabledChainIDs) != 0 {
 		var parseErr error
 		// NOTE: avoid shadowing txVerifierChains here. It should refer to the global variable.
-		txVerifierChains, parseErr = txverifier.ValidateChains(*transferVerifierEnabledChains)
+		txVerifierChains, parseErr = txverifier.ValidateChains(*transferVerifierEnabledChainIDs)
 
 		logger.Debug("validated txVerifierChains", zap.Any("chains", txVerifierChains))
 		if parseErr != nil {
-			logger.Fatal("transferVerifierEnabledChains input is invalid", zap.Error(parseErr))
+			logger.Fatal("transferVerifierEnabledChainIDs input is invalid", zap.Error(parseErr))
 		}
 	}
 
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"runtime"
 	"slices"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -937,6 +938,7 @@ func runNode(cmd *cobra.Command, args []string) {
 	}
 
 	txVerifierChains, err := parseTxVerifierChains(*transferVerifierEnabledChains)
+	logger.Info("parsed txVerifierChains", zap.Any("chains", txVerifierChains))
 	if err != nil {
 		logger.Fatal("Could not parse transferVerifierEnabledChains", zap.Error(err))
 	}
@@ -1751,6 +1753,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *sepoliaRPC,
 				Contract:         *sepoliaContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDSepolia),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1763,6 +1766,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *holeskyRPC,
 				Contract:         *holeskyContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDHolesky),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1775,6 +1779,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *arbitrumSepoliaRPC,
 				Contract:         *arbitrumSepoliaContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDArbitrumSepolia),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1787,6 +1792,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *baseSepoliaRPC,
 				Contract:         *baseSepoliaContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDBaseSepolia),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1799,6 +1805,7 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *optimismSepoliaRPC,
 				Contract:         *optimismSepoliaContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDOptimismSepolia),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1811,6 +1818,20 @@ func runNode(cmd *cobra.Command, args []string) {
 				Rpc:              *polygonSepoliaRPC,
 				Contract:         *polygonSepoliaContract,
 				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDPolygonSepolia),
+			}
+
+			watcherConfigs = append(watcherConfigs, wc)
+		}
+
+		if shouldStart(monadDevnetRPC) {
+			wc := &evm.WatcherConfig{
+				NetworkID:        "monad_devnet",
+				ChainID:          vaa.ChainIDMonadDevnet,
+				Rpc:              *monadDevnetRPC,
+				Contract:         *monadDevnetContract,
+				CcqBackfillCache: *ccqBackfillCache,
+				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDMonadDevnet),
 			}
 
 			watcherConfigs = append(watcherConfigs, wc)
@@ -1928,12 +1949,13 @@ func parseTxVerifierChains(
 	supported := txverifier.SupportedChains()
 	parsed := strings.Split(input, ",")
 	enabled := make([]vaa.ChainID, 0)
-	for _, chain := range parsed {
-		chainId, err := vaa.ChainIDFromString(chain)
-		if err != nil {
-			return nil, fmt.Errorf("%s is not a valid chainId: %w. input: %s", chain, err, input)
+	for _, chainStr := range parsed {
+		chain, parseErr := strconv.Atoi(chainStr)
+		if parseErr != nil {
+			return nil, fmt.Errorf("could not parse chainId from string %s: %w", chainStr, parseErr)
 		}
-		if !slices.Contains(supported, chainId) {
+		chainId := vaa.ChainID(chain)
+		if !slices.Contains(supported, vaa.ChainID(chainId)) {
 			return nil, fmt.Errorf("chainId %d is not supported by Transfer Verifier", chainId)
 		}
 		enabled = append(enabled, chainId)
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1831,18 +1831,6 @@ func runNode(cmd *cobra.Command, args []string) {
 			watcherConfigs = append(watcherConfigs, wc)
 		}
 
-		if shouldStart(monadDevnetRPC) {
-			wc := &evm.WatcherConfig{
-				NetworkID:        "monad_devnet",
-				ChainID:          vaa.ChainIDMonadDevnet,
-				Rpc:              *monadDevnetRPC,
-				Contract:         *monadDevnetContract,
-				CcqBackfillCache: *ccqBackfillCache,
-				TxVerifier:       slices.Contains(txVerifierChains, vaa.ChainIDMonadDevnet),
-			}
-
-			watcherConfigs = append(watcherConfigs, wc)
-		}
 	}
 
 	var ibcWatcherConfig *node.IbcWatcherConfig = nil
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -286,9 +286,10 @@ var (
 
 	subscribeToVAAs *bool
 
-	// CLI parameter value
+	// A list of chain IDs that should enable the Transfer Verifier. If empty, Transfer Verifier will not be enabled.
 	transferVerifierEnabledChains *string
-	// CLI parameter value
+	// Global variable used to store enabled Chain IDs for Transfer Verification. Contents are parsed from
+	// transferVerifierEnabledChains.
 	txVerifierChains []vaa.ChainID
 )
 
@@ -1933,7 +1934,8 @@ func argsConsistent(args []string) bool {
 	return true
 }
 
-// Parse a list of chainIds from a comma separated string and validate that they have a Transfer Verifier implementation.
+// parseTxVerifierChains parses a list of chainIds from a comma separated string and validate that they have a Transfer Verifier implementation.
+// If the input is empty, Transfer Verifier will not be enabled.
 func parseTxVerifierChains(
 	// A comma-separated list of Wormhole ChainIDs.
 	input string,
@@ -1941,19 +1943,31 @@ func parseTxVerifierChains(
 	if len(input) == 0 {
 		return nil, errors.New("could not parse input to parseTxVerifierChains: input empty")
 	}
-	supported := txverifier.SupportedChains()
 	parsed := strings.Split(input, ",")
-	enabled := make([]vaa.ChainID, 0)
+	knownChains := vaa.GetAllNetworkIDs()
+	supportedChains := txverifier.SupportedChains()
+
+	// NOTE: Using a known capacity and counter here avoids unnecessary reallocations compared to using `append()`.
+	enabled := make([]vaa.ChainID, 0, len(parsed))
+	i := uint8(0)
 	for _, chainStr := range parsed {
 		chain, parseErr := strconv.Atoi(chainStr)
 		if parseErr != nil {
 			return nil, fmt.Errorf("could not parse chainId from string %s: %w", chainStr, parseErr)
 		}
+
 		chainId := vaa.ChainID(chain)
-		if !slices.Contains(supported, vaa.ChainID(chainId)) {
-			return nil, fmt.Errorf("chainId %d is not supported by Transfer Verifier", chainId)
+
+		if !slices.Contains(knownChains, chainId) {
+			return nil, fmt.Errorf("chainId %d is not a known Chain ID", chainId)
 		}
-		enabled = append(enabled, chainId)
+
+		if !slices.Contains(supportedChains, chainId) {
+			return nil, fmt.Errorf("chainId %d does not have a Transfer Verifier implementation", chainId)
+		}
+
+		enabled[i] = chainId
+		i++
 	}
 
 	return enabled, nil
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -2,9 +2,7 @@ package guardiand
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"math"
 	"net"
 	_ "net/http/pprof" // #nosec G108 we are using a custom router (`router := mux.NewRouter()`) and thus not automatically expose pprof.
 	"os"
@@ -945,7 +943,7 @@ func runNode(cmd *cobra.Command, args []string) {
 	if cmd.Flags().Changed("transferVerifierEnabledChains") {
 		var parseErr error
 		// NOTE: avoid shadowing txVerifierChains here. It should refer to the global variable.
-		txVerifierChains, parseErr = validateTxVerifierChains(*transferVerifierEnabledChains)
+		txVerifierChains, parseErr = txverifier.ValidateChains(*transferVerifierEnabledChains)
 
 		logger.Debug("validated txVerifierChains", zap.Any("chains", txVerifierChains))
 		if parseErr != nil {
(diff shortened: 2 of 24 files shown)


node/pkg/watchers/evm/watcher.go
```diff
@@ -88,7 +94,7 @@ type (
 		// VAA ChainID of the network we're connecting to.
 		chainID vaa.ChainID
 
-		// Channel to send new messages to.
+		// Channel to send new messages to. Should be wrapped be a call to `publishIfSafe()`
```
banescusebi, line 97:
```suggestion
		// Channel to send new messages to. Each message should be wrapped by a call to `publishIfSafe()`
```

node/pkg/watchers/evm/watcher.go
```diff
@@ -331,7 +387,14 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
 						)
-						w.msgC <- msg
+						pubErr := w.publishIfSafe(msg, ctx, tx, receipt)
+						if pubErr != nil {
+							logger.Error("could not publish message: transfer verification failed",
+								zap.String("msgId", msg.MessageIDString()),
+								zap.String("txHash", msg.TxIDString()),
+								zap.Error(pubErr),
+							)
+						}
```
banescusebi, line 397: This chunk of code is used 4 more times below. Should we put it inside `publishIfSafe` and then have that function not return anything?

banescusebi (commented):
Just a couple of suggestions

node/pkg/watchers/evm/watcher.go
johnsaigle: My preference is to maintain the return value so that the calling code has an idea of whether or not an error occurred. We could think about reducing code duplication by passing a `*logger` as an argument to that function and allowing it to print the error message, but then the calling code would need another log call if someone wants to log data that isn't visible to `publishIfSafe`, e.g. the `blockNumber` which exists for some instances in this file but not others.

node/cmd/guardiand/node.go
```diff
@@ -1818,3 +1866,29 @@ func argsConsistent(args []string) bool {
...
 	return true
 }
+
+// Parse a list of chainIds from a comma separated string and validate that they have a Transfer Verifier implementation
+func parseTxVerifierChains(
+	// A comma-separated list of Wormhole ChainIDs.
+	input string,
+) ([]vaa.ChainID, error) {
+	if len(input) == 0 {
+		return nil, errors.New("could not parse input to parseTxVerifierChains: input empty")
```
djb15, line 1876: If we wanted to disable transfer verification on all chains, wouldn't the empty input be a valid case?

johnsaigle: In that case we would just not pass in the flag at all, similar to how someone might omit the flag that enables the governor or sets a coingecko key. I think having a CLI interface where `--txVerifierChains=''` is a valid input is a bit messy.

djb15: I'm probably misunderstanding how these Go command line flags work, but if you don't provide the flag what does the value default to? My current understanding is it would default to the value you specify in the declaration, which is an empty string; am I greping that wrong?

johnsaigle: The type is `*string` (pointer to a string) so the default is `nil`, I think

johnsaigle: I did some more testing and patched the code. The latest commit does the following:

- if the CLI flag is an empty string, panic (via logger.Fatal)
- if it's absent, txVerifier is disabled
- if it contains a list of chain IDs, enable Tx Verifier for those chains

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +838,38 @@ func canRetryGetBlockTime(err error) bool {
...
+	// and save on RPC requests and parsing.
+	receipt *gethTypes.Receipt,
+) error {
+	if msg == nil {
+		return errors.New("message publication cannot be nil")
+	}
+	if w.txVerifierEnabled {
+		// This should have already been initialized.
+		if w.txVerifier == nil {
+			return errors.New("transfer verifier should be enabled but is nil")
```
pleasew8t, line 859: alternative: "transfer verifier should be instantiated but is nil"

node/pkg/watchers/evm/watcher.go
```diff
@@ -137,6 +145,11 @@ type (
 		ccqBatchSize       int64
 		ccqBackfillCache   bool
 		ccqLogger          *zap.Logger
+		// Whether the Transfer Verifier should be initialized for this watcher.
+		txVerifierEnabled bool
+		// Transfer Verifier instance
+		txVerifier txverifier.TransferVerifierInterface
+		// txVerifier *txverifier.TransferVerifier[*eth_client.Client, connectors.Connector]
```
pleasew8t, line 152: can be removed

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,58 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+				return errors.New("watcher attempted to create Transfer Verifier but this chainId is not supported")
+			}
+
+			// TODO: It would be better to avoid hard-coding the Wrapped Native Addresses.
+			var tbridge eth_common.Address
+			var weth eth_common.Address
+			switch w.env {
+			case common.UnsafeDevNet:
+				tbridge = eth_common.BytesToAddress(sdk.KnownDevnetTokenbridgeEmitters[w.chainID])
+				weth = eth_common.HexToAddress("0xDDb64fE46a91D46ee29420539FC25FD07c5FEa3E")
```
djb15, line 282: Re the TODO, how about creating a set of `Known[X]NetWETHAddress` constants in the SDK alongside the existing constants?

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,58 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+				CoreBridgeAddr:    w.contract,
+				TokenBridgeAddr:   tbridge,
+				WrappedNativeAddr: weth,
+			}
+
+			var tvErr error
+			w.txVerifier, tvErr = txverifier.NewTransferVerifier(
+				w.ethConn,
+				&addrs,
+				20,
```
djb15, line 302: Comment to explain the choice of magic number?

djb15 (commented):
Integration looks good overall! Would be good to deal with the outstanding TODOs if possible

sdk/mainnet_consts.go
```diff
@@ -194,3 +195,9 @@ var KnownAutomaticRelayerEmitters = []struct {
 	{ChainId: vaa.ChainIDSnaxchain, Addr: "00000000000000000000000027428DD2d3DD32A4D7f7C497eAaa23130d894911"},
 	{ChainId: vaa.ChainIDWorldchain, Addr: "0000000000000000000000001520cc9e779c56dab5866bebfb885c86840c33d3"},
 }
+
+// KnownWrappedNativeAddress is a map of wrapped native addresses by chain ID, e.g. WETH for Ethereum
+var KnownWrappedNativeAddress = map[vaa.ChainID]common.Address{
+	// WETH
```
pleasew8t, line 201: It makes sense to try and maintain consistency here by typing the mapping as `map[vaa.ChainID]string`.

sdk/devnet_consts.go
```diff
@@ -41,3 +42,9 @@ var KnownDevnetAutomaticRelayerEmitters = []struct {
 	{ChainId: vaa.ChainIDEthereum, Addr: "000000000000000000000000cc680d088586c09c3e0e099a676fa4b6e42467b4"},
 	{ChainId: vaa.ChainIDBSC, Addr: "000000000000000000000000cc680d088586c09c3e0e099a676fa4b6e42467b4"},
 }
+
+// KnownDevnetWrappedNativeAddress is a map of wrapped native addresses by chain ID, e.g. WETH for Ethereum
+var KnownDevnetWrappedNativeAddresses = map[vaa.ChainID]common.Address{
```
pleasew8t, line 47: Same comment here about typing to `map[vaa.ChainID]string`.

sdk/testnet_consts.go
```diff
@@ -102,3 +105,11 @@ var KnownTestnetAutomaticRelayerEmitters = []struct {
 	{ChainId: vaa.ChainIDOptimismSepolia, Addr: "00000000000000000000000093BAD53DDfB6132b0aC8E37f6029163E63372cEE"},
 	{ChainId: vaa.ChainIDBaseSepolia, Addr: "00000000000000000000000093BAD53DDfB6132b0aC8E37f6029163E63372cEE"},
 }
+
+// KnownTestnetWrappedNativeAddresses is a list of addresses for deployments of wrapped native asssets (e.g. WETH) on various testnets.
+var KnownTestnetWrappedNativeAddresses = map[vaa.ChainID]common.Address{
+	// WETH
```
pleasew8t, line 111: Same comment here about typing to `map[vaa.ChainID]string`. I'd also argue that we can have one less import in this file by doing so.

node/pkg/common/chainlock.go
```diff
@@ -18,6 +18,25 @@ import (
...
 const AddressLength = 32
 
+// This type represents whether a message has been verified for some definition of verification. This may mean different things
+// on a per-application or per-chain basis.
+// NOTE: This status is currently used only by the Transfer Verifier for supported chains.
+type VerificationState uint8
+
+const (
+	// The default state for a message. This can be used before verification occurs. If no verification is required, `NotApplicable` should be used instead.
+	// NOTE: this value is used as the default, zero-value for the type, so this field should not be re-ordered among other variants.
```
pleasew8t, line 28: Can remove this line

node/pkg/common/chainlock.go
```diff
@@ -18,6 +18,25 @@ import (
...
+	NotVerified VerificationState = iota
+	// Represents a "known bad" status where a Message has been validated and the result indicates an erroneous or invalid message. The message should be discarded.
+	Rejected
+	// Represents an unusual state after validation, neither confirmed to be good or bad.
+	Anomalous
+	// Represents a "known good" status where a Message has been validated and the result is good. The message should be process normally.
```
pleasew8t, line 34:
```suggestion
	// Represents a "known good" status where a Message has been validated and the result is good. The message should be processed normally.
```

node/pkg/common/chainlock.go
```diff
@@ -33,6 +52,9 @@ type MessagePublication struct {
 	// Unreliable indicates if this message can be reobserved. If a message is considered unreliable it cannot be
 	// reobserved.
 	Unreliable bool
+	// This type represents whether a message has been verified for some definition of verification. This may mean different things
```
pleasew8t, line 55: Reuse comment above `VerificationState` type definition.

node/pkg/common/chainlock.go
```diff
@@ -250,5 +295,23 @@ func (msg *MessagePublication) ZapFields(fields ...zap.Field) []zap.Field {
 		zap.Uint8("consistency", msg.ConsistencyLevel),
 		zap.String("message_id", string(msg.MessageID())),
 		zap.Bool("unreliable", msg.Unreliable),
+		zap.String("verificationState", msg.verificationState.String()),
 	)
 }
+
+func (v VerificationState) String() string {
```
pleasew8t, line 302: Could move this definition up to the type definition.

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +838,62 @@ func canRetryGetBlockTime(err error) bool {
...
+		if !valid {
+			w.logger.Error("verifyAndPublish: transfer verification failed", zap.String("txHash", txHash.String()))
+			verificationState = common.Rejected
+		} else {
+			w.logger.Debug("verifyAndPublish: transfer verification successful", zap.String("txHash", txHash.String()))
+			verificationState = common.Valid
+		}
+	}
+
+	updateErr := msg.SetVerificationState(verificationState)
```
pleasew8t, line 885: I wonder about this pattern. My thinking is that *before* transfer verification this logic did not matter, and therefore the guardian would continue processing transfers regardless of the transfer verification operations. Should we not have a verification state that indicates this error ocurred, keep logging it, and pass the message to the msg channel anyway? Until we see this succeeding in a live environment, we should probably make transfer verification observent rather than interacting, if that makes sense.

node/pkg/watchers/evm/watcher.go
```diff
@@ -701,14 +783,25 @@ func (w *Watcher) postMessage(logger *zap.Logger, ev *ethabi.AbiLogMessagePublis
 			zap.Uint8("ConsistencyLevel", ev.ConsistencyLevel),
 		)
 
-		w.msgC <- message
+		ctx, cancel := context.WithCancel(context.Background())
```
pleasew8t, line 786: I took a queue from the Sui watcher and made it so the sui tx verifier uses `context.WithTimeout`, which kind of makes more sense to me. Unless there is something specific that cancels the context in the event of an error.

node/pkg/watchers/evm/reobserve.go
```diff
@@ -47,7 +47,14 @@ func (w *Watcher) handleReobservationRequest(ctx context.Context, chainId vaa.Ch
 				zap.Uint64("current_block", finalizedBlockNum),
 				zap.Uint64("observed_block", blockNumber),
 			)
-			w.msgC <- msg
+
+			pubErr := w.verifyAndPublish(msg, ctx, eth_common.BytesToHash(msg.TxID), receipt)
+
+			if pubErr != nil {
+				w.logger.Error("Error when publishing message", zap.Error(err))
+				continue
```
pleasew8t, line 55: If we do decide to continue sending the message publication over the channel, regardless of potential runtime errors, remember to remove the `continue` so the number of observations is still incremented.

devnet/node.yaml
```diff
@@ -146,6 +146,7 @@ spec:
             - --ccqEnabled=true
             - --ccqAllowedRequesters=beFA429d57cD18b7F8A4d91A2da9AB4AF05d0FBe,25021A4FCAf61F2EADC8202D3833Df48B2Fa0D54
             - --ccqAllowedPeers=12D3KooWSnju8zhywCYVi2JwTqky1sySPnmtYLsHHzc4WerMnDQH,12D3KooWM6WqedfR6ehtTd1y6rJu3ZUrEkTjcJJnJZYesjd89zj8
+            # - --transferVerifierEnabledChains=1
```
pleasew8t, line 149: Is this meant to be commented?

node/cmd/guardiand/node.go
```diff
@@ -1853,3 +1909,42 @@ func argsConsistent(args []string) bool {
...
+	}
+	parsed := strings.Split(input, ",")
+	knownChains := vaa.GetAllNetworkIDs()
+	supportedChains := txverifier.SupportedChains()
+
+	// NOTE: Using a known capacity and counter here avoids unnecessary reallocations compared to using `append()`.
+	enabled := make([]vaa.ChainID, len(parsed))
```
pleasew8t, line 1927: Why not just append? This is startup, which I don't think is that performance-sensitive.

node/cmd/guardiand/node.go
```diff
@@ -500,6 +510,8 @@ func init() {
 	gatewayRelayerKeyPassPhrase = NodeCmd.Flags().String("gatewayRelayerKeyPassPhrase", "", "Pass phrase used to unarmor the gateway relayer key file")
 
 	subscribeToVAAs = NodeCmd.Flags().Bool("subscribeToVAAs", false, "Guardiand should subscribe to incoming signed VAAs, set to true if running a public RPC node")
+
+	transferVerifierEnabledChains = NodeCmd.Flags().String("transferVerifierEnabledChains", "", "Transfer Verifier will be enabled for these chain IDs (comma-separated)")
```
pleasew8t, line 514: Why not use Flags().IntSlice()?

node/pkg/watchers/evm/watcher.go
johnsaigle: I think that's what happens generally. This error is more of a type-safety thing to prevent overwriting an existing state. It could log an error instead of returning I suppose.

node/cmd/guardiand/node.go
johnsaigle: Agreed but I don't think there's much benefit to append

node/pkg/watchers/evm/watcher.go
johnsaigle: I think I chose WithCancel here because other parts of the watcher do this, and they can be cancelled if the code panics. Basically I'm going off of this: https://github.com/wormhole-foundation/wormhole/blob/c91313587f715c0f7f6f8a3aaadd51966ec3551d/node/pkg/watchers/evm/watcher.go#L206-L210

node/pkg/watchers/evm/watcher.go
johnsaigle: I moved this inside of the `if` statement in this function. In the previous commit, the state of the message would be changed from NotVerified to NotApplicable, and potentially return an error. Now, the error handling and state change will not occur unless transfer verifier is enabled.

node/pkg/watchers/evm/reobserve.go
johnsaigle: I removed the continue statements in the most recent commit

node/cmd/guardiand/node.go
johnsaigle: Done using UintSlice

node/cmd/guardiand/node.go
```diff
@@ -500,6 +510,8 @@ func init() {
 	gatewayRelayerKeyPassPhrase = NodeCmd.Flags().String("gatewayRelayerKeyPassPhrase", "", "Pass phrase used to unarmor the gateway relayer key file")
 
 	subscribeToVAAs = NodeCmd.Flags().Bool("subscribeToVAAs", false, "Guardiand should subscribe to incoming signed VAAs, set to true if running a public RPC node")
+
+	transferVerifierEnabledChains = NodeCmd.Flags().UintSlice("transferVerifierEnabledChains", make([]uint, 2), "Transfer Verifier will be enabled for these chain IDs (comma-separated)")
```
djb15, line 514: What's the idea behind the default case?

sdk/mainnet_consts.go
```diff
@@ -194,3 +194,9 @@ var KnownAutomaticRelayerEmitters = []struct {
 	{ChainId: vaa.ChainIDSnaxchain, Addr: "00000000000000000000000027428DD2d3DD32A4D7f7C497eAaa23130d894911"},
 	{ChainId: vaa.ChainIDWorldchain, Addr: "0000000000000000000000001520cc9e779c56dab5866bebfb885c86840c33d3"},
 }
+
+// KnownWrappedNativeAddress is a map of wrapped native addresses by chain ID, e.g. WETH for Ethereum
+var KnownWrappedNativeAddress = map[vaa.ChainID]string{
+	// WETH
+	vaa.ChainIDEthereum: "0xc8f93d9738e7Ad5f3aF8c548DB2f6B7F8082B5e8",
```
pleasew8t, line 201: Don't know what I'm missing, but WETH on eth mainnet is at `0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2`

node/pkg/watchers/evm/watcher.go
```diff
@@ -78,18 +84,20 @@ var (
...
 		// Readiness component
 		readinessSync readiness.Component
-		// VAA ChainID of the network we're connecting to.
+		// VAA ChainID of the network monitored by this watcher.
 		chainID vaa.ChainID
 
-		// Channel to send new messages to.
+		// Channel for sending new MesssagePublications. Messages should not be sent
+		// to this channel directly. Instead, they should be wrapped by
+		// a call to `publishIfSafe()`.
```
pleasew8t, line 100: Update to `verifyAndPublish()`

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,60 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+			// from erroneously setting up a Transfer Verifier or
+			// else continuing in state where txVerifierEnabled is
+			// true but the actual Transfer Verifier is nil.
+			if !slices.Contains(txverifier.SupportedChains(), w.chainID) {
+				return errors.New("watcher attempted to create Transfer Verifier but this chainId is not supported")
+			}
+
+			// Fetch the constants for the Token Bridge and the WETH address from the SDK.
+			var tbridge []byte
+			var weth string
```
pleasew8t, line 278: I think we should rename `weth` to `wnative`?

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,60 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+			switch w.env {
+			case common.UnsafeDevNet:
+				tbridge = sdk.KnownDevnetTokenbridgeEmitters[w.chainID]
+				weth = sdk.KnownDevnetWrappedNativeAddresses[w.chainID]
+			case common.TestNet:
+				tbridge = sdk.KnownTestnetTokenbridgeEmitters[w.chainID]
+				weth = sdk.KnownTestnetWrappedNativeAddresses[w.chainID]
+			case common.MainNet:
+				tbridge = sdk.KnownTokenbridgeEmitters[w.chainID]
+				// https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```
pleasew8t, line 288: remove comment

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,60 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+				weth = sdk.KnownWrappedNativeAddress[w.chainID]
+			}
+			addrs := txverifier.TVAddresses{
+				CoreBridgeAddr:    w.contract,
+				TokenBridgeAddr:   eth_common.BytesToAddress(tbridge[:]),
```
pleasew8t, line 293: In my code I added a `KnownMainnetCoreContracts` to the constants file. I think it just makes sense. You don't have to do it here, but I'm going to keep the change on my branch.

node/pkg/watchers/evm/watcher.go
```diff
@@ -241,6 +259,60 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+				WrappedNativeAddr: eth_common.HexToAddress(weth),
+			}
+
+			// The block height difference between the latest block and the oldest block to keep in memory.
+			// Value is arbitrary and can be adjusted if it helps performance.
+			pruneHeightDelta := uint64(20)
+			var tvErr error
```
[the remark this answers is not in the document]
SEJeff: This needs to be a constant!

node/cmd/guardiand/node.go
johnsaigle: Hm my intention is that you can invoke the guardian with e.g.:

`txVerifierChains=1,2`

but not `txVerifierChains=` (blank)

If you mean the `make` here, I gave it a length and capacity of `2` more or less arbitrarily. I guess I was thinking that if we enable Ethereum and Sui, that's two chains. The value could easily be `0` or `10` or anything else.

sdk/mainnet_consts.go
johnsaigle: Yeah you're right, I think this was a bad copy-paste from the testnet or devnet constant

node/pkg/watchers/evm/watcher.go
johnsaigle: Why not use `KnownTokenbridgeEmitters`?

djb15 (approved):
Thanks for all the changes made previously, looks good to me!

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +845,65 @@ func canRetryGetBlockTime(err error) bool {
...
+		// Transfer Verification is enabled. If disabled, the message
+		// will be emitted with the default value `NotVerified`.
+		var verificationState common.VerificationState
+
+		// Verify the transfer by analyzing the transaction receipt. This is a defense-in-depth mechanism
+		// to protect against fraudulent message emissions.
+		valid := w.txVerifier.ProcessEvent(ctx, txHash, receipt)
+		if !valid {
+			w.logger.Error("verifyAndPublish: transfer verification failed", zap.String("txHash", txHash.String()))
+			verificationState = common.Rejected
```
mdulin2, line 887: Why don't we check for the `tokenBridge` on the VAA before doing the verification here? This way, it would prevent a lot of unnecessary computations.

node/pkg/watchers/evm/watcher.go
```diff
@@ -524,7 +603,17 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 							zap.Stringer("current_blockHash", currentHash),
 						)
 						delete(w.pending, key)
-						w.msgC <- pLock.message
+
+						// Note that `tx` here is actually a receipt
+						txHash := eth_common.Hash(pLock.message.TxID)
+						pubErr := w.verifyAndPublish(pLock.message, ctx, txHash, tx)
```
mdulin2, line 609: I like how low-touch this is.

devnet/node.yaml
bruce-riley: Could this be renamed `transferVerifierEnabledChainIDs`? Also, I know it's commented out but chain ID one is Solana. Do you want to make this two for Ethereum?

node/cmd/guardiand/node.go
bruce-riley: Couldn't you just make the default nil here and then at line 938 do `if len(*transferVerifierEnabledChains) != 0 {` instead of using Changed? Seems more clear to me.

node/cmd/guardiand/node.go
```diff
@@ -1857,3 +1915,38 @@ func argsConsistent(args []string) bool {
 
 	return true
 }
+
+// validateTxVerifierChains validates that a slice of uints correspond to chain IDs with a Transfer Verifier implementation.
```
bruce-riley, line 1919: Should this bee in `pkg/txverifier`?

node/go.mod
```diff
@@ -384,3 +384,5 @@ replace github.com/wormhole-foundation/wormchain => ../wormchain
 replace github.com/CosmWasm/wasmd v0.30.0 => github.com/wormhole-foundation/wasmd v0.30.0-wormchain-2
 
 replace github.com/cosmos/cosmos-sdk => github.com/wormhole-foundation/cosmos-sdk v0.45.9-wormhole
+
+replace github.com/certusone/wormhole/node/pkg/txverifier => ./pkg/txverifier
```
bruce-riley, line 388: Why is this necessary when we don't do it for any of the other local packages?

node/pkg/common/chainlock.go
```diff
@@ -18,6 +18,41 @@ import (
...
+
+const (
+	// The default state for a message. This can be used before verification occurs. If no verification is required, `NotApplicable` should be used instead.
+	NotVerified VerificationState = iota
+	// Represents a "known bad" status where a Message has been validated and the result indicates an erroneous or invalid message. The message should be discarded.
+	Rejected
+	// Represents an unusual state after validation, neither confirmed to be good or bad.
+	Anomalous
+	// Represents a "known good" status where a Message has been validated and the result is good. The message should be processed normally.
+	Valid
```
bruce-riley, line 34: Should these enums start with a prefix? For instance, `Valid` is a pretty generic name that may be used elsewhere.

node/pkg/common/chainlock.go
```diff
@@ -250,5 +313,6 @@ func (msg *MessagePublication) ZapFields(fields ...zap.Field) []zap.Field {
 		zap.Uint8("consistency", msg.ConsistencyLevel),
 		zap.String("message_id", string(msg.MessageID())),
 		zap.Bool("unreliable", msg.Unreliable),
+		zap.String("verificationState", msg.verificationState.String()),
```
bruce-riley, line 316: Isn't this going to break the governor pending transfer DB stuff? Maybe this doesn't need to get marshalled and unmarshalled? Otherwise, you need to update `pkg/db/governor.go` to migrate from the old format to the new. 

Or maybe not, since that uses json marshal? Maybe you should at least have a test over there that verifies it can load an old `MessagePublication` from the DB into the new format?

node/pkg/txverifier/utils.go
```diff
@@ -96,3 +98,10 @@ func denormalize(
 
 	return denormalizedAmount
 }
+
+// SupportedChains returns a slice of Wormhole Chain IDs that have a Transfer Verifier implementation.
+func SupportedChains() []vaa.ChainID {
```
bruce-riley, line 103: So for now, you are only supporting Ethereum, not any of the other EVMs (not even Sepolia or Holeksy)?

node/pkg/watchers/evm/config.go
```diff
@@ -20,6 +20,7 @@ type WatcherConfig struct {
 	L1FinalizerRequired    watchers.NetworkID // (optional)
 	l1Finalizer            interfaces.L1Finalizer
 	CcqBackfillCache       bool
+	TxVerifier             bool
```
bruce-riley, line 23: Would `TxVerifierEnabled` be a better name?

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +845,65 @@ func canRetryGetBlockTime(err error) bool {
...
+	receipt *gethTypes.Receipt,
+) error {
+	if msg == nil {
+		return errors.New("verifyAndPublish: message publication cannot be nil")
+	}
+	if msg.VerificationState() != common.NotVerified {
+		return errors.New("verifyAndPublish: message publication already has a verification status")
+	}
+
+	if w.txVerifierEnabled {
```
bruce-riley, line 871: Why even check this? Why not just do `if w.txVerifier != nil {`?

node/pkg/watchers/evm/watcher.go
bruce-riley: I don't think I like the use of `context.Background`. The `Run` routine has a context that should be passed in.

node/pkg/watchers/evm/watcher.go
bruce-riley: I agree. Not only computation but RPC calls.

bruce-riley: Should there be a Prometheus metric for verification failures?

node/go.mod
johnsaigle: IMO we should do it for the other packages but I didn't want to make more changes than necessary.

node/cmd/guardiand/node.go
johnsaigle: Sure, not a bad idea!

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +845,65 @@ func canRetryGetBlockTime(err error) bool {
 	return exists
 }
 
+// verifyAndPublish validates a MessagePublication to ensure that it's safe. If so, it broadcasts the message. This function
```
bruce-riley, line 855: `evm/watcher.go` has gotten too big. Please move this stuff into something like `verifier.go` and `verifier_test.go`.

node/pkg/watchers/evm/watcher.go
johnsaigle: Done in the latest commit, thanks for catching

node/pkg/common/chainlock.go
johnsaigle: I'll look into it. I thought those worked on Transfer objects rather than MessagePublications. I'll look into adding those serialization tests though.

node/pkg/watchers/evm/reobserve.go
```diff
@@ -68,7 +74,13 @@ func (w *Watcher) handleReobservationRequest(ctx context.Context, chainId vaa.Ch
 					zap.Uint64("current_safe_block", safeBlockNum),
 					zap.Uint64("observed_block", blockNumber),
 				)
-				w.msgC <- msg
+
+				pubErr := w.verifyAndPublish(msg, ctx, eth_common.BytesToHash(msg.TxID), receipt)
+
+				if pubErr != nil {
+					w.logger.Error("Error when publishing message", zap.Error(err))
```
bruce-riley, line 81: You should continue so `numObservations` is not pegged.

node/pkg/watchers/evm/watcher.go
johnsaigle: OK, can do

node/pkg/txverifier/utils.go
johnsaigle: Yes, but adding the testnets should be OK.

bruce-riley (commented):
I left a variety of comments, but my big concern is about the additional RPC call on every message observation, even if it is not a token bridge message.

johnsaigle:
@bruce-riley I just patched this based on @mdulin2's comment when you were reviewing. It should now only evaluate messages where the emitter address is equal to the token bridge, so there shouldn't be more RPC calls.

node/pkg/common/chainlock.go
johnsaigle: Right now I think this is limited in a few ways:
- These live only in `common` so they'd need to be used as `common.Rejected`
- They're using a custom type, so they should only be able to be used by other functions/structs etc. using specifically the type `common.VerificationState`
- The comments on the type itself mentions that it's used for MessagePublications and Transfer Verifier specifically.

Is this enough in your view? We could change these to have a prefix like `VerificationState` but I don't think it's very nice to have a enum variant like `VerificationStateNotVerified` or `VerificationStateValid`

node/pkg/watchers/evm/reobserve.go
johnsaigle: Should be done in the latest commit.

node/pkg/common/chainlock.go
pleasew8t: Alternatively, they could be prefixed with TV? `TV_Valid` or `TV_Anomalous` could work? I personally don't like the underscore, but `TVValid` seems clumsy.

johnsaigle: Another layer we can add here is to move this enum to the new `msg_verifier.go` and import it into common. That way there's a much lower risk of other parts of the codebase accidentally using this enum because they would need to import from there instead of from `common` (which is ... pretty _common_)

node/pkg/txverifier/utils.go
johnsaigle: Testnets added in latest commit.

node/pkg/common/chainlock.go
pleasew8t: I've thought about this some more, and personally think the enums are fine as is. I think having the verification state defined alongside the message publication struct is correct in this context. Putting it in `msg_verifier.go` doesn't make sense, since this is a shared enum across all watchers that should implement transfer verification.

johnsaigle: Thanks, yeah I forgot that we'd want to use these enums for Sui and potentially other chains too.

node/pkg/common/chainlock.go
johnsaigle: Bruce and I discussed this more privately. Serializing this field is not necessary as it is currently not used by the Governor which is the only place that makes use of this DB and marshalling logic. 
I'll add a code comment noting this.

pleasew8t (dismissed):

pleasew8t (approved):

node/pkg/txverifier/evm.go
```diff
@@ -41,39 +40,38 @@ const (
...
 ) bool {
 
 	// Use this opportunity to prune old transaction information from the cache.
 	tv.pruneCache()
 
+	if Cmp(txHash, ZERO_ADDRESS) == 0 {
+		tv.logger.Error("txHash is the zero address")
+		return false
+	}
+	if len(txHash) == 0 {
```
bruce-riley, line 55: `ethCommon.Hash` can't have a zero length, it is a fixed length byte array.

node/pkg/watchers/evm/watcher.go
```diff
@@ -745,6 +852,42 @@ func canRetryGetBlockTime(err error) bool {
...
+) error {
+
+	if msg == nil {
+		return errors.New("verifyAndPublish: message publication cannot be nil")
+	}
+
+	if w.txVerifier != nil {
+		verifiedMsg, err := verify(ctx, msg, txHash, receipt, w.txVerifier)
+
+		if err != nil {
```
bruce-riley, line 879: I really think we should be pegging a metric or something if this fails. Otherwise how will anyone know it's happening?

bruce-riley (commented):
Left a couple minor comments and one more major one that I think we should be pegging prom metrics or something when verification fails. Otherwise how will anyone know it's happening?

node/pkg/watchers/evm/watcher.go
johnsaigle: This is a good point and I'll resolve it in a follow-up PR. Now that we have this mechanism to mark messages as say `Anomalous`, the next task will be to figure out what we want to do with them. Based on that design, we can find good metrics for observing all of this.

node/pkg/txverifier/evm.go
```diff
@@ -41,39 +40,38 @@ const (
...
+	txHash common.Hash,
 	// If nil, this code will fetch the receipt using the TransferVerifier's connector.
 	receipt *geth.Receipt,
 ) bool {
 
 	// Use this opportunity to prune old transaction information from the cache.
 	tv.pruneCache()
 
+	if Cmp(txHash, ZERO_ADDRESS) == 0 {
+		tv.logger.Error("txHash is the zero address")
```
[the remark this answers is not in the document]
johnsaigle: Thanks, I clarified this error message

node/pkg/watchers/evm/watcher.go
johnsaigle: The parent context is now used in the latest commit

node/cmd/guardiand/node.go
johnsaigle: I think Jason implemented this in a recent commit

bruce-riley (dismissed):

djb15 (approved):
Minor comment, looks good otherwise. I assume this is going to be tested in testnet/devnet first with some token transfers and performance evaluated?

node/pkg/watchers/evm/watcher.go
```diff
@@ -701,14 +801,25 @@ func (w *Watcher) postMessage(logger *zap.Logger, ev *ethabi.AbiLogMessagePublis
...
+		pubErr := w.verifyAndPublish(msg, verifyCtx, ev.Raw.TxHash, nil)
+		if pubErr != nil {
+			w.logger.Error("could not publish message: transfer verification failed",
+				zap.String("msgId", msg.MessageIDString()),
+				zap.String("txHash", msg.TxIDString()),
+				zap.Error(pubErr),
+			)
+		}
+
 		ethMessagesConfirmed.WithLabelValues(w.networkName).Inc()
```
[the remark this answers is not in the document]
johnsaigle: Yeah good catch, thanks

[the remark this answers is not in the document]
johnsaigle: Fixed in latest commit as well as another instance in the same file

johnsaigle:
@djb15 Yeah I think we can spend some time testing and benchmarking before the feature actually gets enabled by the Guardians.

djb15 (approved):

evan-gray (approved):
Only reviewed `devnet` as that was the only change missing code owner approval.

PR #4237, opened 2025-01-23
Node: Audit chain IDs

bruce-riley:
This PR updates the node / guardian code to make sure all of the known chain IDs are properly covered.


evan-gray (approved):
nice cleanup

johnsaigle:
Is the idea that we should enforce this ChainID-based sorting elsewhere in the codebase going forward?

bruce-riley:
Yes, I think maintaining a consistent ordering (and by chain ID seems the most logical) is a good way to avoid missing things, like has happened over time.

SEJeff (approved):
This should be done in the js sdk. Nice cleanup.

PR #4239, opened 2025-01-23
sdk: add StringToKnownChainID function for VAAs

johnsaigle:
Adds a function to the SDK's vaa structs that parses known ChainIDs from string representations of ints or chain names.

I've found myself wishing that this function exists in the context of working with ChainIDs. It's nice to have this parsing and validation provided by the SDK.

### DIFF
--- a/sdk/vaa/structs.go
+++ b/sdk/vaa/structs.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"math"
 	"math/big"
+	"strconv"
 	"strings"
 	"time"
 
@@ -319,6 +320,28 @@ func KnownChainIDFromNumber[N number](n N) (ChainID, error) {
 
 }
 
+// StringToKnownChainID converts from a string representation of a chain into a ChainID that is registered in the SDK.
+// The argument can be either a numeric string representation of a number or a known chain name such as "solana".
+// Inputs of unknown ChainIDs, including 0, will result in an error.
+func StringToKnownChainID(s string) (ChainID, error) {
+
+	// Try to convert from chain name first, and return early if it's found.
+	id, err := ChainIDFromString(s)
+	if err == nil {
+		return id, nil
+	}
+
+	// Ensure that the string can be parsed into a uint16 in order to avoid overflow issues when converting
+	// to ChainID (which is a uint16).
+	u16, err := strconv.ParseUint(s, 10, 16)
+	if err != nil {
+		return ChainIDUnset, err
+	}
+
+	return KnownChainIDFromNumber(u16)
+}
+
+// ChainIDFromString converts from a chain's full name (e.g. "solana") to its corresponding ChainID.
 func ChainIDFromString(s string) (ChainID, error) {
 	s = strings.ToLower(s)
 
--- a/sdk/vaa/structs_test.go
+++ b/sdk/vaa/structs_test.go
@@ -1183,3 +1183,85 @@ func TestChainIDFromNumber(t *testing.T) {
 		})
 	}
 }
+
+func TestStringToKnownChainID(t *testing.T) {
+
+	happy := []struct {
+		name     string
+		input    string
+		expected ChainID
+	}{
+		{
+			name:     "simple int 1",
+			input:    "1",
+			expected: ChainIDSolana,
+		},
+		{
+			name:     "simple int 2",
+			input:    "3104",
+			expected: ChainIDWormchain,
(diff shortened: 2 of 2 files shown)


sdk/vaa/structs.go
```diff
@@ -262,6 +263,33 @@ func (c ChainID) String() string {
...
+	found := false
+	// NOTE: This function does not contain ChainIDUnset, therefore this function returns an error if the argument is 0.
+	for _, id := range GetAllNetworkIDs() {
+		if ChainID(u16) == ChainID(id) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		err = fmt.Errorf("value %s is not a valid chainId", s)
```
bruce-riley, line 285: Is it worth doing a special check for id == 0 and returning ChainIDUnset?

bruce-riley (dismissed):

sdk/vaa/structs.go
johnsaigle: Curious about your thoughts on that actually. I originally had the function get up like this but given that `GetAllNetworkIDs()` seemed to specifically omit it, I figured I'd follow what the existing code was doing. Maybe they can have different behaviours though?

evan-gray: It depends on the use case. chain ID 0 / unset is used for governance. so if you're using this to parse something that comes from the _payload_ of real VAAs, then it's possible (though I'm not sure why you would have that as a string). If this is more for splitting a VAA ID or something, there should never be a VAA emitted with chain ID 0 as far as I am aware, which is similarly why 0 is not a "Network ID" for `GetAllNetworkIDs`. I guess I'm not sure of the motivation that led to wishing for this 😅

evan-gray: Follow up - what is the case where you have a chain ID (for some reason) but should be told it is invalid? Any uint16 is technically valid as a chain ID, so anything that passes the first line shouldn't necessarily trigger an error. Otherwise the caller would need to update their code any time there is a new chain added, which quickly becomes a nightmare. Maybe that's why I would lean on the side of continuing to _not_ have this function.

johnsaigle: Hey @evan-gray thanks for the review and for the questions. Essentially my motivation here is to get more use out of the ChainID type and wanted a way to go from a string to a valid, registered ChainID within Wormhole. Based on your questions I'm realizing that I was thinking of this code as a "wormhole SDK" rather than a "vaa SDK" and that's perhaps stretching the purpose of the code.

I think you're right that any uint16 is a valid ChainID in a sense (even if it doesn't correspond to any actual instantiated chain); for that reason `Atoi` is not a good name here because that sort of generic conversion should accept any uint16, I feel.

Would you be open to including this function if renamed/redocumented to something like "ValidChainIDFromString"? I would personally find this function useful but if in your view this is outside of the scope of this file then I'll defer to your judgement and close the PR.

evan-gray: My primary hesitation is that it’s a good way to cause a bug where the written code cannot withstand new chain deployments without being upgraded. This is a consistent issue with other tooling. It’s less of an issue in the guardian code (which dictates new chains, at least today) than it is for code elsewhere. Semantically, a chain number is no less “valid” just because a given version of the code doesn’t know what named chain it corresponds to yet.

evan-gray: Perhaps “known” would be a better adjective.

johnsaigle: How about `KnownNetworkIDFromString()`? That way we surface that it's using the NetworkID list under the hood and it becomes more clear that `0`/ChainIDUnset is not a valid use of the function. In this way also it should have the same maintenance requirements/constraints that GetAllNetworkIDs has.

evan-gray: I’d be okay with that as long as it has a doc string describing what (not) to use it for. Not 100% why they’re called NetworkID and not just ChainID there though. 🤷

johnsaigle: OK awesome, happy to add that documentation.

johnsaigle: I've renamed the function, added documentation, and changed the name of this PR and its description.

johnsaigle:
This PR has been refactored to incorporate changes from #4271  now that it has been merged.

bruce-riley (approved):

pleasew8t (approved):

PR #4241, opened 2025-01-24
Node/Solana: Shim support in testnet

bruce-riley:
This PR adds support for the Solana shim contract in testnet (but not mainnet).

To use this feature, set the following parameters:

```
--solanaRPC YOUR_SOLANA_RPC
--solanaContract 3u8hJUVTA4jH1wYAyUur7FFZVQ8H635K3tSHHF4ssjQ5 
--solanaShimContract EtZMZM22ViKMo4r5y4Anovs3wKQ2owUmDpjygnMMcdEX
```

Don't be alarmed by the size of this PR. A thousand lines of it are tests, and most of that is JSON encoded transactions.


evan-gray (approved):
Confirmed that these changes match what I previously reviewed in https://github.com/wormholelabs-xyz/wormhole/pull/11 with the addition of disabling mainnet support.

node/pkg/watchers/solana/client.go
```diff
@@ -345,10 +355,13 @@ func (s *SolanaWatcher) Run(ctx context.Context) error {
 		zap.String("wsUrl", wsUrl),
 		zap.String("contract", contractAddr),
 		zap.String("rawContract", s.rawContract),
+		zap.String("shimContractAddr", s.shimContractStr),
```
panoel, line 358: You have two variables: shimContractStr and shimContractAddr.  Here, the display text is shimContractAddr, but the variable displayed is shimContractStr.  A little confusing.

bruce-riley: Sorry it's confusing but it's intentional. I want to print out something human readable (which is the string). I guess I can change the tag to just "shimContract".

panoel (dismissed):

evan-gray (approved):

panoel (approved):

PR #4243, opened 2025-01-24
Node/Aptos: Don't replay old observation on startup

### LINKED ISSUE #4235: node/aptos: last message replayed on watcher startup
## Description

When the Aptos watcher starts up, it fetches the most recent event and processes it. On subsequent calls, it fetches only events starting at the following sequence. This means that every time the Aptos watcher starts (because the watcher restarts or a guardian restarts), the latest event gets re-observed.

https://github.com/wormhole-foundation/wormhole/blob/1dfeafd867989a140dd4aa1cf18bb89911cac5c6/node/pkg/watchers/aptos/watcher.go#L90-L94

https://github.com/wormhole-foundation/wormhole/blob/1dfeafd867989a140dd4aa1cf18bb89911cac5c6/node/pkg/watchers/aptos/watcher.go#L155-L163

## Recommendation

Skip processing the event on the first pass, when `nextSequence == 0`, and the event sequence number is not 0. That way the very first event on a new chain will not be skipped (and will be re-observed), but after that this will no longer occur.

bruce-riley:
This PR fixes #4235.

### DIFF
--- a/node/pkg/watchers/aptos/watcher.go
+++ b/node/pkg/watchers/aptos/watcher.go
@@ -205,15 +205,22 @@ func (e *Watcher) Run(ctx context.Context) error {
 					continue
 				}
 
+				eventSeq := eventSequence.Uint()
+				if nextSequence == 0 && eventSeq != 0 {
+					// Avoid publishing an old observation on startup. This does not block the first message on a new chain (when eventSeq would be zero).
+					nextSequence = eventSeq + 1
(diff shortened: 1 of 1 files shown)


evan-gray (approved):
confirmed the [http example](https://pkg.go.dev/net/http#example-Get) does close the body after reading.

panoel (approved):

PR #4116, opened 2024-09-10
Node/EVM: Verify EVM chain ID

bruce-riley:
This PR adds a check on start up of the EVM watcher to verify that the EVM chain ID of the RPC endpoint matches what is expected for the configured chain. This is to protect against a guardian connecting to the wrong chain by mistake.

Additionally, the reobservation with custom endpoint feature (#4260) will be updated to perform this check so that a reobservation is not attempted on the wrong chain.

This PR implements two static configuration maps that define information about each EVM watcher (one for mainnet and one for testnet) and adds a utility to verify that the EVM chain IDs specified in the config match what is reported for a well-known public RPC for that chain.

This PR centralizes much of the data associated with an EVM chain, which will simplify the addition of new chains.

PR #4260, opened 2025-02-07
Node: Reobserve with custom endpoint

bruce-riley:
This PR adds support for the `reobserve-with-endpoint` admin command.

Don't be alarmed by the size of this PR. 900+ lines are generated protobuf code. Also the code in `handleReobservationRequest` in Solana and EVM is mostly just code moved into a function so it can be reused.

Tests performed:

- Sepolia new observations still works
- Sepolia standard reobservation still works
- Sepolia with custom endpoint works
- Solana new observations still works
- Solana standard reobservation by account still works
- Solana standard reobservation by transaction ID still works
- Solana by account ID with custom endpoint works
- Solana by transaction ID with custom endpoint works
- Chain that is not enabled returns "chain X does not support reobservation by endpoint"
- Chain other than Solana or EVM that is enabled returns "chain X does not support reobservation by endpoint"
- Pointing at the wrong EVM chain returns "evm chain ID miss match".


node/cmd/guardiand/adminclient.go
```diff
@@ -415,6 +425,51 @@ func runSendObservationRequest(cmd *cobra.Command, args []string) {
 	}
 }
 
+func runReobservetWithEndpoint(cmd *cobra.Command, args []string) {
```
panoel, line 428: Same comment as above

proto/node/v1/node.proto
```diff
@@ -28,6 +28,9 @@ service NodePrivilegedService {
   // using the node's guardian key. The network rate limits these requests to one per second.
   // Requests at higher rates will fail silently.
   rpc SendObservationRequest (SendObservationRequestRequest) returns (SendObservationRequestResponse);
+  
+  // ReobservetWithEndpoint performs a local reobservation request using the specified endpoint.
+  rpc ReobservetWithEndpoint (ReobservetWithEndpointRequest) returns (ReobservetWithEndpointResponse);
```
panoel, line 33: Name misspelling?

node/pkg/watchers/solana/client.go
```diff
@@ -645,9 +634,9 @@ func (s *SolanaWatcher) fetchBlock(ctx context.Context, logger *zap.Logger, slot
 }
 
 // processTransaction processes a transaction and publishes any Wormhole events.
-func (s *SolanaWatcher) processTransaction(ctx context.Context, logger *zap.Logger, tx *solana.Transaction, meta *rpc.TransactionMeta, slot uint64, isReobservation bool) {
+func (s *SolanaWatcher) processTransaction(ctx context.Context, logger *zap.Logger, rpcClient *rpc.Client, tx *solana.Transaction, meta *rpc.TransactionMeta, slot uint64, isReobservation bool) (numObservations uint32) {
```
johnsaigle, line 637: Would it make sense to access `s.rpcClient` here instead? (And `s.logger` now that it exists)

node/pkg/watchers/solana/client.go
```diff
@@ -772,9 +765,11 @@ func (s *SolanaWatcher) processTransaction(ctx context.Context, logger *zap.Logg
 			}
 		}
 	}
+
+	return
 }
 
-func (s *SolanaWatcher) processInstruction(ctx context.Context, logger *zap.Logger, slot uint64, inst solana.CompiledInstruction, programIndex uint16, tx *solana.Transaction, signature solana.Signature, idx int, isReobservation bool) (bool, error) {
+func (s *SolanaWatcher) processInstruction(ctx context.Context, logger *zap.Logger, rpcClient *rpc.Client, slot uint64, inst solana.CompiledInstruction, programIndex uint16, tx *solana.Transaction, signature solana.Signature, idx int, isReobservation bool) (bool, error) {
```
johnsaigle, line 772: Same as above - would it make sense to access `s.rpcClient` here instead? (And `s.logger` now that it exists)

node/pkg/watchers/solana/client.go
```diff
@@ -857,18 +852,18 @@ func (s *SolanaWatcher) retryFetchMessageAccount(ctx context.Context, logger *za
...
 		common.RunWithScissors(ctx, s.errC, "retryFetchMessageAccount", func(ctx context.Context) error {
-			s.retryFetchMessageAccount(ctx, logger, acc, slot, retry+1, isReobservation)
+			s.retryFetchMessageAccount(ctx, logger, rpcClient, acc, slot, retry+1, isReobservation)
 			return nil
 		})
 	}
 }
 
-func (s *SolanaWatcher) fetchMessageAccount(ctx context.Context, logger *zap.Logger, acc solana.PublicKey, slot uint64, isReobservation bool) (retryable bool) {
+func (s *SolanaWatcher) fetchMessageAccount(ctx context.Context, logger *zap.Logger, rpcClient *rpc.Client, acc solana.PublicKey, slot uint64, isReobservation bool) (numObservations uint32, retryable bool) {
```
johnsaigle, line 861: Same again - would it make sense to access `s.rpcClient` here instead? (And `s.logger` now that it exists)

node/pkg/watchers/solana/client.go
```diff
@@ -881,7 +876,7 @@ func (s *SolanaWatcher) fetchMessageAccount(ctx context.Context, logger *zap.Log
 			zap.Uint64("slot", slot),
 			zap.String("commitment", string(s.commitment)),
 			zap.Stringer("account", acc))
-		return true
+		return 0, true
```
johnsaigle, line 879: Since you're using named return values, you could just use the word `return` here and in all the other cases in this function, if you like. Either way is fine, of course.

johnsaigle (commented):
Looks good, made some suggestions to simplify the code.

node/pkg/watchers/solana/client.go
bruce-riley: Would work for `s.logger` but not for `s.rpcClient` because we need to use the custom one when we are reobserving.

bruce-riley:
Note: Once #4116 is merged, the EVM version of `reobserve` should be updated to call `verifyEvmChainID` before performing the reobservation.

node/pkg/watchers/solana/client.go
bruce-riley: I confirmed that I do need to pass in `rpcClient`, but not `logger`. I made that change.

node/pkg/watchers/solana/client.go
bruce-riley: I confirmed that I do need to pass in `rpcClient`, but not `logger`. I made that change.

node/pkg/watchers/solana/client.go
```diff
@@ -1111,7 +1107,7 @@ func ParseMessagePublicationAccount(data []byte) (*MessagePublicationAccount, er
 	return prop, nil
 }
 
-func (s *SolanaWatcher) populateLookupTableAccounts(ctx context.Context, tx *solana.Transaction) error {
+func (s *SolanaWatcher) populateLookupTableAccounts(ctx context.Context, rpcClient *rpc.Client, tx *solana.Transaction) error {
```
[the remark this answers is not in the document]
bruce-riley: I confirmed that I do need to pass in `rpcClient`, but not `logger`. I made that change.

node/pkg/watchers/solana/client.go
```diff
@@ -1082,6 +1077,7 @@ func (s *SolanaWatcher) processMessageAccount(logger *zap.Logger, data []byte, a
 	}
 
 	s.msgC <- observation
+	return 1
```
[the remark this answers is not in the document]
bruce-riley: But if I have to do that, wherever I call it, I have to check the return value and and check it above so I can return one. This seems cleaner to me.

johnsaigle (dismissed):

node/pkg/watchers/evm/reobserve.go
```diff
@@ -0,0 +1,136 @@
...
+	w.logger.Info("received observation request", zap.String("tx_hash", tx.Hex()))
+
+	// SECURITY: Load the block number before requesting the transaction to avoid a
+	// race condition where requesting the tx succeeds and is then dropped due to a fork,
+	// but blockNumberU had already advanced beyond the required threshold.
+	//
+	// In the primary watcher flow, this is of no concern since we assume the node
+	// always sends the head before it sends the logs (implicit synchronization
+	// by relying on the same websocket connection).
+	blockNumberU := atomic.LoadUint64(&w.latestFinalizedBlockNumber)
```
evan-gray, line 33: this does not seem same to assume that the watcher finalized block has any relation to the transaction from another connection. can `Reobserve` create a new watcher instance?

node/pkg/watchers/solana/client.go
```diff
@@ -341,8 +343,17 @@ func (s *SolanaWatcher) Run(ctx context.Context) error {
 		ContractAddress: contractAddr,
 	})
 
+	// Don't overwrite these fields if they are already set in case there is an old go routine still using them.
```
evan-gray, line 346: does this actually happen somewhere or is this just a defense?

evan-gray (changes requested):

node/pkg/watchers/solana/client.go
johnsaigle: This drew my eye as well. Maybe it's worth logging if `s.ctx` isn't nil? Could be helpful to know if there's a Go routine out there that shouldn't be there.

bruce-riley: I have not actually seen it happen, it just seems like a theoretical possibility since the watcher spawns go routines to fetch accounts. Maybe those could still be running (and attempting to log or something) when the watcher restarts.

bruce-riley: These parameters will always be non-null on a watcher restart. It doesn't mean there is a go routine left around. The code previously just blindly overwrote the values on a restart when it didn't need to. The context is the one passed in from node.go, so it is valid for the life of the program, and once the logger is created, there is no reason to recreate it. It *probably* didn't hurt to blindly overwrite them, but it seemed like a good idea not to. 🤷

johnsaigle: Sounds good -- in that case does it make sense to rephrase the code comment? IMO it would be nice to preserve what you just wrote.

node/pkg/watchers/evm/reobserve.go
bruce-riley: I changed it to use the new `GetLatestBlocks` function to get the current finalized and safe blocks and pass them into this function rather than creating a new watcher. Creating a new watcher would be kind of messy since it would need a separate run function that didn't spin up all the runnables, etc.

bruce-riley:
Now that #4116 is merged, I added a check to make sure the custom endpoint URL points at the correct EVM chain.

evan-gray (approved):

johnsaigle (approved):

panoel (approved):

PR #4265, opened 2025-02-11
Wormchain token denom bug fix

mdulin2:
We created a private repository to do the fix. This is upstreaming the changes in the private repo for the fix.


joelsmith-2019 (approved):
Looks good! Did we want to update the block heights at which these changes are implemented as it's already past January 28th?

mdulin2:
Thanks for taking a look! @joelsmith-2019

This patch was already deployed in an emergency patch with these block numbers. I am trying to add it to this repo now.

mdulin2:
I see that my tests are failing now. I'll fix these and see what's going on now.

mdulin2:
I see the issue. The static height used for the cutover on mainnet was different between the test and the actual height. This caused the tests to fail.

mdulin2:
The test is fixed now. We should be good to merge.

johnsaigle (approved):

mdulin2:
@joelsmith-2019 Can you approve this again? I had a test that broke but fixed it.

joelsmith-2019 (approved):

PR #4267, opened 2025-02-18
Node: Don't hang on post to Loki

bruce-riley:
There was apparently a Grafana outage that caused the guardian to hang while posting to Loki because the channel filled up, blocking the caller. This PR adds checking for channel full and metrics to be incremented if the channel post succeeds or fails.

To test this PR I configured an nginx proxy for grafana logging as documented [here ](https://www.notion.so/wormholefoundation/Guardian-Testing-Plan-c786def970264c0995b2c200cedfffcd#78521dce33714cdc87f07e17f1e64aff)and pointed my local guardian at it.

I then added testing code that creates a go routine to generate ten log messages per second. After verifying those showed up in Grafana, I stopped the ngxinx proxy.

Before this fix, logging (even to the terminal) stopped. After this fix, the logging to the terminal continued. (Obviously there is a gap in Grafana.

I also verified that logging to Grafana resumed a few minutes after restarting the proxy service, without needing to restart the guardian.


andreclaro (approved):

lrogana:
Looks good. We have tested this in scenarios where loki is down

lrogana (approved):

evan-gray (approved):

PR #4269, opened 2025-02-18
Node/Solana: Shim watcher cleanup

bruce-riley:


panoel (approved):

node/pkg/watchers/solana/shim.go
```diff
@@ -269,17 +282,25 @@ func (s *SolanaWatcher) shimProcessRest(
 	var verifiedCoreEvent bool
 	var messageEvent *ShimMessageEventData
 	var err error
-	coreEventFound := false
 	for idx := startIdx; idx < len(innerInstructions); idx++ {
 		inst := innerInstructions[idx]
 		if inst.ProgramIDIndex == whProgramIndex {
-			if verifiedCoreEvent, err = shimVerifyCoreMessage(inst.Data); err != nil {
+			if verifiedCoreEvent {
+				return fmt.Errorf("detected multiple inner core instructions when there should not be at instruction %d, %d", outerIdx, idx)
```
johnsaigle, line 289: Right now I believe this check is too general: this returns an error if the inner instruction's program index matches the Wormhole Program's, regardless of the contents. I think instead what should be flagged here is the case where another "verified core event" occurs.

So the more targeted check would be something like:

```
// Ensure that only one verified core event appears in each set of inner instructions
if foundIt && verifiedCoreEvent {
 // return error 
 }
```

In practice this might be fine, but maybe there's a valid case (now or down the line) where the Wormhole program needs to be interacted with twice in the same inner instruction set.

node/pkg/watchers/solana/shim.go
```diff
@@ -269,17 +282,25 @@ func (s *SolanaWatcher) shimProcessRest(
...
+			}
+			foundIt, err := shimVerifyCoreMessage(inst.Data)
+			if err != nil {
 				return fmt.Errorf("failed to verify inner core instruction for shim instruction %d, %d: %w", outerIdx, idx, err)
 			}
-			alreadyProcessed.add(outerIdx, idx)
-			coreEventFound = true
+			if foundIt {
+				alreadyProcessed.add(outerIdx, idx)
```
johnsaigle, line 296: I think it's OK to add this the instruction to the alreadyProcessed set even if `foundIt` is false.

node/pkg/watchers/solana/shim.go
```diff
@@ -269,17 +282,25 @@ func (s *SolanaWatcher) shimProcessRest(
...
+				verifiedCoreEvent = true
+			}
 		} else if inst.ProgramIDIndex == shimProgramIndex {
-			if !coreEventFound {
+			if messageEvent != nil {
+				return fmt.Errorf("detected multiple shim message event instructions when there should not be at instruction %d, %d", outerIdx, idx)
```
johnsaigle, line 301: Same as the case above -- this will catch and throw an error for any interaction with the shim program, even if this isn't a second shim event. So, this code might end up looking like this:

```
		} else if inst.ProgramIDIndex == shimProgramIndex {
			if !verifiedCoreEvent {
				return fmt.Errorf("detected an inner shim message event instruction before the core event for shim instruction %d, %d: %w", outerIdx, idx, err)
			}
			thisEvent, err := shimParseMessageEvent(s.shimMessageEventDiscriminator, inst.Data)
			if err != nil {
				return fmt.Errorf("failed to parse inner shim message event instruction for shim instruction %d, %d: %w", outerIdx, idx, err)
			}
			if messageEvent != nil && thisEvent != nil && len(thisEvent) >  0 {
				return fmt.Errorf("detected multiple shim message event instructions when there should not be at instruction %d, %d", outerIdx, idx)
			}
			alreadyProcessed.add(outerIdx, idx)
		}

```

node/pkg/watchers/solana/shim.go
```diff
@@ -290,16 +311,21 @@ func (s *SolanaWatcher) shimProcessRest(
 		}
 
 		if verifiedCoreEvent && messageEvent != nil {
-			break
+			if !isTopLevel {
+				// We found what we are looking for, and additional shim events in the inner instruction case are allowed, so we can stop looking.
```
johnsaigle, line 315: Shouldn't the loop continue in case there are multiple `verifiedCoreEvent` occurrences? If not, it would be good to reword this to something like "In the integrator (inner instruction) case, additional core event and shim event pairs are allowed."

Basically my concern with this function is to make sure that:
* there's a one-to-one correspondance of core events and shim events in all cases (i.e. they're always matched)
* there are no "extra events" -- e.g. in the Direct/top-level case, there should only be one core event and one shim event

node/pkg/watchers/solana/shim_test.go
```diff
@@ -1007,3 +1003,1284 @@ func TestShimFromIntegratorWithMultipleShimTransactions(t *testing.T) {
 	assert.False(t, msg.IsReobservation)
 	assert.False(t, msg.Unreliable)
 }
+
+func TestShimDirectWithExtraWhEventBeforeShimEventShouldFails(t *testing.T) {
```
johnsaigle, line 1007:
```suggestion
func TestShimDirectWithExtraWhEventBeforeShimEventShouldFail(t *testing.T) {
```

node/pkg/watchers/solana/shim_test.go
```diff
@@ -1007,3 +1003,1284 @@ func TestShimFromIntegratorWithMultipleShimTransactions(t *testing.T) {
...
+
+	alreadyProcessed := ShimAlreadyProcessed{}
+	found, err := s.shimProcessTopLevelInstruction(logger, whProgramIndex, shimProgramIndex, tx, txRpc.Meta.InnerInstructions, 1, alreadyProcessed, false)
+	require.ErrorContains(t, err, "detected multiple inner core instructions when there should not be")
+	require.False(t, found)
+	require.Equal(t, 0, len(s.msgC))
+	require.Equal(t, 2, len(alreadyProcessed)) // The first core and shim events will have been added.
+}
+
+func TestShimTopLevelShouldFailIfNoInstructionsShouldFail(t *testing.T) {
```
johnsaigle, line 1619:
```suggestion
func TestShimTopLevelEmptyInstructionsShouldFail(t *testing.T) {
```
Just a suggestion to remove the duplicate use of `ShouldFail`

node/pkg/watchers/solana/shim_test.go
```diff
@@ -1007,3 +1003,1284 @@ func TestShimFromIntegratorWithMultipleShimTransactions(t *testing.T) {
...
+
+	alreadyProcessed := ShimAlreadyProcessed{}
+	found, err := s.shimProcessTopLevelInstruction(logger, whProgramIndex, shimProgramIndex, tx, txRpc.Meta.InnerInstructions, 1, alreadyProcessed, false)
+	require.ErrorContains(t, err, "topLevelIndex 1 is greater than 0")
+	require.False(t, found)
+	require.Equal(t, 0, len(s.msgC))
+	require.Equal(t, 0, len(alreadyProcessed))
+}
+
+func TestShimInnerShouldFailIfNoInstructionsShouldFail(t *testing.T) {
```
johnsaigle, line 1811:
```suggestion
func TestShimProcessInnerInstructions_OutOfBoundsStartIndexShouldFail(t *testing.T) {
```
This name is super long but the point is just to clarify that what's being tested here is the "start index" value relative to the length of the inner instructions.

node/pkg/watchers/solana/shim.go
```diff
@@ -226,6 +230,10 @@ func (s *SolanaWatcher) shimProcessInnerInstruction(
 	alreadyProcessed ShimAlreadyProcessed,
 	isReobservation bool,
 ) (bool, error) {
+	if startIdx >= len(innerInstructions) {
+		return false, fmt.Errorf("startIdx %d is greater than or equal to %d", startIdx, len(innerInstructions))
```
johnsaigle, line 234:
```suggestion
		return false, fmt.Errorf("startIdx %d is is out of bounds of slice innerInstructions (length: %d)", startIdx, len(innerInstructions))
```
From the error log it's not clear why e.g. "startIdx 3 is greater than or equal to 3" is a problem

johnsaigle (commented):
I think some of the error handling is a little too general and may cause some bugs down the line. Otherwise, I made some suggestions to improve error messages and fix typos in test names.

node/pkg/watchers/solana/shim.go
bruce-riley: I'd like to leave it the way it is. It think it's cleaner since we haven't "consumed" the event.

node/pkg/watchers/solana/shim.go
bruce-riley: I beefed up the comment.

johnsaigle (approved):
Thanks for incorporating the changes!

banescusebi (approved):
lgtm

panoel (approved):

PR #4274, opened 2025-02-21
Node/EVM: Add GetLatestBlocks API

bruce-riley:
It would be helpful (to be used in #4260 for instance) to have a way to get the current latest, finalized and safe block numbers for an EVM chain. This PR adds that ability.

This PR also removes a couple of obsolete poll connectors from the code base.

Note: The code in `block_utils.go` is really just some functions moved from `finalizer_poller.go`. I'm not sure why the diff doesn't properly show that. 🤷


node/pkg/watchers/evm/connectors/logpoller.go
evan-gray, line 1: I'm torn on getting rid of this - it's potentially, conceptually lower resource and could work with HTTP/S instead of WS, but it likely doesn't work for a production use case without significant additional work for edge cases, such as reorgs (which many websocket EVM RPC implementations handle) or load-balancing. 

Given that, I agree with getting rid of this unused code.

evan-gray (dismissed):
looks good, just the one question

node/pkg/watchers/evm/connectors/logpoller.go
bruce-riley: I'm assuming we can pull it back from git history if we ever need it, and I don't want to have to maintain it in the meantime.

node/pkg/watchers/evm/watcher.go
```diff
@@ -249,6 +249,13 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 		}
 	}
 
+	latest, finalized, safe, err := w.ethConn.GetLatest(ctx)
```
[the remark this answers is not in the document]
bruce-riley: This was just to demonstrate that it works so I removed it.

node/pkg/watchers/evm/connectors/logpoller.go
evan-gray: Agreed!

panoel (approved):

PR #4277, opened 2025-02-24
txverifier: fix bugs and improve error handling

johnsaigle:
- Sometimes From _can_ be set to the zero address when a token is minted. This restriction has been removed
- Made error handling around irrelevant receipts less noisy. Now there will not be any WARN logs when a receipt contains a LogMessagePublished event that was sent by someone other than the token bridge.
- Changed functions that process receipts to always return nil rather than the empty struct
- Added more unit tests to ensure that functions don't panic in nil input
- Added Validation function for TransferReceipts
- Add README for running the Transfer Verifier CLI tool
- Fix typo in variable name

### DIFF
--- a/node/pkg/txverifier/evm.go
+++ b/node/pkg/txverifier/evm.go
@@ -86,10 +86,18 @@ func (tv *TransferVerifier[ethClient, Connector]) ProcessEvent(
 
 	// Parse raw transaction receipt into high-level struct containing transfer details.
 	transferReceipt, parseErr := tv.ParseReceipt(receipt)
-	if parseErr != nil || transferReceipt == nil {
-		tv.logger.Warn("error when parsing receipt. skipping validation",
-			zap.String("receipt hash", receipt.TxHash.String()),
-			zap.Error(parseErr))
+	if transferReceipt == nil {
+		if parseErr != nil {
+			tv.logger.Warn("error when parsing receipt. skipping validation",
+				zap.String("receipt hash", receipt.TxHash.String()),
+				zap.Error(parseErr))
+		} else {
+			tv.logger.Debug("parsed receipt did not contain any LogMessagePublished events",
+				zap.String("receipt hash", receipt.TxHash.String()))
+		}
+		// Return true regardless of parsing errors. False is reserved
+		// for confirmed bad scenarios (arbitrary message
+		// publications), not parsing errors or irrelevant receipts.
 		return true
 	}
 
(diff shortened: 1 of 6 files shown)


node/pkg/txverifier/evmtypes.go
```diff
@@ -518,7 +522,27 @@ type TransferReceipt struct {
...
+		return errors.Join(err, errors.New("parsed receipt's Deposits field is nil"))
+	}
+	if r.Transfers == nil {
+		return errors.Join(err, errors.New("parsed receipt's Transfers field is nil"))
+	}
+	if r.MessagePublications == nil {
+		return errors.Join(err, errors.New("parsed receipt's MessagePublications field is nil"))
+	}
+	if len(*r.MessagePublications) == 0 {
+		return errors.Join(err, errors.New("parsed receipt' has no Message Publications"))
```
pleasew8t, line 542:
```suggestion
		return errors.Join(err, errors.New("parsed receipt has no Message Publications"))
```

node/pkg/txverifier/evm_test.go
```diff
@@ -687,6 +687,39 @@ func TestProcessReceipt(t *testing.T) {
...
 
+func TestTransferReceiptValidate(t *testing.T) {
+	transferReceipt := &TransferReceipt{
+		Deposits:            &[]*NativeDeposit{},
+		Transfers:           &[]*ERC20Transfer{},
+		MessagePublications: &[]*LogMessagePublished{},
+	}
+	err := transferReceipt.Validate()
+	require.Error(t, err, "Validate must return error when its fields are nil")
+
```
pleasew8t, line 698: I don't think you are actually testing the function for `nil` input. This probably returns an error (expected) because of the length of `MessagePublications`, because the fields aren't `nil`. I'd recommend a testcase for each error message.

pleasew8t (dismissed):

node/pkg/txverifier/evm.go
```diff
@@ -86,10 +86,18 @@ func (tv *TransferVerifier[ethClient, Connector]) ProcessEvent(
 
 	// Parse raw transaction receipt into high-level struct containing transfer details.
 	transferReceipt, parseErr := tv.ParseReceipt(receipt)
-	if parseErr != nil || transferReceipt == nil {
-		tv.logger.Warn("error when parsing receipt. skipping validation",
-			zap.String("receipt hash", receipt.TxHash.String()),
-			zap.Error(parseErr))
+	if transferReceipt == nil {
```
bruce-riley, line 89: This change assumes that `tv.ParseReceipt` will never return an error when `transferReceipt` is not nil. Is that a safe assumption (maybe now, but what about in the future)?

johnsaigle: That's a good point, I'll add error-handling at this level at well.

bruce-riley (dismissed):

node/pkg/txverifier/evm.go
johnsaigle: Added in the most recent commit.

bruce-riley (approved):

evan-gray (approved):
only reviewed `cspell-custom-words.txt` changes

PR #4280, opened 2025-02-26
txverifier: Increase integration test coverage and simplify Tilt set-up

johnsaigle:
- Add integration test case for Transfer Tokens (in addition to existing test for 'Transfer Tokens With Payload')
- Refactor integration test set-up. Now there is only one YAML file to manage instead of two. The tests have also been expanded to check for a variable number of error strings and include a timeout. They also report an error if too many error strings are found.
- The Tilt configuration now runs this test as a Job rather than a Deployment. This is more appropriate for a test that is meant to run once.


bruce-riley (dismissed):

pleasew8t (approved):

scripts/check-docker-pin.sh
```diff
@@ -12,9 +12,13 @@
 #   - We ignore solana AS (builder|ci_tests) because it's a relative reference to another FROM call
 #   - We ignore cosmwasm_artifacts AS artifacts because it's a local reference only, is built in tilt
 #   - We ignore base AS (ignite-go-build|ignite-vue-build) because the base image is already pinned in wormchain/Dockerfile.proto
+#   - We ignore guardiand-image AS guardian because this is it's a local reference built in tilt
 #
-git ls-files -z | grep -z "Dockerfile*" | xargs -r -0 grep -s "FROM" | egrep -v 'sha256|scratch|solana|aptos|sui|base|cosmwasm_artifacts|cli-gen|const-gen|dev AS (application|base|builder|ci_tests|tests|artifacts|ignite-go-build|ignite-vue-build|cli-export|const-export|build)'
-if [ $? -eq 0 ]; then
+if git ls-files -z \
```
evan-gray, line 17: This change seems to no longer pick up certain unpinned docker files! If I remove the sha from a dockerfile, this still reports "No unpinned docker files". Please revert this change as it is unnecessary (see suggestion).

devnet/tx-verifier/Dockerfile.tx-verifier-evm
```diff
@@ -0,0 +1,23 @@
+# We need the built guardiand image
+FROM guardiand-image as guardiand
```
evan-gray, line 2: This change passes the pin check for me locally.
```suggestion
FROM guardiand-image AS base
```

devnet/tx-verifier/README.md
```diff
@@ -26,39 +26,40 @@ transactions to the `anvil` instance that powers the Ethereum testnet while bein
...
 could be modified to detect that this script has written the success message to figure out whether the whole test completed successfully.
 
-### Pods
+The integration test is considered successful as soon as two instances of the error pattern are detected, one for
+each message type sent by the `transfer-verifier-test.sh`.
 
-The files detailed below each have a primary role and are responsible for running one of the main pieces of the test functionality:
+### YAML File Description
 
+The YAML file that runs the integration tests runs three containers:
```
evan-gray, line 47: Is this accurate? It looks like the yaml only has one container.

evan-gray (changes requested):

evan-gray (approved):
reviewed the tiltfile changes

devnet/tx-verifier/Dockerfile.tx-verifier-evm
```diff
@@ -0,0 +1,23 @@
...
+
+# node is required to install Foundry
+FROM node:19.6.1-slim@sha256:a1ba21bf0c92931d02a8416f0a54daad66cb36a85d2b73af9d73b044f5f57cfc
+
+# prepare cast
+COPY --from=foundry /usr/local/bin/cast /bin/cast
+
+# prepare guardiand
+COPY --from=guardiand /guardiand /bin/guardiand
+COPY --from=guardiand /usr/lib/libwasmvm.*.so /usr/lib/
```
evan-gray, line 17:
```suggestion
COPY --from=base /guardiand /bin/guardiand
COPY --from=base /usr/lib/libwasmvm.*.so /usr/lib/
```

mdulin2:
Looks like a good change. Isolates the testing to the tx-verifier location more and has a failure case for multiple transfers.

PR #4284, opened 2025-02-27
Update go linting rules 1

mdulin2:
Adding in more linter rules to try to catch some foot-guns in Golang.


johnsaigle:
Closing because these lint rules exist on main now.

PR #4285, opened 2025-02-28
Node/EVM: Berachain not instant finality

bruce-riley:
We noticed recently that Berachain reports different block numbers for latest and finalized / safe. This PR updates it to use the batch poller rather than the instant finality poller.

### DIFF
--- a/node/pkg/watchers/evm/chain_config_test.go
+++ b/node/pkg/watchers/evm/chain_config_test.go
@@ -128,6 +128,7 @@ func getFinalityForTest(env common.Environment, chainID vaa.ChainID) (instantFin
 		chainID == vaa.ChainIDArbitrumSepolia ||
 		chainID == vaa.ChainIDBase ||
 		chainID == vaa.ChainIDBaseSepolia ||
+		chainID == vaa.ChainIDBerachain || // Berachain really isn't instant finality, despite what this doc says: https://docs.berachain.com/faq/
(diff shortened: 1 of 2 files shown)


panoel (approved):

evan-gray (approved):

node/pkg/watchers/evm/chain_config_test.go
```diff
@@ -128,6 +128,7 @@ func getFinalityForTest(env common.Environment, chainID vaa.ChainID) (instantFin
 		chainID == vaa.ChainIDArbitrumSepolia ||
 		chainID == vaa.ChainIDBase ||
 		chainID == vaa.ChainIDBaseSepolia ||
+		chainID == vaa.ChainIDBerachain || // Berachain really isn't instant finality, despite what this doc says: https://docs.berachain.com/faq/
```
johnsaigle, line 131: I don't think this comment is accurate. Instead, it's just that the chain config will require Berachain's blocks to have a `finalized` status.

johnsaigle (approved):
Code updates look good. I disagree with the code comment but it doesn't need to block the PR.

node/pkg/watchers/evm/chain_config_test.go
bruce-riley: Close enough, especially since this function will be deleted soon anyway (per the function prologue).

PR #4292, opened 2025-03-04
Node/P2P: Add protected peers list

bruce-riley:
This PR adds the ability to specify a set of P2P peer IDs that should be added to the protected list.

**Guardian**

This is done for the guardian with something like this:

```
--protectedPeers 12D3KooWE4dmZwxhfjCKHLUqSaww96Cf7kmq1ZuKmzPz3MrJgZxp,12D3KooWMkxuNh1abxCRjsfNZ9D2C8ZgcsQ4xxbXBuuqdxVRbasu \
--protectedPeers 12D3KooWCW3LGUtkCVkHZmVSZHzL3C4WRKWfqAiJPz1NR7dT9Bxh,12D3KooWJXA6goBCiWM8ucjzc4jVUBSqL9Rri6UpjHbkMPErz5zK \
--ccqProtectedPeers 12D3KooWE4dmZwxhfjCKHLUqSaww96Cf7kmq1ZuKmzPz3MrJgZxp,12D3KooWMkxuNh1abxCRjsfNZ9D2C8ZgcsQ4xxbXBuuqdxVRbasu \
--ccqProtectedPeers  12D3KooWCW3LGUtkCVkHZmVSZHzL3C4WRKWfqAiJPz1NR7dT9Bxh,12D3KooWJXA6goBCiWM8ucjzc4jVUBSqL9Rri6UpjHbkMPErz5zK \
```

Note that there can be multiple instances of the same parameter, and each instance may contain either a single peer ID or a comma separated list of them.

**Spy and CCQ proxy Server**

The Spy and the CCQ proxy Server also support the `--protectedPeers` parameter.

### DIFF
--- a/node/pkg/p2p/run_params.go
+++ b/node/pkg/p2p/run_params.go
@@ -60,6 +60,7 @@ type (
 		ccqBootstrapPeers      string
 		ccqPort                uint
 		ccqAllowedPeers        string
+		protectedPeers         []string
 	}
 
 	// RunOpt is used to specify optional parameters.
@@ -162,6 +163,14 @@ func WithDisableHeartbeatVerify(disableHeartbeatVerify bool) RunOpt {
 	}
 }
 
+// WithProtectedPeers is used to set the protected peers.
+func WithProtectedPeers(protectedPeers []string) RunOpt {
+	return func(p *RunParams) error {
+		p.protectedPeers = protectedPeers
+		return nil
+	}
+}
+
 // WithGuardianOptions is used to set options that are only meaningful to the guardian.
 func WithGuardianOptions(
 	nodeName string,
@@ -185,6 +194,7 @@ func WithGuardianOptions(
 	ccqBootstrapPeers string,
 	ccqPort uint,
 	ccqAllowedPeers string,
+	protectedPeers []string,
 ) RunOpt {
 	return func(p *RunParams) error {
 		p.nodeName = nodeName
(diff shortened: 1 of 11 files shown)


node/pkg/p2p/run_params.go
```diff
@@ -162,6 +163,14 @@ func WithDisableHeartbeatVerify(disableHeartbeatVerify bool) RunOpt {
 	}
 }
 
+// WithProtectedPeers is used to set the protected peers.
+func WithProtectedPeers(protectedPeers []string) RunOpt {
```
evan-gray, line 168: should there be a test added for this one?

node/pkg/p2p/run_params.go
```diff
@@ -162,6 +164,22 @@ func WithDisableHeartbeatVerify(disableHeartbeatVerify bool) RunOpt {
...
+	return func(p *RunParams) error {
+		p.protectedPeers = protectedPeers
+		return nil
+	}
+}
+
+// WithCcqProtectedPeers is used to set the protected peers for CCQ.
+func WithCcqProtectedPeers(ccqProtectedPeers []string) RunOpt {
```
evan-gray, line 176: ditto

evan-gray (dismissed):
Concur this parameter follows the pattern of something like https://docs.anza.xyz/operations/setup-an-rpc-node. My understanding is that this would inform the connection manager to maintain connections to these peer IDs should the number of connections grow past the high watermark.

evan-gray (approved):

djb15 (approved):

PR #4294, opened 2025-03-06
Node: Buffer log messages

bruce-riley:
This PR addresses an issue where log messages being published to telemetry are getting dropped because the internal channel used to write to Loki is unbuffered (and is not configurable). A recent PR made the writes to that channel non-blocking because a Grafana outage caused the guardian to hang on logging.

This PR adds a buffered channel and a go routine that reads off that channel and writes to the Loki channel in a blocking fashion. 

See the comment below for the tests that were executed to verify this PR.


evan-gray (dismissed):

panoel (dismissed):

node/pkg/telemetry/loki.go
```diff
@@ -142,13 +153,35 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
 		lokiLabels[level] = levLabels
 	}
 
+	// Create a buffered channel so the application does not block in the logger. The size of this
+	// channel should be big enough to handle a burst of log messages from the app.
+	localChan := make(chan api.Entry, 1000)
```
johnsaigle, line 162: It would be better to declare the `1000` as a constant, and maybe take the comment away here and add it to the defintion.

node/pkg/telemetry/loki.go
```diff
@@ -142,13 +153,35 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+
+	// Create a local context with a cancel function so we can signal our go routine to shutdown when the time comes.
+	localContext, localCancel := context.WithCancel(ctx)
+
+	// Kick off a go routine to read from the local buffered channel and write to the Loki unbuffered channel.
+	go func(ctx context.Context, c client.Client, localC chan api.Entry) {
+		for {
+			select {
+			case <-ctx.Done():
+				c.Stop()
```
johnsaigle, line 172: Will buffered messages in the local channel be processed when this occurs? Or when any error occurs? We might encounter a case where the buffered messages are dropped.

node/pkg/telemetry/loki.go
```diff
@@ -142,13 +153,35 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+				return
+			case entry := <-localC:
```
johnsaigle, line 174: Is it worth adding error checking here? I'm not sure why `localC` would get closed early but no harm in checking right?

node/pkg/telemetry/loki.go
```diff
@@ -64,28 +73,30 @@ func (logger *ExternalLoggerLoki) log(time time.Time, message json.RawMessage, l
...
-		logger.c.StopNow()
+		// Signal the go routine to shut down the logger and exit.
+		logger.localCancel()
 	}
 }
 
 func (logger *ExternalLoggerLoki) close() error {
-	logger.c.Stop()
+	// Signal the go routine to shut down the logger and exit.
+	logger.localCancel()
```
johnsaigle, line 101: should this also do `logger.c.Stop()`?

node/pkg/telemetry/loki.go
```diff
@@ -142,13 +153,35 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
 		lokiLabels[level] = levLabels
 	}
 
+	// Create a buffered channel so the application does not block in the logger. The size of this
+	// channel should be big enough to handle a burst of log messages from the app.
+	localChan := make(chan api.Entry, 1000)
+
+	// Create a local context with a cancel function so we can signal our go routine to shutdown when the time comes.
+	localContext, localCancel := context.WithCancel(ctx)
```
johnsaigle, line 165: I'd prefer a name like `bufferedLogChannel` or `lokiConsumerChannel` or something similar.

node/pkg/telemetry/loki.go
```diff
@@ -142,13 +153,35 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+		for {
+			select {
+			case <-ctx.Done():
+				c.Stop()
+				return
+			case entry := <-localC:
+				c.Chan() <- entry // can_block: That's why we are in this separate go routine.
+			}
+		}
+	}(localContext, c, localChan)
```
johnsaigle: I'm wondering about the unwinding here if the cancel context is called. Similar to the Done case, what happens to the buffered messages? Do they get dropped?

johnsaigle (commented):
I had some questions about the error-handling or early cancel cases, and also left some suggestions about comments and variable names. Overall I think the mechanism makes sense.

node/pkg/telemetry/loki.go
johnsaigle: Ah you committed this before I finished writing my review

node/pkg/telemetry/loki.go
evan-gray: Also, I think stop will get called twice now, once here, and again in `close()` which triggered this.

bruce-riley: This code has changed drastically. Please take another look.

node/pkg/telemetry/loki.go
bruce-riley: This code has changed drastically. Please take another look.

bruce-riley:
**Testing**

Here is how I tested this PR.

- First, enabled logging to Grafana, pointing it at the dev instance. Verified that log messages are seen, the `wormhole_loki_messages_sent` metric is pegged and the `wormhole_loki_messages_dropped` one may be pegged a little (just confirming that logging to Grafana is working).
- Created a go routine that writes to the logger at a high rate of speed and monitored the drop metric. Before the fix, this metric was being pegged a lot, after the fix, it is not. Confirmed the messages showed up in Grafana.
- After running for a minute, caused the above routine to do a fatal error and confirmed that the guardian exited cleanly.
- Sent the guardian a sigterm and verified that it shut down cleanly.
- Create an invalid config that causes a `log.Fatal` in node.go (after Loki is enabled) and confirm the message makes it to Grafana.
- With the guardian pointed at an nginx proxy for logging, stopped the proxy server, verified the drop metric eventually gets pegged (after 1000 messages), then restarted the proxy and verified the drop metric stops getting pegged and messages resume showing up in Grafana.

**Testing Code**

Added the following code into the EVM watcher to generate lots of logs.

```
	common.RunWithScissors(ctx, errC, "log_junk", func(ctx context.Context) error {
		t := time.NewTicker(250 * time.Millisecond)
		defer t.Stop()
		count := 0
		interval := 0
		for {
			select {
			case <-ctx.Done():
				return nil
			case <-t.C:
				// if interval == 2*15 {
				// 	logger.Fatal("And here's a fatal error!")
				// }
				for i := 0; i < 100; i++ {
					logger.Info("Logging junk", zap.Int("interval", interval), zap.Int("count", count))
					count++
				}
				interval++
			}
		}
	})
```

evan-gray (approved):

panoel (approved):

node/pkg/telemetry/loki.go
```diff
@@ -64,28 +84,29 @@ func (logger *ExternalLoggerLoki) log(time time.Time, message json.RawMessage, l
...
 	if level == zapcore.FatalLevel {
-		logger.c.StopNow()
+		logger.stopWorkerWithTimeout()
 	}
 }
 
 func (logger *ExternalLoggerLoki) close() error {
-	logger.c.Stop()
+	// Shut down the worker and wait for it to exit. It has a timeout so we won't wait forever.
+	logger.stopWorkerWithTimeout()
```
johnsaigle, line 109: It looks like the `stopWorkerWithTimeout` doesn't return an error, so this function always returns `nil` and never an `error`. I think the function signature here can be changed to not return anything.
It seems to only be called here:
https://github.com/wormhole-foundation/wormhole/blob/a2fbcc682e8ea4506272409b652d4504690167fd/node/pkg/telemetry/telemetry.go#L89-L91
And that function too can be changed to not have a return value. I looked at the incoming calls to that function and it looks like none of them check the error case anyway.

Alternatively, errors could get propagated out of `logWriter` through `stopWorkerWithTimeout` and into this function.

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+	go func(c client.Client) {
+		c.StopNow()
+		stopExitedC <- struct{}{}
+	}(c)
+
+	// Wait for the go routine to exit or the timer to expire.
+	select {
+	case <-stopExitedC:
+		return nil
+	case <-time.After(clientTimeout):
```
johnsaigle, line 278: Would you mind adding a comment here and in the next function why a `time.After()`-style timeout is used here instead of the Context-based pattern in `flushLogsWithTimeout`?

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+		case entry, ok := <-localC:
+			if !ok {
+				logger.Error("Loki log writer is exiting because the buffered channel has been closed")
+				return
+			}
+
+			// Write to Loki in a blocking manner unless we are signaled to shutdown.
+			select {
+			case c.Chan() <- entry:
+				pendingEntry = nil
```
johnsaigle, line 212: Should there be a check here that `pendingEntry` is not equal to nil before writing to it?

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+// and shutdown the Loki client using a timeout for both actions.
+func logWriter(ctx context.Context, logger *zap.Logger, localC chan api.Entry, workerExitedC chan struct{}, c client.Client) {
+	// pendingEntry is used to save the last log message if the write to Loki is interrupted by the context being canceled. We will attempt to flush it.
+	var pendingEntry *api.Entry
+
+	for {
+		select {
+		case entry, ok := <-localC:
+			if !ok {
+				logger.Error("Loki log writer is exiting because the buffered channel has been closed")
```
johnsaigle, line 205: I think this case should continue with the normal cleanup logic even when `localC` is closed: 
* `pendingEntry` should be checked for nil and written to Loki if it's not
* the Loki client should be stopped
* the `workerExitedC` channel should be written to

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+				logger.Error("worker failed to flush logs", zap.Error(err), zap.Int("numEventsRemaining", len(localC)))
+			}
+
+			// Stop the client without blocking indefinitely.
+			if err := stopClientWithTimeout(c); err != nil {
+				logger.Error("worker failed to stop Loki client", zap.Error(err))
+			}
+
+			// Signal that we are done.
+			workerExitedC <- struct{}{}
```
johnsaigle, line 231: Might be worth adding another select statement here in case the channel somehow ends up being full or closed.

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+				return
+			}
+
+			// Write to Loki in a blocking manner unless we are signaled to shutdown.
+			select {
+			case c.Chan() <- entry:
+				pendingEntry = nil
+			case <-ctx.Done():
+				// Time to shutdown. We probably failed to write this message, save it so we can try to flush it.
+				pendingEntry = &entry
```
johnsaigle, line 215: Also here -- what if `pendingEntry` is already populated?

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+			}
+		case <-ctx.Done():
+			logger.Info("Loki log writer shutting down")
+
+			// Flush as much as we can in our allowed time.
+			if err := flushLogsWithTimeout(localC, c, pendingEntry); err != nil {
+				logger.Error("worker failed to flush logs", zap.Error(err), zap.Int("numEventsRemaining", len(localC)))
```
johnsaigle, line 222: Should this include information about `pendingEntry` in the case where that write fails? Right now this could log information about `localC` but pendingEntry might get ignored.

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
...
+func flushLogsWithTimeout(localC chan api.Entry, c client.Client, pendingEntry *api.Entry) error {
+	// Create a timeout context. Base it on the background one since ours has been canceled.
+	timeout, cancel := context.WithTimeout(context.Background(), clientTimeout)
+	defer cancel()
+
+	if pendingEntry != nil {
+		select {
+		case c.Chan() <- *pendingEntry:
+		case <-timeout.Done():
+			return errors.New("timeout")
```
johnsaigle, line 249: I think it's worth disambiguating the error messages here. e.g. this could be "timeout: could not write pending entry" and the other one "timeout: could not completely flush buffered logs".

It might be helpful also write out the configured timeout so that it's visible in the logs as something like "timeout after X ms"

node/pkg/telemetry/loki.go
```diff
@@ -142,15 +163,133 @@ func NewLokiCloudLogger(ctx context.Context, logger *zap.Logger, url string, pro
 		lokiLabels[level] = levLabels
 	}
 
+	// Create a buffered channel so the application does not block in the logger.
+	bufferedChan := make(chan api.Entry, bufferSize)
+
+	// Create a local context with a cancel function so we can signal our worker to shutdown when the time comes.
```
johnsaigle, line 169: Maybe also "Cancelling the worker will also stop the connection to the Loki client."

johnsaigle (changes requested):
I think there are a few edge-cases worth considering and handling here.

node/pkg/telemetry/loki.go
bruce-riley: I make the text unique. I don't think it's necessary to log the timeout since it is a constant in the code.

node/pkg/telemetry/loki.go
bruce-riley: Removed the return value.

node/pkg/telemetry/loki.go
bruce-riley: Same comment as above.

johnsaigle (approved):

evan-gray (approved):

PR #4242, opened 2025-01-24
node: Add Transfer Verifier whitepaper

johnsaigle:
Adds a whitepaper for the Transfer Verifier.

Future work: the Sui details should be added to the whitepaper once #4324 is stable and merged.

PR #4304, opened 2025-03-19
docs: Add transfer verifier instructions for guardians

johnsaigle:
This document serves as configuration instructions for the Guardians for enabling the Transfer Verifier feature.

## Related

- #4242 (The actual whitepaper is here)
- #4233

### DIFF
--- a/docs/transfer-verifier.md
+++ b/docs/transfer-verifier.md
@@ -1,0 +1,15 @@
+# Transfer Verifier
+
+Below are admin controls surfaced to the Guardians for the Governor plugin. For
+a background on the feature and its objectives, see [the whitepaper](whitepapers/0014_transfer_verifier.md)
+
+## How To Enable Transfer Verifier
+The Transfer Verifier feature is disabled by default. Guardians can enable it by passing the following flag to the `guardiand` command when starting it up:
+
+```bash
+# Enable Transfer Verifier for chain with ID 2 (Ethereum)
+--transferVerifierEnabledChainIDs=2
+```
+
+This parameter is a comma-separated list of Wormhole Chain IDs for which transfer verification will be enabled.
+Only some chains support the Transfer Verifier.


docs/transfer-verifier.md
```diff
@@ -0,0 +1,16 @@
...
+The Transfer Verifier feature is disabled by default. Guardians can enable it by passing the following flag to the `guardiand` command when starting it up:
+
+```bash
+# Enable Transfer Verifier for chain with ID 2 (Ethereum)
+--transferVerifierEnabledChainIDs=2
+```
+
+This parameter is a comma-separated list of Wormhole Chain IDs for which transfer verification will be enabled.
+
+Only some chains support the Transfer Verifier.
```
djb15, line 16:
```suggestion
Only some chains support the Transfer Verifier. Enabling Transfer Verifier for Chain IDs via the command line that do not have the underlying support will have no effect. 
```

johnsaigle: This suggestion isn't accurate. It will cause the node to fail to start.

https://github.com/wormhole-foundation/wormhole/blob/bce6b859f4f95b012619d08fcf84394885fa11d8/node/cmd/guardiand/node.go#L960-L968

djb15: Oh I misremembered! Could you add the inverse of what I suggested then? (i.e. what you described in your comment)

johnsaigle:
```suggestion
Only some chains support the Transfer Verifier. If an unsupported chain is specified, the node will not start. It will also display an error message stating that the chain does not have a Transfer Verifier implementation.
```

djb15 (approved):

mdulin2 (approved):

johnsaigle:
@mdulin2 @djb15 thanks for the reviews. I realized this file was missing an example for configuring more than one chain so I've added that. Please re-review 🙏🏻

docs/transfer-verifier.md
```diff
@@ -0,0 +1,19 @@
...
+# Example 1: Enable Transfer Verifier for chain with ID 2 (Ethereum)
+--transferVerifierEnabledChainIDs=2
+
+# Example 2: Enable Transfer Verifier for both Ethereum and Sui
+--transferVerifierEnabledChainIDs=2,21
+```
+
+This parameter is a comma-separated list of Wormhole Chain IDs for which transfer verification will be enabled.
+
+Only some chains support the Transfer Verifier.
```
johnsaigle:
```suggestion
Only some chains support the Transfer Verifier. If an unsupported chain is specified, the node will not start. It will also display an error message stating that the chain does not have a Transfer Verifier implementation.
```

djb15 (approved):

djb15 (approved):

PR #4307, opened 2025-03-20
lint: add linters with no violations

johnsaigle:
This PR adds new linters to golangci-lint, specifically ones that the current code does not violate. The idea here is to get these new linters "for free" and prevent any future code from violating these rules.

Related:
This PR has different scope from these other open PRs, as they aim to resolve violations:
- https://github.com/wormhole-foundation/wormhole/pull/4284 which adds new linters and fixes their violations
- #4224 which updates the golanci-lint version and fixes violations that arose from the rules being updated

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,14 +15,15 @@ linters:
     - typecheck
     - unused
 
-  # additional linters
-    - gosec
+    # additional linters
     - forcetypeassert
+    - gosec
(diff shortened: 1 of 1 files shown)


johnsaigle:
Converting this to draft for now. We might as well update the major version of golangci-lint first as it changes to yml format. #4322

PR #4312, opened 2025-03-21
node: Drop MessagePublications with Rejected status

johnsaigle:
This PR adds a changes that causes the processor to drop messages marked as Rejected and logs that it has done so.

Messages can only be marked as Rejected if:
- Transfer Verification is enabled via a CLI flag
- The originating chain is Ethereum (or an Ethereum testnet)
- The message fails transfer verification

Follow-up work to: #4233


johnsaigle:
Closing this in favour of a new, more robust PR that will be opened soon.

PR #4314, opened 2025-03-25
node: add predeclared lint and fix violations

johnsaigle:
Go allows you to shadow built-in keyword functions which can lead to all kinds of weird errors. 

This PR adds a lint that helps us detect these cases and fixes existing violations in Wormhole by renaming variables.

More info on the lint here: https://github.com/nishanths/predeclared

Here's an example that you can run locally, showing a case where `len` and `int` are shadowed in this way. Note that this code compiles and runs.

```go

package main

import "fmt"

type int string

func len(s string) uint {
	return 3
}

func main() {
	fmt.Printf("test\n")
	fmt.Printf("num=%d\n", int(1_000_000)+int(3))
	fmt.Printf("length=%d\n", len("foo"))
	fmt.Printf("length2=%d\n", len("foobar"))
}
```

### DIFF
--- a/node/pkg/query/response.go
+++ b/node/pkg/query/response.go
@@ -1025,12 +1025,12 @@ func (sar *SolanaAccountQueryResponse) UnmarshalFromReader(reader *bytes.Reader)
 			return fmt.Errorf("failed to read owner [%d]: %w", n, err)
 		}
 
-		lenData := uint32(0)
-		if err := binary.Read(reader, binary.BigEndian, &lenData); err != nil {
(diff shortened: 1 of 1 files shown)


bruce-riley (approved):

djb15 (approved):

johnsaigle:
Moving this to draft for the moment as it adds a linter that is also added in #4323. I may close this depending on how that progresses.

johnsaigle:
#4323 added this lint, so this PR is not necessary.

Issue #4029, opened 2024-07-19
node: Make iteration over Governor information more efficient


johnsaigle:
## Description and context

In the context of the Governor, we normally we work with `chainEntry`s via the `chains` map which maps chain IDs to pointers to `chainEntry`s. This is nice because we can look up info by chain entry.

https://github.com/wormhole-foundation/wormhole/pull/4016 adds a slice of sorted chain IDs `chainIds` as a field to the Governor struct. The idea is to provide a stable way to iterate over `chainEntry` in a deterministic way. (Note that iterating over maps is non-deterministic in Go).

This works just fine but is possibly inefficient. Is there a better way to do this?

### Store a slice of pointers to chainEntrys

@bruce-riley recommended that we could modify `chainIds` to a slice of pointers to `chainEntry`s. If we do this, we avoid the need to store the chainIds in a slice and can simply iterate over the pointers. This may be more efficient than iterating over a slice of chainIds and then doing a map lookup for each one.

Pros: likely more efficient
Cons: A developer needs to carefully manage the `slice` and the map in this case instead one of the pointers is modified. (not the data that the pointer is pointing too). There is no mechanism that ensures they stay in sync. 

### Make use of iterators in Go 1.23

In August Go will have support for iterators. We could use this to iterate over sorted keys:

https://pkg.go.dev/iter#hdr-Standard_Library_Usage
```go
for _, key := range slices.Sorted(maps.Keys(m)) {
	...
}
```

Pros: no need to store a second, redundant data structure to get deterministic iteration ordering. No desync risk. Easy to reason about.
Cons: may not be more efficient in terms of execution. Need to wait until the Guardians support Go 1.23

## Definition of done

* Iterate on both approaches and possibly others
* Measure performance of potential approaches
* Evaluate performance gains against potential for mistakes
* Decide whether to replace existing implementation or replace it with a new one

PR #4315, opened 2025-03-26
node: Add Notary package for verifying MessagePublications

johnsaigle:
# Overview

## New Notary package
This PR adds a "Notary" package that can evaluates the verification state of a Message Publication. It provides a Verdict on what should occur, e.g. it reports that an `Anomalous` message should be delayed.

It stores messages that should be Delayed or Blackholed in a database (BadgerDB)

Details are in #4452 

## Changes to the Processor
The processor now inspects the Notary's queue of message to delay and skips publishing them until their ReleaseTime has passed. Once they are ready to release, it passes them to the Governor and Accountant as normal (assuming Guardians have them enabled).

# Design decisions + supporting tasks

## Database

The package uses the BadgerDB key-value store to delay and persist messages, similar to the Governor. This allows the Notary's decisions to persist across restarts.

An implementation is possible that would use only in-memory representations, but this has some drawbacks:
* We would lose messages that are marked as delayed on a node restart
* Messages with a `Rejected` status should never be processed until the end of time. We need a way to track this.

Rather than use the existing `Database` struct (like Governor and Accountant), a new `NotaryDB` struct was added that is a wrapper for the BadgerDB connection. The goal here is to isolate the Notary-related API from other modules.

## Creation of min-heap data structure PendingMessageQueue

Conceptually, it makes sense to store delayed messages according to their release time. If the collection is always sorted, we can access elements from one side or the other until one message is too recent to process. In that case, all of the other messages will also be too new, and we can exit early.

This also allows us to be flexible with our delay time, compared with storing the time _at which_ the message is stored rather than when it should be delayed. We can delay a message for one minute, a day, or a year. As long as the sorting holds, we don't need to iterate over many items.

In order to achieve this, I've added a min-heap data structure with a safe API that allows for storing messages this way.

# Background
Follow-up task on: #4233 

# Future Work

## Add drop/release/reset admin commands

This package should allow a super-minority of Guardians to forcibly drop, release, or reset the delay time of delayed or blackholed messages. This can help when the Notary has delayed an Anomalous message that is in fact fraudulent - in this case, the message can be moved to the 'blackhole' deny-list. Alternatively, if the Transfer Verifier or Notary has a bug, this allows the Guardians to release a harmless message.

## Add Guardian p2p feature flags
Similar to: https://github.com/wormhole-foundation/wormhole/pull/4317

This change will be done in a follow-up PR. 

## Use the priority message queue min-heap for the Governor

The new data structure could be made generic or copied for the types used by the Governor and could grant similar benefits. See also #4029.

### DIFF
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -45,7 +45,7 @@ func (p *PendingMessage) MarshalBinary() ([]byte, error) {
 // UnmarshalBinary implements BinaryUnmarshaler for [PendingMessage].
 func (p *PendingMessage) UnmarshalBinary(data []byte) error {
 
-	if len(data) < marshaledMsgSizeMin {
+	if len(data) < marshaledMsgLenMin {
 		return ErrInputSize{Msg: "data too short"}
 	}
 
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -237,20 +237,3 @@ func (q *PendingMessageQueue) ContainsMessagePublication(msgPub *MessagePublicat
 		return bytes.Equal(pMsg.Msg.MessageID(), msgPub.MessageID())
 	})
 }
-
-// // RangeElements provides a way to iterate over the queue. Because queue is a
-// // min-heap, the last item that can be accessed via Pop will have the earliest timestamp.
-// func (q *PendingMessageQueue) Iter() func(yield func(index int, value *PendingMessage) bool) {
-// 	// implements the range-over function pattern.
-// 	return func(yield func(index int, value *PendingMessage) bool) {
-// 		if q == nil {
-// 			return // Safely handle nil pointers
-// 		}
-//
-// 		for i, v := range q.heap {
-// 			if !yield(i, v) {
-// 				break
-// 			}
-// 		}
-// 	}
-// }
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -5,6 +5,7 @@ import (
 	"cmp"
 	"container/heap"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"slices"
 	"time"
@@ -138,11 +139,11 @@ func NewPendingMessageQueue() *PendingMessageQueue {
 
 // Push adds an element to the heap. If msg is nil, nothing is added.
 // Returns nil if the heap is empty or if the value is not a *[PendingMessage].
-func (q *PendingMessageQueue) Push(msg *PendingMessage) {
-	if msg == nil {
+func (q *PendingMessageQueue) Push(pMsg *PendingMessage) {
+	if pMsg == nil {
 		return
 	}
-	heap.Push(&q.heap, msg)
+	heap.Push(&q.heap, pMsg)
 }
 
 // Pop removes the last element from the heap and returns its value.
@@ -173,7 +174,33 @@ func (q *PendingMessageQueue) Peek() *PendingMessage {
 	return &last
 }
 
-// Contains determines whether the queue contains a [MessagePublication] (not a [PendingMessage]).
+// RemoveItem removes target MessagePublication from the heap. Returns the element that was removed or nil
+// if the item was not found. No error is returned if the item was not found.
+func (q *PendingMessageQueue) RemoveItem(target *MessagePublication) (*PendingMessage, error) {
+	if target == nil {
+		return nil, errors.New("pendingmessage: nil argument for RemoveItem")
+	}
+	if len(target.TxID) == 0 {
+		return nil, errors.New("pendingmessage: argument has empty TxID")
+	}
+
+	var removed *PendingMessage
+	for i, item := range q.heap {
+		// Assumption: TxIDs are unique across MessagePublications.
+		if bytes.Equal(item.Msg.TxID, target.TxID) {
+			pMsg, ok := heap.Remove(&q.heap, i).(*PendingMessage)
+			if !ok {
+				return nil, errors.New("pendingmessage: item removed from heap is not PendingMessage")
+			}
+			removed = pMsg
+			break
+		}
+	}
+
+	return removed, nil
+}
+
+// Contains determines whether the queue contains a [PendingMessage].
 func (q *PendingMessageQueue) Contains(pMsg *PendingMessage) bool {
 	if pMsg == nil {
 		return false
@@ -181,7 +208,7 @@ func (q *PendingMessageQueue) Contains(pMsg *PendingMessage) bool {
 	return slices.Contains(q.heap, pMsg)
 }
 
-// Contains determines whether the queue contains a [MessagePublication] (not a [PendingMessage]).
+// ContainsMessagePublication determines whether the queue contains a [MessagePublication] (not a [PendingMessage]).
 func (q *PendingMessageQueue) ContainsMessagePublication(msgPub *MessagePublication) bool {
 	if msgPub == nil {
 		return false
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -13,6 +13,12 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
+const ( 
+	// marshaledPendingMsgLenMin is the minimum length of a marshaled pending message.
+	// It includes 8 bytes for the timestamp.
+	marshaledPendingMsgLenMin = 8 + marshaledMsgLenMin
+)
+
 // PendingMessage is a wrapper type around a [MessagePublication] that includes the time for which it
 // should be released.
 type PendingMessage struct {
@@ -29,7 +35,7 @@ func (p *PendingMessage) MarshalBinary() ([]byte, error) {
 	buf := new(bytes.Buffer)
 
 	// Compare with [PendingTransfer.Marshal].
-	//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
+	// #nosec G115  -- int64 and uint64 have the same number of bytes, and Unix time won't be negative.
 	vaa.MustWrite(buf, binary.BigEndian, uint64(p.ReleaseTime.Unix()))
 
 	bz, err := p.Msg.MarshalBinary()
@@ -45,13 +51,13 @@ func (p *PendingMessage) MarshalBinary() ([]byte, error) {
 // UnmarshalBinary implements BinaryUnmarshaler for [PendingMessage].
 func (p *PendingMessage) UnmarshalBinary(data []byte) error {
 
-	if len(data) < marshaledMsgLenMin {
-		return ErrInputSize{Msg: "data too short"}
+	if len(data) < marshaledPendingMsgLenMin {
+		return ErrInputSize{Msg: "pending message too short"}
 	}
 
 	// Compare with [UnmarshalPendingTransfer].
 	p.ReleaseTime = time.Unix(
-		//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
+		// #nosec G115  -- int64 and uint64 have the same number of bytes, and Unix time won't be negative.
 		int64(binary.BigEndian.Uint64(data[0:8])),
 		0,
 	)
@@ -76,12 +82,6 @@ func (p *PendingMessage) UnmarshalBinary(data []byte) error {
 // timestamps must be greater. This should allow for constant-time lookups when
 // looking for messages to release.
 //
-// SECURITY:  Only the functions labelled Safe should be called directly. The other
-// functions must be public to satisfy the heap interface but do not perform
-// safety checks.
-// This follows the recommendations for using heap: heap.Pop() and heap.Push()
-// should be used rather than calling the below functions directly, and the
-// Safe functions do this.
 // See: https://pkg.go.dev/container/heap#Interface
 type pendingMessageHeap []*PendingMessage
 
@@ -95,7 +95,7 @@ func (h pendingMessageHeap) Swap(i, j int) {
 	h[i], h[j] = h[j], h[i]
 }
 
-// Push dangerously pushes a value to the heap. Use [pendingMessageHeap.SafePush] instead.
+// Push dangerously pushes a value to the heap. Use [pendingMessageHeap.Push] instead.
 func (h *pendingMessageHeap) Push(x any) {
 	// Push and Pop use pointer receivers because they modify the slice's length,
 	// not just its contents.
@@ -112,7 +112,7 @@ func (h *pendingMessageHeap) Push(x any) {
 	*h = append(*h, item)
 }
 
-// Pops dangerously pops a value from the heap. Use [pendingMessageHeap.SafePop] instead.
+// Pops dangerously pops a value from the heap. Use [pendingMessageHeap.Pop] instead.
 func (h *pendingMessageHeap) Pop() any {
 	old := *h
 	n := len(old)
@@ -138,7 +138,6 @@ func NewPendingMessageQueue() *PendingMessageQueue {
 }
 
 // Push adds an element to the heap. If msg is nil, nothing is added.
-// Returns nil if the heap is empty or if the value is not a *[PendingMessage].
 func (q *PendingMessageQueue) Push(pMsg *PendingMessage) {
 	if pMsg == nil {
 		return
@@ -165,6 +164,7 @@ func (q *PendingMessageQueue) Len() int {
 	return q.heap.Len()
 }
 
+// Peek returns the element at the top of the heap without removing it.
 func (q *PendingMessageQueue) Peek() *PendingMessage {
 	if q.heap.Len() == 0 {
 		return nil
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -31,7 +31,6 @@ func (p *PendingMessage) MarshalBinary() ([]byte, error) {
 	// Compare with [PendingTransfer.Marshal].
 	vaa.MustWrite(buf, binary.BigEndian, uint32(p.ReleaseTime.Unix()))
 
-
 	b, err := p.Msg.MarshalBinary()
 	if err != nil {
 		return buf.Bytes(), fmt.Errorf("failed to marshal pending message: %w", err)
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -13,10 +13,14 @@ import (
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
-const ( 
+const (
 	// marshaledPendingMsgLenMin is the minimum length of a marshaled pending message.
 	// It includes 8 bytes for the timestamp.
 	marshaledPendingMsgLenMin = 8 + marshaledMsgLenMin
+)
+
+var (
+	ErrAlreadyExsits = errors.New("pending message queue: message publication is already queued")
 )
 
 // PendingMessage is a wrapper type around a [MessagePublication] that includes the time for which it
@@ -125,6 +129,7 @@ func (h *pendingMessageHeap) Pop() any {
 }
 
 // PendingMessageQueue is a min-heap that sorts [PendingMessage] in descending order of Timestamp.
+// It also prevents duplicate [MessagePublication]s from being added to the queue.
 type PendingMessageQueue struct {
 	// pendingMessageHeap exposes dangerous APIs as a necessary consequence of implementing [heap.Interface].
 	// Wrap it and expose only a safe API.
@@ -139,10 +144,9 @@ func NewPendingMessageQueue() *PendingMessageQueue {
 
 // Push adds an element to the heap. If msg is nil, nothing is added.
 func (q *PendingMessageQueue) Push(pMsg *PendingMessage) {
-	if pMsg == nil {
-		return
+	if pMsg != nil && !q.ContainsMessagePublication(&pMsg.Msg) {
+		heap.Push(&q.heap, pMsg)
 	}
-	heap.Push(&q.heap, pMsg)
 }
 
 // Pop removes the last element from the heap and returns its value.
@@ -180,14 +184,11 @@ func (q *PendingMessageQueue) RemoveItem(target *MessagePublication) (*PendingMe
 	if target == nil {
 		return nil, errors.New("pendingmessage: nil argument for RemoveItem")
 	}
-	if len(target.TxID) == 0 {
-		return nil, errors.New("pendingmessage: argument has empty TxID")
-	}
 
 	var removed *PendingMessage
 	for i, item := range q.heap {
-		// Assumption: TxIDs are unique across MessagePublications.
-		if bytes.Equal(item.Msg.TxID, target.TxID) {
+		// Assumption: MsgIDs are unique across MessagePublications.
+		if bytes.Equal(item.Msg.MessageID(), target.MessageID()) {
 			pMsg, ok := heap.Remove(&q.heap, i).(*PendingMessage)
 			if !ok {
 				return nil, errors.New("pendingmessage: item removed from heap is not PendingMessage")
@@ -213,8 +214,9 @@ func (q *PendingMessageQueue) ContainsMessagePublication(msgPub *MessagePublicat
 	if msgPub == nil {
 		return false
 	}
+	// Relies on MessageIDString to be unique.
 	return slices.ContainsFunc(q.heap, func(pMsg *PendingMessage) bool {
-		return pMsg.Msg.TxIDString() == msgPub.TxIDString()
+		return pMsg.Msg.MessageIDString() == msgPub.MessageIDString()
 	})
 }
 
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -29,14 +29,15 @@ func (p *PendingMessage) MarshalBinary() ([]byte, error) {
 	buf := new(bytes.Buffer)
 
 	// Compare with [PendingTransfer.Marshal].
-	vaa.MustWrite(buf, binary.BigEndian, uint32(p.ReleaseTime.Unix()))
+	// nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
+	vaa.MustWrite(buf, binary.BigEndian, uint64(p.ReleaseTime.Unix()))
 
-	b, err := p.Msg.MarshalBinary()
+	bz, err := p.Msg.MarshalBinary()
 	if err != nil {
-		return buf.Bytes(), fmt.Errorf("failed to marshal pending message: %w", err)
+		return buf.Bytes(), fmt.Errorf("marshal pending message: %w", err)
 	}
 
-	vaa.MustWrite(buf, binary.BigEndian, b)
+	vaa.MustWrite(buf, binary.BigEndian, bz)
 
 	return buf.Bytes(), nil
 }
@@ -50,14 +51,15 @@ func (p *PendingMessage) UnmarshalBinary(data []byte) error {
 
 	// Compare with [UnmarshalPendingTransfer].
 	p.ReleaseTime = time.Unix(
-		int64(binary.BigEndian.Uint32(data[0:4])),
+		// nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
+		int64(binary.BigEndian.Uint64(data[0:8])),
 		0,
 	)
 
-	err := p.Msg.UnmarshalBinary(data[4:])
+	err := p.Msg.UnmarshalBinary(data[8:])
 
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal pending transfer msg: %w", err)
+		return fmt.Errorf("unmarshal pending message: %w", err)
 	}
 
 	return nil
@@ -167,7 +169,8 @@ func (q *PendingMessageQueue) Peek() *PendingMessage {
 	if q.heap.Len() == 0 {
 		return nil
 	}
-	last := *q.heap[q.heap.Len()-1]
+	// container/heap stores the "next" element at the first offset.
+	last := *q.heap[0]
 	return &last
 }
 
--- a/node/pkg/common/pendingmessage.go
+++ b/node/pkg/common/pendingmessage.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
@@ -17,10 +18,6 @@ const (
 	// marshaledPendingMsgLenMin is the minimum length of a marshaled pending message.
 	// It includes 8 bytes for the timestamp.
 	marshaledPendingMsgLenMin = 8 + marshaledMsgLenMin
-)
-
-var (
-	ErrAlreadyExsits = errors.New("pending message queue: message publication is already queued")
 )
 
 // PendingMessage is a wrapper type around a [MessagePublication] that includes the time for which it
@@ -128,12 +125,13 @@ func (h *pendingMessageHeap) Pop() any {
 	return last
 }
 
-// PendingMessageQueue is a min-heap that sorts [PendingMessage] in descending order of Timestamp.
+// PendingMessageQueue is a thread-safe min-heap that sorts [PendingMessage] in descending order of Timestamp.
 // It also prevents duplicate [MessagePublication]s from being added to the queue.
 type PendingMessageQueue struct {
 	// pendingMessageHeap exposes dangerous APIs as a necessary consequence of implementing [heap.Interface].
 	// Wrap it and expose only a safe API.
 	heap pendingMessageHeap
+	mu   sync.RWMutex
 }
 
 func NewPendingMessageQueue() *PendingMessageQueue {
@@ -144,9 +142,15 @@ func NewPendingMessageQueue() *PendingMessageQueue {
 
 // Push adds an element to the heap. If msg is nil, nothing is added.
 func (q *PendingMessageQueue) Push(pMsg *PendingMessage) {
-	if pMsg != nil && !q.ContainsMessagePublication(&pMsg.Msg) {
-		heap.Push(&q.heap, pMsg)
+	// noop if the message is nil or already in the queue.
+	if pMsg == nil || q.ContainsMessagePublication(&pMsg.Msg) {
+		return
 	}
+
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	heap.Push(&q.heap, pMsg)
 }
 
 // Pop removes the last element from the heap and returns its value.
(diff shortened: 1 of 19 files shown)


mdulin2 (dismissed):

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+
+	"go.uber.org/zap"
+)
+
+type (
+	// Verdict is an enum that reports what action the Notary has taken after processing a message.
+	Verdict uint8
+)
+
+const (
```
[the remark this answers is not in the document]
johnsaigle: I disagree, I think verificaiton state is a status and the verdict is an action. For example, in the future we may want to Blackhole Anomalous messages instead. It's also possible that the Notary could be extended to give different verdicts on different types of messages.

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+func (v Verdict) String() string {
+	switch v {
+	case Approve:
+		return "Approve"
+	case Delay:
+		return "Delay"
+	case Blackhole:
+		return "Blackhole"
+	default:
+		return "Unknown"
```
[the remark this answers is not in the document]
johnsaigle: Yeah good catch!

node/pkg/processor/processor.go
```diff
@@ -312,7 +365,50 @@ func (p *Processor) Run(ctx context.Context) error {
...
+			// in serial in the following order: Notary -> Governor -> Accountant.
+			// NOTE: The Accountant can signal to a channel that it is ready to publish a message via
+			// writing to acctReadC so it is not handled separately here.
+
+			if p.notary != nil {
+				p.notary.ProcessReadyMessages()
+
+				// Iterate over all ready messages. Hand-off to the Governor or the Accountant
+				// if they're enabled. If not, publish.
+				for msg := range slices.Values(p.notary.Ready()) {
```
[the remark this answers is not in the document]
johnsaigle: Yeah I'm a bit nervous to dig into a big refactor of the processor, might want to do that in a follow-up PR instead.

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+}
+
+const (
+	// How long a message should be held in the pending list before being processed.
+	defaultDelay = time.Hour * 24
```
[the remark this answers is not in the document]
johnsaigle: Sure, I just picked a number. It can be longer

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	n.logger.Debug(
+		"notary: begin process ready message",
+		zap.Int("readyCount", len(n.ready)),
+		zap.Int("delayedCount", n.delayed.Len()),
+	)
+	now := time.Now()
+	for n.delayed.Len() != 0 {
+		next := n.delayed.Peek()
```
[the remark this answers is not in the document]
johnsaigle: It'll go from Delayed to Approved automatically when the time runs out, but yeah the admin command will have to handle the case from moving manually from one place to the other.

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+	// state.
+	//
+	// In particular, the following invariants must hold:
+	// - When a message is released, it must be deleted from the database.
+
+	delayed *common.PendingMessageQueue
+	// ready contains message publications that have been delayed but are now ready to release.
+	ready []*common.MessagePublication
+	// All of the messages that have been black-holed due to being rejected by the Transfer Verifier.
+	blackholed []*common.MessagePublication
```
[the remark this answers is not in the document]
johnsaigle: Yeah this makes sense for the Anomalous case where basically the message will be eventually Blackholed or Approved. The notary should query the blackholed list and mark the incoming message as blackholed if it already exists there.

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+
+	return nil
+}
+
+func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err error) {
+
+	n.logger.Debug("notary: processing message", msg.ZapFields()...)
+
+	// Only token transfers are currently supported.
+	if !vaa.IsTransfer(msg.Payload) {
```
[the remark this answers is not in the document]
johnsaigle: Yeah it does, I was thinking this could be an extra guard in case the Transfer Verifier logic gets changed but the Notary gets forgotten.

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,428 @@
...
+
+	switch msg.VerificationState() {
+	case common.Anomalous:
+		err = n.delay(msg, defaultDelay)
+		v = Delay
+	case common.Rejected:
+		err = n.blackhole(msg)
+		v = Blackhole
+	case common.NotVerified, common.NotApplicable, common.Valid:
+		// NOTE: All other statuses are simply approved for now. In the future, it may be
```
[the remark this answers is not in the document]
johnsaigle: What do you mean by bail? Do you want the function to return an error? My idea here was to fail open because ultimately we only care about Anomalous and Rejected right now.

node/pkg/notary/notary.go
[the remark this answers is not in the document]
johnsaigle: I've added a RemoveItem function to the heap and added unit tests to show that it preserves the ordering of elements even if we remove an arbitrary one.

johnsaigle:
Converting to draft because the behaviour of the Transfer Verifier is changing to add a new Verification State that will be used for the EVM implementation. Given this, we may want to change how the Processor responds to those messages.

https://github.com/wormhole-foundation/wormhole/pull/4362

node/pkg/processor/processor.go
```diff
@@ -286,17 +292,73 @@ func (p *Processor) Run(ctx context.Context) error {
 			)
 			p.gst.Set(p.gs)
 		case k := <-p.msgC:
+			// This is the main message processing loop. It is responsible for handling messages that are
+			// received on the message channel. Depending on the configuration, a message may be processed
+			// by the Notary, the Governor, and/or the Accountant.
+			// This loop effectively causes each of these components to be process messages in a modular
```
djb15:
```suggestion
			// This loop effectively causes each of these components to process messages in a modular
```

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,236 @@
...
+func (p PendingMessage) Compare(other PendingMessage) int {
+	return cmp.Compare(p.ReleaseTime.Unix(), other.ReleaseTime.Unix())
+}
+
+// MarshalBinary implements BinaryMarshaler for [PendingMessage].
+func (p *PendingMessage) MarshalBinary() ([]byte, error) {
+	buf := new(bytes.Buffer)
+
+	// Compare with [PendingTransfer.Marshal].
+	//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
```
djb15: `nosec` pls 🙏 Like the pattern of ignoring as little as possible

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,236 @@
...
+
+	vaa.MustWrite(buf, binary.BigEndian, bz)
+
+	return buf.Bytes(), nil
+}
+
+// UnmarshalBinary implements BinaryUnmarshaler for [PendingMessage].
+func (p *PendingMessage) UnmarshalBinary(data []byte) error {
+
+	if len(data) < marshaledMsgLenMin {
```
djb15: Should add 8 to account for the release time?

```
len(data) < marshaledMsgLenMin + 8
```

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,236 @@
...
+// timestamps must be greater. This should allow for constant-time lookups when
+// looking for messages to release.
+//
+// SECURITY:  Only the functions labelled Safe should be called directly. The other
+// functions must be public to satisfy the heap interface but do not perform
+// safety checks.
+// This follows the recommendations for using heap: heap.Pop() and heap.Push()
+// should be used rather than calling the below functions directly, and the
+// Safe functions do this.
+// See: https://pkg.go.dev/container/heap#Interface
```
djb15: I can't find any `SafePush` functions? Is this out of date?

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func NewPendingMessageQueue() *PendingMessageQueue {
+	q := &PendingMessageQueue{heap: pendingMessageHeap{}}
+	heap.Init(&q.heap)
+	return q
+}
+
+// Push adds an element to the heap. If msg is nil, nothing is added.
+// Returns nil if the heap is empty or if the value is not a *[PendingMessage].
```
djb15: There is no return value for this function

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Pops dangerously pops a value from the heap. Use [pendingMessageHeap.SafePop] instead.
+func (h *pendingMessageHeap) Pop() any {
+	old := *h
+	n := len(old)
+	if n == 0 {
+		panic("PendingMessageHeap: cannot Pop from empty heap")
+	}
+	last := old[n-1]
```
djb15, line 123: Can you zero out `old[n-1]` after you read from it please?

node/pkg/db/notary.go
```diff
@@ -0,0 +1,235 @@
+// TODO explain the meaning of pending and blackholed.
```
djb15: Can you remove this todo now that the whitepaper exists?

node/pkg/db/notary.go
```diff
@@ -0,0 +1,235 @@
...
+
+	key := delayKey(p)
+	if updateErr := d.update(key, b); updateErr != nil {
+		return &DBError{Op: OpUpdate, Key: key, Err: updateErr}
+	}
+
+	return nil
+}
+
+func (d *NotaryDB) StoreBlackhole(m *common.MessagePublication) error {
```
djb15:
```suggestion
func (d *NotaryDB) StoreBlackholed(m *common.MessagePublication) error {
```

node/pkg/db/notary.go
```diff
@@ -0,0 +1,235 @@
...
+						ErrUnmarshal,
+						unmarshalErr,
+					)
+				}
+				result.Delayed = append(result.Delayed, &pMsg)
+			case Unknown:
+				// Should only occur in case of an error.
+			default:
+				// The key-value store is shared across other modules and message types (e.g. Governor, Accountant).
+				// If another key is discovered, just ignore it.
```
djb15, line 163: If the key-value store is shared, doesn't that mean that all other keys will be of type `Unknown`? And we'll never fall into the `default` case?

node/pkg/db/notary.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+type NotaryLoadResult struct {
+	Delayed    []*common.PendingMessage
+	Blackholed []*common.MessagePublication
+}
+
+// LoadAll retrieves all keys from the database.
+func (d *NotaryDB) LoadAll(logger *zap.Logger) (*NotaryLoadResult, error) {
+	const defaultResultCapacity = 10
```
djb15: What's the idea behind this default capacity? Isn't it likely we're going to exceed this and have to do a reallocation anyway?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,483 @@
...
+
+	return nil
+}
+
+func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err error) {
+
+	n.logger.Debug("notary: processing message", msg.ZapFields()...)
+
+	// Only token transfers are currently supported.
+	if !vaa.IsTransfer(msg.Payload) {
```
djb15, line 163: Should we also check here that the sender is actually a token bridge?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	switch msg.VerificationState() {
+	case common.Anomalous:
+		err = n.delay(msg, DelayFor)
+		v = Delay
+	case common.Rejected:
+		err = n.blackhole(msg)
+		v = Blackhole
+	case common.CouldNotVerify, common.NotVerified, common.NotApplicable, common.Valid:
```
djb15: Worth splitting these into 2 sets and adding some debug logging for testnet?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,483 @@
...
+			continue // Skip if Pop returns nil (shouldn't happen if Peek worked)
+		}
+
+		// Append return value.
+		readyMsgs = append(readyMsgs, &pMsg.Msg)
+
+		// Update database.
+		err := n.database.DeleteDelayed(pMsg)
+		if err != nil {
+			n.logger.Error("delete pending message from notary database", pMsg.Msg.ZapFields(zap.Error(err))...)
```
djb15, line 251: If this delete failed, the next time we release ready messages, the same message will be included again. Haven't checked yet myself, but would this have any downstream side effects of double processing?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,483 @@
...
+	// message equality before and after loading to the database will fail.
+	release := time.Unix(time.Now().Unix(), 0)
+
+	pMsg := &common.PendingMessage{
+		Msg:         *msg,
+		ReleaseTime: release.Add(dur),
+	}
+
+	// Store in in-memory slice. This should happen even if a database error occurs.
+	n.delayed.Push(pMsg)
```
djb15, line 304: Should we be checking the database/heap first to see if the message has already been delayed? For example, we don't want the same message to get delayed twice, where one then gets manually blackholed, and then the other gets released as normal after the delay, effectively circumventing the blackhole

node/pkg/processor/processor.go
```diff
@@ -286,17 +292,73 @@ func (p *Processor) Run(ctx context.Context) error {
...
+						p.logger.Error("message will not be processed", k.ZapFields(zap.String("verdict", verdict.String()))...)
+					} else {
+						// Delayed messages are added to a separate queue and processed elsewhere.
+						p.logger.Error("message will be delayed", k.ZapFields(zap.String("verdict", verdict.String()))...)
+					}
+					// We're done processing the message.
+					continue
+				case guardianNotary.Unknown:
+					p.logger.Error("notary returned Unknown verdict", k.ZapFields(zap.String("verdict", verdict.String()))...)
+				case guardianNotary.Approve:
```
djb15, line 336: Also add a default case that logs an error message? We shouldn't hit the default case so would be good to know if that happens

node/pkg/processor/processor.go
```diff
@@ -318,7 +380,49 @@ func (p *Processor) Run(ctx context.Context) error {
...
+							continue
+						}
+					}
+
+					// Hand-off to accountant. If we get here, both the Notary and the Governor
+					// have signalled that the message is OK to publish.
+					if p.acct != nil {
+						shouldPub, err := p.acct.SubmitObservation(msg)
+						if err != nil {
+							return fmt.Errorf("accountant: failed to process message `%s`: %w", msg.MessageIDString(), err)
```
djb15, line 412: Do we want to return here? It'll mean we never get to the Governor pending messages?

node/pkg/common/pendingmessage.go
johnsaigle: Thanks yeah this was authored before you left the similar comment on #4428 and i didn't come back and also apply that suggestion here. I'll fix this

node/pkg/common/pendingmessage.go
johnsaigle: It must be, I'll take a look and rework this comment. thanks

node/pkg/common/pendingmessage.go
johnsaigle: what's the concern here?

node/pkg/db/notary.go
johnsaigle: I can delete it if you like. It probably doesn't matter much because this will only run once on startup anyway

node/pkg/db/notary.go
johnsaigle: Yeah I think I may have added two different branches to satisfy the `exhaustruct` lint but I can try to consolidate them

node/pkg/processor/processor.go
johnsaigle: This should mimic the existing logic for Accountant errors.
The Governor is already processed just above in this branch. The governor-related code below is what runs if the Notary is not enabled.
I don't like how this code is structured (hence the TODO above) but I didn't want to mess around with the processor logic too much.

node/pkg/notary/notary.go
johnsaigle: This iterates over and consumes elements from the in-memory queue rather than the database. So, that situation would only happen when the Guardian restarts and it loads the delayed messages from the DB into the queue.

In general, if a message is processed more than once it would need to flow through the Governor and the Accountant successfully in order to be processed into a VAA. If a VAA has already been produced for that message, then making it all the way through those checks should result in a duplicated VAA.

We should definitely think through any new edge-cases that occur here, but with the caveat that the Governor and the Guardians themselves already do not have a permanent record of already-processed messages or VAAs, so I think this new code doesn't change that model.

node/pkg/notary/notary.go
johnsaigle: In the latest commit I added a uniqueness check in the queue's Push method so that duplicate MessagePublications can't be added.

node/pkg/common/pendingmessage.go
johnsaigle: OK so in an earlier phase of this I was going to write a SafePush and SafePop in order to avoid the panics in the normal Push and Pop that you can see here.
Instead the approach I took is to make all the relevant methods private and wrap them in the PendingMessageQueue struct. That struct's Push and Pop methods do the relevant checks before calling the unsafe methods.

I updated the comment.

node/pkg/notary/notary.go
johnsaigle: Done, let me know what you think

node/pkg/processor/processor.go
```diff
@@ -318,7 +383,49 @@ func (p *Processor) Run(ctx context.Context) error {
...
+			// When more than one of these features are enabled, messages should be processed
+			// serially in the order: Notary -> Governor -> Accountant.
+			// NOTE: The Accountant can signal to a channel that it is ready to publish a message via
+			// writing to acctReadC so it is not handled here.
+			if p.notary != nil {
+				readyMsgs := p.notary.ReleaseReadyMessages()
+
+				// Iterate over all ready messages. Hand-off to the Governor or the Accountant
+				// if they're enabled. If not, publish.
+				for msg := range slices.Values(readyMsgs) {
```
djb15: My IDE shows an error on this line fyi: `cannot range over slices.Values(readyMsgs)`. Just flagging as I'm not sure if this flow is tested or if this is just a golang version specific thing

node/pkg/processor/processor.go
djb15: Gave this another look and I agree on both counts, I think what you've added is correct + the structure is a bit weird

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,238 @@
...
+	return len(h)
+}
+func (h pendingMessageHeap) Less(i, j int) bool {
+	return h[i].ReleaseTime.Before(h[j].ReleaseTime)
+}
+func (h pendingMessageHeap) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+}
+
+// Push dangerously pushes a value to the heap. Use [pendingMessageHeap.Push] instead.
```
djb15: Assume this comment and the one below should be referring to `pendingMessageQueue`?

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// ContainsMessagePublication determines whether the queue contains a [MessagePublication] (not a [PendingMessage]).
+func (q *PendingMessageQueue) ContainsMessagePublication(msgPub *MessagePublication) bool {
+	if msgPub == nil {
+		return false
+	}
+	// Relies on MessageIDString to be unique.
+	return slices.ContainsFunc(q.heap, func(pMsg *PendingMessage) bool {
+		return pMsg.Msg.MessageIDString() == msgPub.MessageIDString()
```
djb15: Any reason why you're doing a string comparison here vs a byte comparison above in `RemoveItem`?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,493 @@
...
+	n.logger.Info("notary ready")
+
+	return nil
+}
+
+func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err error) {
+
+	n.logger.Debug("notary: processing message", msg.ZapFields()...)
+
+	if tokenBridge, ok := sdk.KnownTokenbridgeEmitters[msg.EmitterChain]; !ok {
```
djb15, line 168: I think you need to move this first check down a little. First we should let non-token transfers through, and only if the message is a token transfer check then token bridge for that chain. Some chains don't have the token bridge deployed, so the current behaviour would make all messages from those chains "Unknown".

i.e. if not transfer -> approve, if transfer check if the emitter is the token bridge

node/pkg/notary/notary.go
djb15: Thinking some more about this, does it make sense to attempt the delete from database first before appending to `readyMsgs`? That way we don't have any possibility of double processing

node/pkg/common/pendingmessage.go
djb15: Yeah fair enough, it's pretty hard to properly zero out memory with Go so lets just let the garbage collector deal with it

node/pkg/processor/processor.go
johnsaigle: Are you able to see what Go version your IDE is using locally, as well as the LSP info (likely `gopls`)? I'll look into it in the meantime but it would be good to track version differences to help sort out the difference. I also have in-line hints enabled but I don't think this raised a problem.

node/pkg/common/pendingmessage.go
johnsaigle: At one point I changed the comparison to work over MessageIDString instead of TxID, so one of them is likely left over from that. I'll change both to do a byte comparison.

node/pkg/notary/notary.go
johnsaigle: Sure I can make that change

node/pkg/common/pendingmessage.go
johnsaigle: I deleted the second sentence in both comments in the latest commit.

node/pkg/processor/processor.go
djb15: Using `1.24.2` locally

johnsaigle: I'm on `1.24.6`. Weird.

Anyway, it's easy enough to use the other syntax form so I'll just do that so we don't need to worry.

Also see: https://pkg.go.dev/slices#Values. I'd be surprised if this is an issue because I think this is typical usage for the function. It was added in Go 1.23.0 which is why I was asking about versioning.

node/pkg/notary/notary.go
```diff
@@ -155,12 +155,21 @@ func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err erro
...
+		// Return Unknown if the token bridge is not registered in the SDK.
 		n.logger.Error("notary: unknown token bridge emitter", msg.ZapFields()...)
 		return Unknown, errors.New("unknown token bridge emitter")
 	} else {
-		// Only token transfers are currently supported and the transfer must be from the token bridge.
-		if !vaa.IsTransfer(msg.Payload) || !bytes.Equal(msg.EmitterAddress.Bytes(), tokenBridge) {
+		// Approve if the token transfer is not from the token bridge.
+		// For now, the notary only rules on token transfers from the token bridge.
+		if !bytes.Equal(msg.EmitterAddress.Bytes(), tokenBridge) {
+			n.logger.Debug("notary: automatically approving message publication because it is not from the token bridge", msg.ZapFields()...)
```
djb15, line 176: Is it possible to have a transfer type VAA that isn't from the token bridge? Should this be anomalous/unknown instead?

djb15 (commented):
Last comment, looks good otherwise. Thanks for all the changes

node/pkg/notary/notary.go
johnsaigle: https://github.com/wormhole-foundation/wormhole/blob/6e191a01f7277444d8d00908b5ac801ae81876df/sdk/vaa/structs.go#L722-L725

All it does it check the first byte of the payload, and it's annotated to show that the caller is responsible for checking the rest.

So the relationship is: All token transfers have the first byte as 0x01 or 0x03, but not all payloads beginning with those bytes are token transfers.

I think it's fine to let these through for now as the Notary is essentially only used to act on the Transfer Verifier. If we wanted to extend Notary to do more in the future, we could be more strict about this. Right now though I think that it's more likely to cause problems than to solve them.

What do you think?

djb15: That's fair, probably better to err on the side of being more permissive to start with

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+	// Verdict is an enum that reports what action the Notary has taken after processing a message.
+	Verdict uint8
+)
+
+const (
+	Unknown Verdict = iota
+	// Approve means a message should be processed normally. All messages that are not Token Transfers
+	// must always be Approve, as the Notary does not support other messasge types.
+	Approve
+	// Approved means a message should be temporarily delayed so that it can be manually inspected.
```
bemic: Typo: "Approved" -> "Delay"

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	n.mutex.Lock()
+
+	// Store in in-memory slice. This should happen even if a database error occurs.
+	n.blackholed.Add(msg.VAAHash())
+
+	// Store in database.
+	dbErr := n.database.StoreBlackholed(msg)
+	if dbErr != nil {
+		return dbErr
```
bemic, line 336: If we exit the function here, we do not call `n.mutex.Unlock()`.
Is it deadlock?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+		// Update database. Do this before adding the message to the ready list so that we don't
+		// accidentally add the same message twice if deleting the message from the database fails.
+		err := n.database.DeleteDelayed(pMsg)
+		if err != nil {
+			n.logger.Error("delete pending message from notary database", pMsg.Msg.ZapFields(zap.Error(err))...)
+			continue
+		}
+
+		// Append return value.
+		readyMsgs = append(readyMsgs, &pMsg.Msg)
```
bemic, line 264: I think there should be a check like: `IsBlackholed(&pMsg.Msg)` **before** this append.

There is a small time/race window where Msg can possibly be saved both as `blackholed` and `ready` after it was delayed. It is possible only after ReleaseTime which makes the likelihood fairly small but still.

The window is here in blackhole() function

```
	// Unlock mutex before calling removeDelayed, which also acquires the mutex.
	n.mutex.Unlock()

	// When a message is blackholed, it should be removed from the delayed list and database.
	err := n.removeDelayed(msg)
```
Goroutine A calling `blackhole()`
Goroutine B calling `ReleaseReadyMessages()` 

A: Saves blackholed msg in db -> unlocks mutex
B: `DeleteDelayed` msg -> append msg ready (THIS IS IN THE WINDOW)
A: entering `removeDelayed` -> msg already deleted

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+			return Approve, nil
+		}
+	}
+
+	// Return early if the message has already been blackholed. This is important in case a message
+	// is reobserved or otherwise processed here more than once. An Anomalous message that becomes
+	// delayed and later blackholed should not be able to be re-added to the Delayed queue.
+	if n.blackholed.Contains(msg.VAAHash()) {
+		n.logger.Warn("notary: got message publication that is already blackholed",
+			msg.ZapFields(zap.String("verdict", v.String()))...,
```
bemic: `v` is empty at this point and will return `unknown`. We can probably just assign it in the block `v:=Blackhole` ?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+		ReleaseTime: release.Add(dur),
+	}
+
+	// Store in in-memory slice. This should happen even if a database error occurs.
+	n.delayed.Push(pMsg)
+
+	// Store in database.
+	dbErr := n.database.StoreDelayed(pMsg)
+	if dbErr != nil {
+		return dbErr
```
bemic, line 309: Here the DB store failed but `n.delayed.Push(pMsg)` was already performed. Should we move the Push after this check to make sure the state is consistent?

node/pkg/notary/notary.go
```diff
@@ -0,0 +1,509 @@
...
+		if !bytes.Equal(msg.EmitterAddress.Bytes(), tokenBridge) {
+			n.logger.Debug("notary: automatically approving message publication because it is not from the token bridge", msg.ZapFields()...)
+			return Approve, nil
+		}
+	}
+
+	// Return early if the message has already been blackholed. This is important in case a message
+	// is reobserved or otherwise processed here more than once. An Anomalous message that becomes
+	// delayed and later blackholed should not be able to be re-added to the Delayed queue.
+	if n.blackholed.Contains(msg.VAAHash()) {
```
bemic: We can maybe use `IsBlackholed()`. It also has mutex inside to avoid potential error/panic in case of `blackhole()` function accessing the same memory.

node/pkg/notary/notary.go
bemic: *In the comment

node/pkg/notary/notary.go
johnsaigle: Hm good question, maybe it would be safer to call Unlock here too

node/pkg/notary/notary.go
johnsaigle: Sounds good, I'll add the check

node/pkg/notary/notary.go
johnsaigle: This is just for logging and the actual Verdict is returned explicitly, so I'll just change this to `Blackhole.String()`

node/pkg/notary/notary.go
johnsaigle: Hm I think that I'd favour including the message in the in-memory list even if it doesn't get written to the DB. That way we still delay weird messages via the processor code.

The risk of the transfer existing only in-memory in the DB is that it would not persist if the node restarted. But this is true anyway in case storing the message in the DB fails. 

What do you think? I can add a comment to explain this.

In theory it would be possible to make this more robust by e.g. including a loop that checks to see if any of the in-memory transfers are missing a DB entry (or vice-versa) and trying to fix that. I feel like we could consider that as an addition in a future PR though.

djb15 (approved):
Happy with all the changes following my last review

node/pkg/notary/notary.go
bemic: > Hm I think that I'd favour including the message in the in-memory list even if it doesn't get written to the DB. That way we still delay weird messages via the processor code.

This is a good point. I agree with you.

node/pkg/notary/notary.go
johnsaigle: Added an Unlock call in the error path

bemic (approved):
Everything looks good to me.

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,238 @@
...
+)
+
+const (
+	// marshaledPendingMsgLenMin is the minimum length of a marshaled pending message.
+	// It includes 8 bytes for the timestamp.
+	marshaledPendingMsgLenMin = 8 + marshaledMsgLenMin
+)
+
+var (
+	ErrAlreadyExsits = errors.New("pending message queue: message publication is already queued")
```
panoel:
```suggestion
	ErrAlreadyExists = errors.New("pending message queue: message publication is already queued")
```

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+)
+
+const (
+	// marshaledPendingMsgLenMin is the minimum length of a marshaled pending message.
+	// It includes 8 bytes for the timestamp.
+	marshaledPendingMsgLenMin = 8 + marshaledMsgLenMin
+)
+
+var (
```
panoel: Is this used anywhere?

node/pkg/common/pendingmessage.go
```diff
@@ -0,0 +1,238 @@
...
+	err := p.Msg.UnmarshalBinary(data[8:])
+
+	if err != nil {
+		return fmt.Errorf("unmarshal pending message: %w", err)
+	}
+
+	return nil
+}
+
+// A pendingMessageHeap is a min-heap of [PendingMessage] and uses the heap interface
```
panoel, line 75: Does the heap need to be thread safe?

node/pkg/db/notary.go
```diff
@@ -0,0 +1,231 @@
+// SECURITY: The calling code is responsible for handling mutex operations when
+// working with this package.
+package db
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/certusone/wormhole/node/pkg/common"
```
panoel, line 10: Is certusone still a thing to reference?

node/pkg/db/notary_test.go
```diff
@@ -0,0 +1,105 @@
+package db
+
+import (
+	"encoding/hex"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/certusone/wormhole/node/pkg/common"
```
panoel, line 10: Same comment about certusone

node/pkg/common/pendingmessage.go
johnsaigle: Nope, leftover from an earlier version. Thanks!

node/pkg/common/pendingmessage.go
johnsaigle: That's a good question. I added a mutex for the Notary generally but maybe an internal one could be used here for the main operations. That way the structure can manage its own state in a thread-safe way instead of relying on the module to call it.

node/pkg/db/notary.go
johnsaigle: That's still how the top-level `go.mod` file is defined for `node/`, so I believe so. Let me know if you think differently.

panoel (approved):

node/pkg/common/pendingmessage.go
```diff
@@ -214,25 +228,29 @@ func (q *PendingMessageQueue) ContainsMessagePublication(msgPub *MessagePublicat
...
+// 			return // Safely handle nil pointers
+// 		}
+//
+// 		for i, v := range q.heap {
+// 			if !yield(i, v) {
+// 				break
+// 			}
+// 		}
+// 	}
+// }
```
djb15: If this is unused please remove rather than commenting out

johnsaigle: Oops thanks, I forgot about this

PR #4317, opened 2025-03-27
node: Add Transfer Verifier to guardian p2p features list

johnsaigle:
Adds a label for Guardian p2p features that signals when a Transfer Verifier implementation is enabled, and for what chains.


bruce-riley:
I think it would be better to format this feature string once on start up and pass that into p2p as a string, rather than building it on every heartbeat.

Also, you can't use commas in your string because that is what separates feature fields. Use "|" instead.

bruce-riley:
Hey, I think the change I made in PR #4332 (after you'd already created this PR) would make this much easier. I think you'd only need to change node.go to create your feature string and then append it to `featureFlags`. Something like [this](https://github.com/wormhole-foundation/wormhole/blob/7349f6ad7ed4d49cb2867fceea88da21c73ad325/node/cmd/guardiand/node.go#L1779-L1781). You should not need to touch any other files.

johnsaigle:
Converting this to draft until I can look into Bruce's comments. I might close this and open a different PR with that solution.

PR #4318, opened 2025-03-27
Node: Alternate publishers

bruce-riley:
Please see the file README.md for an overview of this feature.

Note that this PR includes moving `url_verification.go` from `node/pkg/node` to `node/pkg/common` and updating all references to it. This was necessary so the `altpub` package could use `ValidateURL` without causing a circular reference.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -323,76 +323,76 @@ func init() {
 	fogoContract = NodeCmd.Flags().String("fogoContract", "", "Address of the Fogo program (required if fogoRpc is specified)")
 	fogoShimContract = NodeCmd.Flags().String("fogoShimContract", "", "Address of the Fogo shim program")
 
-	ethRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "ethRPC", "Ethereum RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	ethRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "ethRPC", "Ethereum RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	ethContract = NodeCmd.Flags().String("ethContract", "", "Ethereum contract address")
 
-	bscRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "bscRPC", "Binance Smart Chain RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	bscRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "bscRPC", "Binance Smart Chain RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	bscContract = NodeCmd.Flags().String("bscContract", "", "Binance Smart Chain contract address")
 
-	polygonRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "polygonRPC", "Polygon RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	polygonRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "polygonRPC", "Polygon RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	polygonContract = NodeCmd.Flags().String("polygonContract", "", "Polygon contract address")
 
-	avalancheRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "avalancheRPC", "Avalanche RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	avalancheRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "avalancheRPC", "Avalanche RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	avalancheContract = NodeCmd.Flags().String("avalancheContract", "", "Avalanche contract address")
 
-	oasisRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "oasisRPC", "Oasis RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	oasisRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "oasisRPC", "Oasis RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	oasisContract = NodeCmd.Flags().String("oasisContract", "", "Oasis contract address")
 
-	fantomRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "fantomRPC", "Fantom Websocket RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	fantomRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "fantomRPC", "Fantom Websocket RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	fantomContract = NodeCmd.Flags().String("fantomContract", "", "Fantom contract address")
 
-	karuraRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "karuraRPC", "Karura RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	karuraRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "karuraRPC", "Karura RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	karuraContract = NodeCmd.Flags().String("karuraContract", "", "Karura contract address")
 
-	acalaRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "acalaRPC", "Acala RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	acalaRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "acalaRPC", "Acala RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	acalaContract = NodeCmd.Flags().String("acalaContract", "", "Acala contract address")
 
-	klaytnRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "klaytnRPC", "Klaytn RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	klaytnRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "klaytnRPC", "Klaytn RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	klaytnContract = NodeCmd.Flags().String("klaytnContract", "", "Klaytn contract address")
 
-	celoRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "celoRPC", "Celo RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	celoRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "celoRPC", "Celo RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
 	celoContract = NodeCmd.Flags().String("celoContract", "", "Celo contract address")
 
-	moonbeamRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "moonbeamRPC", "Moonbeam RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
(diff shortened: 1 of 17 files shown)


node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+	case err := <-errC:
+		return err
+	}
+}
+
+// PublishObservation publishes a signed observation to all the endpoints that care about it. It handles both immediate and delayed publishers.
+func (ap *AlternatePublisher) PublishObservation(emitterChain vaa.ChainID, obs *gossipv1.Observation) {
+	var data []byte
+	for _, ep := range ap.endpoints {
+		if ep.shouldPublish(emitterChain) {
```
johnsaigle, line 359: If this was modified to `if !ep.shouldPublish(emitterChain) { continue }`, a level of indentation could be removed for the rest of the code.

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+}
+
+// httpWorker is the entrypoint for an HTTP worker. Each httpWorker is responsible for posting an HTTP request and waiting for the response.
+func (w *HttpWorker) httpWorker(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case req := <-w.ap.httpWorkerChan:
```
johnsaigle, line 425: Should this part have error handling or a default case?

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+	channelDelay.WithLabelValues(req.ep.label).Observe(float64(time.Since(req.start).Microseconds()))
+
+	// Create the HTTP POST request using our context so it can be interrupted.
+	// Note that we are not using a timeout context because the HTTP client already has a timeout.
+	// Note that we make a copy of the payload because `bytes.NewBuffer` takes ownership of the data, and the same data might be in multiple requests.
+	r, err := http.NewRequestWithContext(ctx, "POST", req.url, bytes.NewBuffer(bytes.Clone(req.data)))
+	if err != nil {
+		ap.logger.Error("Failed to create post request", zap.Error(err))
+		requestFailed.WithLabelValues(req.ep.label, "create_failed").Inc()
+		return
```
johnsaigle, line 442: Why not return the errors in this function?

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+
+	if len(observations) != 0 {
+		batch := gossipv1.SignedObservationBatch{
+			Addr:         w.ap.guardianAddr,
+			Observations: observations,
+		}
+
+		data, err := proto.Marshal((&batch))
+		if err != nil {
+			panic("failed to marshal batch")
```
johnsaigle, line 522: Same as above, I think normal error handling is better here.

node/pkg/altpub/alternate_pub_test.go
```diff
@@ -0,0 +1,279 @@
...
+		{label: "With_delay_and_one_chain_name", input: "test;http://localhost:3333;200us;solana", delayUs: 200, chains: []vaa.ChainID{vaa.ChainIDSolana}},
+		{label: "With_delay_and_two_chain_numbers", input: "test;http://localhost:3333;500ms;1,2", delayUs: 500000, chains: []vaa.ChainID{vaa.ChainIDSolana, vaa.ChainIDEthereum}},
+		{label: "With_delay_and_chain_number_and_name", input: "test;http://localhost:3333;500ms;1,ethereum", delayUs: 500000, chains: []vaa.ChainID{vaa.ChainIDSolana, vaa.ChainIDEthereum}},
+		{label: "With_no_delay_and_one_chain", input: "test;http://localhost:3333;0;1", chains: []vaa.ChainID{vaa.ChainIDSolana}},
+		{label: "With_no_delay_and_except_one_chain", input: "test;http://localhost:3333;0;-pythnet", chains: []vaa.ChainID{vaa.ChainIDPythNet}, exceptFor: true},
+		{label: "With_no_delay_and_except_two_chains", input: "test;http://localhost:3333;0;-pythnet,1", chains: []vaa.ChainID{vaa.ChainIDSolana, vaa.ChainIDPythNet}, exceptFor: true},
+
+		// Error cases
+		{label: "Empty", input: "", errText: "not enough fields"},
```
johnsaigle, line 89: It could be good to add a length == 3 case here, e.g. input "wormholescan;http://localhost:3333;0". This is in reference to the README where specifying a delay means we need to specify a chain.

node/pkg/processor/message.go
```diff
@@ -93,7 +93,7 @@ func (p *Processor) handleMessage(ctx context.Context, k *common.MessagePublicat
 	}
 
 	// Broadcast the signature.
-	ourObs, msg := p.broadcastSignature(v.MessageID(), k.TxID, digest, signature, shouldPublishImmediately)
+	ourObs, msg := p.broadcastSignature(v.MessageID(), k.TxID, digest, signature, shouldPublishImmediately, k.EmitterChain)
```
johnsaigle, line 96: Thoughts on just passing `k` (MessagePublication) instead of `k.TxID` and `k.EmitterChain`?

node/cmd/guardiand/node.go
```diff
@@ -311,76 +314,76 @@ func init() {
 	solanaContract = NodeCmd.Flags().String("solanaContract", "", "Address of the Solana program (required if solanaRpc is specified)")
 	solanaShimContract = NodeCmd.Flags().String("solanaShimContract", "", "Address of the Solana shim program")
 
-	ethRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "ethRPC", "Ethereum RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
+	ethRPC = common.RegisterFlagWithValidationOrFail(NodeCmd, "ethRPC", "Ethereum RPC URL", "ws://eth-devnet:8545", []string{"ws", "wss"})
```
johnsaigle, line 317: Moving the URL validation to `common` makes sense, but I think the flag validation functions should stay in `node` as they're not going to be used by other parts of the code.

johnsaigle (changes requested):
Overall the PR looks good, and the documentation and tests are very clear.

I left some comments about potential error-handling and restructuring. There are also some style notes that can be ignored if you want.

johnsaigle:
Oh one more thing -- should this PR include p2p options so that it's possible to see when Guardians have enabled this feature?

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+		guardianAddr:   guardianAddr,
+		endpoints:      endpoints,
+		httpWorkerChan: make(chan *HttpRequest, PubChanSize),
+	}, nil
+}
+
+// parseEndpoint parses an `--additionalPublishEndpoint` parameter into an Endpoint object. It returns an error if the string is invalid.
+func parseEndpoint(config string) (*Endpoint, error) {
+	fields := strings.Split(config, ";")
+	if len(fields) < 2 {
```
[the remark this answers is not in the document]
bruce-riley: That is not correct. You can have three fields (a delay with no chains, which might be the Wormholescan case). There is a test case for this.

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,534 @@
...
+				if data == nil {
+					batch := gossipv1.SignedObservationBatch{
+						Addr:         ap.guardianAddr,
+						Observations: []*gossipv1.Observation{obs},
+					}
+
+					var err error
+					data, err = proto.Marshal((&batch))
+					if err != nil {
+						panic("failed to marshal batch")
```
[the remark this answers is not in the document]
bruce-riley: I'm not sure I agree with this. There are numerous places in the guardian that panic in cases like this (and actually this exact case). The function `proto.Marshal` **should not** fail, but it does return an error, so we need to check it. I don't think it's appropriate to return an error to the processor (which calls this) since it shouldn't need to know or care if the alternate publisher encounters an error.

I think whether or not the guardian code should contain panics is a valid thing to discuss outside of this PR, but for now, I am following the established pattern.

node/pkg/altpub/alternate_pub.go
bruce-riley: Same comment as above.

node/pkg/altpub/alternate_pub.go
bruce-riley: Added error handling. Not sure what a default would do.

node/pkg/altpub/alternate_pub.go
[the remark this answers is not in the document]
johnsaigle: Sounds good, in general my preference would be to never panic on marshal errors. Agreed that it's out of scope for this PR.

bruce-riley:
> Oh one more thing -- should this PR include p2p options so that it's possible to see when Guardians have enabled this feature?

I like this. I'll look at adding it.

node/pkg/altpub/README.md
```diff
@@ -0,0 +1,113 @@
...
+--additionalPublishEndpoint label;url;delay;chains
+```
+
+<!-- cspell:enable -->
+
+The fields are defined as follows:
+
+- **label** is a string that is used to tag this endpoint in log messages and Prometheus metrics.
+- **url** is the http server endpoint to which the guardian should connect and publish.
+- **delay**, if specified (or non-zero) is the time the guardian should delay in order to batch observations. Zero / not set means publish immediately.
```
panoel, line 35: Can you add the units of the delay?  seconds or mS?

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,559 @@
...
+		} else {
+			requestFailed.WithLabelValues(req.ep.label, "post_failed").Inc()
+		}
+		return fmt.Errorf("post failed: %w", err)
+	}
+	stop := time.Now()
+	resp.Body.Close()
+
+	// Peg the appropriate metric, based on the request result.
+	if resp.StatusCode < 300 {
```
panoel, line 484: Isn't 200 -> 299 the actual range for success?

panoel (dismissed):

node/pkg/altpub/alternate_pub.go
```diff
@@ -0,0 +1,559 @@
...
+		ap *AlternatePublisher
+
+		// ep provides access to the endpoint in the worker.
+		ep *Endpoint
+	}
+)
+
+// All of the Prometheus metrics are indexed by the endpoint label.
+var (
+	obsvDropped = promauto.NewCounterVec(
```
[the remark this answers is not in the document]
bruce-riley: No, this is actually counting observations that don't get enqueued. (If we are batching, the number of requests is the number of batches, not observations.)

node/pkg/altpub/README.md
```diff
@@ -0,0 +1,113 @@
...
+- **chains**, if specified, is a comma-separated list of emitter chain IDs or names for which observations should be forwarded. If not set, all chains will be published.
```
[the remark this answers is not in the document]
bruce-riley: Yes. Since we are in the Wormhole repo, I didn't think I needed to say that. 😁

node/pkg/altpub/alternate_pub.go
bruce-riley: The 100 range is info. Are you suggesting I treat those as errors?

evan-gray (dismissed):

node/pkg/altpub/alternate_pub.go
evan-gray: FWIW, this is in the go example https://pkg.go.dev/net/http#example-Get

node/cmd/guardiand/node.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/certusone/wormhole/node/pkg/guardiansigner"
 	"github.com/certusone/wormhole/node/pkg/watchers"
 	"github.com/certusone/wormhole/node/pkg/watchers/ibc"
+	"github.com/ethereum/go-ethereum/crypto"
```
evan-gray, line 20: how did this dependency get in here twice?

evan-gray (approved):

node/cmd/guardiand/node.go
bruce-riley: I guess VSC was doing me a favor. Removed it.

evan-gray (approved):

panoel (approved):

PR #4319, opened 2025-03-28
lint: prevent hiding lint violations in sdk/ when node/ has errors

johnsaigle:
Because `set -e` is used at the top of this script, any command that fails causes the script to exit. 

As a result, we can't see golangci-lint violations for `sdk/` when there are also violations in `node/`, because the `sdk/` doesn't get linted. It makes more sense to view these issues together so that we can see all Go issues at once.

This change disables errexit when calling golangci-lint. It restores it after the commands run. The function will still return an error if either call to golangci-lint returns a non-zero exit code.

This PR also adds the `sdk/` path prefix, similar to what is done for `node/`.


scripts/lint.sh
```diff
@@ -60,12 +60,25 @@ lint(){
...
 
     cd "${ROOT}/sdk"
     golangci-lint run --timeout=10m $GOLANGCI_LINT_ARGS ./...
+    sdk_lint_status=$?
+
+    # Restore errexit
+    set -e
+
+    # Return failure if either linting operation failed
+    [[ $node_lint_status -eq 0 && $sdk_lint_status -eq 0 ]]
```
djb15, line 81: Should this be something like:

```
[[ $node_lint_status -eq 0 && $sdk_lint_status -eq 0 ]] || exit 1
```

instead?

johnsaigle: Yeah I think so. I don't think `[[ ]]`  will trigger an exit even with `set -e`

johnsaigle:
```suggestion
    [[ $node_lint_status -eq 0 && $sdk_lint_status -eq 0 ]] || exit 1
```
I'll make the change. Right now the script when run with the "lint" command will return 1 when the linters fail, as expected. That happens I think because the return value of `lint()` is returned to the caller (the `case` statement at the end of the script), which is the last statement in the code. Exiting early here more closely matches what `set -e` is probably intending to do. Without the exit, the script could theoretically continue running after a failing lint if it was modified to have other statements after the `case` part.

scripts/lint.sh
```diff
@@ -60,12 +60,25 @@ lint(){
...
+    node_lint_status=$?
 
     cd "${ROOT}/sdk"
     golangci-lint run --timeout=10m $GOLANGCI_LINT_ARGS ./...
+    sdk_lint_status=$?
+
+    # Restore errexit
+    set -e
+
+    # Return failure if either linting operation failed
```
johnsaigle, line 80:
```suggestion
    # Exit if either linting operation failed
```

scripts/lint.sh
djb15: Did this change get made and fall out with the force push?

johnsaigle: Yeah sorry

djb15 (approved):

PR #4320, opened 2025-03-28
node: Enable transfer verifier for the node in devnet

johnsaigle:
* Enables Transfer Verifier in devnet. This allows a full integration test between the Transfer Verifier package and its initial implementation of the node.
* adds some additional logging to make the results of transfer verification more clear.
* corrects an error message that implied a message won't be published if it fails transfer verification. Messages are always published in the current implementation, regardless of whether Transfer Verifier deems them to be bad. The error case is only triggered for `nil` or other parsing/programming errors.

Related work:
* #4233 added code to the watcher to include the Transfer Verifier
* https://github.com/wormhole-foundation/wormhole/pull/4169/files added CI tests for the Transfer Verifier as a standalone binary package


node/pkg/watchers/evm/watcher.go
```diff
@@ -882,8 +882,16 @@ func (w *Watcher) verifyAndPublish(
 			return err
 		}
 		msg = &verifiedMsg
+		w.logger.Info(
+			"verified transfer",
+			msg.ZapFields()...,
+		)
 	}
 
+	w.logger.Info(
```
bruce-riley, line 891: Not sure I like that we will now be logging two or three messages for every observation (the original "message observed" plus these two). Also, bear in mind that by default info logging in disabled in CI so you won't see these anyway. But we will see them in prod.

bruce-riley (commented):
One comment to be considered before I approve. Thanks.

node/pkg/watchers/evm/watcher.go
johnsaigle: I mostly had these in mind for local testing and testnet. I think they're warranted because they're reporting different parts of the process (initial observation, evaluation, and actual publishing). I'd be happy to change these to Debug level so that they're filtered out of prod. Would that work?

bruce-riley (approved):

pleasew8t (approved):

djb15 (approved):

PR #4322, opened 2025-04-01
ci: golangci-lint major version upgrade

johnsaigle:
While we had #4224 on-the-go, golangci-lint released a new major version.

This PR bumps the version in Ci. It also modifies .golangci.yml using the `golangci-lint migrate` command. https://golangci-lint.run/product/migration-guide/

New violations for staticcheck were excluded here so that we don't make a giant PR with a ton of small changes across many files. Instead, each of the excluded rules can be enabled separately for a more pleasant review experience.

Any of the rules with the `ST` prefix refer to style-related lints so they're less important for us anyway.


mdulin2 (approved):

djb15:
Looks like we got rid of `typecheck`, is that deliberate?

johnsaigle:
@djb15 It looks like `typecheck` was removed, and so wasn't copied over with the `migrate` command.

https://golangci-lint.run/product/migration-guide/#typecheck -- there's a note here though it's pretty confusing imo.

If you add `typecheck` back to the yml file, this error occurs:

```
Error: can't load config: typecheck is not a linter, it cannot be enabled or disabled
Failed executing command with error: can't load config: typecheck is not a linter, it cannot be enabled or disabled
```

So I think removing it here is correct.

djb15 (approved):

evan-gray (approved):
only reviewed scripts change

PR #4324, opened 2025-04-02
node: Integrate Transfer Verifier into Sui Watcher

pleasew8t:


devnet/tx-verifier/sui/sui_config/client.yaml
```diff
@@ -0,0 +1,8 @@
+keystore:
+  File: /.sui/sui_config/sui.keystore
+envs:
+  - alias: sui-devnet
+    rpc: "http://sui:9000"
+    ws: ~
```
johnsaigle, line 6: Can you add a comment for this `ws` key? I'm not sure what it does by looking.

node/cmd/txverifier/sui.go
```diff
@@ -18,6 +18,7 @@ import (
 
 const (
 	INITIAL_EVENT_FETCH_LIMIT = 25
+	EVENT_QUERY_INTERVAL      = 2 * time.Second
```
johnsaigle, line 21: nit: The Go convention for constant names is CamelCase rather than SCREAMING_SNAKE_CASE

devnet/tx-verifier/Dockerfile.tx-verifier-sui
```diff
@@ -0,0 +1,47 @@
+# Apply guardiand-image
+FROM guardiand-image as base
+
+FROM cli-gen AS cli-export
+FROM const-gen AS const-export
+FROM sui-node AS sui-node-base
+# FROM ghcr.io/wormhole-foundation/sui:1.19.1-mainnet@sha256:97e2e62d43262883b7a5bed5c9a9dd83d766a00a9476473eb275948fb53d376e AS sui
```
johnsaigle, line 7: Delete this? Seems to be commented out

devnet/tx-verifier/README.md
```diff
@@ -14,42 +14,30 @@ The basic idea is as follows:
...
 
-### Scripts
+| Script/Component | Description |
+|------------------|-------------|
+| `Dockerfile.tx-verifier-evm` | The dockerfile representing the image used for EVM transfer verifier tests. |
+| `tx-verifier-evm-runner.sh` | Runs the guardiand binary which contains the transfer verifier tool, which monitors the local Ethereum network for events. |
+| `tx-verifier-evm-tests.sh` | Contains the `cast` commands that simulate malicious interactions with the Token Bridge and Core Bridge. It is able to broadcast transactions to the `anvil` instance that powers the Ethereum testnet while being able to impersonate arbitrary senders. <br>This allows performing actions that otherwise should be impossible, like causing a Publish Message event to be emitted from the Core Bridge without a corresponding deposit or transfer into the Token Bridge. <br>The current integration test sends exactly two messages, each one corresponding to a different Token Bridge endpoint (Transfer and Transfer With Payload). |
 
-#### tx-verifier-evm-tests.sh
+## Sui Integration Tests
```
johnsaigle, line 25: Maybe the dependence on `worm` could be added here?

node/pkg/txverifier/sui.go
```diff
@@ -2,8 +2,10 @@ package txverifier
 
 // TODOs:
 //	* balances on Sui are stored as u64's. Consider using uint64 instead of big.Int
+//  * replace errors with error.join() like EVM
```
johnsaigle, line 5: Can you remove these TODOs, or address them if you still want to fix them?

node/pkg/txverifier/sui.go
```diff
@@ -23,24 +25,29 @@ var (
...
-	suiCoreContract        string
-	suiTokenBridgeEmitter  string
+	// Used to create the event filter.
+	suiCoreContract string
+	// Used to check the emitter of the `WormholeMessage` event.
+	suiTokenBridgeEmitter string
+	// Used to match the owning package of native and wrapped asset types.
 	suiTokenBridgeContract string
 	suiEventType           string
+	suiApiConnection       SuiApiInterface
```
johnsaigle, line 35: It might be worth adding a *zap.logger here. This way you can remove it as a parameter for each of the functions, and use the Watcher's logger later when initializing this struct.

node/pkg/txverifier/sui.go
```diff
@@ -164,40 +183,61 @@ func (s *SuiTransferVerifier) processObjectUpdates(objectChanges []ObjectChange,
...
+
+	if err != nil {
+		logger.Error("Failed to process digest", zap.Error(err))
+		return false
+	}
+
+	return true
+}
+
+func (s *SuiTransferVerifier) ProcessDigest(ctx context.Context, digest string, logger *zap.Logger) (uint, error) {
```
johnsaigle, line 199: extreme nit: maybe use a specific uint here rather than the platform-specific one

node/pkg/txverifier/sui.go
```diff
@@ -164,40 +183,61 @@ func (s *SuiTransferVerifier) processObjectUpdates(objectChanges []ObjectChange,
...
-	// TODO: Using `Warn` for testing purposes. Update to Fatal? when ready to go into PR.
-	// TODO: Revisit error handling here.
 	for key, amountOut := range requestedOutOfBridge {
+		keyParts := strings.Split(key, "-")
 
 		if _, exists := transferredIntoBridge[key]; !exists {
-			logger.Warn("transfer-out request for tokens that were never deposited",
-				zap.String("tokenAddress", key))
-			// TODO: Is it better to return or continue here?
+			// This implies that a token leaving the bridge was never deposited into it.
```
johnsaigle, line 225: nit: You could consider writing the error string once here, either as a string or in errors.New, and then log and return that value. That way both will have the same error string and it doesn't need to be duplicated here

node/pkg/txverifier/sui.go
```diff
@@ -164,40 +183,61 @@ func (s *SuiTransferVerifier) processObjectUpdates(objectChanges []ObjectChange,
...
 			return 0, errors.New("transfer-out request for tokens that were never deposited")
-			// continue
 		}
 
 		amountIn := transferredIntoBridge[key]
 
 		if amountOut.Cmp(amountIn) > 0 {
-			logger.Warn("requested amount out is larger than amount in")
+			// Implies that more tokens are being requested out of the bridge than were deposited into it.
+			logger.Error("token bridge transfer requested for an amount larger than what was deposited",
```
johnsaigle, line 236: nit: Same note here re: duplicated error strings.

node/pkg/txverifier/sui.go
```diff
@@ -215,14 +255,14 @@ type SuiApiResponse interface {
 	GetError() error
 }
 
-func suiApiRequest[T SuiApiResponse](rpc string, method string, params string) (T, error) {
+// TODO: add context
```
johnsaigle, line 258: Looks like this has been done, so the comment can be removed

node/pkg/txverifier/sui.go
```diff
@@ -215,14 +255,14 @@ type SuiApiResponse interface {
...
+func suiApiRequest[T SuiApiResponse](ctx context.Context, rpc string, method string, params string) (T, error) {
 	var defaultT T
 
 	// Create the request
 	requestBody := fmt.Sprintf(`{"jsonrpc":"2.0", "id": 1, "method": "%s", "params": %s}`, method, params)
 
-	//nolint:noctx
-	req, err := http.NewRequest("POST", rpc, strings.NewReader(requestBody))
+	req, err := http.NewRequestWithContext(ctx, "POST", rpc, strings.NewReader(requestBody))
```
johnsaigle, line 265: nit: On L282 (not shown here because of folding), there's a call to `io.ReadAll(resp.Body)`

It might be better to use io.Copy() with a bounded buffer or setting a size limit. This could be done in a follow-up PR though

node/pkg/txverifier/suitypes.go
```diff
@@ -12,10 +13,11 @@ import (
 const SUI_CHAIN_ID = 21
 
 // The SuiApi interface defines the functions that are required to interact with the Sui RPC.
+// TODO: add context
```
johnsaigle, line 16: Can be deleted?

node/pkg/txverifier/suitypes.go
```diff
@@ -12,10 +13,11 @@ import (
 const SUI_CHAIN_ID = 21
```
johnsaigle, line 13: Might be better to just use vaa.ChainID in this file instead (and change the types from uint16)

node/pkg/watchers/sui/config.go
```diff
@@ -11,10 +11,11 @@ import (
 )
 
 type WatcherConfig struct {
-	NetworkID        watchers.NetworkID // human readable name
-	ChainID          vaa.ChainID        // ChainID
-	Rpc              string
-	SuiMoveEventType string
+	NetworkID         watchers.NetworkID // human readable name
+	ChainID           vaa.ChainID        // ChainID
```
johnsaigle, line 15: nit
```suggestion
	ChainID           vaa.ChainID
```

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/certusone/wormhole/node/pkg/common"
+	"go.uber.org/zap"
+)
+
+func (e *Watcher) verify(
```
johnsaigle, line 11: I'm not a huge fan of adding a method here vs in the watcher code directly. The EVM implementation also uses a function (not a method on the watcher) -- we should be consistent here. I don't really have a strong preference for which way we do it, but we should do it the same way.

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,58 @@
...
+	ctx context.Context,
+	msg *common.MessagePublication,
+	txDigest string,
+	logger *zap.Logger,
```
johnsaigle, line 15: If this is a method on the Watcher, you can use the Watcher's logger instead of adding a parameter here.

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	localMsg := msg
+
+	var verificationState common.VerificationState
+
+	// If the emitter address of the message does not match the token bridge emitter,
+	// mark the message's verification state as NotApplicable.
+	if localMsg.EmitterAddress.String() != e.suiTxVerifier.GetTokenBridgeEmitter() {
+		verificationState = common.NotApplicable
```
johnsaigle, line 37: This function should also return `NotApplicable` if the payload isn't a token transfer. (via `vaa.IsTransfer`). I see this is missing in the EVM side too, so we can make a follow-up PR to add it there.

node/pkg/watchers/sui/watcher.go
```diff
@@ -46,6 +49,10 @@ type (
 		loopDelay                 time.Duration
 		queryEventsCmd            string
 		postTimeout               time.Duration
+
+		// Sui transaction verifier
+		suiTxVerifier     *txverifier.SuiTransferVerifier
+		txVerifierEnabled bool
```
johnsaigle, line 55: nit: I think it's worth adding a zap.Logger here as well rather than passing it as a parameter to all of the methods. Could be done in a cleanup PR later.

johnsaigle (changes requested):
I reviewed the watcher integration parts and some of the testnet structure, but I don't really know how Sui works.

I noted some things to cleanup as well as some nits that can be ignored and/or handled in future PRs. Feel free to just resolve them

node/pkg/watchers/sui/msg_verifier.go
mdulin2: Good callout!

devnet/tx-verifier/sui/sui_config/client.yaml
pleasew8t: I just removed it.

node/pkg/txverifier/sui.go
pleasew8t: Removed

node/pkg/txverifier/sui.go
pleasew8t: This is an improvement I'll consider if we ever decide to replace the existing API helpers in the watcher with the ones we wrote for Sui tx verification.

node/pkg/txverifier/sui.go
pleasew8t: going to ignore this one

node/pkg/txverifier/suitypes.go
pleasew8t: I like this suggestion. Thanks.

node/pkg/watchers/sui/watcher.go
pleasew8t: Agreed that it's better suited for a later PR, seeing as the scope is beyond tx verification.

node/pkg/watchers/sui/msg_verifier.go
pleasew8t: The Sui watcher in general is structured a bit different to EVM, one difference being the logger that gets passed around instead of being part of the watcher struct. I think there is some refactoring that can be done in the Sui watcher, but I'm going to avoid making it part of this PR.

node/pkg/watchers/sui/msg_verifier.go
pleasew8t: The reason I did it this way, was because of personal preference - it seems silly to take a member of a struct (`suiTxVerifier`) and pass that to a function that is in some sense closely related to the struct itself, instead of just adding a method to the struct itself.

node/pkg/txverifier/sui.go
pleasew8t: Deferring this to a later update, seeing as the Sui watcher does not yet have a logger when the `SuiTransferVerifier` struct is instantiated.

devnet/tx-verifier/sui/sui_config/client.yaml
pleasew8t: Actually, put this back in. It's part of the local sui client config, copied from the devnet setup.

node/pkg/watchers/sui/msg_verifier.go
johnsaigle: Yeah makes sense. I do want these to be consistent so I'll make a cleanup PR to change the EVM side to also use a method.

node/pkg/watchers/sui/watcher.go
```diff
@@ -275,24 +319,55 @@ func (e *Watcher) inspectBody(logger *zap.Logger, body SuiResult, isReobservatio
 		IsReobservation:  isReobservation,
 	}
 
+	err = e.verifyAndPublish(ctx, observation, *body.ID.TxDigest, logger)
+
+	if err != nil {
+		logger.Error("Message publication error",
```
bruce-riley, line 325: There should be a Prometheus metric for this so that guardians can easily track it.

bruce-riley (approved):

johnsaigle (dismissed):

djb15 (commented):
Still need to review the devnet stuff but the integration into the watcher looks good to me, very similar to what I was envisioning

devnet/tx-verifier/sui/tx-verifier-sui-runner.sh
djb15, line 1: Is it worth adding a test for a PTB that contains both a legit and fake transfer?

node/pkg/txverifier/suitypes.go
```diff
@@ -130,34 +131,35 @@ func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success
...
-	Result []SuiTryMultiGetPastObjectsResult `json:"result"`
+	Result [2]SuiTryMultiGetPastObjectsResult `json:"result"`
 }
 
 // Gets the balance difference of the two result objects.
 func (r SuiTryMultiGetPastObjectsResponse) GetBalanceDiff() (*big.Int, error) {
 
 	if len(r.Result) != 2 {
-		return big.NewInt(0), fmt.Errorf("incorrect number of results received")
+		return big.NewInt(0), fmt.Errorf("incorrect number of objects in Result")
```
johnsaigle, line 141: Since you're using a pointer return type here, I think it's better to return `nil` in this function than a BigInt of 0.
```suggestion
		return nil, fmt.Errorf("incorrect number of objects in Result")
```

node/pkg/txverifier/suitypes.go
```diff
@@ -130,34 +131,35 @@ func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success
...
 	if err != nil {
-		return big.NewInt(0), fmt.Errorf("error in getting old balance: %w", err)
+		return big.NewInt(0), fmt.Errorf("error in getting new balance: %w", err)
 	}
 
 	difference := newBalance.Sub(newBalance, oldBalance)
+
 	// If the asset is wrapped, it means that the balance was burned, so the difference should be negative.
 	if isWrapped {
 		difference = difference.Mul(difference, big.NewInt(-1))
```
johnsaigle, line 165: Just FYI you can use the `.Neg()` function on BigInt to flip the sign.

node/pkg/txverifier/suitypes.go
```diff
@@ -130,34 +131,35 @@ func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success
...
-	oldBalance, err := r.Result[1].GetVersionBalance(isWrapped)
+	newBalance, err := r.Result[0].GetVersionBalance(isWrapped)
 	if err != nil {
-		return big.NewInt(0), fmt.Errorf("error in getting old balance: %w", err)
+		return big.NewInt(0), fmt.Errorf("error in getting new balance: %w", err)
 	}
 
 	difference := newBalance.Sub(newBalance, oldBalance)
+
 	// If the asset is wrapped, it means that the balance was burned, so the difference should be negative.
```
johnsaigle, line 163: Maybe an error check should be added if the sign is negative but the asset isn't wrapped?

node/pkg/txverifier/suitypes.go
```diff
@@ -375,24 +377,24 @@ func (r SuiTryMultiGetPastObjectsResult) GetTokenChain() (uint16, error) {
...
 
-	chain, err := extractFromJsonPath[float64](*r.Details, path)
+	chain, err := extractFromJsonPath[float64](r.Details, path)
 
 	if err != nil {
 		return 0, fmt.Errorf("error in extracting chain: %w", err)
 	}
 
-	return uint16(chain), nil
+	return vaa.ChainID(chain), nil
```
johnsaigle, line 391: Maybe this could be changed to use `KnownChaindIDFromNumber` https://github.com/wormhole-foundation/wormhole/blob/main/sdk/vaa/structs.go#L312 ?
As written this will accept any number and truncate it to uint16. Maybe that's fine in practice, but the check doesn't hurt.

node/pkg/txverifier/utils.go
```diff
@@ -23,6 +23,10 @@ const (
 func extractFromJsonPath[T any](data json.RawMessage, path string) (T, error) {
 	var defaultT T
 
+	if data == nil {
+		return defaultT, fmt.Errorf("supplied JSON data is nil")
```
johnsaigle, line 27: Later in this function, could you modify the error messages `return defaultT, fmt.Errorf("can't convert to type T")`? It will always just return the letter `T`, not the actual type

node/pkg/txverifier/suitypes.go
```diff
@@ -416,21 +418,21 @@ func (r SuiTryMultiGetPastObjectsResult) GetVersion() (string, error) {
...
-	return version, nil
+	return objectType, nil
 }
 
 // Definition of the WormholeMessage event
 type WormholeMessage struct {
-	ConsistencyLevel *uint8  `json:"consistency_level"`
-	Nonce            *uint64 `json:"nonce"`
+	ConsistencyLevel uint8   `json:"consistency_level"`
+	Nonce            uint64  `json:"nonce"`
```
johnsaigle, line 433: Nonce should be a uint32, is it different for Sui?

node/pkg/txverifier/suitypes.go
```diff
@@ -416,21 +418,21 @@ func (r SuiTryMultiGetPastObjectsResult) GetVersion() (string, error) {
...
 	Payload          []byte  `json:"payload"`
 	Sender           *string `json:"sender"`
-	Sequence         *string `json:"sequence"`
-	Timestamp        *string `json:"timestamp"`
+	Sequence         string  `json:"sequence"`
```
johnsaigle, line 436: Should be a uint64

node/pkg/txverifier/utils.go
```diff
@@ -107,6 +111,7 @@ func SupportedChains() []vaa.ChainID {
 	return []vaa.ChainID{
 		// Mainnets
 		vaa.ChainIDEthereum,
+		vaa.ChainIDSui,
```
johnsaigle, line 114: Does Sui have a testnet value?

node/pkg/watchers/sui/watcher.go
```diff
@@ -275,24 +324,49 @@ func (e *Watcher) inspectBody(logger *zap.Logger, body SuiResult, isReobservatio
...
+		}
+
+		msg = &verifiedMsg
+	}
+
+	e.msgChan <- msg
+
 	suiMessagesConfirmed.Inc()
-	if isReobservation {
+	if msg.IsReobservation {
```
johnsaigle, line 362: The `NewWatcher` in `config.go` always returns `nil` for its `interfaces.Reobserver` field. Are reobservations possible on Sui? @bruce-riley maybe you would know?

node/cmd/txverifier/sui.go
```diff
@@ -164,7 +164,7 @@ func runTransferVerifierSui(cmd *cobra.Command, args []string) {
 			logger.Info("Context cancelled")
 		case <-ticker.C:
 			// Fetch new events
-			resp, err := suiApiConnection.QueryEvents(eventFilter, "null", 25, true)
+			resp, err := suiApiConnection.QueryEvents(ctx, eventFilter, "null", InitialEventFetchLimit, true)
```
johnsaigle, line 167: `"null"` is used everywhere for this parameter -- is this correct or should it update under some conditions?

If it's always `"null"` then maybe the QueryEvents function should remove it as an argument.

node/pkg/watchers/sui/watcher.go
pleasew8t: The Sui watcher does monitor a re-observation requests channel [here](https://github.com/pleasew8t/wormhole/blob/70363a2817f1a72bebd3ae09368a2a494ebc72a8/node/pkg/watchers/sui/watcher.go#L468).

node/pkg/txverifier/suitypes.go
pleasew8t: Good call out. Changed to uint32.

node/pkg/txverifier/sui.go
```diff
@@ -164,48 +179,75 @@ func (s *SuiTransferVerifier) processObjectUpdates(objectChanges []ObjectChange,
...
+
+	if err != nil {
+		logger.Error("Failed to process digest", zap.Error(err))
+		return false
+	}
+
+	return true
+}
+
+func (s *SuiTransferVerifier) ProcessDigest(ctx context.Context, digest string, logger *zap.Logger) (uint, error) {
```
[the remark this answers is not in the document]
pleasew8t: Going to ommit this change for now. The check isn't expensive, but I'm happy to have the RPC fail if the tx digest is invalid.

node/pkg/txverifier/suitypes.go
```diff
@@ -130,34 +131,35 @@ func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success
...
 	}
 
-	oldBalance, err := r.Result[1].GetVersionBalance(isWrapped)
+	newBalance, err := r.Result[0].GetVersionBalance(isWrapped)
 	if err != nil {
-		return big.NewInt(0), fmt.Errorf("error in getting old balance: %w", err)
+		return big.NewInt(0), fmt.Errorf("error in getting new balance: %w", err)
 	}
 
 	difference := newBalance.Sub(newBalance, oldBalance)
```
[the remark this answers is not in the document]
pleasew8t: Added a comment to highlight that, in case of future modifications

devnet/tx-verifier/sui/tx-verifier-sui-runner.sh
```diff
@@ -0,0 +1,228 @@
+#!/bin/sh
```
johnsaigle:
```suggestion
#!/bin/bash
```
This script uses many bash-specific bits that aren't compatible with regular `sh`. You're also invoking it using `bash` in the `yaml` file. 
(If you're curious about details, try running `shellcheck`)

node/pkg/txverifier/suitypes.go
```diff
@@ -416,21 +426,30 @@ func (r SuiTryMultiGetPastObjectsResult) GetVersion() (string, error) {
...
 	ConsistencyLevel *uint8  `json:"consistency_level"`
-	Nonce            *uint64 `json:"nonce"`
+	Nonce            *uint32 `json:"nonce"`
 	Payload          []byte  `json:"payload"`
 	Sender           *string `json:"sender"`
 	Sequence         *string `json:"sequence"`
 	Timestamp        *string `json:"timestamp"`
 }
+
+// FieldsData struct {
```
johnsaigle: Delete?

node/pkg/txverifier/suitypes.go
```diff
@@ -366,36 +370,42 @@ func (r SuiTryMultiGetPastObjectsResult) GetTokenAddress() (tokenAddress string,
...
 
 	if !wrapped {
-		return SUI_CHAIN_ID, nil
+		return vaa.ChainIDSui, nil
 	}
 
 	path := "content.fields.value.fields.info.fields.token_chain"
 
-	chain, err := extractFromJsonPath[float64](*r.Details, path)
+	chain, err := extractFromJsonPath[float64](r.Details, path)
```
johnsaigle, line 388: Is chain encoded as a float64 here? This looks strange, as does the truncation in the call to `KnownChainIDFromNumber`. Can you pass uint16 here?

node/pkg/txverifier/suitypes.go
```diff
@@ -366,36 +370,42 @@ func (r SuiTryMultiGetPastObjectsResult) GetTokenAddress() (tokenAddress string,
...
 
 	if err != nil {
-		return 0, fmt.Errorf("error in extracting chain: %w", err)
+		return 0, fmt.Errorf("Error extracting chain: %w", err)
 	}
 
-	return uint16(chain), nil
+	chainId, err := vaa.KnownChainIDFromNumber(uint16(chain))
```
johnsaigle, line 398: I don't think `float64 --> uint16` is safe

node/pkg/txverifier/utils.go
```diff
@@ -16,13 +16,30 @@ import (
...
+func (i InvariantError) Error() string {
+	return fmt.Sprintf("invariant violated: %s", i.Msg)
+}
+
 // Extracts the value at the given path from the JSON object, and casts it to
 // type T. If the path does not exist in the object, an error is returned.
 func extractFromJsonPath[T any](data json.RawMessage, path string) (T, error) {
 	var defaultT T
 
+	if data == nil {
```
johnsaigle, line 112:
```suggestion
	if len(data) == 0 {
```
Can `json.RawMessage` (alias for `[]byte`) be `nil` here? In any case the length check will also cover the `nil` case.

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,67 @@
...
+	if !vaa.IsTransfer(msg.Payload) || "0x"+localMsg.EmitterAddress.String() != e.suiTxVerifier.GetTokenBridgeEmitter() {
+		verificationState = common.NotApplicable
+	} else {
+		// Validate the transfers in the transaction block associated with the
+		// transaction digest.
+		valid, err := e.suiTxVerifier.ProcessDigest(ctx, txDigest, logger)
+
+		fmt.Println(valid, err)
+
+		if valid {
```
johnsaigle: This allows for a case where we have `true, error` and the error isn't handled. I think this should be rewritten to check the error first.

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,67 @@
...
+			verificationState = common.Valid
+		} else if err != nil {
+			// If an error was returned with valid = false, then it's a signal that an internal error occurred,
+			// and the validity of the transfer cannot be determined.
+			logger.Error("an internal tx verifier error occurred: ", zap.Error(err))
+			verificationState = common.NotApplicable
```
johnsaigle: Let's change this to `CouldNotVerify` when the other PR is merged:
https://github.com/wormhole-foundation/wormhole/pull/4362/files#diff-3b96484e558bdbaba23ef8da2847b69f926e9d8dc67f6b2f2d941be3590ed394R37-R38

node/pkg/watchers/sui/watcher.go
```diff
@@ -176,9 +189,44 @@ func NewWatcher(
...
+		case common.MainNet:
+			suiCoreBridgeAddress = sdk.KnownMainnetCoreContracts[vaa.ChainIDSui]
+			suiTokenBridgeAddress = sdk.KnownMainnetTokenBridgeContracts[vaa.ChainIDSui]
+			suiTokenBridgeEmitter = hex.EncodeToString(sdk.KnownTokenbridgeEmitters[vaa.ChainIDSui])
+		case common.TestNet:
+			suiCoreBridgeAddress = sdk.KnownTestnetCoreContracts[vaa.ChainIDSui]
+			suiTokenBridgeAddress = sdk.KnownTestnetTokenBridgeContracts[vaa.ChainIDSui]
+			suiTokenBridgeEmitter = hex.EncodeToString(sdk.KnownTestnetTokenbridgeEmitters[vaa.ChainIDSui])
+
+		case common.UnsafeDevNet, common.AccountantMock, common.GoTest:
```
johnsaigle, line 228: May be better to use `default` here instead.
EDIT: my suggestion might actually be blocked by the exhaustruct linter. You might need both?

johnsaigle (commented):
Some minor changes, but nearly done

node/pkg/txverifier/sui.go
```diff
@@ -164,64 +195,94 @@ func (s *SuiTransferVerifier) processObjectUpdates(objectChanges []ObjectChange,
...
+	} else {
+		return count, verified, err
+	}
+}
+
+// Return conditions:
+//
+//	_, true, nil - verification succeeded
+//	_, false, nil - verification failed
+//	_, false, err - verification could not be performed due to an internal error
```
djb15: This last comment is misleading as errors are returned also when there are invariant violations

node/pkg/txverifier/suitypes.go
pleasew8t: So this was a weird quirk of parsing the JSON. Interestingly enough, if you change float64 here to some other type, like uint16, the error [here](https://github.com/pleasew8t/wormhole/blob/1816a695603a0a073a5f76a0fffb57775933e476/node/pkg/txverifier/utils.go#L71) will fire.

johnsaigle: Ah weird. In practice this is probably fine, maybe worth a comment though.

node/pkg/txverifier/suitypes.go
pleasew8t: Made an update here by adding the following check:

```
	if float64(uint16(chain)) != chain {
		return 0, fmt.Errorf("failed to cast chain ID: %f", chain)
	}
```

pleasew8t: I did a quick test of this to make sure it works:

```
package main
import "fmt"
import "math"

func main() {
  var f float64
  f = 0
  var i uint16
  for i = 1; i < math.MaxUint16; i++ {
	if i%1000 == 0 {
		fmt.Println("at ", i)
	}
	f += 1

	if uint16(f) != i {
		panic("could not convert")
	}
  }

}
```

node/pkg/txverifier/utils.go
pleasew8t: Happy to change it if there's a specific convention, but I prefer the `nil` check.

node/pkg/watchers/sui/msg_verifier.go
pleasew8t: made the change. checking for error, then validity, and then reporting as anomalous

node/pkg/watchers/sui/msg_verifier.go
pleasew8t: With the other PR merged, we should have some conflicts anyway, so will do that after all current comments have been resolved

node/pkg/watchers/sui/watcher.go
pleasew8t: Will investigate if CI fails on the linter.

node/pkg/watchers/sui/msg_verifier.go
johnsaigle: Unresolving this conversation as the `CouldNotVerify` state should be available now.

node/pkg/txverifier/utils.go
johnsaigle: My hunch is that this should return an error if the JSON is empty, regardless of the nil check. We could do both if you want to be extra clear about it.

johnsaigle (commented):
Ready to approve as soon as the outstanding comments are addressed.

node/pkg/txverifier/suitypes.go
```diff
@@ -130,37 +131,40 @@ func (o ObjectChange) ValidateTypeInformation(expectedPackageId string) (success
...
 	}
 
+	// NOTE: newBalance is also modified by this operation, so care should be taken if future changes
+	// relies on the original value of newBalance.
 	difference := newBalance.Sub(newBalance, oldBalance)
+
 	// If the asset is wrapped, it means that the balance was burned, so the difference should be negative.
 	if isWrapped {
-		difference = difference.Mul(difference, big.NewInt(-1))
+		difference.Neg(difference)
```
johnsaigle, line 168: The value coming out of this function can be negative. It's used in `extractTransfersIntoBridgeFromObjectChanges` to populate `Amount` for a transfer, which in turn is used in `validateSolvency` to determine the validity of a message.

The key logic for that is here:
```
		// Subtract the requested amount from the transfer amount
		transfer.Amount = new(big.Int).Sub(transfer.Amount, request.Amount)

		// The moment the transfer amount goes negative, the asset is insolvent.
		if transfer.Amount.Sign() < 0 {
			insolventAssetKeys = append(insolventAssetKeys, request.AssetKey)
		}
```

What happens if the Amount is negative here? If it's handled safely, do you mind adding comments/logging/assertions and/or tests to show that it's safe?

node/pkg/txverifier/utils.go
```diff
@@ -308,3 +308,98 @@ func upsert(
...
+
+type MsgIdToRequestOutOfBridge map[string]*RequestOutOfBridge
+
+// Represents a request to move assets out of the Sui token bridge
+type RequestOutOfBridge struct {
+	AssetKey string
+	Amount   *big.Int
+
+	// Validation parameters
+	DepositMade    bool // True of `assetKey` was deposited into the bridge
```
djb15: Typo "of" -> "if"

node/pkg/txverifier/utils.go
```diff
@@ -308,3 +308,98 @@ func upsert(
...
+
+type AssetKeyToTransferIntoBridge map[string]*TransferIntoBridge
+
+// Represents a transfer of assets into the Sui token bridge
+type TransferIntoBridge struct {
+	Amount  *big.Int
+	Solvent bool
+}
+
+// validateSolvency checks wether or not the transfers into the brigde are sufficient to cover the
```
djb15: "whether"

node/pkg/txverifier/utils.go
```diff
@@ -308,3 +308,98 @@ func upsert(
...
+// requests to transfer assets out of the bridge.
+//   - if an asset is considered involvent, all requests out of the bridge for that asset are marked
```
djb15: "insolvent"

node/pkg/txverifier/utils.go
```diff
@@ -308,3 +308,98 @@ func upsert(
...
+	// This function assumes that amounts are non-nil, so an error is returned if
+	// any amount is nil. An empty resolution map is also returned.
+	for msgIdStr, request := range requests {
+		if request.Amount == nil {
+			return resolved, fmt.Errorf("nil amount in request out of bridge for msgID %s", msgIdStr)
+		}
+	}
+
+	for assetKey, transfer := range transfers {
+		if transfer.Amount == nil {
```
djb15, line 360: Is it possible to perform a 0 amount transfer in Sui? Is there a possibility that someone could hit this case and induce the early return?

djb15: Do both these 0 amount checks guard against anything anyway?

node/pkg/txverifier/sui.go
```diff
@@ -229,47 +245,82 @@ func (s *SuiTransferVerifier) processDigestInternal(ctx context.Context, digest
...
-			// This implies that a token leaving the bridge was never deposited into it.
-			invariantError := &InvariantError{Msg: INVARIANT_NO_DEPOSIT}
+	if err != nil {
+		logger.Error("Error validating solvency", zap.Error(err))
+		return false, err
+	}
 
-			return 0, false, invariantError
-		}
+	// If msgIdStr is found in the resolved map, check only that request. Otherwise, check all requests.
```
djb15, line 285: Can you explain why we need to switch here between checking a single request vs all requests? I'm struggling to grep that

node/pkg/txverifier/suitypes.go
pleasew8t: So this is actually how burning on Sui works. We are checking the total supply of the asset, which means the total supply change is negative if a wrapped asset was burned == transferred into the bridge. I tried to say that in the comment above `if wrapped`, but I can state that better.

node/pkg/txverifier/utils.go
pleasew8t: So it's not a zero-amount check. It's rather a check to ensure that the *big.Int is not nil, which would obviously result in a nil pointer deference later. I documented this earlier in the function, but let me know if you think this should change.

node/pkg/txverifier/suitypes.go
johnsaigle: Doesn't the negative value mess with the solvency math though?

node/pkg/txverifier/sui.go
pleasew8t: So this is sort of the same behaviour as the EVM transfer verifier. The difference is that in the Sui transfer verifier I opted to verify the entire transaction's transfers, instead of just a single one, if the message ID can't be found. In the EVM transfer verifier this is done if the message ID is an empty string.

djb15: Gotcha, thanks

node/pkg/txverifier/suitypes.go
pleasew8t: The negative value comes from the API. I'm negativing the negative.

johnsaigle: Ah OK, I didn't get that from the comment. Maybe worth rewording the comment, or else e.g. checking `isWrapped` along with https://pkg.go.dev/math/big#Int.Sign

pleasew8t: We can't really do that, because we still need to account for redeemed transfers

devnet/tx-verifier/sui/tx-verifier-sui-runner.sh
pleasew8t: Yes it is. I've noted this, and will do it in a follow-up PR.

node/pkg/txverifier/suitypes.go
johnsaigle: OK do you mind rewording the comment then? I'm not really clear on how this works or what you're working around, so I think the extra explanation could be really helpful.

node/pkg/watchers/sui/watcher.go
```diff
@@ -275,24 +327,49 @@ func (e *Watcher) inspectBody(logger *zap.Logger, body SuiResult, isReobservatio
...
+		verifiedMsg, err := e.verify(ctx, msg, txDigest, logger)
+
+		if err != nil {
+			return err
+		}
+
+		msg = &verifiedMsg
+	}
+
+	e.msgChan <- msg
```
evan-gray: was was pushing to the channel moved from where it was in the code and why was the comment removed?

node/pkg/watchers/sui/watcher.go
```diff
@@ -275,24 +327,49 @@ func (e *Watcher) inspectBody(logger *zap.Logger, body SuiResult, isReobservatio
...
 	logger.Info("message observed",
-		zap.String("txHash", observation.TxIDString()),
-		zap.Time("timestamp", observation.Timestamp),
-		zap.Uint32("nonce", observation.Nonce),
-		zap.Uint64("sequence", observation.Sequence),
-		zap.Stringer("emitter_chain", observation.EmitterChain),
-		zap.Stringer("emitter_address", observation.EmitterAddress),
-		zap.Binary("payload", observation.Payload),
-		zap.Uint8("consistencyLevel", observation.ConsistencyLevel),
+		msg.ZapFields()...,
```
evan-gray, line 386: these are not equivalent. `ZapFields` does not include the sequence, emitter_chain, emitter_address, or payload. I would recommend against changing this unless it was needed for some reason, and then not lose any information in doing so.

sdk/mainnet_consts.go
```diff
@@ -189,3 +189,14 @@ var KnownWrappedNativeAddress = map[vaa.ChainID]string{
 	// WETH
 	vaa.ChainIDEthereum: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
 }
+
+// KnownMainnetCoreContracts is a map of known core contract addresses.
+var KnownMainnetCoreContracts = map[vaa.ChainID]string{
+	vaa.ChainIDSui: "5306f64e312b581766351c07af79c72fcb1cd25147157fdc2f8ad76de9a3fb6a",
```
evan-gray: how is this being used? this was the originally published package, I think, but that is not typically how we refer to Sui packages because they, themselves are non-upgradeable. This may not be the current version of the package in the future, so it is not safe to use in all cases. I would appreciate a more specific name if it is only used for its type, which would be safe. Otherwise, you should reconsider your approach and use the state object like the TS SDK and most others do.

sdk/testnet_consts.go
```diff
@@ -101,6 +101,17 @@ var KnownTestnetAutomaticRelayerEmitters = []struct {
...
 
+// KnownTestnetCoreContracts is a map of known core contract addresses.
+var KnownTestnetCoreContracts = map[vaa.ChainID]string{
+	vaa.ChainIDSui: "31358d198147da50db32eda2562951d53973a0c0ad5ed738e9b17d88b213d790",
+}
+
+// KnownTestnetTokenBridgeContracts is a map of known token bridge contract addresses. In some cases,
+// there might be a difference between the emitter address and the token bridge contract (such as for Sui).
+var KnownTestnetTokenBridgeContracts = map[vaa.ChainID]string{
+	vaa.ChainIDSui: "a6a3da85bbe05da5bfd953708d56f1a3a023e7fb58e5a824a3d4de3791e8f690",
```
evan-gray: The token bridges are also incorrect as the current version. For example, the testnet state is https://suiscan.xyz/testnet/object/0x6fb10cdb7aa299e9a4308752dadecb049ff55a892de92992a1edbd7912b3d6da/fields and the current version is `11`, package `0x71a0a23c9496d90f294fd3d3e383dd21df8819667cc6e94107ac39c8bb98ccd9` (see Upgrade cap).
Object `0xa6a3da85bbe05da5bfd953708d56f1a3a023e7fb58e5a824a3d4de3791e8f690` was deleted according to the explorer, I'm not sure where this address came from.

node/pkg/watchers/sui/watcher.go
pleasew8t: I put the comment back.

Moving the push to the channel was more of a choice of convention to be similar to the EVM watcher's [`verifyAndPublish`](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/node/pkg/watchers/evm/watcher.go#L898).

node/pkg/watchers/sui/watcher.go
pleasew8t: The `msgID` contains the sequence, emitter chain and emitter address. I can add the payload to this? Maybe something like:

```go
		w.logger.Debug(
			"verified transfer",
			append(msg.ZapFields(), zap.Binary("payload", msg.Payload))...,
		)
```

I like the compactness. Let me know what you think. Otherwise, I'll revert back to the old code.

sdk/mainnet_consts.go
pleasew8t: The package ID is used for querying core bridge events. It's part of the `type` field in the `suix_queryEvents`'s `MoveEventType` parameter. I used the same API in the transfer verifier as the watcher does (see [here](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/node/pkg/watchers/sui/watcher.go#L243), and an example watcher `--suiMoveEventType` [here](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/Tiltfile#L236)).

There are two scenarios in which this is used:
* The transfer verifier run in standalone mode to [query events](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/node/cmd/txverifier/sui.go#L213)
* To [verify](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/node/pkg/txverifier/sui.go#L95-L96) that the `MessagePublication` from an event in a PTB originates from the core bridge.

Let me know if this isn't 100% correct, and I can make changes. I'm also happy to remove the changes from the `*_consts.go` files and move them to be local to transfer verifier, or just parse out the package ID from the `suiMoveEventType` argument (which, in hindsight, seems better).

sdk/testnet_consts.go
pleasew8t: The token bridge package ID is used to [validate the type information](https://github.com/pleasew8t/wormhole/blob/495867e6199927edf5c1d1035b44a96016f1a09a/node/pkg/txverifier/suitypes.go#L97) for object changes within a PTB. It's the way we identify balance changes on the token bridge.

I did some digging, and it seems like the token bridge on testnet is practically dead. I can follow a similar approach and localise the package IDs to the transfer verifier package, if you'd prefer.

sdk/mainnet_consts.go
evan-gray: I agree I would parse this from the suiMoveEventType parameter

sdk/testnet_consts.go
evan-gray: For now, please move this to the transfer verifier package and call it something more accurate like originalTokenBridgePackageId

node/pkg/watchers/sui/watcher.go
evan-gray: I still would generally lean towards making the minimally invasive code change. There was no need to change this code for this feature PR, but I’m alright with the change you suggested. Perhaps a better suggestion would be to save this for a separate PR where you make this log consistent across all the watchers.

node/pkg/watchers/sui/watcher.go
evan-gray: Consistency with the EVM watcher is a good argument. I’ll resolve this after the comment is put back

node/pkg/watchers/sui/watcher.go
evan-gray: Just read the above comment about consistency with the EVM watcher which seems reasonable, though I then realized this log is an info and not a debug - I’m confident which is preferable so again I’d lean towards no change, but I’ll resolve this for the consistency otherwise.

node/pkg/txverifier/sui.go
```diff
@@ -57,171 +66,274 @@ func (s *SuiTransferVerifier) GetEventFilter() string {
...
-		resp, err := suiApiConnection.TryMultiGetPastObjects(objectChange.ObjectId, objectChange.Version, objectChange.PreviousVersion)
+		if objectChange.PreviousVersion == "" {
+			logger.Warn("No previous version of asset available",
+				zap.String("objectId", objectChange.ObjectId),
+				zap.String("currentVersion", objectChange.Version))
+			continue
+		}
 
+		// Get the previous version of the object. This makes a call to the Sui API.
+		resp, err := s.suiApiConnection.TryMultiGetPastObjects(ctx, objectChange.ObjectId, objectChange.Version, objectChange.PreviousVersion)
```
mdulin2, line 175: There is no error handling here that gets propagated up. After reasoning about it, I think it's okay though. 

If the `balanceChange` is skipped for a WH message transfer, then it would be caught on the insolvency checks. So, this failing will fail later anyway.

pleasew8t: My entire thinking behind error handling with the Sui transfer verifier is that if it's malformed/invalid it gets ignored. Practically speaking, there is no good response to bad replies from the API.

node/pkg/txverifier/suitypes.go
```diff
@@ -52,6 +54,62 @@ type SuiQueryEventsResponse struct {
...
+	SuiApiStandardResponse
+	Result *json.RawMessage `json:"result"`
+}
+
+func (s SuiGetObjectResponse) TokenBridgePackageId() (string, error) {
+	if s.Result == nil {
+		return "", fmt.Errorf("sui_getObject response is nil, indicating an invalid state object ID or api error")
+	}
+
+	path := "data.content.fields.token_registry.type"
```
evan-gray: This is not the correct path for getting the current package ID. The current package ID is found on the Upgrade Cap. If you were looking for the original package ID, you can simply get the type of the state or simply put the original package ID in code since it will never change.

node/pkg/txverifier/suitypes.go
```diff
@@ -52,6 +54,62 @@ type SuiQueryEventsResponse struct {
 	Result SuiQueryEventsResult `json:"result"`
 }
 
+type SuiGetObjectResponse struct {
+	SuiApiStandardResponse
+	Result *json.RawMessage `json:"result"`
+}
+
+func (s SuiGetObjectResponse) TokenBridgePackageId() (string, error) {
```
evan-gray: I would recommend being explicit in Sui about whether a function is getting the original package ID (for type purposes) or the current package ID (for calling purposes)

node/pkg/txverifier/sui.go
```diff
@@ -16,31 +14,48 @@ import (
...
+)
+
 // Global variables
 var (
 	suiModule    = "publish_message"
 	suiEventName = "WormholeMessage"
+
+	// The state objects used for the different environments. These are used to extract the token bridge
+	// package ID as well as the emitter address.
+	SuiMainnetStateObjectId = "0xc57508ee0d4595e5a8728974a4a93a787d38f339757230d441e895422c07aba9"
```
evan-gray: might be worth a better variable name here that encompasses what state object ID it is (e.g. `SuiMainnetTokenBridgeStateObjectId`)

node/pkg/watchers/sui/watcher.go
```diff
@@ -176,9 +187,71 @@ func NewWatcher(
...
+		if err != nil {
+			return nil, err
+		}
+
+		suiTokenBridgeEmitter, err = stateObject.TokenBridgeEmitter()
+		if err != nil {
+			return nil, err
+		}
+
+		suiTokenBridgeAddress, err = stateObject.TokenBridgePackageId()
```
evan-gray: if this is intended to be the current package ID, what is the expected behavior when the package is upgraded? does the verifier just stop verifying until the watcher or guardian is restarted? again, worth specifying whether this is the _current_ or _original_ ID

node/pkg/txverifier/suitypes.go
pleasew8t: I was aware of using the state object's type, but since I want the token registry's package ID (which I guess is the original package ID?) I wanted to be explicit about it.

node/pkg/txverifier/suitypes.go
pleasew8t: I'll do some renaming

node/pkg/txverifier/sui.go
pleasew8t: Will do some renaming here too

node/pkg/watchers/sui/msg_verifier.go
```diff
@@ -0,0 +1,64 @@
...
+		return common.MessagePublication{}, fmt.Errorf("transfer verifier is nil")
+	}
+
+	localMsg := *msg
+
+	var verificationState common.VerificationState
+
+	// If the payload does not represent a transfer, or if the emitter address of the message does
+	// not match the token bridge emitter, mark the message's verification state as NotApplicable.
+	if !vaa.IsTransfer(msg.Payload) || "0x"+localMsg.EmitterAddress.String() != e.suiTxVerifier.GetTokenBridgeEmitter() {
```
panoel, line 37: Do you have to toLowerCase() the strings before comparing?

PR #4330, opened 2025-04-04
Node: Fogo support

bruce-riley:
This PR adds guardian support for Fogo. This PR assumes an HTTP connection like Solana. However, since Fogo has a very short block time (20 ms?), we should investigate using a websocket rather than HTTP. In the meantime, this should be good enough for testnet.

To enable Fogo in testnet, use the following parameters:

```
--fogoRPC YOUR_HTTP_RPC
--fogoContract BhnQyKoQQgpuRTRo6D8Emz93PvXCYfVgHhnrR4T3qhw4
--fogoShimContract EtZMZM22ViKMo4r5y4Anovs3wKQ2owUmDpjygnMMcdEX
```


node/cmd/guardiand/node.go
```diff
@@ -405,6 +411,7 @@ func init() {
 	suiMoveEventType = NodeCmd.Flags().String("suiMoveEventType", "", "Sui move event type for publish_message")
 
 	solanaRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "solanaRPC", "Solana RPC URL (required)", "http://solana-devnet:8899", []string{"http", "https"})
+	fogoRPC = node.RegisterFlagWithValidationOrFail(NodeCmd, "fogoRPC", "Fogo RPC URL (required)", "http://solana-devnet:8899", []string{"http", "https"})
```
panoel, line 414: The default is the solana rpc?

evan-gray: That's not a default, it's an example. https://github.com/wormhole-foundation/wormhole/blob/33dd6a56541c2d15d3e20faa1330ba542a5fa727/node/pkg/node/url_verification.go#L73

PR #4332, opened 2025-04-04
Node: Add Solana shim feature flag

bruce-riley:
This PR adds a the Solana shim to the feature flags published in heartbeat messages (as "solshim") so that we can see in the dashboard if a guardian has it enabled. This feature can be used for other features, such as the Fogo shim.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1726,7 +1726,7 @@ func runNode(cmd *cobra.Command, args []string) {
 		watcherConfigs = append(watcherConfigs, wc)
 
 		if *solanaShimContract != "" {
-			featureFlags = append(featureFlags, "solshim")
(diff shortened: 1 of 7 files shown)


node/cmd/guardiand/node.go
```diff
@@ -1721,6 +1724,10 @@ func runNode(cmd *cobra.Command, args []string) {
 			Commitment:    rpc.CommitmentFinalized,
 		}
 		watcherConfigs = append(watcherConfigs, wc)
+
+		if *solanaShimContract != "" {
+			featureFlags = append(featureFlags, "solshim")
```
evan-gray, line 1729: Would whoever is monitoring want to know the address of the shim that is set?

bruce-riley: Welp, you're the one who was worried about the length of the feature string, but I can add it. 😁

panoel (approved):

panoel (approved):

PR #4335, opened 2025-04-07
github: Add djb15 as root reviewer

djb15:
Proposal to add `djb15` as a second root reviewer


tonyjin (approved):
Approved after discussion with current root approver and Asymmetric Research security leadership.

evan-gray (approved):

PR #4337, opened 2025-04-09
Tilt/CI: Pin foundry release

bruce-riley:
It looks like some of the nightly releases of foundry are disappearing. This PR pins foundry to a specific official release.

### DIFF
--- a/devnet/tx-verifier/Dockerfile.cast
+++ b/devnet/tx-verifier/Dockerfile.cast
@@ -1,7 +1,7 @@
 # These versions are pinned to match the Dockerfile in the `ethereum/`
 # directory. Otherwise, there is nothing special about them and they can be
 # updated alongside the other Dockerfile.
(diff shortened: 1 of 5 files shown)


johnsaigle (approved):
LGTM with respect to the Transfer Verifier changes. I don't believe a specific version of Foundry is required there, and it's meant to be pegged to the main one used for Ethereum anyway.

johnsaigle:
@pleasew8t This will probably conflict with the changes in https://github.com/wormhole-foundation/wormhole/pull/4324, just a heads up

kcsongor (approved):

evan-gray (approved):
confirmed this matches the sha at the time of this commit https://github.com/foundry-rs/foundry/pkgs/container/foundry/391890852?tag=v1.0.0

also, it seems that these images are now [multi-platform](https://docs.docker.com/build/building/multi-platform/) so it should be possible (and encouraged) to drop the `--platform` flag, but that can be done in a follow-up PR.

PR #4338, opened 2025-04-09
cleanup: Remove use of tendermint/random for non-Cosmos code

johnsaigle:
- Removes references to tendermint/random. 
This package was being used to generate random Uint32 and Uint64 values for sequence and nonce values. This dependency was added to the admintemplate many years ago and there isn't a justification in the commit for adding it instead of using std library functions. The original motivation may have been that the Tendermint library claims that it is thread-safe; however Go's standard cryptography libraries have changed a lot and there's no reason to think they are not thread safe.

- Adds new random functionality in the form of Uint32 and Uint64 functions.

These changes make it so that the only parts of the node that reference Tendermint are the ones directly related to Cosmos code. 

Note that both Tendermint and the CometBFT fork have either deleted or deprecated this functionality anyway.
- https://github.com/tendermint/tendermint/issues/2956 
- https://github.com/cometbft/cometbft/blob/main/libs/rand/random.go

### DIFF
--- a/node/cmd/guardiand/admintemplate.go
+++ b/node/cmd/guardiand/admintemplate.go
@@ -14,7 +14,6 @@ import (
 	"github.com/holiman/uint256"
 	"github.com/mr-tron/base58"
 	"github.com/spf13/pflag"
-	"github.com/tendermint/tendermint/libs/rand"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 
 	"github.com/ethereum/go-ethereum/crypto"
@@ -23,6 +22,7 @@ import (
 
 	"github.com/certusone/wormhole/node/pkg/devnet"
 	nodev1 "github.com/certusone/wormhole/node/pkg/proto/node/v1"
+	"github.com/certusone/wormhole/node/pkg/random"
 )
 
 var setUpdateNumGuardians *int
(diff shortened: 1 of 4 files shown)


node/cmd/guardiand/admintemplate.go
```diff
@@ -1252,3 +1273,16 @@ func isValidUint256(s string) (bool, error) {
 
 	return true, nil
 }
+
+// randSeqNonce returns a new random sequence and nonce. It exits if the random numbers can't be generated.
+func randSeqNonce() (uint64, uint32) {
+	seq, err := random.Uint64()
+	if err != nil {
+		log.Fatal("random number: ", err)
```
bruce-riley, line 1281: Not sure it matters, but can you make the log text unique between these two cases.

bruce-riley (approved):

node/cmd/guardiand/admintemplate.go
johnsaigle: The log text will appear as `"random number: failed to generate random <type>: <actual error>"` due to the formatting in `Error()`, so the complete log will distinguish which function failed.

Is that OK?

pleasew8t (approved):

PR #4341, opened 2025-04-10
ci: add golangci-lint linters with no violations

johnsaigle:
Rework of #4307 after the major version upgrade for golangci-lint.

The rules have also been sorted alphabetically.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
     - gosec
     - govet
     - ineffassign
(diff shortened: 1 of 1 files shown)


djb15 (approved):

mdulin2 (approved):
All of these seem like good practices! 

Given that CI still passed when you added these with no additional changes, they seem like things to avoid and not prone to false positives.

PR #4348, opened 2025-04-11
Node/Solana: Minor watcher cleanup

bruce-riley:
This PR does some minor clean up of the Solana watcher. Notably:

- Optimizes the check to see if a transaction might include a wormhole `PostMessage`
- Can't save the `ctx` across a restart since it may have previously been canceled
- Removes logging of the commitment level since it is in the logger name
- Cleans up various other log messages
- A few other minor things
- Don't count the wrong commitment level as an observation

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -819,7 +819,7 @@ func (s *SolanaWatcher) processInstruction(ctx context.Context, rpcClient *rpc.C
 				zap.String("watcher commitment", string(s.commitment)),
 			)
 		}
-		return true, nil
+		return false, nil
 	}
 
 	// The second account in a well-formed Wormhole instruction is the VAA program account.
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -1172,6 +1172,8 @@ func isPossibleWormholeMessage(whLogPrefix string, logMessages []string) bool {
 					return true
 				}
 			}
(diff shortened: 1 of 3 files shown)


node/pkg/watchers/solana/client.go
```diff
@@ -1183,3 +1161,18 @@ func (s *SolanaWatcher) checkCommitment(commitment rpc.CommitmentType, isReobser
...
+
+// isPossibleWormholeMessage searches the logs on a transaction to see if it contains a Wormhole core PostMessage.
+// It looks for a log for the core contract, followed by one for the sequence number.
+func isPossibleWormholeMessage(whLogPrefix string, logMessages []string) bool {
+	for idx := 0; idx < len(logMessages); idx++ {
+		if strings.HasPrefix(logMessages[idx], whLogPrefix) {
+			for idx1 := idx + 1; idx1 < len(logMessages); idx1++ {
+				if strings.HasPrefix(logMessages[idx1], "Program log: Sequence:") {
+					return true
+				}
```
evan-gray, line 1173: since this loops over the remainder of the list, you can add an `else { return false }` here to end earlier (as is, this would go back and continue looping over the rest of the log messages again)

evan-gray (approved):

node/pkg/watchers/solana/client.go
evan-gray: meant to put this comment two lines down :grimacing:

panoel (dismissed):

panoel (approved):

evan-gray (approved):

PR #4350, opened 2025-04-15
Node/Solana: TX for account

bruce-riley:
This PR updates the Solana watcher so that it can use the `getSignaturesForAddress` RPC call to poll for Wormhole observations, and enables the feature for Fogo. (This is the technique used in the dashboard monitor.)

Note that this PR does not affect the behavior for either Solana or Pyth.

The hope is that this technique will work better than the current technique of polling for every block, given Fogo's short block time.

Each interval (one second), it calls `getSignaturesForAddress` to read all transactions involving the Core contract since the last detected transaction. If it finds any transactions, it uses the standard transaction processing code to observe any messages.

This PR also adds the transaction signature to the "message observed" log message, to help with debugging and reobservation.

### DIFF
--- a/node/pkg/watchers/solana/tx_for_addr.go
+++ b/node/pkg/watchers/solana/tx_for_addr.go
@@ -1,0 +1,323 @@
+package solana
+
+// This code queries for Wormhole observations by reading transactions involving the core contract.
+// It takes the start and end blocks to be checked, using one plus the previous current slot as the start,
+// and the current slot as the end. It gets the first transaction in the oldest block and the last transaction
+// in the newest block and uses the `getSignaturesForAddress` RPC call to query for all transactions in that
+// range that involved the Wormhole core contract. It then reads each of those transactions and uses the standard
+// transaction processing code to observe any messages found in those transactions.
+
+// TODO: Get rid of "TEST:" log messages.
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
+	"github.com/gagliardetto/solana-go/rpc/jsonrpc"
+	"go.uber.org/zap"
+)
+
+const (
+	// MaxSignaturesPerQuery is the maximum number of transactions to be returned by `GetSignaturesForAddressWithOpts`.
+	// The max / default is 1000. You can set this to something smaller (like 5) to test pagination.
+	MaxSignaturesPerQuery = 1000
+
+	// NumGetBlockRetries is how many times we will try to query for a slot, allowing for skipped / missing slots.
+	NumGetBlockRetries = 25
+)
+
+/* TODO: Delete this code if we end up not needing it.
+// transactionProcessor is the entry point of the runnable that periodically queries for new Wormhole observations.
+// It uses the standard `DefaultPollDelay`, although the timing will vary based on query delays. Each interval, it gets
+// the latest slot. It uses that slot and one plus the latest slot of the previous interval to determine a range of slots.
+// It then invokes the function that queries for Wormhole transactions in slots in that range.
+func (s *SolanaWatcher) transactionProcessor(ctx context.Context) error {
+	timer := time.NewTicker(DefaultPollDelay)
+	defer timer.Stop()
+
+	// Keep track of the last slot of the previous interval which determines the oldest slot we need to query next time.
+	var oldestSlot uint64
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
+			if newestSlot, err := s.processLatestTransactions(ctx, oldestSlot); err != nil {
+				s.logger.Error("failed to get transactions", zap.Error(err))
+				s.errC <- err
+				return err
+			} else {
+				oldestSlot = newestSlot
+			}
+		}
+	}
+}
+
+// processLatestTransactions gets the latest slot and then invokes the function that queries for Wormhole transactions in slots in the specified range.
+func (s *SolanaWatcher) processLatestTransactions(ctx context.Context, oldestSlot uint64) (uint64, error) {
+	rCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	newestSlot, err := s.rpcClient.GetSlot(rCtx, s.commitment)
+	cancel()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get latest slot: %w", err)
+	}
+
+	if oldestSlot == 0 {
+		// This is the startup scenario.
+		oldestSlot = newestSlot - 1
+	} else if oldestSlot == newestSlot {
+		s.logger.Debug("not getting transactions, slot number has not advanced", zap.Uint64("slotNum", oldestSlot))
+		return oldestSlot, nil
+	} else {
+		oldestSlot++
+	}
+
+	err = s.processTransactionsForSlots(oldestSlot, newestSlot)
+	if err != nil {
+		return 0, err
+	}
+
+	return newestSlot, nil
+}
+*/
+
+// processTransactionsForSlots queries for the transactions for a range of slots and processes any core events.
+func (s *SolanaWatcher) processTransactionsForSlots(oldestSlot uint64, newestSlot uint64) error {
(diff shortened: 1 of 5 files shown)


node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,317 @@
...
+
+// processTransactionsForSlots queries for the transactions for a range of slots and processes any core events.
+func (s *SolanaWatcher) processTransactionsForSlots(oldestSlot uint64, newestSlot uint64) error {
+	if newestSlot < oldestSlot {
+		// We probably got load balanced. Just wait until next time.
+		s.logger.Debug("not getting transactions, slot number went backwards", zap.Uint64("lastSlot", oldestSlot), zap.Uint64("newestSlot", newestSlot))
+		return nil
+	}
+
+	newestBlock, err := s.findNextValidBlock(newestSlot, true, NumGetBlockRetries)
```
evan-gray: if this decrements, wouldn't it be possible to have misses because in `processLatestTransactions` it returns the `newestSlot` based on `s.lastSlot` as opposed to the "next valid block" that this resolves to (which is actually the previous valid block)? this may also be an issue with the dashboard code this was based on.

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,317 @@
+package solana
+
+// This code queries for Wormhole observations by reading transactions involving the core contract.
+// It takes the start and end blocks to be checked, using one plus the previous current slot as the start,
+// and the current slot as the end. It gets the first transaction in the oldest block and the last transaction
```
evan-gray: I'm not 100% that this will work correctly, based on my recollection / the comments in the dashboard watcher, `getSignaturesForAddress` is _exclusive_ so we did (first block - 1) and (last block + 1)

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,317 @@
...
+
+	// Get the block for the slot, retrying if the block is not yet available (probably because we are behind a proxy).
+	var block *rpc.GetBlockResult
+	var err error
+	for retries := maxRetries; retries > 0; retries -= 1 {
+		rewards := false
+		maxSupportedTransactionVersion := uint64(0)
+		block, err = s.rpcClient.GetBlockWithOpts(s.ctx, uint64(slot), &rpc.GetBlockOpts{
+			Encoding:                       solana.EncodingBase64, // solana-go doesn't support json encoding.
+			TransactionDetails:             "full",
```
evan-gray: do you need `full` or can you use only `signatures`? if changed, the above functions would have to change to use `Signatures` instead of `Transactions`

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,317 @@
...
+	return tx.Signatures[len(tx.Signatures)-1], nil
+}
+
+// findNextValidBlock queries for a block associated with a slot. If the slot was skipped or is missing,
+// it looks for the "next" one, based on the `decrement` flag. If that is true, we decrement the slot number.
+// If it is false, we increment it. This process continues until we find a valid block or the number of
+// retries is exhausted.
+func (s *SolanaWatcher) findNextValidBlock(slot uint64, decrement bool, retries int) (*rpc.GetBlockResult, error) {
+	if retries == 0 {
+		return nil, errors.New("no block found after exhausting retries")
```
evan-gray: what will effectively happen if we have 25 skipped slots in a row? will the watcher restart?

evan-gray (changes requested):
I'm a little concerned about the edge case around `findNextValidBlock`. It would be good if we could find a way to test the various edge cases to ensure we're doing the right thing.

node/pkg/watchers/solana/tx_for_addr.go
bruce-riley: Yes, this will kill the watcher. Does that make sense?

evan-gray: I suppose the resulting gaps are no different than the existing watcher, at least as long as that is the case. I'm assuming that the watcher will restart and pick the current slot as the last slot (and not get stuck in a loop with the same last slot failing this look back).

That said, if there were 25 skipped slots, all of the guardian's fogo watchers might then restart and that delay could result in misses. I don't know enough about the logistics / distribution of fogo's block generation / voting to determine the likelihood of that so that makes me iffy on this behavior. I'm wondering if there's a cleaner way where we can be more certain of getting a result. Perhaps by always looking backwards for the prior block with transactions _and_ storing the last transaction of the first block we were interested in... something like that. Again, this might be worth a document with some more annotation because the exclusivity of the signature search makes this a bit tricky. There also might be a cleaner way to achieve this whole thing... https://solana.com/docs/rpc/http/getsignaturesforaddress doesn't require a `before` so could we do one query with no `before` or `until` with a limit of `1` to get our starting point and then use that signature as our `until` on the next query (or something like that) which feels akin to how we handled one of the other watchers.

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,183 @@
...
+
+// getPrevWormholeSignature reads the most recent transaction involving the Wormhole core contract and returns it.
+func (s *SolanaWatcher) getPrevWormholeSignature() (solana.Signature, error) {
+	limit := int(1)
+	signatures, err := s.rpcClient.GetSignaturesForAddressWithOpts(s.ctx, s.contract, &rpc.GetSignaturesForAddressOpts{
+		Commitment: s.commitment,
+		Limit:      &limit,
+	})
+
+	if err != nil || len(signatures) == 0 {
```
evan-gray, line 76: just noting that I *think* this would be okay even for a new core bridge deployment because the core bridge must at least have been initialized first, meaning there would realistically always be at least 1 most recent transaction.

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,183 @@
...
+		return solana.Signature{}, err
+	}
+
+	return signatures[len(signatures)-1].Signature, nil
```
evan-gray: obviously the limit is `1` above, but aren't the transactions returned in descending order so the most recent would be `signatures[0]`? (`getTransactionSignatures` reverses them, but this is using the direct RPC result)

node/pkg/watchers/solana/tx_for_addr.go
bruce-riley: Yes, that was my assumption. Don't start up the watcher until after you initialize the core.

node/pkg/watchers/solana/tx_for_addr.go
bruce-riley: This code has been completely redesigned, so I'm marking this resolved.

node/pkg/watchers/solana/tx_for_addr.go
bruce-riley: This code has been completely redesigned, so I'm marking this resolved.

node/pkg/watchers/solana/tx_for_addr.go
bruce-riley: We discussed this offline and came up with a better approach that does not require tracking blocks. Marking this issue as resolved.

node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -0,0 +1,183 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	// MaxSignaturesPerQuery is the maximum number of transactions to be returned by `GetSignaturesForAddressWithOpts`.
+	// The max / default is 1000. You can set this to something smaller (like 5) to test pagination.
+	MaxSignaturesPerQuery = 1000
+
+	// NumGetBlockRetries is how many times we will try to query for a slot, allowing for skipped / missing slots.
+	NumGetBlockRetries = 25
```
evan-gray: I think this is unused now

evan-gray (changes requested):
just a couple nits, overall I think it's cleaner now

evan-gray (approved):

panoel (approved):

PR #4352, opened 2025-04-15
Node: Clean up feature flags

bruce-riley:
This feature cleans up the guardian feature flags published in heartbeat messages and displayed in the dashboard.

It does the following:

- Removes the individual chains from the IBC flag (since they are hard coded for a release). It keeps the wormchain version. Listing all the chains was taking up too much space in the dashboard.
- Adds a feature flag for the TX Verifier feature.
- Introduces the concept of static and dynamic feature flags. The static ones are built once on initialization and passed into `p2p.go` via the `featureFlags` parameter. Dynamic feature flags are generated each heartbeat interval and are passed into `p2p.go` via the `featureFlagFuncs` parameter. These functions are evaluated each interval and the results added to the features list.
- Sorts the feature flags so the order in the dashboard makes more sense.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -966,6 +966,13 @@ func runNode(cmd *cobra.Command, args []string) {
 		if parseErr != nil {
 			logger.Fatal("transferVerifierEnabledChainIDs input is invalid", zap.Error(parseErr))
 		}
+
+		// Format the feature string in the form "txverifier:ethereum|sui" and append it to the feature flags.
+		chainNames := make([]string, 0, len(txVerifierChains))
+		for _, cid := range txVerifierChains {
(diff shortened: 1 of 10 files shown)


johnsaigle (approved):
Looks good. Is there a place in the documentation that lists the supported chains though, or is it possible to find out only by reading the source code?

bruce-riley:
> Looks good. Is there a place in the documentation that lists the supported chains though, or is it possible to find out only by reading the source code?

You're so speed! I'm not done making changes yet. 😁 

The list of chains is hard coded, so if you have a release number (in the dashboard), you can look at the code to see what chains are using IBC.

johnsaigle (approved):
Looks good. I'm wondering though whether Transfer Verifier should be a static flag given this definition?

panoel (approved):

PR #4353, opened 2025-04-15
ci: detect unused function arguments with unparam

johnsaigle:
# Overview
Enables unparam, flagging unused arguments, and fixes violations. In most cases I renamed the argument to `_` instead of doing deeper refactoring.

To reproduce the warnings, apply the golangci-lint changes to origin/main and run `make lint`.

There is also some minor cleanup in this PR, updating a handful of comments and moving magic numbers into constants.

# Further Work

## Delete deterministic key code?

`node/pkg/devnet/deterministic_bridge_key.go` used to generate keys deterministically but now just reads them from a fixed slice. It seems like it would be worth moving this functionality to the SDK, as that's where a lot of the more modern devnet code lives.

## Refactor GetGovernorChainData

This function is set-up to work dynamically, but it has only one caller that hard-codes a `time.Now` argument, and the callee ignores this value anyway. Right now I've marked this with nolint/TODO but this code should be reworked.


sdk/vaa/payloads.go
```diff
@@ -276,6 +276,9 @@ type (
 	}
 )
 
+// However, this would be a breaking change for integrators that use this function
```
bruce-riley, line 279: The "However" sounds odd here.

bruce-riley (dismissed):

sdk/vaa/payloads.go
johnsaigle: Ah yeah thanks, the order of these lines is backward. I can fix this

PR #4362, opened 2025-04-29
node: Fix wrapped asset query errors for the Transfer Verifier

johnsaigle:
# Bug fixes
Fixes several bugs in the Transfer Verifier concerning the querying of wrapped assets:

## Wrapped asset information was not parsed correctly

This line is wrong: https://github.com/wormhole-foundation/wormhole/blob/71127ab1e86db07b1a02e6103c9db305b2601157/node/pkg/txverifier/evmtypes.go#L788

`binary.BigEndian.Uint16()` only grabs the first two bytes of a slice, but we need the last 2 (of the total 32 from the EVM word length). So this returned 0 for every query to chainId()

## Request In was not using Origin Address

This resulted in the wrapped address being compared with the origin address when calculating amounts coming in vs. amounts going out. We didn't encounter this issue because (un)wrapping assets always failed.

## Wrong type for Origin Address 

Origin Address was being represented as a go-ethereum address. This is incorrect for non-EVM addresses. It is now a `vaa.Address` which is an alias for `[32]byte`.

# Better testing

* It's now possible to run the standalone binary on a single receipt, after which the program exits.
This should allow us to more quickly test real data. Previously, one had to leave the binary running for a long time and pray that you found a relevant transaction.
In general, this form of testing should be preferred over e.g. extensive mocking and closing testnet data. That approach will simply take a long time to get the relevant data configured, and the complexity involved with mocking all of the relevant calls and data is error prone.

# Cleanup/refactor

* Origin Address added to ERC20 transfers for convenience of comparison
* Instead of using the results of a `wrappedAsset()` call to get decimal and unwrapped address information, the `decimals()` and `nativeContract()` (origin address) are queried directly. 
* I changed the names of several of the RPC query methods to more closely match the RPC calls made to the Token Bridge. This should (hopefully) make the code easier to follow and check against Token Bridge Info.


node/pkg/txverifier/evm.go
```diff
@@ -639,43 +643,56 @@ func parseLogMessagePublishedPayload(
...
+	var (
+		originAddress vaa.Address
+		newDetails    = *details
+		decimals      uint8
+		// The token address that will be queried for decimal information.
+		queryAddr common.Address
+	)
+	// At this point, details.TokenChain should be set to the chain ID in the token transfer's payload header.
+	// TODO: I'm not sure that this is correct. The address should be validated either way.
 	if details.TokenChain == tv.chainIds.wormholeChainId {
```
mdulin2, line 674: I'm confused about the necessity of this code. Here's my understanding of it currently: 
- If the TokenChain is the same as the current chain, the origin address is set to the current `OriginAddressRaw` from the payload. That makes sense. 
- In the case that the TokenChain is *different* than the current chain, we do some other logic:
    - Query the OriginAddressRaw and the TokenChain with `wrappedAddr` to fetch its locally deployed token. Again, that makes sense. 
    - Call `nativeContract` with the address from the previous step - the token implementation. 
    - Set the result of the `nativeContact` call to the `originAddress`. 

What I'm confused about is the circular usage in the non-native case. The result of the nativeContract is the `OriginAddressRaw`, right?

mdulin2 (commented):
Just a few things to think about.

node/pkg/txverifier/evm.go
```diff
@@ -19,18 +17,17 @@ import (
...
 )
 
-// ProcessEvent processes a LogMessagePublished event, and is either called
-// from a watcher or from the transfer verifier standalone process. It fetches
+// TODO: More errors should be converted into sentinel errors instead of being created and returned in-line.
+var (
+	ErrNotTransfer = errors.New("payload is not a token transfer")
+)
+
+// ProcessEvent processes a LogMessagePublished event. It fetches
```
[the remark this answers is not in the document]
johnsaigle: > are any of the other errors happening in ProcessEvent acceptable to fail open in production or just for testnet?

Everything in this package should be considered permanent. Anything related to prod vs. mainnet is a consequence of whether the Notary is merged.

> , there's a difference between Valid with transfers, no transfers and invalid processing that's not being considered here.

The function is meant to _only_ return false if the receipt is _invalid_ with transfers. Everything else returns true. I can update the code comment here to make this more clear.

node/pkg/txverifier/evm.go
```diff
@@ -155,17 +152,21 @@ func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
 // Do additional processing on the raw data that has been parsed. This
```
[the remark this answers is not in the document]
johnsaigle: Yeah that line should probably return an error instead of just logging one.

node/pkg/txverifier/evm.go
johnsaigle: Hmmmm yeah I think you're right, I don't remember why this was written this way. I'll try deleting it and see if something jumps out.

node/pkg/txverifier/evm.go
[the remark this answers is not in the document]
johnsaigle: @mdulin2 I refactored the function so now it doesn't fail open or fail closed. Check out the updated PR description for details and please re-review when you have a chance.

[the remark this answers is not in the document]
mdulin2: I like this refactor - it is less ambiguous now. Thanks!

node/pkg/txverifier/evm.go
mdulin2: Okay, I'm glad I'm not crazy. Thanks for considering.

node/pkg/txverifier/evm.go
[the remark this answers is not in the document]
mdulin2: > Everything in this package should be considered permanent. Anything related to prod vs. mainnet is a consequence of whether the Notary is merged.

Cool, I'll keep this in mind.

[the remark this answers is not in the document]
mdulin2: I read through the rewrite of it. I _assume_ that the notary would probably let a receipt in this invalid state to pass for the time being? 

My hope is that we can effectively guarantee that this edge case is never reached. Like the bug report, the end result would still be the same—bypassing the transfer verifier. 

To me, this looks like a really good opportunity to run it on the token bridge and see if anything unexpected fails. If it does, then we're doing something wrong again. This should effectively always succeed except in the case of a major hack.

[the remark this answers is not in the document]
johnsaigle: Yeah I haven't updated the Notary yet, I think I'll put that PR in draft mode until this is merged. After this PR is merged, we can decide what we want to do when messages can't be parsed properly. Maybe we can reuse `Anomalous` for this case.

node/pkg/txverifier/evm.go
johnsaigle: This part of the code is totally different in the latest commit.

devnet/tx-verifier/tx-verifier-evm-runner.sh
```diff
@@ -22,10 +22,10 @@ guardiand \
 current_count=$(grep -c "$error_pattern" "$log_file")    
 echo "Found ${current_count} of ${TARGET} instances"
 
-# if we found the requisite number of error messages, we can exit
-if [ $current_count -ne $TARGET ]; then 
-    echo "Tests failed. Only found ${current_count} of ${TARGET} required log messages"
+# if we found at least the requisite number of error messages, we can exit
+if [ $current_count -lt $TARGET ]; then
```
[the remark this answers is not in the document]
johnsaigle: The standalone CLI tool basically has two levels of logging. There's an "outer" layer that is reporting information about the tool itself, and the "inner" layer is the Transfer Verifier code. Both outputs appear in the STDOUT of the process.

CLI was failing because it was expecting exactly two error messages. That's logically correct, but each error message got logged twice because of the two layers.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,212 @@ import (
...
 	if receipt == nil {
 		tv.logger.Debug("receipt was not passed as an argument. fetching it using the connector")
 		var txReceiptErr error
 		receipt, txReceiptErr = tv.evmConnector.TransactionReceipt(ctx, txHash)
+
 		if txReceiptErr != nil {
-			tv.logger.Warn("could not find core bridge receipt", zap.Error(txReceiptErr))
-			return true
+			tv.addToCache(txHash, &evaluation{receipt, false, txReceiptErr})
+			return false, txReceiptErr
```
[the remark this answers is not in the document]
johnsaigle: Yeah I think you're right. We could prune fairly often for these look ups and/or not cache errors. What do you think?

node/pkg/txverifier/evm.go
```diff
@@ -420,6 +537,8 @@ func (tv *TransferVerifier[evmClient, connector]) ParseReceipt(
 				continue
 			}
 
+			tv.logger.Debug("parsed TransferDetails from LogMessagePublished", zap.String("transferDetails", transferDetails.String()))
+
 			// If everything went well, append the message publication
 			messagePublications = append(messagePublications, &LogMessagePublished{
 				EventEmitter:    log.Address,
```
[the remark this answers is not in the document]
johnsaigle: Good question. I do like the idea of reworking this so that the struct is only built when everything is guaranteed to be correct. I'm a bit hesitant to make this PR bigger though. Could we do it in a follow-up PR instead?

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,212 @@ import (
...
 }
 
-// Do additional processing on the raw data that has been parsed. This
-// consists of checking whether assets are wrapped for both ERC20
-// Transfer logs and LogMessagePublished events. If so, unwrap the
-// assets and fetch information about the native chain, native address,
-// and token decimals. All of this information is required to determine
+// updateReceiptDetails performs additional processing on the raw data that has been parsed. This
+// consists of checking whether assets are wrapped. If so, unwrap the
+// assets and fetch information about the origin chain, origin address,
```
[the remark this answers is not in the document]
johnsaigle: Yeah the assumption is that the core bridge might be untrustworthy, so this is looking up the ERC20 information with respect to the Token Bridge instead and comparing that with the core bridge's message.

node/pkg/txverifier/evm.go
```diff
@@ -183,95 +229,129 @@ func (tv *TransferVerifier[ethClient, Connector]) UpdateReceiptDetails(
...
-	tv.logger.Debug("finished updating receipt details")
-	return nil
-}
+			// Update ChainID
+			var fetchErr error
+			originChain, fetchErr = tv.chainId(queryAddr)
+			if fetchErr != nil {
+				return errors.Join(errors.New("chainId: Can't get chainId for token. Can't continue to process this receipt"), fetchErr)
+			}
+			if originChain == vaa.ChainIDUnset {
```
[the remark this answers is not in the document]
djb15: That's fair, it's not an expensive check

node/pkg/txverifier/evm.go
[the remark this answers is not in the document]
djb15: Yeah not an issue to leave it in, it's probably just a bit cleaner to rework it. Happy for it to go in a follow up PR

node/pkg/txverifier/evm.go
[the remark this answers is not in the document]
djb15: I was leaning towards not caching errors but not sure if that opens up a DoS risk? Could an attacker specify an arbitrary `txHash` that will always fail or is the input controlled/bounded by whatever comes from the watchers?

node/pkg/txverifier/evm.go
```diff
@@ -581,23 +695,34 @@ func (tv *TransferVerifier[evmClient, connector]) ProcessReceipt(
...
-			tv.logger.Error("transfer-out request for tokens that were never deposited",
-				zap.String("key", key))
-			localErr = &InvariantError{Msg: "transfer-out request for tokens that were never deposited"}
+			invariantErrors = errors.Join(invariantErrors, &InvariantError{Msg: "transfer-out request for tokens that were never deposited"})
 		} else {
+			// TODO: We may want to add an allow-list for fee-on-transfer and rebasing tokens, or else
+			// they will fail validation here. The reason is that the Core Bridge's amount out is a function
+			// of the Token Bridge calling `balanceOf()` when a token transfer occurs. If this method is implemented
+			// in a strange way in the token contract, the amount out may be greater than the amount sent in
+			// and hence break the invariant testing below.
```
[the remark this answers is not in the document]
johnsaigle: This is the only one that I've encountered so far.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,212 @@ import (
...
-	return true
+	// Cache results
+	eval.Err = nil
+	eval.Result = true
+	tv.addToCache(txHash, &eval)
+
+	return true, nil
 }
 
 func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
```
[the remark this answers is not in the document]
johnsaigle: Do you feel very strongly about this change? It should prune on a rolling window based on block height, currently set to `20` -- so a receipt should be pruned after 20 blocks.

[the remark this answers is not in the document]
djb15: For Ethereum you'll be getting hundreds of transactions max per block so 20,000+ max transactions in the cache. Other EVM chains usually have faster but smaller blocks, so the situation improves. So I think it's sensibly bounded, but would be nice to keep an eye on what the cache is doing if we ever have issues; not tied to using prometheus here, just having the metrics exposed somewhere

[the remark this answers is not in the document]
djb15: I kinda like that idea, would be nice not to reinvent the wheel here and have a go-to caching primitive. But (I'm sure you'd agree) this feels like feature creep on this PR, so would consider that better placed for a future refactor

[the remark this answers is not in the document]
johnsaigle: Sure, so for now I could add a log message and we could consider the LRU cache as something to do in the future?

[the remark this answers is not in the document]
djb15: Yep that's fine by me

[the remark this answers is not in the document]
johnsaigle: @djb15 I added more lines to clear the caches in this area of the code. When they're cleared, an Info-level log is emitted.

node/pkg/txverifier/evm.go
```diff
@@ -181,12 +181,101 @@ func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
...
 	}
-	// TODO: The other caches should be pruned here too.
+
+	// For the other caches, just prune them if they're too big.
+	if len(tv.chainIdCache) >= CacheMaxSize {
+		var (
+			i = 0
+			// Slice to store keys to delete in order to avoid deleting keys from maps that are being
+			// iterated over.
+			keysToDelete = make([]string, CacheDeleteCount)
```
djb15: Instead of having the delete count being a fixed number, should the count be `max(CacheDeleteCount, len(tv.chainIdCache) - CacheMaxSize)`? Or simply the latter, where you trim to `CacheMaxSize`? Otherwise this could theoretically grow indefinitely?

node/pkg/txverifier/evm.go
```diff
@@ -181,12 +181,101 @@ func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
...
-		tv.logger.Debug("pruned cached receipt evaluations",
+		tv.logger.Info("pruned receipt evaluations cache",
 			zap.Int("numPruned", numPrunedEvals),
 			zap.Int("cacheSize", len(tv.evaluations)),
 		)
 	}
-	// TODO: The other caches should be pruned here too.
+
+	// For the other caches, just prune them if they're too big.
+	if len(tv.chainIdCache) >= CacheMaxSize {
```
djb15, line 185: Worth making this a helper function instead of copying multiple times for each cache?

node/pkg/txverifier/evm.go
```diff
@@ -181,12 +181,101 @@ func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
...
+		var (
+			i = 0
+			// Slice to store keys to delete in order to avoid deleting keys from maps that are being
+			// iterated over.
+			keysToDelete = make([]string, CacheDeleteCount)
+		)
+		for key := range tv.chainIdCache {
```
djb15: This is where the LRU cache would be good to avoid a scenario where the first X keys are the most common and get pruned every time. Nice suggestion to add that in a follow up PR!

node/pkg/txverifier/evm.go
johnsaigle: It gets a little annoying because the keys have different types (string vs common.Address) but I could do this

node/pkg/txverifier/evm.go
johnsaigle: My idea was basically like "if this gets to 100 or above, just remove 10 entries". (arbitrary numbers)
How would this grow indefinitely?

node/pkg/txverifier/evm.go
djb15: Would consider this a nice to have if you're squeezed for time

node/pkg/txverifier/evm.go
djb15: I commented in the wrong place, doesn't apply to the chain id cache. But for the decimal cache for instance, it's theoretically possible to have more than 10 new decimals (i.e. tokens) per receipt right? So unless I'm misunderstanding I think you could grow by more than you prune on each iteration

johnsaigle: This only fires if there are over `CacheMaxSize` (100) entries in the cache though, e.g. see L213. And when it does, it only removes 10 entries. I'm not sure how the scenario you're imaging could happen.

node/pkg/txverifier/utils.go
```diff
@@ -154,3 +154,53 @@ func ValidateChains(
...
+
+	if currentLen <= CacheMaxSize {
+		return 0, nil // Nothing to delete, cache is within limits.
+	}
+
+	// Bounds check
+	// NOTE: cache length must be greater than or equal to CacheMaxSize.
+	if len(cache) < CacheMaxSize {
+		return 0, nil
+	}
```
djb15: Aren't you doing practically the same check immediately above?

node/pkg/txverifier/utils.go
```diff
@@ -154,3 +154,53 @@ func ValidateChains(
...
+
+	// Delete either a fixed number of entries or else delete enough so that the cache
+	// will be at CacheMaxSize after the operation.
+	deleteCount := max(CacheDeleteCount, len(cache)-CacheMaxSize)
```
djb15: Use `currentLen` instead of `len(cache)`?

node/pkg/txverifier/utils.go
johnsaigle: Oops yes I think this is a bad copy paste. I'll clean it up

djb15 (approved):
Thanks for making those changes 🙏

devnet/tx-verifier-evm.yaml
```diff
@@ -17,7 +17,9 @@ spec:
           image: tx-verifier-evm
           env:
             - name: ERROR_PATTERN
-              value: "invalid receipt: no deposits and no transfers"
+              # NOTE: This value must match the string used in the invariant violation branch
+              # // of the Transfer Verifier CLI tool.
```
pleasew8t: nit: Can remove the `//` sequence from the comment

node/cmd/txverifier/evm.go
```diff
@@ -151,7 +206,92 @@ func runTransferVerifierEvm(cmd *cobra.Command, args []string) {
...
+					ErrInvariant.Error(),
+					zap.String("txHash", vLog.Raw.TxHash.String()),
+				)
+			}
+			logger.Debug("done processing", zap.Bool("result", valid), zap.String("txHash", vLog.Raw.TxHash.String()))
 		}
 	}
 }
+
+type sanityCheck struct {
```
pleasew8t: style: move this code to the top of the file, so it's "defined before using it". functionally doesn't matter, so feel free to ignore.

node/cmd/txverifier/evm.go
```diff
@@ -151,7 +206,92 @@ func runTransferVerifierEvm(cmd *cobra.Command, args []string) {
...
+		true,
+		nil,
+	},
+	// An NFT transfer. Ensures that ERC721 transfers are not interpreted as ERC20 transfers that the program should analyze.
+	{
+		common.HexToHash(`0x5550571b9e7cee04db0e93b75cd6df655d356e3a9913c392a075d5e50dda1f2c`),
+		false,
+		txverifier.ErrNoMsgsFromTokenBridge,
+	},
+	// This is a deflationary token, and in this case the amount out of the token bridge is greater than
```
pleasew8t, line 292: I wonder what the correct way of handling this is. I checked Wormholescan, and the transfer was completed. That makes sense, because there is no current mechanism similar to transfer verifier that inspects the incoming deposit. I'll spend some time thinking about this, but we should probably try and do something about it.

node/pkg/txverifier/README.md
```diff
@@ -1,43 +1,46 @@
...
+├── sui.go
+├── suitypes.go
+├── utils_test.go
+└── utils.go
+
+1 directory, 10 files
 ```
 
 The package is organized by runtime environment. Currently there are implementations for the Ethereum and Sui blockchains.
 Because the Ethereum implementation is (hopefully) generalizable to other EVM-chains, it is referred to as
```
pleasew8t, line 6: formatting: newline wonkiness going on here. Some sentences are broken up by newline halfway through, and new sentences, like the one here, I assume are meant to be part of the same paragraph as the previous line.

node/pkg/txverifier/evmtypes.go
```diff
@@ -37,6 +35,26 @@ const (
 	EVENTHASH_WETH_DEPOSIT = "0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c"
 )
 
+// Constants related to caching RPC calls and receipt evaluations.
+const (
+	// CacheMaxSize is the maximum number of entries that should be in any of the caches.
+	CacheMaxSize = 100
+	// CacheDeleteCount specifies the number of entries to delete from a cache once it reaches CacheMaxSize.
```
pleasew8t: Shouldn't `CacheDeleteCount` and `CacheMaxSize` be moved to `utils.go`? Seems like a more logical place to put it, since this isn't EVM specific.

node/pkg/txverifier/utils_test.go
```diff
@@ -279,3 +281,251 @@ func TestValidateChains(t *testing.T) {
 		})
 	}
 }
+
+func Test_deleteEntries_StringKeys(t *testing.T) {
+	tests := []struct {
+		name         string
+		setupCache   func() *map[string]vaa.Address
+		want         int
```
pleasew8t: nit: rename to something more descriptive, like `wantNumPruned`

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
 	}
 
-	// Caching: record a new lastBlockNumber.
-	tv.lastBlockNumber = receipt.BlockNumber.Uint64()
-	tv.processedTransactions[txHash] = receipt
+	eval := evaluation{Receipt: receipt}
+
+	// Caching: record a new lastBlockNumber and add the receipt to the cache.
+	if receipt.BlockNumber != nil {
+		tv.lastBlockNumber = receipt.BlockNumber.Uint64()
```
pleasew8t, line 97: Is there a case here for this to be monotonically increasing, rather than always setting it to the last inspected block number?

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
+	}
 
 	// Parse raw transaction receipt into high-level struct containing transfer details.
-	transferReceipt, parseErr := tv.ParseReceipt(receipt)
+	transferReceipt, parseErr := tv.parseReceipt(receipt)
+
+	if parseErr != nil {
+		eval.Err = parseErr
```
pleasew8t: These evals are never written to the cache. The function returns short of calling `tv.addToCache()`. If that is expected, then I'd remove the `Err` logging from the`evaluation` struct, or just avoid setting the property alltogether.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
+		}
+
+		// This represents an invariant violation in token transfers.
+		tv.logger.Error("invariant violation", zap.Error(processErr), zap.String("receipt summary", summary.String()))
+		// The error is deliberately discarded in this case, but is logged above.
+		return false, nil
 	}
 
 	// Update statistics
 	if summary.logsProcessed == 0 {
```
pleasew8t: Just noting that we only log that the summary is `nil`, earlier. If summary was `nil`, this line would cause a `nil`-pointer exception.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
-			zap.Error(updateErr))
-		return true
+		eval.Err = updateErr
+		return false, updateErr
 	}
 
 	// Ensure that the amount coming in is at least as much as the amount requested out.
-	summary, processErr := tv.ProcessReceipt(transferReceipt)
-	tv.logger.Debug("finished processing receipt", zap.String("summary", summary.String()))
+	summary, processErr := tv.processReceipt(transferReceipt)
```
pleasew8t: nit: Does it make sense to rename this function to `validateReceipt`? As I was reading this, all the different functions that handle receipts kinda sounded the same.

node/pkg/txverifier/evm.go
```diff
@@ -637,88 +777,37 @@ func parseLogMessagePublishedPayload(
...
-	denormalized := denormalize(details.AmountRaw, decimals)
-
-	newDetails = details
-	newDetails.OriginAddress = originAddress
-	newDetails.Amount = denormalized
-	return newDetails, nil
 }
 
 // upsert inserts a new key and value into a map or update the value if the key already exists.
 func upsert(
```
pleasew8t: style: Better placed in the utils.go file perhaps?

node/pkg/txverifier/evm.go
```diff
@@ -637,88 +777,37 @@ func parseLogMessagePublishedPayload(
...
 	dict *map[string]*big.Int,
 	key string,
 	amount *big.Int,
-) {
+) error {
+	if dict == nil || amount == nil {
```
pleasew8t: shouldn't there be a check here to ensure `*dict != nil` as well?

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
+	var numPrunedEvals int
+
 	// Iterate over recorded transaction hashes, and clear receipts older than `pruneDelta` blocks
-	for hash, receipt := range tv.processedTransactions {
-		if receipt.BlockNumber.Uint64() < tv.lastBlockNumber-tv.pruneHeightDelta {
-			numPrunedReceipts++
-			delete(tv.processedTransactions, hash)
+	for hash, eval := range tv.evaluations {
+		if eval.Receipt != nil {
+			if eval.Receipt.BlockNumber.Uint64() < tv.lastBlockNumber-tv.pruneHeightDelta {
```
pleasew8t, line 165: Another reason to potentially make `lastBlockNumber` monotonically increasing is to ensure that processing old transactions don't cause memory usage to blow up

node/pkg/txverifier/README.md
```diff
@@ -1,43 +1,46 @@
 # Transfer Verifier - Package Documentation
 
-## Package Structure
+## Overview
 
 ```
+
```
pleasew8t: nit: I would just remove this directory structure. If we wanted to document the different files, a table would look nicer and be easier to maintain.

node/pkg/txverifier/evm_test.go
```diff
@@ -761,20 +770,20 @@ func TestNoPanics(t *testing.T) {
 	defer mocks.ctxCancel()
 
 	require.NotPanics(t, func() {
```
pleasew8t, line 772: Nothing to change. Just giving this an upvote. Did not realise `require.NotPanics()` was a thing.

node/pkg/txverifier/evm_test.go
```diff
@@ -22,14 +22,22 @@ import (
...
 var (
 	coreBridgeAddr  = common.HexToAddress("0x98f3c9e6E3fAce36bAAd05FE09d375Ef1464288B")
 	tokenBridgeAddr = common.HexToAddress("0x3ee18B2214AFF97000D974cf647E7C347E8fa585")
-	nativeAddr      = common.HexToAddress("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2") // weth
-	usdcAddr        = common.HexToAddress("0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48")
-	eoaAddrGeth     = common.HexToAddress("0xbeefcafe")
-	eoaAddrVAA, _   = vaa.BytesToAddress([]byte{0xbe, 0xef, 0xca, 0xfe})
+
+	// WETH
+	nativeAddrGeth   = common.HexToAddress("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2")
```
pleasew8t, line 31: What's with the reference to `Geth`, here? You mean Ethereum?

node/pkg/txverifier/evm.go
```diff
@@ -431,14 +572,7 @@ func (tv *TransferVerifier[evmClient, connector]) ParseReceipt(
...
 	if len(messagePublications) == 0 {
-		if receiptErr == nil {
-			// There are no valid message publications, but also no recorded errors.
-			// This occurs when the core bridge emits a LogMessagePublished event but it is not sent by
-			// the Token Bridge. In this case, just return nil for both values.
-			return nil, nil
-		}
-		// If other errors occurred, also mention that there were no valid LogMessagePublished events.
-		receiptErr = errors.Join(receiptErr, errors.New("parsed receipt but received no LogMessagePublished events"))
+		receiptErr = errors.Join(receiptErr, ErrNoMsgsFromTokenBridge)
```
pleasew8t, line 581: nit: rename error to `ErrNoLogsFromCoreBridge`

node/pkg/txverifier/evm.go
```diff
@@ -494,22 +618,25 @@ func (tv *TransferVerifier[evmClient, connector]) ProcessReceipt(
 		zap.String("receipt", receipt.String()),
 	)
 
-	summary = NewReceiptSummary()
+	summary := NewReceiptSummary()
 
 	if len(*receipt.MessagePublications) == 0 {
-		return summary, errors.New("no message publications in receipt")
+		return nil, ErrNoMsgsFromTokenBridge
```
pleasew8t: This check can be removed, since we already checking this in `receipt.Validate()`

node/pkg/txverifier/evm.go
```diff
@@ -494,22 +618,25 @@ func (tv *TransferVerifier[evmClient, connector]) ProcessReceipt(
...
 	}
 
 	if len(*receipt.Deposits) == 0 && len(*receipt.Transfers) == 0 {
-		return summary, errors.New("invalid receipt: no deposits and no transfers")
+		// This should result in an invariant error below but this is helpful context.
+		tv.logger.Warn("receipt is a token transfer but has no deposits or transfers",
```
pleasew8t, line 631: Can this be moeved to `receipt.Validate()`?

devnet/tx-verifier/tx-verifier-evm-runner.sh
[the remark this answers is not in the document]
pleasew8t: Doesn't it then make sense to add a comment about that behaviour and up the required count to be exactly what we are expecting? x2?

node/cmd/txverifier/README.md
```diff
@@ -7,9 +7,37 @@ modify the package code at `node/pkg/txverifier/` and run this tool to test the
 
 ### Ethereum
 
-Ensure that you have a valid API key connected with a **WebSockets** URL. 
+_Ensure that you have a valid API key connected with a **WebSockets** URL._
```
pleasew8t: nit: This requires a valid API key connected with...

Third person writing

node/cmd/txverifier/README.md
```diff
@@ -7,9 +7,37 @@ modify the package code at `node/pkg/txverifier/` and run this tool to test the
...
 
-The following command runs the Transfer Verifier against mainnet.
+#### Devnet
+
+The script at `scripts/transfer-verifier-localnet.sh` runs the transfer verifier against devnet.
+
+#### Automated testing
+
+This command provides a "sanity check" mode that runs the package against mainnet data using a hard-coded
+list of tx hashes and their expected return values. It checks that ruling of true/false matches what's
```
pleasew8t: nit: `list of transaction hashes`

node/cmd/txverifier/README.md
```diff
@@ -7,9 +7,37 @@ modify the package code at `node/pkg/txverifier/` and run this tool to test the
...
+expected, and ensures that the expected error codes match.
+
+The code will log a fatal error and exit as soon as one of the sanity checks fail.
+
+This functions as a quick way to do regression testing using real data, avoiding the need for extensive mocking
```
pleasew8t, line 24: `This feature provides a quick way`

node/cmd/txverifier/README.md
```diff
@@ -23,3 +51,28 @@ The following command runs the Transfer Verifier against mainnet.
 
 To test against a forked local network, change the RPC URL to anvil's default (also used by the Tilt network), and update
 the contract addresses.
```
pleasew8t, line 53: nit: more newline wonkiness

node/cmd/txverifier/README.md
```diff
@@ -23,3 +51,28 @@ The following command runs the Transfer Verifier against mainnet.
...
+
+##### Testing a single receipt
+
+A single receipt can be evaluated by adding the `--hash` flag and passing an Ethereum receipt hash.
```
pleasew8t: `and passing an Ethereum transaction hash.`

node/cmd/txverifier/evm.go
```diff
@@ -33,8 +34,14 @@ var (
 	wrappedNativeContract *string
 	// Height difference between pruning windows (in blocks).
 	pruneHeightDelta *uint64
+	// Receipt hash to analyze.
```
pleasew8t: nit: Transaction hash

node/cmd/txverifier/evm.go
```diff
@@ -131,6 +144,48 @@ func runTransferVerifierEvm(cmd *cobra.Command, args []string) {
...
+	if *sanity {
+		// Ensure each check gives the correct result and error type.
+		// The (false, nil) case is not tested here because it requires a core bridge exploit; this can
+		// be tested using the integration tests in Tilt.
+		for i, check := range sanityChecks {
+			logger.Info(fmt.Sprintf("Running sanity check %d for txHash %s", i, check.txHash))
+			valid, err := transferVerifier.TransferIsValid(ctx, check.txHash, nil)
+			logger.Debug("done processing", zap.Bool("result", valid), zap.String("txHash", check.txHash.String()))
+
+			if err != nil {
```
pleasew8t, line 165: Shouldn't we also check that if `err` is `nil`, that the exepcted error is also `nil`? I feel like we can just match against the exepected error, and `logger.Fatal` if it doesn't?

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
-) bool {
+) (bool, error) {
 
-	// Use this opportunity to prune old transaction information from the cache.
 	tv.pruneCache()
 
 	if Cmp(txHash, ZERO_ADDRESS) == 0 {
-		tv.logger.Error("txHash is all zeroes")
-		return false
+		tv.logger.Error("txHash is the zero address")
```
pleasew8t: When logging errors like this, isn't it better to just reuse the sentinal error? `ErrTxHashIsZeroAddr.Error()`? Makes updating errors more central.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +14,252 @@ import (
...
+
+		} else {
+			tv.logger.Info("pruned cached calls to chainId()",
+				zap.Int("numDeleted", numDeleted),
+				zap.Int("cacheSize", len(tv.chainIdCache)),
+			)
+		}
+	}
+	if len(tv.decimalsCache) >= CacheMaxSize {
+		numDeleted, err := deleteEntries(&tv.nativeContractCache)
```
pleasew8t: This needs to be `tv.decimalsCache`

pleasew8t: The log should also log `len(tv.decimalsCache)`

node/pkg/txverifier/evm.go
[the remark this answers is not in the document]
pleasew8t: So not entirely correct. The problem is as follows:
* The `LogMessagePublished` event has the origin chain and address
* The `ERC20` transfers are the wrapped assets

We need to "unwrap" the addresses of each ERC20 token so we can match that token to the token in the `LogMessagePublished` event.

node/pkg/txverifier/evm.go
```diff
@@ -183,95 +269,109 @@ func (tv *TransferVerifier[ethClient, Connector]) UpdateReceiptDetails(
...
-			// It's somewhat common for transfers to be made across the bridge for assets
-			// that are not properly registered. In this case, calls to isWrappedAsset() on
-			// the Token Bridge will return true but the calls to wrappedAsset() will return
-			// the zero address. In this case it's impossible to determine the decimals and
-			// therefore there is no way to compare the amount transferred or burned with
-			// the LogMessagePublished payload. In this case, we can't process this receipt.
-
-			return errors.Join(errors.New("error when fetching native info for ERC20 Transfer. Can't continue to process this receipt"), fetchErr)
-		}
+	if len(*receipt.Deposits) > 0 {
```
pleasew8t, line 267: nit: I think we can really just remove this `if-else`. We are already logging the processing of each deposit.

node/pkg/txverifier/evm.go
```diff
@@ -183,95 +269,109 @@ func (tv *TransferVerifier[ethClient, Connector]) UpdateReceiptDetails(
...
-		if logFetchErr != nil {
-			// The unwrapped address and the denormalized amount are necessary for checking
-			// that the amount matches.
-			return errors.Join(errors.New("error when populating wormhole details. cannot verify receipt"), logFetchErr)
-		}
-		message.TransferDetails = newDetails
+	// Populate the origin chain ID and address fields for ERC20 transfers so that they can be compared against the destination asset in the
+	// LogMessagePublished payload.
+	if len(*receipt.Transfers) > 0 {
+		tv.logger.Debug("populating details for ERC20 Transfers")
```
pleasew8t, line 287: nit: same as previous comment

node/pkg/txverifier/evm.go
```diff
@@ -183,95 +269,109 @@ func (tv *TransferVerifier[ethClient, Connector]) UpdateReceiptDetails(
...
-		zap.String("receipt", receipt.String()),
-	)
+		var (
+			queryAddr   = transfer.TokenAddress
+			originChain vaa.ChainID
+			originAddr  vaa.Address
+		)
+		// Handle and return early for native assets.
+		if !isWrapped {
+			tv.logger.Debug("token is native (not wrapped)")
```
pleasew8t, line 305: should we disambiguate the wording here a bit? When talking about "native" assets, the protocol token (Ethereum/Sol/whatever) comes to mind. We should rename this throughout the code to reflect that the token is on it's origin chain. `token address is it's origin address`, in this case.

node/pkg/txverifier/evm.go
```diff
@@ -183,95 +269,109 @@ func (tv *TransferVerifier[ethClient, Connector]) UpdateReceiptDetails(
...
+			}
+			if originChain == vaa.ChainIDUnset {
+				return errors.Join(errors.New("chainId: chainId for token is unset after query. Can't continue to process this receipt"))
+			}
 
-	tv.logger.Debug("finished updating receipt details")
-	return nil
-}
+			// Find the origin address by querying the wrapped asset address.
+			nativeAddr, err := tv.nativeContract(queryAddr)
```
pleasew8t, line 320: Follow-up on the previous comment regarding native tokens: I think this is still correct, calling `nativeContract`, since that's what the token contract offers us. But I don't think the token address should be referred to as its native address. Maybe better to call it its origin address?

pleasew8t: Why not directly set `originAddr, err := tv.nativeContract...`?

node/pkg/txverifier/evm.go
```diff
@@ -339,11 +444,23 @@ func (tv *TransferVerifier[evmClient, connector]) ParseReceipt(
 
 			tv.logger.Debug("adding deposit", zap.String("deposit", deposit.String()))
 			deposits = append(deposits, deposit)
+
```
pleasew8t, line 451: I think we need to add a check in the `EVENTHASH_WETH_DEPOSIT` case that this event was emitted by the native token contract. Any contract that emits this event could insert themselves in the deposit list. Might be an issue down the line.

node/pkg/txverifier/evmtypes.go
```diff
@@ -960,10 +1008,9 @@ func validate[L TransferLog](tLog TransferLog) error {
 			return &InvalidLogError{Msg: "destination is not set"}
 		}
 
-		// TODO is this valid for assets that return the zero address from unwrap?
-		// if Cmp(log.OriginAddress(), ZERO_ADDRESS_VAA) == 0 {
-		// 	return errors.New("origin cannot be zero")
-		// }
+		if Cmp(log.OriginAddress(), ZERO_ADDRESS_VAA) == 0 {
```
pleasew8t: This check is used in all three cases. Consider moving it to the top of the function with the other generic checks.

node/pkg/txverifier/evmtypes.go
```diff
@@ -973,22 +1020,13 @@ func validate[L TransferLog](tLog TransferLog) error {
 			return &InvalidLogError{Msg: "target address cannot be zero"}
 		}
 
-		if len(log.TransferDetails.OriginAddressRaw) == 0 {
-			return &InvalidLogError{Msg: "origin address raw cannot be empty"}
+		if log.TransferDetails.Amount.Sign() == -1 {
```
pleasew8t, line 1013: Is it really necessary to check the sign across the code? AFAIK marshalling bytes to a big.Int cannot ever influence the sign.

node/pkg/txverifier/evm.go
johnsaigle: I think native is better because it's more consistent with the wrapped vs. native distinction used elsewhere in Wormhole, including the Token Bridge code.
I can try to rename some instances of Origin though

node/pkg/txverifier/evm_test.go
johnsaigle: It's to contrast this format to the VAA address format. I think maybe e.g `nativeAddrEVM` could be better than `Geth` as a name

node/pkg/txverifier/evm.go
johnsaigle: Do you mean for the values of `dict`? This is checking that the pointer is nil which seems right to me

node/pkg/txverifier/evm.go
johnsaigle: Do you mean we should add a check here to only update `lastBlockNumber` if the new one is greater?

node/cmd/txverifier/evm.go
johnsaigle: Agreed, we should think this over and resolve it later.

node/pkg/txverifier/README.md
johnsaigle: The lines are wrapped to ~120 characters in this file and they render as full paragraphs in Markdown so I'm not going to change this. I have a strong preference for wrapping lines in documentation so that no horizontal scrolling is necessary.

In general, there's no markdown formatter or style guide for the repo, so if you feel strongly about this we should define and automate policy rather than sorting it out ad-hoc in PR comments.

node/pkg/txverifier/evm.go
johnsaigle: Nevermind, I'll add it

node/pkg/txverifier/evm.go
johnsaigle: This isn't correct. The error signals that the Token Bridge isn't the sender of any of the message publications from the core bridge. It's possible to get several logs from the core bridge and still reach this case.

johnsaigle: I'll add a comment to clarify this

node/pkg/txverifier/evm.go
johnsaigle: No that method doesn't have access to the `tv.logger`. I think it's better to keep the logging here and have the checks in `Validate()`

node/cmd/txverifier/README.md
johnsaigle: Same response as above, line-wrapping is for editor-viewing not semantic meaning.

node/pkg/txverifier/evm.go
johnsaigle: It's a debug-level log so I'm fine to leave this.

node/pkg/txverifier/evm.go
johnsaigle: It's a debug-level log so I'm fine to leave this.

node/pkg/txverifier/evm.go
johnsaigle: `originAddr` is defined outside of this block so that would create a local shadow variable with the same name.

node/pkg/txverifier/evmtypes.go
johnsaigle: I'd rather leave it in as an invariant, seems like a good thing to explicitly deny in case the code ever changes in the future.

node/pkg/txverifier/evmtypes.go
johnsaigle: Done

node/pkg/txverifier/utils_test.go
johnsaigle: Done

node/pkg/txverifier/evm.go
johnsaigle: Done

node/cmd/txverifier/evm.go
johnsaigle: Added

node/pkg/txverifier/evm.go
johnsaigle: done

node/pkg/txverifier/evm.go
johnsaigle: Absolutely, nice catch. Added this.

node/pkg/txverifier/evmtypes.go
johnsaigle: Done

node/pkg/txverifier/evm.go
johnsaigle: Yeah Dirk and I discussed this here.
https://github.com/wormhole-foundation/wormhole/pull/4362#discussion_r2112735424

I'll make this change.

node/pkg/txverifier/evm.go
johnsaigle: Thanks, I'll move this log statement into the non-nil branch

node/pkg/txverifier/evm.go
johnsaigle: Nevermind I see what you're saying. I'll add a condition that this is only updated if the new value is greater, and add comments.

node/pkg/txverifier/evm.go
```diff
@@ -19,161 +13,248 @@ import (
...
 		}
 	}
 
-	// Caching: record a new lastBlockNumber.
-	tv.lastBlockNumber = receipt.BlockNumber.Uint64()
-	tv.processedTransactions[txHash] = receipt
+	eval := evaluation{Receipt: receipt}
+
+	// Caching: if the new receipt is newer than the last one, update the last block number.
+	// This is important to prevent the Transfer Verifier from recording a block number that older
```
pleasew8t: that _is_ older than

PR #4369, opened 2025-05-02
node/admin: add template for core bridge set message

fergarrui:
A new template command has been added:

```
message CoreBridgeSetMessageFee {
  // ID of the chain of the core bridge contract where the message fee should be updated (uint16).
  uint32 chain_id = 1;
  // New message fee.
  uint64 message_fee = 2;
}
```


bruce-riley (dismissed):
This looks good, but don't you also need to update `adminserver.go` and `sdk/vaa/payload.go`?

fergarrui:
> This looks good, but don't you also need to update `adminserver.go` and `sdk/vaa/payload.go`?

Updated both files, thank you @bruce-riley 
Please let me know if I need to update any other file

node/cmd/guardiand/admintemplate.go
```diff
@@ -204,6 +207,13 @@ func init() {
 	AdminClientAccountantModifyBalanceCmd.Flags().AddFlagSet(moduleFlagSet)
 	TemplateCmd.AddCommand(AdminClientAccountantModifyBalanceCmd)
 
+	// flags for the core-bridge-set-message-fee command
+	coreBridgeSetMessageFeeFlagSet := pflag.NewFlagSet("core-bridge-set-message-fee", pflag.ExitOnError)
+	coreBridgeSetMessageFeeChainId = coreBridgeSetMessageFeeFlagSet.String("chain-id", "", "Chain ID")
```
johnsaigle, line 212: Does it makes sense to use a uint32 flag instead of a string here?

node/cmd/guardiand/admintemplate.go
```diff
@@ -204,6 +207,13 @@ func init() {
...
+	coreBridgeSetMessageFeeMessageFee = coreBridgeSetMessageFeeFlagSet.String("message-fee", "", "New message fee")
```
johnsaigle, line 213: Same here, is it possible to use a uint64 value instead of a string?

fergarrui: Done!

node/cmd/guardiand/admintemplate.go
fergarrui: The problem here is that we're using a string as an argument, e.g:

```
build/bin/guardiand template core-bridge-set-message-fee --chain-id base_sepolia --message-fee 0
```

node/pkg/adminrpc/adminserver.go
```diff
@@ -663,6 +663,23 @@ func wormholeRelayerSetDefaultDeliveryProvider(req *nodev1.WormholeRelayerSetDef
 	return v, nil
 }
 
+func coreBridgeSetMessageFeeToVaa(req *nodev1.CoreBridgeSetMessageFee, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+	if req.ChainId > math.MaxUint16 {
```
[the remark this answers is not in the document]
fergarrui: Good point, done

bruce-riley (changes requested):
Please add a test in `sdk/vaa/payloads_test.go`.

bruce-riley (approved):

johnsaigle (approved):

PR #4375, opened 2025-05-14
node(governor): reobservations should not flow cancel

johnsaigle:
Reobservations are complicated to reason about in a context where flow cancel is enabled. Disabling reobservations is simpler, with only a minor downside that otherwise-eligible reobservations of token transfers will not flow cancel.

### DIFF
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -657,7 +657,8 @@ func (gov *ChainGovernor) ProcessMsgForTime(msg *common.MessagePublication, now 
 	// - This will cause the summed value of Sui to decrease.
 	emitterChainEntry.transfers = append(emitterChainEntry.transfers, transfer)
 
-	if gov.flowCancelEnabled {
+	// Reobservation messages must not flow cancel.
+	if gov.flowCancelEnabled && !msg.IsReobservation {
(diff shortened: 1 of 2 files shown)


mdulin2:
Should we be checking these in other spots? 

- [Database reloading](https://github.com/wormhole-foundation/wormhole/blob/bce6b859f4f95b012619d08fcf84394885fa11d8/node/pkg/governor/governor_db.go#L249). 
- [GetNotionalAvailableByChain](https://github.com/wormhole-foundation/wormhole/blob/bce6b859f4f95b012619d08fcf84394885fa11d8/node/pkg/governor/governor_monitoring.go#L298)

johnsaigle:
Putting this on pause for now until we have a longer discussion about it.

johnsaigle:
#4428 should ensure that `IsReobservation` can be loaded from the Governor codebase which should help this feature assuming we still want to go ahead with it.

johnsaigle:
This patch needs to be reworked following https://github.com/wormhole-foundation/wormhole/pull/4805. This marks the reobservation as always true for canonical message publications, which means that they will be unable to do _any_ flow cancelling according to the changes in this PR. 

We'll need to build out a different configuration or check for when a canonical guardian is the one processing flow cancelling, which may be somewhat complicated if the delegate/canonical membership status is not visible to the governor package.

PR #4390, opened 2025-05-29
ci: add depguard linter to create a deny-list for dependencies

johnsaigle:
This PR adds a new linter that should preserve some of the work from [recent dependency-related PRs](https://github.com/wormhole-foundation/wormhole/pulls?q=is%3Apr+author%3Ajohnsaigle+label%3Adependencies+) into the future by adding the removed dependencies to a deny-list.

I only deny-listed certain dependencies for which I'm fairly confident that there is no legitimate use case.

### Testing

Do one or both:
- Add one of the deny-listed dependencies to `go.mod` and include it in a file, then run `make lint`.
- Add a dependency we're currently using to the deny-list in order to observe the warnings.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
     # Ensure that the body is closed on HTTP and websocket conns
     - bodyclose
     - contextcheck
+    - depguard
     # Duplicate word usage, such as 'and and' in a comment.
     - dupword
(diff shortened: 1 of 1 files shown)


mdulin2 (approved):

djb15 (approved):
Nice!

PR #4399, opened 2025-06-11
fix: moved message fee from int64 to string and uint256

fergarrui:
* message fee should be `uint256` and not `int64`, or else we'll get an `toUint256_outOfBounds` error when submitting the VAA to the core bridge


node/pkg/adminrpc/adminserver.go
```diff
@@ -669,9 +669,19 @@ func coreBridgeSetMessageFeeToVaa(req *nodev1.CoreBridgeSetMessageFee, timestamp
...
+	}
+	new_fee, overflow := uint256.FromBig(new_fee_big)
+	if overflow {
+		return nil, errors.New("new fee overflow")
+	}
+
 	body, err := vaa.BodyCoreBridgeSetMessageFee{
 		ChainID:    chainId,
-		MessageFee: uint64(req.MessageFee),
+		MessageFee: new_fee,
```
johnsaigle, line 687: Is it worth ensuring that the sign isn't negative before setting this value?

fergarrui: Just added a new check @johnsaigle

johnsaigle (approved):

bruce-riley (approved):

PR #4402, opened 2025-06-16
sdk: Auto-generate ChainID methods

johnsaigle:


bruce-riley:
This is very cool! Thanks for doing it! Two quick comments:

- Should there be either an `sdk/vaa/README.md` or a prologue in `sdk/vaa/structs.go` that describes how to add a chain ID?
- Can we just go ahead and kill the `TestChainIDFromString` and `TestChainId_String`? Otherwise, people still need to manually update them. With everything being autogenerated, I don't think those tests add any value anyway.

sdk/README.md
```diff
@@ -3,6 +3,28 @@
...
+2. **Regenerate methods** by running:
+   ```bash
+   make go-generate
+   ```
+   This runs `chainid_generator.go` which auto-generates `String()`, `ChainIDFromString()`, and `GetAllNetworkIDs()` methods.
+
+3. **Update other components** as needed:
+   - Add to governor chain lists (`node/pkg/governor/mainnet_chains.go`)
+   - Add manual tokens if required (`node/pkg/governor/manual_tokens.go`)
+   - Update any chain-specific configuration files
```
bruce-riley, line 26: - Add to the `ChainID` in `proto/publicrpc/v1/publicrpc.proto`.
- If watcher support is necessary, update the guardian code.

djb15 (approved):
Cool!

PR #4404, opened 2025-06-19
Node: Specify config file

bruce-riley:
This PR allows a guardian operator to load config parameters from an arbitrary config file when running `guardiand node` by specifying the `--config <fname>` command line argument. This file may be in any of the formats supported by [Viper](https://pkg.go.dev/github.com/spf13/viper#section-readme), including .yaml and .json.

Parameters from this file are loaded first, then overridden by environment variables starting with `GUARDIAND`, and finally command line arguments.

**NOTE**: This PR changes existing behavior. Previously, if the file `node/config/guardiand.yaml` existed, it was loaded. However, this did not support specifying an arbitrary file location.

If you were previously using `node/config/guardiand.yaml` for your configuration, please add `--config node/config/guardiand.yaml` to your command line.

If a config file is loaded, a line like this will appear at the start of the log file:

```
INFO    loaded config file      {"filePath": "node/config/guardiand.yaml"}
```

### DIFF
--- a/devnet/node.yaml
+++ b/devnet/node.yaml
@@ -80,6 +80,8 @@ spec:
           command:
             - /guardiand
             - node
+            - --config
+            - node/config/guardiand.yaml
             # - --ethRPC
             # - ws://eth-devnet:8545
             # - --wormchainURL
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -45,6 +45,7 @@ import (
 	libp2p_crypto "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"github.com/wormhole-foundation/wormhole/sdk"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 	"go.uber.org/zap"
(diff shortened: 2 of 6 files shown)


johnsaigle (dismissed):
I did not test locally but the changes seem sensible.

devnet/node.yaml
```diff
@@ -80,6 +80,8 @@ spec:
           command:
             - /guardiand
             - node
+            - --config
```
panoel, line 83: Do you want these lines to be commented out initially?  What if the file does not exist?

panoel (approved):

devnet/node.yaml
bruce-riley: It does exist. (It's created at the bottom of this file.) Without these lines, tilt failed to come up because `ethRPC` is defined in the config file (to verify reading the config works.)

johnsaigle (dismissed):

node/cmd/guardiand/node.go
```diff
@@ -582,8 +579,7 @@ var Build = "prod"
 // initConfig initializes the file configuration.
 func initConfig(cmd *cobra.Command, args []string) error {
 	return node.InitFileConfig(cmd, node.ConfigOptions{
-		FilePath:  configPath,
-		FileName:  configFilename,
+		FilePath:  viper.ConfigFileUsed(),
```
evan-gray, line 582: please update the [operations document](https://github.com/wormhole-foundation/wormhole/blob/main/docs/operations.md#config-file) either in this PR or a follow up.

evan-gray (dismissed):

evan-gray (approved):
thanks!

PR #4406, opened 2025-06-20
Node: Custom consistency level

bruce-riley:
This PR implements the guardian watcher changes to support the custom consistency level feature.

Note that the branch `node_custom_consistency_level` is based on `eth_custom_consistency_level` but this PR includes both branches, so that we can run the tests in CI.


node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+		pe.additionalBlocks = uint64(req.additionalBlocks)
+	default:
+		w.cclLogger.Error("invalid custom handling type, treating as finalized", zap.Stringer("emitterAddress", emitterAddr), zap.Uint8("reqType", uint8(req.Type())), zap.Error(err))
+		pe.message.ConsistencyLevel = vaa.ConsistencyLevelFinalized
+	}
+}
+
+// cclShouldPublish is called to see if an observation can be released because the custom conditions have been met.
+func (w *Watcher) cclShouldPublish(pe *pendingMessage, latestBlockNum uint64) bool {
+	if pe.additionalBlocks == 0 {
```
evan-gray: is `0` additional blocks not a valid setting? it seems like a message would never get published if this was set to `0`. perhaps this check is not needed?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+		return data, nil
+	}
+
+	data, err := CCLReadContract(ctx, w.ethConn.Client(), w.cclAddr, emitterAddr)
+	if err != nil {
+		return cclEmptyData, err
+	}
+
+	w.cclCacheUpdate(emitterAddr, data)
+	w.cclLogger.Info("TEST: read contract", zap.Stringer("emitterAddr", emitterAddr))
```
evan-gray: should this be a debug?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+	return data, nil
+}
+
+// CCLReadContract calls into the contract to read the configuration for a given emitter.
+// This is a free function so it can be called by the config verification tool.
+func CCLReadContract(ctx context.Context, ethClient *ethclient.Client, cclAddr ethCommon.Address, emitterAddr ethCommon.Address) ([32]byte, error) {
+	timeout, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
+
+	caller, err := cclAbi.NewCclCaller(ethCommon.BytesToAddress(cclAddr.Bytes()), ethClient)
```
evan-gray: why does `cclAddr` need to be converted to bytes back to address?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+func (w *Watcher) cclCacheLookUp(emitterAddr ethCommon.Address) ([32]byte, bool) {
+	w.cclCacheLock.Lock()
+	defer w.cclCacheLock.Unlock()
+
+	if entry, exists := w.cclCache[emitterAddr]; exists {
+		if time.Since(entry.readTime) < CCLCacheTimeoutInterval {
+			return entry.data, true
+		}
+
+		w.cclLogger.Info("TEST: cache entry has expired", zap.Stringer("emitterAddr", emitterAddr))
```
evan-gray: debug?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+
+	return cclEmptyData, false
+}
+
+// cclCacheUpdate updates the entry in the cache for a given emitter, including the read time.
+func (w *Watcher) cclCacheUpdate(emitterAddr ethCommon.Address, data [32]byte) {
+	w.cclCacheLock.Lock()
+	w.cclCache[emitterAddr] = CCLCacheEntry{data, time.Now()}
+	w.cclCacheLock.Unlock()
+	w.cclLogger.Info("TEST: cache entry added", zap.Stringer("emitterAddr", emitterAddr))
```
evan-gray: debug?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,374 @@
...
+		pe.message.ConsistencyLevel = req.consistencyLevel
+		pe.additionalBlocks = uint64(req.additionalBlocks)
+	default:
+		w.cclLogger.Error("invalid custom handling type, treating as finalized", zap.Stringer("emitterAddress", emitterAddr), zap.Uint8("reqType", uint8(req.Type())), zap.Error(err))
+		pe.message.ConsistencyLevel = vaa.ConsistencyLevelFinalized
+	}
+}
+
+// cclShouldPublish is called to see if an observation can be released because the custom conditions have been met.
+func (w *Watcher) cclShouldPublish(pe *pendingMessage, latestBlockNum uint64) bool {
```
evan-gray: looks like this code is only used by the `latest` event condition in the watcher, maybe this can be removed.

node/pkg/watchers/evm/watcher.go
```diff
@@ -489,6 +500,14 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 					stats.Height = int64(blockNumberU) // #nosec G115 -- This conversion is safe indefinitely
 					w.updateNetworkStats(&stats)
 					w.ccqAddLatestBlock(ev)
+
+					w.pendingMu.Lock()
+					for key, pLock := range w.pending {
+						if w.cclShouldPublish(pLock, blockNumberU) {
```
evan-gray: instead of using `cclShouldPublish`, perhaps just check if `pLock.message.ConsistencyLevel` is latest and do the same `if pLock.height+pLock.additionalBlocks <= blockNumberU` condition as below. I agree with doing a separate loop here though and keeping the `continue` statement as the minimally invasive change, though I'd be open to updating the if/else and `safe` check below to avoid the `publishObservation` refactor instead (that conditional on 531 always confused me) :shrug:

node/pkg/watchers/evm/custom_consistency_level.go
bruce-riley: I'm not sure zero additional blocks makes sense? (Wouldn't you just use the usual finalized, safe or immediate then?) But I guess it won't do any harm to allow it. Anyway, as you point out below, this function is no longer needed, so this will go away.

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,335 @@
...
+		data     [32]byte
+		readTime time.Time
+	}
+
+	// CCLCache is the layout of the cache of config data for emitters
+	CCLCache map[ethCommon.Address]CCLCacheEntry
+)
+
+// CCLCacheTimeoutInterval is the lifetime of a cache entry. After that time, we delete the entry and re-read the config data.
+// TODO: This time is arbitrary. Does it sound okay?
```
bruce-riley, line 98: @evan-gray does a five minute cache timeout sound reasonable? 🙏

evan-gray: Yes, I think so.

node/pkg/watchers/evm/watcher.go
bruce-riley: I restructured the code to hopefully make it more readable.

evan-gray (approved):
this looks cleaner, thanks!

whitepapers/0001_generic_message_passing.md
```diff
@@ -189,12 +189,28 @@ level meanings and enforcing them.
...
+it reads the `CustomConsistencyLevel` on-chain contract, using the emitter address as the key to determine the
+special handling to be performed. If the emitter address is not configured in the contract, then the observation
+is treated as though it had consistency level 202 (finalized).
+
+#### Additional Blocks handling
+
+Currently, the only supported custom handling is to wait the configured number of additional blocks after the specified
+consistency level. For instance, if the integrator has configured their emitter address in the `CustomConsistencyLevel`
+contract as consistencyLevel == `201` and additional blocks == 5, then the watcher will not approve the observation until
+five blocks after when the observation block is marked safe (note, not five safe blocks, but five blocks after safe).
```
evan-gray: per our discussion, let's update the parenthetical note and add a couple examples like Ethereum latest + 2 and HyperEVM finalized + 5 (which *would* be 5 _finalized_ blocks)

evan-gray (approved):

testing/contract-integrations/custom_consistency_level/__tests__/test_custom_consistency_level.ts
```diff
@@ -0,0 +1,307 @@
...
+} from "@certusone/wormhole-sdk";
+
+import * as CustomConsistencyLevel from "../../../../ethereum/build-forge/CustomConsistencyLevel.sol/CustomConsistencyLevel.json";
+import * as TestCustomConsistencyLevel from "../../../../ethereum/build-forge/TestCustomConsistencyLevel.sol/TestCustomConsistencyLevel.json";
+
+const ci = process.env.CI == "true";
+
+const ETH_NODE_URL = ci ? "http://eth-devnet:8545" : "http://localhost:8545";
+
+const ETH_PRIVATE_KEY10 =
```
evan-gray: you may have an intermittent failure using this key because it is also used by `sui-integration.ts`

bruce-riley: Added a couple of additional keys and switched to using one of those.

evan-gray (dismissed):

whitepapers/0001_generic_message_passing.md
```diff
@@ -189,12 +189,45 @@ level meanings and enforcing them.
 
 - `200` - publish immediately
 - `201` - `safe`, if available, otherwise falls back to `finalized`
+- `202` - `finalized`
+- `203` - custom handling
 - anything else is treated as `finalized`
```
panoel, line 194: Does this statement still make sense?  Otherwise, why have `202`?

panoel (dismissed):

whitepapers/0001_generic_message_passing.md
evan-gray: I think so, it’s just stating that 202 is the actual enum for finalized but finalized is also the fallback value.

ethereum/contracts/custom_consistency_level/CustomConsistencyLevel.sol
```diff
@@ -0,0 +1,32 @@
...
+    string public constant VERSION = customConsistencyLevelVersion;
+
+    mapping(address => bytes32) private _configurations;
+
+    // ==================== External Interface ===============================================
+
+    /// @inheritdoc ICustomConsistencyLevel
+    function configure(
+        bytes32 config
+    ) external override {
```
djb15, line 21: Why does this method have the `override` decorator? Same with `getConfiguration`?

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,335 @@
...
+func cclParseAdditionalBlocksConfig(reader *bytes.Reader) (CCLRequest, error) {
+	consistencyLevel := uint8(0)
+	if err := binary.Read(reader, binary.BigEndian, &consistencyLevel); err != nil {
+		return nil, fmt.Errorf("failed to read consistency level: %w", err)
+	}
+
+	blocks := uint16(0)
+	if err := binary.Read(reader, binary.BigEndian, &blocks); err != nil {
+		return nil, fmt.Errorf("failed to read num blocks: %w", err)
+	}
```
djb15, line 271: Worth adding a defensive check here that the remaining bytes of the reader are zero?

ethereum/contracts/custom_consistency_level/libraries/ConfigMakers.sol
```diff
@@ -0,0 +1,23 @@
...
+
+    uint8 public constant TYPE_ADDITIONAL_BLOCKS = 1;
+
+    /// @notice Encodes an additional blocks custom consistency level configuration.
+    /// @param consistencyLevel The consistency level to wait for.
+    /// @param blocksToWait The number of additional blocks to wait after the consistency level is reached.
+    /// @return bytes The encoded config.
+    function makeAdditionalBlocksConfig(
+        uint8 consistencyLevel,
+        uint16 blocksToWait
```
djb15, line 17: Should we make `blocksToWait` a `uint8` or cap at some upper bound? On a chain with slow block times with a `type(uint16).max` extra wait, is it possible for an attacker to grow the pending message queue in the watcher to a size that causes a performance degradation? I'm not confident on the feasibility given the gas costs, just raising the question for input

evan-gray: I didn’t choose uint8 because there were previously and are currently chains with significantly longer finality times (e.g. previously polygon, currently linea)

ethereum/contracts/custom_consistency_level/CustomConsistencyLevel.sol
bruce-riley: It won't compile without it. Maybe this is related to the old version of the compiler being used here (to be consistent with the other wormhole contracts) and how interfaces worked in that version? I see that, in the token bridge, `Bridge` does not have `is ITokenBridge`. 

I tried removing `is ICustomConsistencyLevel`, which lets me remove the override, but then I have to duplicate things. Please take a look at the latest commit and let me know what you think.

ethereum/contracts/custom_consistency_level/libraries/ConfigMakers.sol
bruce-riley: @djb15, unless you have a strong objection, I would like to leave it as is. 

If we are really concerned about the attack you mentioned, a better approach might be to limit how many outstanding observations a given emitter can have (I have no idea what that limit should be). A problem with that is all the watcher can do is drop the excess, which may just lead to a reobservation storm.

If we really feel this is a serious attack vector, maybe we should open an issue and handle it with a follow-on PR?

ethereum/contracts/custom_consistency_level/TestCustomConsistencyLevel.sol
```diff
@@ -0,0 +1,47 @@
...
+contract TestCustomConsistencyLevel {
+    string public constant VERSION = testCustomConsistencyLevelVersion;
+
+    ICustomConsistencyLevel public immutable customConsistencyLevel;
+    IWormhole public immutable wormhole;
+    uint32 public nonce;
+
+    constructor(
+        address _wormhole,
+        address _customConsistencyLevel,
```
djb15, line 22: I think you can avoid the need for the interface by removing this as an input argument, and just deploying an instance of the `CustomConsistencyLevel` contract from the constructor? Instead of wrapping the address in an interface

ethereum/contracts/custom_consistency_level/CustomConsistencyLevel.sol
djb15: Added another comment below that I think will address the duplication concerns. But the change to the contract itself looks good

ethereum/contracts/custom_consistency_level/TestCustomConsistencyLevel.sol
bruce-riley: But is that what we want integrators to do? I think the interface is mostly for use by integrators.

ethereum/contracts/custom_consistency_level/libraries/ConfigMakers.sol
djb15: I think I'm ok to leave it as is, having slept on it I think gas costs should make it infeasible. No strong feelings but we could add metrics to track the size of the pending queue if that doesn't already exist

ethereum/contracts/custom_consistency_level/TestCustomConsistencyLevel.sol
djb15: Oh of course, good point! In that case I think it's probably preferable to revert the latest change and go back to using the `override`. Pains me to see `override` when it doesn't seem needed but if the compiler is complaining I guess we have no choice 😢

ethereum/contracts/custom_consistency_level/CustomConsistencyLevel.sol
bruce-riley: Per the discussion below, I am leaving the override as the less unpleasant solution.

ethereum/contracts/custom_consistency_level/libraries/ConfigMakers.sol
evan-gray: agree, like all wormhole core messages, gas costs already provide a spam deterrent (however thin).

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// The config data is 32 bytes and the AdditionalBlocks request uses the first four, so we should have 28 left.
```
evan-gray, line 273: my impression of the earlier comment was to check that these 28 bytes were zeroes as well, but I could be mistaken (also I personally don't think that's critical to check since the spec doesn't specify what needs to be in the unused bytes)

evan-gray (approved):

ethereum/contracts/custom_consistency_level/CustomConsistencyLevel.sol
```diff
@@ -0,0 +1,32 @@
...
+        _configurations[msg.sender] = config;
```
kcsongor, line 22: might be worth suggesting in the docs that integrator contracts expose an admin-only function that invokes this `configure` method, so that the integrator can perform config changes without needing to do contract upgrades.

alternatively, we could consider
a) handling config update authority here in this contract
b) allowing the `contract.owner()` to configure this for the contract

these approaches have some downsides though, notably a) is a lot of extra complexity, while b) is maybe too opinionated (only works for `Ownable` contracts)

so just calling out in the docs might be sufficient.

evan-gray: I agree it's something I'd like to see in the docs. My recommendation would be to keep the admin configuration up to the integrator. From what I understand, most integrators already have some form of admin only calls. For example [NTT Transceiver](https://github.com/wormhole-foundation/native-token-transfers/blob/988c6dc8e7606b4af82a477e65860b6679c44e84/evm/src/Transceiver/Transceiver.sol#L76) already has `onlyOwner` available.

ethereum/contracts/custom_consistency_level/interfaces/ICustomConsistencyLevel.sol
```diff
@@ -0,0 +1,24 @@
...
+    function configure(
+        bytes32 config
+    ) external;
+
+    /// @notice Returns the configuration for a given emitter address.
+    /// @param emitterAddress The emitter address for which the config has been set.
+    /// @return The configuration.
+    function getConfiguration(
+        address emitterAddress
+    ) external returns (bytes32);
```
nonergodic, line 23: Misses view modifier.

PR #4407, opened 2025-06-23
node: Reduce limits to zero for recently deprecated chains

johnsaigle:
Compare with
* the consensus for these chains on mainnet: https://wormhole-foundation.github.io/wormhole-dashboard/#/?endpoint=Mainnet
* the entry for Aurora, which was also deprecated


djb15 (dismissed):
Looks good. I think you could also set Terra2 and Snaxchain to 0

djb15 (approved):
Sad times given the effort of putting together the manual tokens list 😭

PR #4413, opened 2025-06-30
near: remove signer check

MagicGordon:
Remove signer check to support asset bridging from contracts and keyless accounts.


johnsaigle:
Hello, I'm closing this because it represents a potential major change to the security assumptions of this token bridge.
Please re-open this as an Issue or Discussion with justifications for the change and an explanation of other considerations and workarounds. As it stands, we cannot accept this change.

PR #4419, opened 2025-07-02
Node: Audit protobuf chain IDs

bruce-riley:
This PR adds a test to verify that the chain IDs defined in the protobuf match the ones defined in the go SDK.

Special thanks to @johnsaigle as this PR borrows code from his PR #4402.

This PR also disables some logging in the node key tests since it's just noise.

### DIFF
--- a/node/pkg/common/chain_id_test.go
+++ b/node/pkg/common/chain_id_test.go
@@ -23,7 +23,6 @@ func TestVerifyProtoBufChainIDs(t *testing.T) {
 	// Get all of the chain IDs from the SDK and put them in a map keyed by chain ID.
 	vaaChainList := vaa.GetAllNetworkIDs()
 	require.NotEmpty(t, vaaChainList)
-	assert.NotEmpty(t, vaaChainList)
 
 	vaaChains := ChainMap{}
 	for _, chainId := range vaaChainList {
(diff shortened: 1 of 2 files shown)


node/pkg/common/chain_id_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+type ChainMap map[vaa.ChainID]string
+
+// The purpose of this test is to verify that the `ChainID` definitions in the protobuf are in sync with the ones in vaa.structs.
+// This test parses the generated protobuf file (thereby also verifying that generated files have been updated) and compares the
+// list of chain IDs with the ones returned by `vaa.GetAllNetworkIDs`, making sure that everything matches and there are no duplicates.
+func TestVerifyProtoBufChainIDs(t *testing.T) {
+	// Get all of the chain IDs from the SDK and put them in a map keyed by chain ID.
+	vaaChainList := vaa.GetAllNetworkIDs()
+	require.NotEmpty(t, vaaChainList)
```
johnsaigle, line 25: Two identical asserts here.

evan-gray:
@johnsaigle there is [a separate CI test](https://github.com/wormhole-foundation/wormhole/blob/befd9a6b29b51cf6c89e16ab1400aca77a249277/.github/workflows/build.yml#L303-L307) that ensures the committed go files match the ones generated by the proto, so that test would fail in the case you describe, covering this case.

PR #4421, opened 2025-07-03
ci: add cspell to lint script

johnsaigle:
Runs `cspell` (if installed) as part of local linting. This allows developers to spellcheck locally and catch errors, which helps avoid doing an entire CI run that fails just because of a typo.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -55,6 +55,15 @@ format(){
 }
 
 lint(){
+    # === Spell check
+    if ! command -v cspell >/dev/null 2>&1; then
+        printf "%s\n" "cspell is not installed. Skipping spellcheck"
+    else
+    # NOTE: Keep this command in sync with `.github/workflows/build.yml`
+        cspell -c cspell.config.yaml --dictionary cspell-custom-words.txt "**.*md"
+    fi
+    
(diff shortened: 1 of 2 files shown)


scripts/lint.sh
```diff
@@ -55,6 +55,15 @@ format(){
 }
 
 lint(){
+    # === Spell check
+    if ! command -v cspell >/dev/null 2>&1; then
+        printf "%s\n" "cspell is not installed. Skipping spellcheck"
+    else
+    # NOTE: Keep this command in sync with `.github/workflows/build.yml`
```
djb15, line 62: Should you have the inverse comment on the workflow so both point to each other?

johnsaigle: Sure I can add that.

scripts/lint.sh
```diff
@@ -55,6 +55,15 @@ format(){
...
+        cspell -c cspell.config.yaml --dictionary cspell-custom-words.txt "**.*md"
```
evan-gray, line 63: the config contains the dictionary, and [I'm pretty sure](https://cspell.org/docs/getting-started#1-create-a-configuration-file) it will automatically find that config.

PR #4423, opened 2025-07-03
Node/Solana: Tweak tx_for_addr functionality

bruce-riley:
This PR addresses a couple of issues that came out of an AR security review of the Fogo watcher changes.

The first issue ([Linear PROT-635](https://linear.app/wormholelabs/issue/PROT-635/svm-watcher-explicit-no-signatures-error)) expressed concern if `getPrevWormholeSignature` returned zero transactions, meaning we would do the `getSignaturesForAddress` query with a null starting signature. This actually works fine, and would be what happened if we started the watcher before there were any transactions involving the core contract. The existing code worked fine, but I have added some logging and comments to make this behavior more clear.

The second issue ([Linear PROT-636](https://linear.app/wormholelabs/issue/PROT-636/svm-watcher-timeout-for-tx-for-addr-loop)) expressed that the series of queries in `getTransactionSignatures` could run indefinitely, hanging the watcher. As recommended in the PR, I added a check to give up after a total of one minute. Giving up requires restarting the watcher because we cannot risk processing the transactions read so far. Since transactions are returned in reverse time order (newest to oldest), attempting to process what we've seen so far would mean leaving a gap, causing missed observations.


node/pkg/watchers/solana/tx_for_addr.go
```diff
@@ -109,7 +116,15 @@ func (s *SolanaWatcher) getTransactionSignatures() ([]*rpc.TransactionSignature,
 	numSignatures := MaxSignaturesPerQuery
 	currSig := solana.Signature{}
 
+	startTime := time.Now()
 	for numSignatures == MaxSignaturesPerQuery {
+		if time.Since(startTime) > time.Minute {
+			// It shouldn't take this long to query transactions. Return an error and kill the watcher. We can't
+			// attempt to process the transactions because we read them from most recent backwards, so there could
+			// be a gap and we might miss older transactions.
+			return results, fmt.Errorf("getTransactionSignatures has taken too long, numResults: %d", len(results))
```
evan-gray, line 125: just noting that this behavior should probably be noted in the release notes and someone should be monitoring for these errors once Fogo is live in mainnet.

evan-gray (approved):

johnsaigle (approved):
Thanks for the changes!

PR #4424, opened 2025-07-07
ci: Enable staticcheck lints and fix violations

johnsaigle:
- Ensure `error` types are the final return value
- Ensure error messages begin with a lowercase letter (official Go style recommendation)
- Enable style rule such that Duration variable names don't end with time units
- Remove exception for a file that has since been deleted


node/pkg/node/node.go
```diff
@@ -53,6 +53,23 @@ const (
...
+	return fmt.Sprintf("component %s is already configured and cannot be configured a second time", e.componentName)
+}
+
+type ComponentDependencyError struct {
+	componentName          string
+	dependentComponentName string
+}
+
+func (e ComponentDependencyError) Error() string {
+	return fmt.Sprintf("component %s requires %s to be configured first, check the order of your options", e.componentName, e.dependentComponentName)
```
djb15: Are the variables the wrong way round here or need renaming? I would expect `dependentComponentName` to require  `componentName` to be configured first since it is "dependent" on it? Could rename to `dependencyComponentName`?

johnsaigle: Yeah you're right, I'll reword this

bruce-riley (approved):

PR #4425, opened 2025-07-07
ci: fix cspell command to also run on subdirectories

johnsaigle:
Previously `make lint` was invoking `cspell` only on the markdown files in the current directory instead of also checking subdirectories.


scripts/lint.sh
```diff
@@ -60,7 +60,7 @@ lint(){
         printf "%s\n" "cspell is not installed. Skipping spellcheck"
     else
     # NOTE: Keep this command in sync with `.github/workflows/build.yml`
-        cspell -c cspell.config.yaml --dictionary cspell-custom-words.txt "**.*md"
+        cspell -c cspell.config.yaml --dictionary cspell-custom-words.txt "*/**.*md"
```
evan-gray: I still think you could remove the config and dictionary flags, did you try?

evan-gray (approved):

scripts/lint.sh
johnsaigle: You're right, thanks for the reminder. I ran it with these arguments removed and the command worked. Then I deleted the yaml and dictionary files and got errors, so everything is working correctly.

I removed the flags in the latest commit as well as the new comments.

djb15 (approved):

PR #4428, opened 2025-07-11
node: Fix marshaling for MessagePublications

johnsaigle:
- Correct marshaling for MessagePublication to include the newer fields added to this struct (Unreliable, VerificationState)
- Implement marshal/unmarshal in new methods on MessagePublication that conform to the standard Go marshaling interfaces
- Adds tests:
    - round-trip unit tests for marshaling
    - checks for each of the unmarshaling error conditions
    - fuzz testing for unmarshaling, including test cases that caused panics in previous iterations of the unmarshal function
- Reformats some existing tests to use a table-driven test approach
- Add README to Governor including a short guide on how to do DB upgrades
- Add comments

### Background
- Previously, these fields were not marshaled but this had little effect as the MessagePublication was only unmarshaled by the Governor which did not makes use of the new fields.
- However, the Verification State will be required by the Transfer Verifier and Notary, so it's now necessary to fix marshaling.
- This work was originally included in #4315 but I've extracted it here to make reviewing both simpler.

### Reviewing

It's worth having a look at previous Governor migrations, e.g. https://github.com/wormhole-foundation/wormhole/pull/4219/files#diff-8eec9902996736b121d6ee0317870d9f156d407f690871a05a62ca3777e8a854.

You can consider this PR as changing the `MessagePublication` serialization format to version `N+1`. As a result:
- Functions that (un)marshaled version `N-1` have been removed
- Functions that (un)marshaled version `N` have been relabelled "old" and/or deprecated and/or are now only executed in the `isOld` branches
- New functions are added to support (un)marshaling version `N+1`

### Future work

The old marshal/unmarshal methods can be removed from the codebase. This will likely require performing an upgrade for the Governor's key-value storage schema to ensure that the new message publications do not conflict with the old ones. Scaffolding code for this upgrade is already present in the Governor package.


node/pkg/common/chainlock.go
```diff
@@ -207,7 +256,115 @@ func UnmarshalOldMessagePublicationWithTxHash(data []byte) (*MessagePublication,
...
+		buf     = make([]byte, 0, bufSize)
+	)
+
+	// TxID (and length)
+	buf = append(buf, uint8(txIDLen))
+	buf = append(buf, msg.TxID...)
+
+	// Timestamp
+	tsBytes := make([]byte, 8)
+	//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
```
djb15: Can you use `nosec` here instead of `nolint`?

node/pkg/common/chainlock.go
```diff
@@ -272,6 +429,105 @@ func UnmarshalMessagePublication(data []byte) (*MessagePublication, error) {
...
+
+	// Unreliable
+	if !validBinaryBool(data[pos]) {
+		return ErrInvalidBinaryBool
+	}
+	msg.Unreliable = data[pos] != 0
+	pos++
+
+	// verificationState
+	if data[pos] > NumVariantsVerificationState {
```
djb15: Should this be `>=` as the highest value in the enum is 5?

node/pkg/common/chainlock.go
```diff
@@ -207,7 +256,115 @@ func UnmarshalOldMessagePublicationWithTxHash(data []byte) (*MessagePublication,
...
+	// - EmitterAddress
+	// - IsReobservation
+	// - Unreliable
+	// - verificationState
+	// - Payload length
+	// - Payload
+
+	const (
+		// TODO: is this big enough?
+		PayloadSizeMax = math.MaxUint16
```
djb15: What's the reasoning for adding a size cap at all?

node/pkg/common/chainlock.go
```diff
@@ -207,7 +256,115 @@ func UnmarshalOldMessagePublicationWithTxHash(data []byte) (*MessagePublication,
...
+	)
+
+	// TxID (and length)
+	buf = append(buf, uint8(txIDLen))
+	buf = append(buf, msg.TxID...)
+
+	// Timestamp
+	tsBytes := make([]byte, 8)
+	//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
+	be.PutUint64(tsBytes, uint64(msg.Timestamp.Unix()))
```
djb15, line 265: Reason for moving away from `uint32`?

node/pkg/common/chainlock.go
```diff
@@ -272,6 +429,105 @@ func UnmarshalMessagePublication(data []byte) (*MessagePublication, error) {
...
+
+	// Read TxID
+	msg.TxID = make([]byte, txIDLen)
+	copy(msg.TxID, data[pos:pos+int(txIDLen)])
+	pos += int(txIDLen)
+
+	// Timestamp
+	timestamp := be.Uint64(data[pos : pos+8])
+	// Nanoseconds are not serialized
+	//nolint:gosec // uint64 and int64 have the same number of bytes, and Unix time won't be negative.
```
djb15: Same re `nosec`

node/pkg/common/chainlock.go
johnsaigle: IMO it's good to always have an explicit upper bound rather than doing unlimited reads. I guess there's some implicit constraint which would be the max of the payload size that any chain could send? I'd rather have one configured though

node/pkg/common/chainlock.go
johnsaigle: What about the year 2107? 😄 I can change it back if you prefer

node/pkg/common/chainlock.go
johnsaigle: Hm I think it should actually be
```suggestion
	if data[pos] > NumVariantsVerificationState-1 {
```
Because the enum starts at 0 and there are 6 variants. So the valid range for `data[pos]` here is 0-5 inclusive

node/pkg/common/chainlock.go
bruce-riley: I think if you want to cap the size of the payload, you need to socialize this more widely to make sure it will not impact any integrators.

node/pkg/common/chainlock.go
```diff
@@ -15,8 +15,54 @@ import (
...
-const HashLength = 32
-const AddressLength = 32
+const (
+	HashLength    = 32
+	AddressLength = 32
+
+	// The minimum size of a marshaled message publication. It is the sum of the sizes of each of
+	// the fields plus length information for fields with variable lengths (TxID and Payload).
+	marshaledMsgSizeMin = 1 + // TxID length (uint8)
+		20 + // TxID ([]byte), minimum length of 20 bytes
```
bruce-riley: Any reason not to use `HashLength` here? I think the TxID is either an `ethCommon.Hash` (32 bytes) or `solana.Signature` (64 bytes).

node/pkg/common/chainlock.go
johnsaigle: That's a good idea, thanks. For now I can delete this check and we could come back to it later after socializing it.

node/pkg/common/chainlock.go
johnsaigle: thanks that makes sense. HashLength is 32 though, so how does that work with Solana.Signatures?

node/pkg/common/chainlock.go
johnsaigle: This should be fixed now and I added a unit test

node/pkg/common/chainlock.go
```diff
@@ -148,7 +197,7 @@ func (msg *MessagePublication) Marshal() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-// UnmarshalOldMessagePublicationWithTxHash deserializes a MessagePublication from when the TxHash was a fixed size ethCommon.Hash.
+// Deprecated: UnmarshalOldMessagePublicationWithTxHash deserializes a MessagePublication from when the TxHash was a fixed size ethCommon.Hash.
```
bruce-riley: Although this function is currently obsolete (since all guardians have long since updated to the latest format), don't you need to do something similar where this function matches what's currently in prod and the standard one is the new format, then update the DB code to use a different key prefix for the new one? And then when you read the DB, look for both old and new, and if you find any old ones, rewrite them in the new format and delete the old ones. I know it's a pain, but I think it's necessary. . .

node/pkg/common/chainlock.go
bruce-riley: Actually, reading farther down in the code, it looks like you add the txID length to this min, so should it even be included here?

node/pkg/common/chainlock.go
johnsaigle: I think you're right but I was wondering if I should do that in this PR or in a separate one. What do you think?

bruce-riley: If you use a follow on PR, there is the risk that this will get deployed without that, and database records in the old format will not get properly reloaded.

johnsaigle: I thought it would be OK since the new functions aren't actually called, but at the same time I agree that it would be better to not risk the gap.

node/pkg/common/chainlock.go
johnsaigle: I did some research and there's not a consistent maximum transaction or block size across all of the supported chains, and so I don't think it makes sense to set a limit here.
It's possible to do this in upstream the watchers instead if we want, but that's a task for another time.

node/pkg/common/chainlock.go
johnsaigle: I removed that extra length check. Any more info on the Solana signatures bit?

bruce-riley: Yeah, the TxID is no longer fixed length. It is currently either 32 or 64, but I guess in the future it might be a different length. I guess the min it should be now is 32. But since you add the exact length in later, I'm not sure why you need to add anything here? (And why 20 anyway?)

Bottom line, can you just delete line 25 because you add in the lenght below?

johnsaigle: The 20 was to represent an EVM address but I realize that was a mitskae.

Line 25 is the timestamp, did you mean a different line?

The goal with this context is to do a quick failure for byte slices less than this length.

johnsaigle: I made some updates, let me know what you think. It turns out `HashLength` was only used by the previous "old" method (which this PR removes), so I have renamed it to be more specific.

node/pkg/common/chainlock.go
johnsaigle: This PR now includes a migration so I'll mark this conversation as resolved.

node/pkg/common/chainlock.go
```diff
@@ -148,66 +214,106 @@ func (msg *MessagePublication) Marshal() ([]byte, error) {
...
-	msg.Timestamp = time.Unix(int64(unixSeconds), 0)
 
-	if err := binary.Read(reader, binary.BigEndian, &msg.Nonce); err != nil {
-		return nil, fmt.Errorf("failed to read nonce: %w", err)
-	}
+	payloadLen := len(msg.Payload)
+	// Set up for serialization
+	var (
+		be      = binary.BigEndian
+		bufSize = marshaledMsgLenMin + txIDLen + payloadLen
```
djb15: Should you also subtract `TxIDLenMin` from the sum for `bufSize`? Since `marshaledMsgLenMin` already includes `TxIDLenMin`?

node/pkg/common/chainlock.go
```diff
@@ -272,6 +378,141 @@ func UnmarshalMessagePublication(data []byte) (*MessagePublication, error) {
 	return msg, nil
 }
 
+// UnmarshalBinary implements the BinaryUnmarshaler interface for MessagePublication.
+func (m *MessagePublication) UnmarshalBinary(data []byte) error {
+
+	// fixedFieldsLen is the minimum length of the fixed portion of a message publication.
+	// It is the sum of the sizes of each of the fields plus length information for fields with variable lengths (TxID and Payload).
```
djb15: You don't include the `txIdLength` so I would update this comment

node/pkg/common/chainlock.go
```diff
@@ -272,6 +378,141 @@ func UnmarshalMessagePublication(data []byte) (*MessagePublication, error) {
...
+	pos += int(txIDLen)
+
+	// TxID has a dynamic length, so now that we've read it, check that the remaining data is long enough for the rest of the message. This means that all fixed-length fields can be parsed with a payload of 0 or more bytes.
+	if len(data)-pos < fixedFieldsLen {
+		return ErrInputSize{msg: "data too short after reading TxID", got: len(data)}
+	}
+
+	// Timestamp
+	timestamp := be.Uint64(data[pos : pos+8])
+	// Nanoseconds are not serialized
```
djb15: What do you mean by "Nanoseconds are not serialized"?

node/pkg/common/chainlock_test.go
```diff
@@ -29,7 +48,325 @@ func encodePayloadBytes(payload *vaa.TransferPayloadHdr) []byte {
...
+			if tt.errorChecker != nil {
+				tt.errorChecker(t, err)
+			} else if tt.expectedErr != nil {
+				require.ErrorIs(t, err, tt.expectedErr, "expected specific error type for test case: %s", tt.name)
+			}
+		})
+	}
+}
+
+func FuzzMessagePublicationUnmarshalBinary(f *testing.F) {
```
djb15, line 296: Love to see it!

node/pkg/db/governor.go
```diff
@@ -383,7 +340,7 @@ func (d *Database) GetChainGovernorDataForTime(logger *zap.Logger, now time.Time
 				transfers = append(transfers, v)
 
 			} else if isOldTransfer(key) {
```
djb15, line 342: Do you still need this `else if` block if you're using the same unmarshaller as "new/current" transfers? Or is the idea here to keep the block for the future case where we do another upgrade and the format changes and we want to use a different unmarshaller?

node/pkg/common/chainlock.go
johnsaigle: ~~Before we were storing the timestamps in `u32` (4 bytes) but a timestamp in Go is actually 8 bytes. My understanding is that the other 4 bytes were storing nanoseconds, but it's been a while since I wrote this.~~

node/pkg/db/governor.go
johnsaigle: Yeah it's mostly just to keep the scaffolding in place. I can add a comment explaining this

node/pkg/common/chainlock.go
johnsaigle: Yeah I think so, I'll revisit this

johnsaigle: Added the subtraction.

node/pkg/common/chainlock.go
johnsaigle: I've added this in and made the comment more descriptive.

node/pkg/common/chainlock.go
johnsaigle: I've updated the code comment to be more clear.

johnsaigle: The comment is motivated by this handling of time data in the old unmarshal method: https://github.com/wormhole-foundation/wormhole/blob/c21afca3ff09f11d0b4fb0d383f83308c9747321/node/pkg/common/chainlock.go#L338-L342

johnsaigle: Worth comparing the Marshal implementations also:
https://github.com/wormhole-foundation/wormhole/blob/c21afca3ff09f11d0b4fb0d383f83308c9747321/node/pkg/common/chainlock.go#L203

https://github.com/wormhole-foundation/wormhole/blob/c21afca3ff09f11d0b4fb0d383f83308c9747321/node/pkg/common/chainlock.go#L262-L266

Basically it seems like the old way to Marshal was truncating the underlying timestamp value (int64) to uint32, and I guess this is because some of those bits hold nanoseconds. Then when the data is loaded, the nanosecond param of `time.Unix()` is set to 0.

In the new approach, all of the underlying bytes of the timestamp are written to the structure, but the nanoseconds are still set to 0.

Let me know what you think.

djb15: Gotcha, this makes sense thanks for explaining!

djb15 (dismissed):

node/pkg/governor/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Performing a database upgrade
+
+A database upgrade is required whenever the serialized format of a `MessagePublication` changes.
+This occurs if a field is added, removed, or changed or if the marshaling format changes.
+
+The Governor has to handle this in a special way because it must be able to read the old format and write the new format.
+This is required so that the Governor can continue to monitor transfers and messages that have already been recorded.
+It only needs to support the old format for the duration of its sliding window as the old transfers and messages will
+automatically dropped after the duration of the window.
```
bruce-riley: Nit: "be dropped".

node/pkg/governor/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### Database APIs
+
+The actual database CRUD calls should be kept version-agnostic to avoid an explosion in the number of versioned
+methods and the accompanying maintenance burden.
+
+### Live Migration
+
+When the Governor restarts and loads its data from the key-value store, it marks which transfers and/or messages
+are in the old format and saves them in the new format right away. This minimizes the time needed to support
```
bruce-riley: It also deletes the old versions, right?

node/pkg/governor/README.md
```diff
@@ -0,0 +1,54 @@
+# Governor
```
bruce-riley, line 1: Thanks for documenting all this.

node/pkg/governor/README.md
johnsaigle: Yes, I'll reword this

djb15 (approved):

node/pkg/db/governor.go
```diff
@@ -383,7 +340,9 @@ func (d *Database) GetChainGovernorDataForTime(logger *zap.Logger, now time.Time
...
-				v, err := unmarshalOldTransfer(val)
+				// NOTE: This is the intentionally the same as IsTransfer branch
```
bruce-riley: Nit: "is intentionally the same", drop the first "the".

node/pkg/db/governor_test.go
```diff
@@ -102,37 +98,291 @@ func TestTransferMsgID(t *testing.T) {
...
+
+// TestIsTransferV4 tests the IsTransfer function for the current transfer format.
+// The V4 suffix matches the "GOV:XFER4:" prefix used by the current transfer implementation.
+func TestIsTransferV4(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    []byte
+		expected bool
+	}{
+		{
```
bruce-riley, line 139: Add a case for "old transfer"?

bruce-riley (approved):

node/pkg/db/governor_test.go
johnsaigle: Added an `XFER3` case to make sure it fails

node/pkg/common/chainlock.go
```diff
@@ -66,12 +130,14 @@ type MessagePublication struct {
 	ConsistencyLevel uint8
 	EmitterChain     vaa.ChainID
 	EmitterAddress   vaa.Address
-	Payload          []byte
-	IsReobservation  bool
+	// NOTE: there is no upper bound on the size of the payload. Wormhole supports arbitrary payloads
+	// due to the variance in transaction and block sizes between chains. However, during deserialization,
+	// payload lengths are bounds-checked against math.MaxInt to prevent makeslice panics from malformed input.
```
johnsaigle:
```suggestion
	// payload lengths are bounds-checked against [PayloadLenMax] to prevent makeslice panics from malformed input.
```

PR #4430, opened 2025-07-11
node(governor): Remove duplication in unit test set-up

johnsaigle:
- Pass a testing instance to the set-up function and exit early on nil/err conditions rather than making each individual test perform this action

### DIFF
--- a/node/pkg/governor/governor_test.go
+++ b/node/pkg/governor/governor_test.go
@@ -649,19 +649,14 @@ func newChainGovernorForTestWithLogger(t *testing.T, ctx context.Context, logger
 	gov := NewChainGovernor(logger, &db, common.GoTest, true, "")
 
 	err := gov.Run(ctx)
-	if err != nil {
-		require.NoError(t, err)
-	}
+	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


node/pkg/governor/governor_test.go
```diff
@@ -656,31 +638,29 @@ func TestTrimmingAllTransfersShouldReturnZero(t *testing.T) {
...
-		return nil, fmt.Errorf("ctx is nil")
-	}
+func newChainGovernorForTestWithLogger(t *testing.T, ctx context.Context, logger *zap.Logger) *ChainGovernor {
+	require.NotNil(t, ctx)
 
 	var db guardianDB.MockGovernorDB
 	gov := NewChainGovernor(logger, &db, common.GoTest, true, "")
 
 	err := gov.Run(ctx)
 	if err != nil {
```
djb15: You can remove the `if` statements for all of these now where you're calling into `require.NoError`

djb15 (approved):

mdulin2 (approved):
Seems like a good change. Removes some duplicated code.

PR #4431, opened 2025-07-11
node(governor): refactor `dayLength` and `dailyLimit` fields and change wording

johnsaigle:
- Remove the `daylength` field configured on the Governor in favour of using a constant. This was always configured to be the same value, and it unclear conceptually to change the length of minutes in a day, when in fact we want to change the duration of a sliding window of time.
- Rename `dailyLimit` to `USDLimit` and add code comments explaining the relationship to the sliding window. Update all code and documentation references accordingly.
- Change wording in code comments to refer to a sliding window. This more accurate reflects that the Governor works with a rolling window rather than as discrete calendar days of 24h.
- Use separate constants for mainnet and devnet deployments

Note: because `dailyLimit` is an exported field, and is included in log ouput, **this likely constitutes a breaking change** for packages that import the Governor code as well as for monitoring tools. They will need to update their code to use the new variable name.


node/pkg/governor/governor.go
```diff
@@ -110,8 +112,10 @@ type (
 	// `transfers` with positive Value represent outgoing transfers from the emitterChainId. Transfers with negative
 	// Value represent incoming transfers of Assets that can Flow Cancel.
 	chainEntry struct {
-		emitterChainId          vaa.ChainID
-		emitterAddr             vaa.Address
+		emitterChainId vaa.ChainID
+		emitterAddr    vaa.Address
+		// The amount of notional value that can leave the chain during the period
+		// defined by the sliding window.
 		dailyLimit              uint64
```
djb15: Makes sense to change this variable name too imo

johnsaigle: @djb15 I changed it to USDLimit and updated all the code and documentation references.

johnsaigle: This will be breaking change so I updated the PR description.
If this ends up not being worth it I'm fine to leave `dailyLimit` as-is and just remove the day length field.

djb15: Ahhh ok yeah if Bruce and co don't think it's worth making a breaking change here then I agree you can revert back to the commit pre this change

bruce-riley:
For historical reference, `dayLengthInMinutes` was made a variable rather than a constant so we could use a shorter "day" in testing. 24 hours is a long time to wait in tilt. 😁 

But I guess we don't really do that much governor testing in tilt, so it doesn't matter.

node/pkg/governor/governor.go
bruce-riley: I think it's reasonable to change it if you want. Anyone who uses it will just need to make a change whenever they decide to pick up a new version of the package. 🤷

bruce-riley (approved):

djb15 (dismissed):

johnsaigle:
Adding draft mode until I get around to fixing the conflicts

node/pkg/governor/governor_prices.go
```diff
@@ -152,7 +152,7 @@ func (gov *ChainGovernor) queryCoinGecko(ctx context.Context) error {
 	// between 1000 and 2000 tokens.
 	throttle := make(chan int, 1)
 	go func() {
-		ticker := time.NewTicker(time.Duration(15) * time.Second)
+		ticker := time.NewTicker(15 * time.Second)
```
pleasew8t, line 155: Any chance this 15 could become a constant? Should it be `coinGeckoQueryIntervalInMins`?

pleasew8t (approved):

PR #4446, opened 2025-07-25
Node: Remove deprecated chains

bruce-riley:
This chain removes references to Terra, Oasis, Aurora, Karura, Acala and XLPA from the guardian.

This includes removing them from the go SDK and the protobuf. This involves leaving comments indicating that they are obsolete so no one reuses the chain IDs.


node/hack/repair_terra/repair.go
```diff
@@ -31,21 +31,15 @@ import (
 )
 
 var fcdMap = map[vaa.ChainID]string{
```
johnsaigle: Maybe this whole file should be deleted?

node/pkg/watchers/evm/chain_config.go
```diff
@@ -66,16 +66,12 @@ var (
 		vaa.ChainIDPolygon: {Finalized: true, Safe: false, EvmChainID: 137, PublicRPC: "https://polygon-bor-rpc.publicnode.com", ContractAddr: "0x7A4B5a56256163F07b2C80A7cA55aBE66c4ec4d7"},
 
 		vaa.ChainIDAvalanche: {InstantFinality: true, Finalized: true, Safe: true, EvmChainID: 43114, PublicRPC: "https://avalanche-c-chain-rpc.publicnode.com", ContractAddr: "0x54a8e5f9c4CbA08F9943965859F6c34eAF03E26c"},
-		vaa.ChainIDOasis:     {InstantFinality: true, Finalized: false, Safe: false, EvmChainID: 42262, PublicRPC: "https://emerald.oasis.dev/", ContractAddr: "0xfe8cd454b4a1ca468b57d79c0cc77ef5b6f64585"},
-		// vaa.ChainIDAurora:    Not supported in the guardian.
```
johnsaigle, line 74: Maybe this line should stay, alongside the Gnosis, Btc comments and so on? Or else all of them could be removed

bruce-riley:
> Isn't Snaxchain, also, deprecated?

Is it? It's hard to keep up. I'll remove it.

panoel (approved):

sdk/vaa/structs.go
```diff
@@ -210,26 +210,22 @@ const (
 	ChainIDSolana ChainID = 1
 	// ChainIDEthereum is the ChainID of Ethereum
 	ChainIDEthereum ChainID = 2
-	// ChainIDTerra is the ChainID of Terra
+	// WARNING: ChainIDTerra is only supported in devnet / Tilt.
 	ChainIDTerra ChainID = 3
```
johnsaigle, line 214: I'm wonder if the methods like `GetAllNetworkIDs()` should specifically exclude these ChainIDs from their results as a workaround? I think as written they'll still return Terra 1 and 2 in results even though they're not actually supported.

bruce-riley: That would require passing the env or a bool into those functions, because they would need to return the Terras if we are running in tilt.

That seems like a lot of work for something that should be temporary.

FYI, per @panoel's request, I created an issue for the Core Protocol Team to fix the tilt issue and finish removing the Terras.

johnsaigle: Sounds reasonable, thanks for the explanation.

Issue #4447, opened 2025-07-28
After stopping the spy and restarting it immediately, the spy cannot receive the VAA received by the guardian


liuxiaomeiG:
## Description and context

After stopping the spy and restarting it immediately, the spy cannot receive the VAA received by the guardian

## Steps to reproduce

1. Start guardian and spy, send cross-chain transactions, spy can receive vaa broadcast by guardian;
2. Stop spy and restart immediately, send cross-chain transactions

## Experienced behavior

spy cannot receive vaa received by guardian

## Expected behavior

spy can receive vaa received by guardian

PR #4448, opened 2025-07-28
node: fix after stopping the spy and restarting it immediately, the s…

liuxiaomeiG:
…py cannot receive the VAA received by the guardian #4447


node/cmd/common/utils.go
```diff
@@ -0,0 +1,21 @@
+package common
```
bruce-riley: Is there some reason that this can't go in `node/pkg/common`?

bruce-riley (dismissed):

node/cmd/common/utils.go
liuxiaomeiG: I didn't realize it could be put in this package. It has been moved in.

bruce-riley (approved):

PR #4450, opened 2025-07-28
ci: Minor updates for successful run in fork

djb15:
Kudos to @yevshev for these:

1. Slightly increase the readiness timeout from 5 minutes (default) to 10 minutes. We were seeing the following when running CI in a fork:
```
Error: Readiness timeout after 5m0s: target query-sdk-ci-tests:runtime not ready
```
2. Debian Buster repos [have been moved](https://lists.debian.org/debian-devel-announce/2024/03/msg00003.html), so to build the images from scratch need to update the sources. We don't see this normally as the images are cached.
3. Fix lint in yaml, has no net effect on behaviour

With these changes and https://github.com/wormhole-foundation/wormhole/pull/4443, running CI from private forks should be resolved.


djb15:
`readiness_timeout` was only added in Tilt `v0.35.0` so we can't make that patch here. Will look to downgrade the Tilt installation on the fork to `v0.33.3` to see if it resolves the build timeouts. No need to review for now

djb15:
Confirmed using Tilt `v0.33.3` also works in fork with self hosted runner

PR #4451, opened 2025-07-29
node(transfer verifier): Update TransferIsValid method to return results per Message rather than per Receipt

johnsaigle:
- Update TransferIsValid method to support per-message validation

- Add `msgID` parameter (equivalent to VAA ID) to `TransferIsValid` method
  signature to enable validation of specific messages within transaction
  receipts. Update caching mechanism to store per-message validation
  results instead of per-transaction results. Modify receipt processing
  to track message-specific evaluations and validate transfer amounts on
  a per-message basis.

Key changes:
- Add msgID parameter to TransferIsValid method for targeted message validation
- Replace single Result field in evaluation struct with Results map keyed by message ID
- Update receipt summary to track transfers per message ID rather than aggregated amounts
- Modify invariant checking to validate each message's transfers individually
- Add MsgID method to generate unique message identifiers from LogMessagePublished events
- Add sequence field to LogMessagePublished struct for message identification

### DIFF
--- a/node/pkg/txverifier/evm.go
+++ b/node/pkg/txverifier/evm.go
@@ -20,15 +20,16 @@ const (
 
 // TODO: More errors should be converted into sentinel errors instead of being created and returned in-line.
 var (
-	ErrCouldNotGetNativeDetails = errors.New("could not parse native details for transfer")
-	ErrDepositFromWrongAddress  = errors.New("parsed Deposit event has wrong address")
-	ErrInvalidReceiptArgument   = errors.New("invalid TransferReceipt argument")
-	ErrInvalidUpsertArgument    = errors.New("nil argument passed to upsert")
-	ErrNoMsgsFromTokenBridge    = errors.New("no messages published from Token Bridge")
-	ErrNotTransfer              = errors.New("payload is not a token transfer")
-	ErrParsedReceiptIsNil       = errors.New("nil receipt after parsing")
-	ErrReceiptHasNoMsgPub       = errors.New("receipt has no LogMessagePublished events")
-	ErrTxHashIsZeroAddr         = errors.New("txHash is the zero address")
+	ErrCachedReceiptHasNoDataForMessage = errors.New("cached receipt has no data for this message")
+	ErrCouldNotGetNativeDetails         = errors.New("could not parse native details for transfer")
+	ErrDepositFromWrongAddress          = errors.New("parsed Deposit event has wrong address")
+	ErrInvalidReceiptArgument           = errors.New("invalid TransferReceipt argument")
+	ErrInvalidUpsertArgument            = errors.New("nil argument passed to upsert")
+	ErrNoMsgsFromTokenBridge            = errors.New("no messages published from Token Bridge")
+	ErrNotTransfer                      = errors.New("payload is not a token transfer")
+	ErrParsedReceiptIsNil               = errors.New("nil receipt after parsing")
+	ErrReceiptHasNoMsgPub               = errors.New("receipt has no LogMessagePublished events")
+	ErrTxHashIsZeroAddr                 = errors.New("txHash is the zero address")
 )
 
 // Custom error type used to signal that a core invariant of the token bridge has been violated.
@@ -54,6 +55,10 @@ func (i InvariantError) Error() string {
 //	false and err: the receipt could not be properly processed or is not a token transfer.
 func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 	ctx context.Context,
+	// The message ID is a combination of the emitter chain, emitter address, and sequence.
+	// If empty, the code will return true if all of the messages are valid, and false otherwise.
+	// This should only be done for testing.
+	msgIDStr string,
 	txHash common.Hash,
 	// If nil, this code will fetch the receipt using the TransferVerifier's connector.
 	receipt *geth.Receipt,
@@ -69,11 +74,35 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 	tv.logger.Debug("detected LogMessagePublished event",
 		zap.String("txHash", txHash.String()))
 
-	// Caching: record used/inspected tx hash.
+	// Remains empty if the caller wants to check all messages in a receipt.
+	var msgID msgID
+	if msgIDStr != "" {
+		// Validate the validMsgID parameter if provided.
+		validMsgID, err := NewMsgID(msgIDStr)
+		if err != nil {
+			return false, err
+		}
+		if validMsgID.EmitterChain != tv.chainIds.wormholeChainId {
+			return false, errors.New("msgID's emitter chain does not match the chain being monitored")
+		}
+		msgID = validMsgID
+	}
+
+	// Check if the message has already been verified. If so, skip receipt parsing and return the cached result.
 	if eval, exists := tv.evaluations[txHash]; exists {
 		tv.logger.Debug("skip: transaction hash already processed",
 			zap.String("txHash", txHash.String()))
-		return eval.Result, nil
+
+		if isValidTransfer, ok := eval.Results[msgID]; ok {
+			return isValidTransfer, nil
+		}
+
+		// Should never happen, probably indicates a bug in the code.
+		tv.logger.Warn(
+			ErrCachedReceiptHasNoDataForMessage.Error(),
+			zap.String("msgID", msgID.String()),
+		)
+		return false, ErrCachedReceiptHasNoDataForMessage
 	}
 
 	// Get the full transaction receipt for this txHash if it was not provided as an argument.
@@ -86,8 +115,6 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 			return false, txReceiptErr
 		}
 	}
-
-	eval := evaluation{Receipt: receipt}
 
 	// Caching: if the new receipt is newer than the last one, update the last block number.
 	// This is important to prevent the Transfer Verifier from recording a block number that is older
@@ -124,14 +151,10 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 
 	// Ensure that the amount coming in is at least as much as the amount requested out.
 	summary, processErr := tv.validateReceipt(transferReceipt)
-	if summary != nil {
-		tv.logger.Debug("finished processing receipt", zap.String("txHash", txHash.String()), zap.String("summary", summary.String()))
-
-		if summary.logsProcessed == 0 {
-			tv.logger.Warn("receipt logs empty for tx", zap.String("txHash", txHash.Hex()))
-		}
-	} else {
-		tv.logger.Debug("finished processing receipt (but summary is nil)", zap.String("txHash", txHash.String()))
+	tv.logger.Debug("finished processing receipt", zap.String("txHash", txHash.String()), zap.String("summary", summary.String()))
+	if summary == nil {
+		tv.logger.Warn("receipt summary is nil", zap.String("txHash", txHash.String()))
+		return false, processErr
 	}
 
 	if processErr != nil {
@@ -147,9 +170,12 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 		return false, nil
 	}
 
-	// Cache successful results.
-	eval.Result = true
-	tv.addToCache(txHash, &eval)
+	cacheEvaluation := NewEvaluation(receipt.BlockNumber.Uint64())
+	for msgID, isValid := range summary.msgPubResult {
+		cacheEvaluation.Results[msgID] = isValid
+	}
+
+	tv.addToCache(txHash, &cacheEvaluation)
 
 	return true, nil
 }
@@ -160,17 +186,17 @@ func (tv *TransferVerifier[ethClient, Connector]) pruneCache() {
 	var numPrunedEvals int
 
 	// Iterate over recorded transaction hashes, and clear receipts older than `pruneDelta` blocks
-	for hash, eval := range tv.evaluations {
-		if eval.Receipt != nil && eval.Receipt.BlockNumber != nil {
-			if eval.Receipt.BlockNumber.Uint64() < tv.lastBlockNumber-tv.pruneHeightDelta {
+	for txHash, eval := range tv.evaluations {
+		if eval != nil {
+			if eval.blockNumber < tv.lastBlockNumber-tv.pruneHeightDelta {
 				numPrunedEvals++
-				delete(tv.evaluations, hash)
+				delete(tv.evaluations, txHash)
 			}
 		} else {
 			// NOTE: Kind of sloppy to delete these right away, but we shouldn't be caching
 			// many nil receipts anyway.
 			numPrunedEvals++
-			delete(tv.evaluations, hash)
+			delete(tv.evaluations, txHash)
 		}
 	}
 
@@ -569,6 +595,7 @@ func (tv *TransferVerifier[evmClient, connector]) parseReceipt(
 			messagePublications = append(messagePublications, &LogMessagePublished{
 				EventEmitter:    log.Address,
 				MsgSender:       logMessagePublished.Sender,
+				Sequence:        logMessagePublished.Sequence,
 				TransferDetails: transferDetails,
 			})
 
@@ -693,7 +720,6 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
 
 	// Process LogMessagePublished events.
 	for _, message := range *receipt.MessagePublications {
-		td := message.TransferDetails
 
 		validateErr := validate[*LogMessagePublished](message)
 		if validateErr != nil {
@@ -705,18 +731,12 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
 			tv.logger.Debug("skip: irrelevant LogMessagePublished event")
 			continue
 		}
-
-		upsertErr := upsert(&summary.out, key, message.TransferAmount())
-		if upsertErr != nil {
-			return nil, upsertErr
+		msgID, err := tv.MsgID(message)
+		if err != nil {
+			return nil, err
 		}
 
-		tv.logger.Debug("successfully parsed a LogMessagePublished event payload",
-			zap.String("tokenAddress", td.OriginAddress.String()),
-			zap.String("tokenChain", td.TokenChain.String()),
-			zap.String("amount", td.Amount.String()))
-
-		summary.logsProcessed++
+		summary.addTransferOut(msgID, key, message.TransferAmount())
 	}
 
 	tv.logger.Debug("done building receipt summary", zap.String("summary", summary.String()))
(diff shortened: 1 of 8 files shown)


node/pkg/txverifier/evm.go
```diff
@@ -40,20 +41,24 @@ func (i InvariantError) Error() string {
 	return fmt.Sprintf("invariant violated: %s", i.Msg)
 }
 
-// TransferIsValid processes a token transfer receipt based on a LogMessagePublished event. It fetches
-// the full transaction receipt associated with the txHash, and parses all
+// TransferIsValid processes a Message Publications representing a token transfer.
```
pleasew8t: nit: "a message publication" (lowercase, and singular?)

node/pkg/txverifier/evm.go
```diff
@@ -40,20 +41,24 @@ func (i InvariantError) Error() string {
...
-// transfers and token deposits into the token bridge as inbound transfers. It then
+// transfers and token deposits/transfers into the token bridge as inbound transfers. It then
 // verifies that the sum of the inbound transfers is at least as much as the sum of
 // the outbound transfers.
 //
 // Return values:
 //
-//	true:  the token transfer's receipt is valid.
+//	true: the token transfer's in the messagepublication is valid.
 //	false and nil: the token transfer has violated an invariant and is unsafe.
```
pleasew8t: question: are we guaranteeing that `true, err` is never possible? I'm of the opinion our current logic around the return from this function (including Sui here) is quite complicated. No need to change, but we could probably just document and say it explicitely/add some form of check for this.

pleasew8t: also, `messagepublication` -> `message publication`

node/pkg/txverifier/evm.go
```diff
@@ -40,20 +41,24 @@ func (i InvariantError) Error() string {
...
 //	false and err: the receipt could not be properly processed or is not a token transfer.
 func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
 	ctx context.Context,
+	// The message ID is a combination of the emitter chain, emitter address, and sequence.
```
pleasew8t, line 51: nit: we could add here that it's similar to the VAA ID that would eventually get generated? That way there's no confusion about what this is.

node/pkg/txverifier/evm.go
```diff
@@ -69,11 +74,35 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
-	if eval, exists := tv.evaluations[txHash]; exists {
+	// Remains empty if the caller wants to check all messages in a receipt.
+	var msgID msgID
+	if msgIDStr != "" {
+		// Validate the validMsgID parameter if provided.
+		validMsgID, err := NewMsgID(msgIDStr)
+		if err != nil {
+			return false, err
+		}
+		if validMsgID.EmitterChain != tv.chainIds.wormholeChainId {
```
pleasew8t: If we are checking the emitter chain we should be consistent and also check the emitter address.

node/pkg/txverifier/evm.go
```diff
@@ -123,33 +150,36 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
+	tv.logger.Debug("finished processing receipt", zap.String("txHash", txHash.String()), zap.String("summary", summary.String()))
+	if summary == nil {
+		tv.logger.Warn("receipt summary is nil", zap.String("txHash", txHash.String()))
+		return false, validateReceiptErr
 	}
 
-	if processErr != nil {
-		// Check if the error type is an invariant error. If not, it's just a parsing error.
+	if validateReceiptErr != nil {
+		// Check if the error type is an invariant error. If not, it's just a parsing error. It's important to
```
pleasew8t: I'm not entirely sure how this responds differently to a single invalid transfer in the batch invalidating the entire transaction. I think the flow needs to change here:
* check if a non-invariant-violation error ocurred, and return `false, err` if so
* if the errors contain an invariant violation && the msg ID is empty, return `false, nil`

Basically, if a MsgId did exist, we continue with caching the evaluations, and return from the cached evaluation the result for the specific MsgId.

node/pkg/txverifier/evm.go
johnsaigle: Yeah I kind of hate `bool, err` as a return type in Go. We could return a custom struct or something, but it feels important to distinguish between parsing and invariant errors, idk.
`true, err` should never be possible but it's worth making that super clear.

node/pkg/txverifier/evm.go
johnsaigle: Yeah I could add something like "See the SDK" that has MessageID function that produces the same output

node/pkg/txverifier/evmtypes.go
```diff
@@ -506,19 +535,22 @@ func parseERC20TransferEvent(logTopics []common.Hash, logData []byte) (common.Ad
 
 // Abstraction over a LogMessagePublished event emitted by the core bridge.
 type LogMessagePublished struct {
-	// Which contract emitted the event.
+	// Which contract emitted the event. (not to be confused with EmitterAddress in a VAA)
 	EventEmitter common.Address
 	// Which address sent the transaction that triggered the message publication.
 	MsgSender common.Address
```
pleasew8t: nit: "The address of the account that initiated the bridge transfer."

node/pkg/txverifier/evmtypes.go
```diff
@@ -639,42 +673,56 @@ func (r *TransferReceipt) String() string {
 }
 
 // Summary of a processed TransferReceipt. Contains information about relevant
-// transfers requested in and out of the bridge.
+// transfers requested in and out of the bridge. Message Publications within a receipt
```
pleasew8t, line 682: nit: "publications" (lowercase)

node/pkg/txverifier/evmtypes.go
```diff
@@ -1002,6 +1041,11 @@ func validate[L TransferLog](tLog TransferLog) error {
 			return &InvalidLogError{Msg: "destination is not set"}
 		}
 
+		// It's invalid for the core bridge to send a message to itself.
+		if Cmp(log.Sender(), log.Emitter()) == 0 {
```
pleasew8t: Honestly, I'd say if this happens it's crazy suspicious and not just an invalid error. Something to alert on perhaps?

node/pkg/txverifier/evmtypes_test.go
```diff
@@ -532,6 +529,19 @@ func TestValidateLogMessagePublished(t *testing.T) {
 				},
 			},
 		},
+		"invalid: msg.sender cannot be equal to emitter": {
```
pleasew8t, line 532: Shouldn't this be "emitter cannot be core bridge"? Remember to remove if you decide to remove the core bridge check.

node/pkg/txverifier/evmtypes.go
johnsaigle: I'm going to leave this as-is because it refers to a specific type `MessagePublication` that is used elsewhere in the Guardian

node/pkg/txverifier/evmtypes.go
johnsaigle: I'll add this check to `TransferIsValid` so we get access to the logger. I'm going to leave it here so we can keep unit testing it.

node/pkg/txverifier/evmtypes_test.go
johnsaigle: I think the test checks any sender against any emitter rather than checking emitter against the hard-coded core bridge address, so I think the existing name is more accurate. Do you feel otherwise?

node/pkg/txverifier/evm.go
johnsaigle: I like the capitalization to tie it directly to the same struct used in the watchers. I'll remove the plural

node/pkg/txverifier/evm.go
johnsaigle: I've added more comments and rewritten (hopefully: simplified) the return values of this function. I believe it's impossible to hit `true, err` now but let me know if I made a mistake.

node/pkg/txverifier/evm.go
johnsaigle: Hey I rewrote this section with the idea that it could be easier to follow. Let me know what you think about the new version.

node/pkg/txverifier/evmtypes.go
```diff
@@ -639,44 +679,154 @@ func (r *TransferReceipt) String() string {
...
+		}
+	}
+	return true
+}
+
+func (s *ReceiptSummary) InvariantErrors() string {
+	if len(*s.problems) == 0 {
+		return ""
+	}
+	var errs = make([]string, 2)
```
djb15: Why size 2?

node/pkg/txverifier/evm.go
```diff
@@ -69,11 +69,39 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
+		if err != nil {
+			return false, err
+		}
+
+		msgID = validMsgID
+	}
+
+	// Check if the message has already been verified, and if so, return the cached result.
+	if receiptEvaluation, exists := tv.evaluations[txHash]; exists {
+		if !receiptEvaluation.msgInCache(msgID) {
```
djb15: If `checkWholeReceipt` is true, won't this do a lookup of an empty message id and fall into this branch?

node/pkg/txverifier/evm.go
```diff
@@ -69,11 +69,39 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
+			return false, ErrCachedReceiptHasNoDataForMessage
+		}
+
+		// Return true if the entire receipt is safe.
+		if receiptEvaluation.ReceiptSummary.isSafe() {
+			return true, nil
+		}
+
+		//  Return true if the message is safe. (This function also checks that the receipt is safe.)
+		return receiptEvaluation.ReceiptSummary.msgSafe(msgID), nil
```
djb15: Surely this will never be true if `receiptEvaluation.ReceiptSummary.isSafe() == false`. Does `msgSafe` actually add any additional value vs `isSafe`?

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +749,132 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+		zap.Bool("receiptIsSolvent?", receiptIsSolvent),
+		zap.Bool("receipt isSafe?", receiptSummary.isSafe()),
+	)
+
+	return receiptSummary, nil
+}
+
+// isSolvent() reports whether the sum of the outbound transfers is at least as much as the sum of the inbound transfers.
+// It is used to examine the solvency of the entire receipt, without considering individual messages.
+func (receiptSummary *ReceiptSummary) isSolvent() bool {
```
djb15: Do you need to use some state here to decrement from `receiptSummary.in[transferOut.tokenID]` for each outbound transfer of a given token id? Since the `in` amounts are accumulated, but the `out` transfers are keyed by message id?

e.g. if amount in was 10 for token X, two out transfers of 8 of token X would be considered solvent (as both are <10)

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +749,132 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+		//   an invalid message that interferes with legitimate messages. For example, a rebasing token might
+		//   break the Token Bridge's invariant. If an attacker initiates a transfer of a rebasing asset along with another user's
+		//   transfer (e.g. in the case of a contract that allows permissionless claims), then the whole receipt
+		//   might be marked as invalid if we're not careful.
+		for msgID, transferOut := range receiptSummary.out {
+			if amountIn, exists := receiptSummary.in[transferOut.tokenID]; !exists {
+				// This is never OK, even in the edge cases documented above.
+				receiptSummary.msgPubResult[msgID] = false
+				receiptSummary.addProblem(ErrInvariantMissingCollateral, MsgInvariant)
+			} else {
```
djb15: What is this doing that `isSolvent` isn't already doing?

node/pkg/txverifier/evm_test.go
```diff
@@ -672,8 +911,190 @@ func TestValidateReceipt(t *testing.T) {
...
+							TokenChain:    2,
+							Amount:        big.NewInt(1),
+						},
+					},
+				},
+			},
+			// Parsing error, so the summary should be nil.
+			// expected: ReceiptSummary{},
+		},
+		"unsafe receipt: total amount in is too small even though messages are valid when considered individually": {
```
djb15, line 942: This test case is targeting the wrong thing I think; the input tokens are native, output are USDC

djb15: Based on my earlier comments I can get this test case to pass when it shouldn't after correcting the tokens with the following diff. We deposit 3 tokens and withdraw 4, and no errors are thrown.

```
diff --git a/node/pkg/txverifier/evm_test.go b/node/pkg/txverifier/evm_test.go
index f10e3be63..36498a24c 100644
--- a/node/pkg/txverifier/evm_test.go
+++ b/node/pkg/txverifier/evm_test.go
@@ -974,8 +974,8 @@ func TestValidateReceipt(t *testing.T) {
 						TokenChain:   vaa.ChainIDEthereum,
 						From:         eoaAddrGeth,
 						To:           tokenBridgeAddr,
-						Amount:       big.NewInt(1),
-						OriginAddr:   nativeAddrVAA,
+						Amount:       big.NewInt(3),
+						OriginAddr:   usdcAddrVAA,
 					},
 				},
 				MessagePublications: &[]*LogMessagePublished{
@@ -985,10 +985,10 @@ func TestValidateReceipt(t *testing.T) {
 						Sequence:     1,
 						TransferDetails: &TransferDetails{
 							PayloadType:   TransferTokens,
-							OriginAddress: nativeAddrVAA,
+							OriginAddress: usdcAddrVAA,
 							TargetAddress: eoaAddrVAA,
 							TokenChain:    2,
-							Amount:        big.NewInt(1),
+							Amount:        big.NewInt(2),
 						},
 					},
 					{
@@ -997,37 +997,27 @@ func TestValidateReceipt(t *testing.T) {
 						Sequence:     2,
 						TransferDetails: &TransferDetails{
 							PayloadType:   TransferTokens,
-							OriginAddress: nativeAddrVAA,
+							OriginAddress: usdcAddrVAA,
 							TargetAddress: eoaAddrVAA,
 							TokenChain:    2,
-							Amount:        big.NewInt(1),
+							Amount:        big.NewInt(2),
 						},
 					},
 				},
 			},
 			expected: ReceiptSummary{
 				in: map[string]*big.Int{
-					fmt.Sprintf("%s-%s", nativeAddrVAA, "2"): big.NewInt(1),
+					fmt.Sprintf("%s-%s", usdcAddrVAA, "2"): big.NewInt(1),
 				},
 				out: map[msgID]transferOut{
-					{Sequence: 1, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: {fmt.Sprintf("%s-%s", nativeAddrVAA, "2"), big.NewInt(1)},
-					{Sequence: 2, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: {fmt.Sprintf("%s-%s", nativeAddrVAA, "2"), big.NewInt(1)},
+					{Sequence: 1, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: {fmt.Sprintf("%s-%s", usdcAddrGeth, "2"), big.NewInt(1)},
+					{Sequence: 2, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: {fmt.Sprintf("%s-%s", usdcAddrGeth, "2"), big.NewInt(1)},
 				},
 				msgPubResult: map[msgID]bool{
 					{Sequence: 1, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: true,
 					{Sequence: 2, EmitterChain: vaa.ChainIDEthereum, EmitterAddress: VAAAddrFrom(tokenBridgeAddr)}: true,
 				},
-				problems: &Problems{
-					// This manifests as two problems because each individual message out is ruled insolvent relative to the whole receipt.
-					&ReceiptProblem{
-						violationKind: ReceiptInvariant,
-						err:           *ErrInvariantReceiptInsolvent,
-					},
-					&ReceiptProblem{
-						violationKind: ReceiptInvariant,
-						err:           *ErrInvariantReceiptInsolvent,
-					},
-				},
+				problems: &Problems{},
 			},
 		},
 		"unsafe receipt: one valid message, one invalid message": {

```

node/pkg/txverifier/evmtypes.go
johnsaigle: The `prealloc` lint said I should add a value so I just picked one. Happy to use a different one.

node/pkg/txverifier/evm.go
johnsaigle: Yes I'll add a check here that checkWholeReceipt has to be false

node/pkg/txverifier/evm_test.go
johnsaigle: Thanks I've added an adapted version of this test in the latest commit.

node/pkg/txverifier/evm.go
johnsaigle: I added the state check in this function in the latest commit

node/pkg/txverifier/evm.go
johnsaigle: I removed the extra receipt.isSafe check from the msgSafe method.

node/pkg/txverifier/evm.go
```diff
@@ -69,11 +72,41 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
+			)
+			return false, ErrCachedReceiptHasNoDataForMessage
+		}
+
+		// Return true if the entire receipt is safe.
+		if receiptEvaluation.ReceiptSummary.isSafe() {
+			return true, nil
+		}
+
+		//  Return true if the message is safe. (This function also checks that the receipt is safe.)
```
djb15:
```suggestion
		//  Return true if the message is safe.
```

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +754,179 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+	}
+
+	for _, transferOut := range receiptSummary.out {
+		if amountIn, exists := receiptSummary.in[transferOut.tokenID]; !exists {
+			// Some outbound asset is not present in the inbound transfers.
+			insolventAssets = append(insolventAssets, transferOut.tokenID)
+		} else {
+			// Some asset is insolvent.
+			if transferOut.amount.Cmp(amountIn) == 1 {
+				insolventAssets = append(insolventAssets, transferOut.tokenID)
```
djb15: I think you can get rid of this? If you just leave the decrementing in you're basically doing the same below when you check the solvency for each token

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +754,179 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+		// If any token balance is less than zero, then the entire receipt is insolvent.
+		if adjustedTotal.Cmp(big.NewInt(0)) < 0 {
+			insolventAssets = append(insolventAssets, tokenID)
 		}
 	}
 
-	return summary, invariantErrors
+	// Remove duplicates.
+	slices.Sort(insolventAssets)
+	insolventAssets = slices.Compact(insolventAssets)
```
djb15: I think you can also get rid of the duplicate check then too

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +754,179 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+		//
+		// Individual messages can be evaluated independently of each other and the receipt as a whole.
+		// A message is valid if and only if the amount of its outbound asset is less than or equal to the amount
+		// of the corresponding inbound asset.
+		//
+		// Taking all of the above into account, there are two important edge-cases to consider:
+		// - Each individual message might be satisfied by the inbound amount, but the sum of the inbound transfers
+		//   could be less than the sum of the outbound transfers. In this case, the messages ARE valid but the receipt
+		//   is NOT valid. For example, if the inbound sum is 10 tokens and there are two messages each of 10 tokens,
+		//   then each individual message is valid but the receipt is not.
```
djb15: Might be worth revisiting this docstring with the latest changes as it feels a bit misleading to me

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +754,179 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
+		)
 	}
+	tv.logger.Debug("receipt summary after solvency check", zap.String("summary", receiptSummary.String()))
+	if len(*receipt.MessagePublications) == 1 && len(insolventAssets) > 0 {
+		// Only one message publication exists in the receipt. Check for solvency.
+		// There is only one message publication in the receipt, but we don't have the msgID to do a lookup.
+		for msgID := range receiptSummary.out {
+			receiptSummary.msgPubResult[msgID] = false
+		}
+	} else {
```
djb15, line 796: Might be worth adding another branch here for receipts with a single message publication and no insolvent assets, which would make that `msgId` valid

johnsaigle: Made an equivalent change in the latest commit, though I didn't create another branch. Let me know if it looks OK.

johnsaigle:
@djb15 I removed the Problem struct in the latest commit. I felt like it was possible to get the same thing done even without a boolead, instead relying on the the `isSafe()` and `isMsgSafe()` methods. Let me know what you think.

node/pkg/txverifier/evm.go
```diff
@@ -20,40 +23,40 @@ const (
...
-	ErrTxHashIsZeroAddr         = errors.New("txHash is the zero address")
+	ErrCachedReceiptHasNoDataForMessage = errors.New("cached receipt has no data for this message")
+	ErrCouldNotGetNativeDetails         = errors.New("could not parse native details for transfer")
+	ErrDepositFromWrongAddress          = errors.New("parsed Deposit event has wrong address")
+	ErrInvalidReceiptArgument           = errors.New("invalid TransferReceipt argument")
+	ErrInvalidUpsertArgument            = errors.New("nil argument passed to upsert")
+	ErrNoMsgsFromTokenBridge            = errors.New("no messages published from Token Bridge")
+	ErrNotTransfer                      = errors.New("payload is not a token transfer")
+	ErrParsedReceiptIsNil               = errors.New("nil receipt after parsing")
+	ErrReceiptHasNoMsgPub               = errors.New("receipt has no LogMessagePublished events")
```
djb15: Not used

node/pkg/txverifier/evm.go
```diff
@@ -705,62 +754,159 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
...
-				tv.logger.Warn(
-					"requested amount out is larger than amount in for this token. It may be a deflationary or rebasing asset. Review is required.",
-					zap.String("key", key),
-					zap.String("amountIn", amountIn.String()),
-					zap.String("amountOut", amountOut.String()),
-				)
-				invariantErrors = errors.Join(invariantErrors, &InvariantError{Msg: "requested amount out is larger than amount in"})
+				}
+
+				receiptSummary.msgPubResult[msgID] = msgIsSolvent
```
djb15: I think we can get rid of this `else` block and just use the `insolventAssets` array? If a token is insolvent any transfers of that token are invalid, otherwise the transfer is valid

djb15: Not sure why this comment has included the deleted code, I'm just referring to the new `else` block logic starting from line `821` (technically line `806` if we include the comments + logger calls)

johnsaigle:
@djb15 I removed the else block you flagged.

I noticed after this though that some msgPubs were missing in the unit tests. It turns out that this loop was the part that ensured each message publication has a `msgPub` entry in the summary.

I changed this so that, as the LogMessagePublished parts are being parsed/validated, it also adds msgPub entry to the the map.

It has a default entry of `true` which the insolvency and missing assets checks then can mark as false. Another way to do this would be to start everything as `false`, then later after all the checks are done, go back and find the _solvent_ assets, and mark those as true.

node/pkg/txverifier/evm.go
```diff
@@ -69,11 +70,41 @@ func (tv *TransferVerifier[ethClient, Connector]) TransferIsValid(
...
+		validMsgID, err := tv.ParseMsgID(msgIDStr)
+		if err != nil {
+			return false, err
+		}
+
+		msgID = validMsgID
+	}
+
+	// Check to see if the receipt is cached.
+	if receiptEvaluation, exists := tv.evaluations[txHash]; exists {
```
pleasew8t: Could we rename this to `evaluation`? `receiptEvaluation` is the same name used for a type, and seems like the kind of thing we want to avoid.

pleasew8t: or `rcptEvaluation`

node/pkg/txverifier/evm.go
```diff
@@ -612,7 +651,7 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
 		return nil, ErrInvalidReceiptArgument
 	}
 
-	invalidErr := receipt.Validate()
+	invalidErr := receipt.SanityCheck()
 	if invalidErr != nil {
```
pleasew8t, line 655: I noticed that we are already calling this in `updateDetails()`. I don't really understand why, since both these methods are ultimately "private" and should never be called in isolation. Can `receipt.SanityCheck()` be called once, prior to calling `updateReceipt()` in `TransferIsValid`?

johnsaigle: Maybe we could save optimizing the code for another follow-up PR? I'm happy if you want to collect this and other feedback in a GH issue that we can come back to.

pleasew8t: Yeah I'm happy to defer this to a future change.

node/pkg/txverifier/utils.go
```diff
@@ -26,6 +27,69 @@ const (
...
+	}
+
+	chainID, err := vaa.StringToKnownChainID(parts[0])
+	if err != nil {
+		return msgID{}, err
+	}
+
+	supported := IsSupported(chainID)
+	if !supported {
+		return msgID{}, err
```
pleasew8t: If we return here, the error is going to be `nil`. Is this intentional (return empty msgID + nil error), or did you want to return an actual error here?

johnsaigle: nice catch, no this should be an error, not nil.

djb15 (approved):

evan-gray (approved):
only reviewed code owner changes in watchers/evm

PR #4456, opened 2025-08-01
ci: Add lint rule that ensures a mutex guards "AlreadyLocked" functions

johnsaigle:
Implements a custom rule using the `revive` linter framework. (Ordinarily `revive` could be added and configured in the usual `golangci-lint` way but not if you use custom rules.)

This is done by creating a thin binary wrapper around the `revive` CLI in accordance with their official documentation. (https://github.com/mgechev/revive/tree/master?tab=readme-ov-file#using-revive-as-a-library)

This specific rule checks that the use of our `*AlreadyLocked` functions are only called when a mutex is locked within the same function scoped. This should help us protect against data corruption issues for parallel access to e.g. the database.

This PR also adds documentation and linting configuration so that we can continue to extend revive if we want and easily incorporate it into our existing lint set-up.


johnsaigle:
This will be refactored into the infrastructure created by #4841

PR #4474, opened 2025-08-18
notary: Add admin rescue commands

johnsaigle:
- Add commands to change the status of delayed and blackholed messages
  in the Notary
- Change Notary commands to use MessageID consistently instead of a
  combination of pointers to MessagePublication and VAAHash
  
This PR is based on #4315 and will stay in draft until that PR is merged.

### DIFF
--- a/node/pkg/notary/admincommands.go
+++ b/node/pkg/notary/admincommands.go
@@ -253,6 +253,7 @@ func createTestMessagePublication() *common.MessagePublication {
 	}
 	payloadBytes := encodePayloadBytes(payload)
 
+	//#nosec G404: this value is only used in devnet mode and only for synthetic data. Cryptographic randomness is not required.
 	sequence := rand.Uint64()
 	msgPub := &common.MessagePublication{
 		TxID:             eth_common.HexToHash("0x06f541f5ecfc43407c31587aa6ac3a689e8960f36dc23c332db5510dfc6a4063").Bytes(),


node/pkg/adminrpc/adminserver.go
```diff
@@ -1112,6 +1121,66 @@ func (s *nodePrivilegedService) ChainGovernorResetReleaseTimer(_ context.Context
...
+		return nil, ErrInvalidMsgID
+	}
+
+	err := s.notary.BlackholeDelayedMsg(req.VaaId)
+	if err != nil {
+		return nil, err
+	}
+
+	return &nodev1.NotaryBlackholeDelayedMessageResponse{
+		Response: "Blackholed message",
```
djb15, line 1140: Can you add in the `vaaId` string into the responses?

node/pkg/notary/admincommands.go
```diff
@@ -0,0 +1,50 @@
...
+	err := n.release([]byte(msgID))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// RemoveBlackholedMsg removes a message from the blackholed list and deletes it from the database.
+func (n *Notary) RemoveBlackholedMsg(msgID string) error {
```
djb15, line 72: Should we add such messages that are removed from the blackholed list to the pending queue with 0 release time to ensure they are processed? I'm going to guess a reobservation would most likely lead to the message being blackholed again

proto/node/v1/node.proto
```diff
@@ -405,6 +407,42 @@ message ChainGovernorResetReleaseTimerResponse {
...
+
+message NotaryRemoveBlackholedMessageRequest {
+  string vaa_id = 1;
+}
+
+message NotaryRemoveBlackholedMessageResponse {
+  string response = 1;
+}
+
+message NotaryResetReleaseTimerRequest {
```
djb15, line 469: Missing the admin command for this message

node/pkg/notary/notary.go
```diff
@@ -329,7 +331,7 @@ func (n *Notary) blackhole(msg *common.MessagePublication) error {
 	n.mutex.Unlock()
 
 	// When a message is blackholed, it should be removed from the delayed list and database.
-	err := n.removeDelayed(msg)
+	err := n.removeDelayed(msg.MessageID())
```
djb15: Probably applies to the notary PR rather than this one, but should we attempt to remove from the delayed list + database before adding to the blackholed list? Just in case the removal fails we don't want the message in both places

djb15 (commented):
Just reviewed the latest commit

johnsaigle:
Now that the Notary has been merged, this is in my court to rebase and fix-up. In doing so I'll go back and respond to the comments

node/pkg/adminrpc/adminserver.go
djb15: @johnsaigle I don't think you've added these to the responses yet?

johnsaigle: Ah oops I added them in the protobuf definitions but not in the actual responses.

node/pkg/notary/notary.go
```diff
@@ -181,7 +184,7 @@ func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err erro
 	// Return early if the message has already been blackholed. This is important in case a message
 	// is reobserved or otherwise processed here more than once. An Anomalous message that becomes
 	// delayed and later blackholed should not be able to be re-added to the Delayed queue.
-	if n.IsBlackholed(msg) {
+	if n.blackholed.Contains(msg.MessageID()) {
```
bemic: Did not we change this line to `n.IsBlackholed(msg)` In the latest PR ([link](https://github.com/wormhole-foundation/wormhole/pull/4315#discussion_r2331360357))

node/pkg/common/pendingmessage.go
```diff
@@ -33,27 +33,24 @@ func (p PendingMessage) Compare(other PendingMessage) int {
...
 		return nil, fmt.Errorf("marshal pending message: %w", err)
 	}
 
-	vaa.MustWrite(buf, binary.BigEndian, bz)
-
+	buf := new(bytes.Buffer)
+	// Compare with [PendingTransfer.Marshal].
+	// #nosec G115  -- int64 and uint64 have the same number of bytes, and Unix time won't be negative.
+	vaa.MustWrite(buf, binary.BigEndian, uint64(p.ReleaseTime.Unix()))
+	buf.Write(msgPubBz)
```
djb15, line 45: What's the difference between `buf.Write` and `vaa.MustWrite`? Why the change?

node/pkg/notary/notary.go
johnsaigle: Ah yeah thanks, I think I lost that line in a rebase

node/pkg/notary/admincommands.go
johnsaigle: It's added below.

node/pkg/notary/admincommands.go
```diff
@@ -0,0 +1,100 @@
...
+	delayErr := n.delay(removedMsgPub, time.Duration(0))
+	if delayErr != nil {
+		return delayErr
+	}
+
+	return nil
+}
+
+// ResetReleaseTimer resets the release timer for a delayed message.
+func (n *Notary) ResetReleaseTimer(msgID string, delayDays uint32) error {
```
djb15: Do we not also need the hook in `adminserver` to call into this?

johnsaigle: Yeah I just realized that, added into the latest commit

node/pkg/common/pendingmessage.go
johnsaigle: During testing the previous implementation yielded a buffer that was 8 bytes shorter than expected. (This came up when trying to Unmarshal retrieved `PendingMessage`s from the DB). 
I think that the buffer was being overwritten by `MustWrite` rather than being appended to, so the timestamp was being lost? Not totally sure.

Also, `binary.Write` used by `vaa.MustWrite` apparently only works for fixed-size types 😨 

Compare [bytes.Write](https://pkg.go.dev/bytes#Buffer.Write) used here with [binary.Write](https://pkg.go.dev/encoding/binary#Write) that is used by `vaa.MustWrite`.

See also https://github.com/wormhole-foundation/wormhole/blob/bd4207984b8a9e9e30f847c37a7361269d94fc95/sdk/vaa/payloads.go#L303-L320 which uses a mixture of both methods.

This might be something to look into more deeply outside the scope of this PR. Doing ad-hoc mixing of the methods here feels like an easy source of mistakes.

djb15: Thanks for the detailed explanation! Agree it would be nice review further as part of future work

node/cmd/guardiand/adminclient.go
djb15, line 69: Do you not also need to do the same with the new commands?

johnsaigle:
@djb15 The latest changes now include a new `script/` under scripts. I ran it against `tilt` locally. It should cover the process of moving messages back and forth between the delayed and blackholed lists, as well as resetting the release timer.

In the process I also added some utility commands to inject synthetic messages for testing. These are only accessible in "unsafe dev mode". There's also an ability for admins to list the messages in both lists and introspect the release time for delayed messages.

It turns out this getting this to work involved modifying a few more files but the changes should be minimal.

Interestingly, I also noticed and fixed an existing bug in the admin service runnable. The Governor is supposed to be optional, but it was listed as a strict dependency for the admin service command. If it was disabled, the current admin server could not be launched, which would have prevented other admin actions like reobservations!

djb15 (approved):
Great script, thanks for taking the time to put this test together! Also tested myself and worked as expected. And nice catch on the strict dependency!

node/pkg/notary/admincommands.go
```diff
@@ -0,0 +1,294 @@
+package notary
+
+// Admin commands for the Notary. This defines a public API for the Notary to be used by the Guardians.
+// This structure allows the admin-level commands to be implemented separately from the core Notary code.
+//
+// TODO: This file uses LoadAll() for reads which is not efficient. However, the DB is exists mostly to facilitate
```
panoel, line 6: Nit
```suggestion
// TODO: This file uses LoadAll() for reads which is not efficient. However, the DB exists mostly to facilitate
```

panoel (approved):

PR #4476, opened 2025-08-19
feat: Plasma chain id reservation

aapclark:


aadam-10:
was included in https://github.com/wormhole-foundation/wormhole/pull/4494

PR #4480, opened 2025-08-26
Add CODEOWNERS for Guardian Dependency Upgrades

mdulin2:
Allows the security team to review dependencies for major components before being upgraded. This was discussed in the security meeting this week.

Review the added `go.mod` files to ensure these are the correct ones to add. I intentionally left out Wormchain's `go.mod` , but I would consider adding it and others if others were passionate about it.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -96,6 +96,10 @@
 
 /node/pkg/watchers @evan-gray @panoel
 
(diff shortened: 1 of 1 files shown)


djb15 (approved):

evan-gray (approved):
I expected all go.mod files, but this is fine

PR #4482, opened 2025-08-27
solana/token_bridge: fix metaplex metadata handling and add support for token2022 tokens (including metadata pointers)

kcsongor:
Following the evolution of the Solana ecosystem, we are proposing changes to the Solana Token Bridge contract that add support for tokens whose Metaplex metadata has been resized and Token2022 tokens including metadata extensions. We detail these changes and their security implications.

## Rationale

Before diving into the specifics of the proposed upgrades, we want to motivate why these upgrades are necessary. There is a wider ecosystem sentiment (echoed by some of our competitors) that contracts should be immutable, and upgrades should be rolled out in the form of new immutable contracts with migration.

This argument has merit, but is unrealistic in an ecosystem such as Solana, where key ecosystem primitives like Metaplex token metadata and the official token standard are themselves mutable. What we are proposing here are upgrades to the Token Bridge contract specifically addressing the evolution of those primitives. No upgrades to the core Wormhole contract messaging contract is proposed here, which itself has no such mutable dependencies.

## Change 1: Metaplex metadata

Metaplex rolled out an upgrade to their metadata program on 10 October, 2024 ([[link](https://developers.metaplex.com/token-metadata/guides/account-size-reduction)](https://developers.metaplex.com/token-metadata/guides/account-size-reduction)), which resized the token metadata account to be smaller. Their intention of doing this was to save on rent costs, but the downstream effect is that any code (whether on-chain or off-chain) that expects the metadata to be a certain size will fail to deserialise the account. The Portal Token Bridge contract is affected by this.

To summarise, the token bridge validates the metadata account’s size to match the expected bytes, which now fails for the smaller metadata accounts. This means that any code path that accesses the metadata account (e.g. token attestation) fails for these tokens. The token bridge is effectively broken for new tokens.

### Proposed change

https://github.com/wormhole-foundation/wormhole/pull/4482/commits/aba6719c9d824b9cad2e16a65aa2711f8c65237c

The length check has been relaxed to accept both the old account size (679 bytes) and the new one (607 bytes).

### Security impact

The account length check is a redundant sanity check performed among other checks by the `deserialize_and_verify_metadata` function. This function verifies that the metadata account corresponds to the expected token. Since Metaplex metadata accounts are PDAs whose address is derived from the token's mint address, the critical security check is the PDA derivation verification. The length check provides no additional security guarantee beyond this PDA verification, making it safe to relax.

### Impact of change

With this patch rolled out, the token bridge will be able to support new tokens (such as those created by pump.fun) to be bridged out of Solana.

## Change 2: Token2022 support

The Token Bridge now supports the Token2022 standard in addition to legacy SPL tokens. [[Token2022](https://spl.solana.com/token-2022)](https://spl.solana.com/token-2022) supports [[token extensions](https://spl.solana.com/token-2022/extensions)](https://spl.solana.com/token-2022/extensions) which provide additional functionality beyond legacy SPL tokens. This implementation supports most Token2022 tokens, with specific exclusions for tokens that would require significant changes to transfer logic.

Token2022 tokens support custom metadata through the Metadata Pointer extension, which can point to an external account or embed metadata directly in the mint account. Both cases are handled.

### Proposed change

The implementation consists of several components:

**Core Token2022 Support:** https://github.com/wormhole-foundation/wormhole/pull/4482/commits/67e87f13678a831e446e83ffeb2ab5158921df4a

- Relaxes ownership checks on mint and token accounts to allow Token2022 program in addition to legacy SPL token program
- Updates token operations to use the appropriate token program instead of hardcoding legacy SPL

**SPL Token Library Changes:** https://github.com/wormhole-foundation/wormhole/pull/4482/commits/5b650eca73127810fcb7fb41bc6c87d11b95f652

- Vendors spl-token library with relaxed program ID checks (commit 7ae1b55)
- Removes hardcoded checks against only the legacy SPL token program

**Zero-Dependency Metadata Parser:** https://github.com/wormhole-foundation/wormhole/pull/4482/commits/265b30e10fd4f23ef476e968fdad2efcd87f4ab6

- Adds a dependency-free parser for Token2022 metadata extensions
- Parses both Metadata Pointer and embedded Token Metadata extensions

**Token2022 Metadata Support:** https://github.com/wormhole-foundation/wormhole/pull/4482/commits/d708cd13e845a8c6cb6ce913b2b30e2c311bd6ae

- Integrates Token2022 metadata parsing into the token bridge
- Handles external metadata accounts and embedded metadata

### Security impact

The changes to the Solitaire framework are localised and follow a similar pattern to how Anchor implements [[InterfaceAccounts](https://www.anchor-lang.com/docs/references/account-types#interfaceaccountinfo-t)](https://www.anchor-lang.com/docs/references/account-types#interfaceaccountinfo-t).

The critical security requirement is ensuring consistent token programs are used between related accounts. For example, when the token bridge creates a custody account to lock tokens during outbound transfers, this custody account must use the same token program as the token's mint account. Program consistency is enforced by always deriving the token program from the provided mint account in instructions, **never** using the client-provided token program parameter which would be unconstrained. This prevents DoS attacks where inconsistent token programs would make custody accounts unable to store tokens.

For metadata handling, the implementation uses a tiered verification approach:

- Metaplex metadata accounts undergo full PDA derivation verification
- Token2022 metadata pointers are validated against the provided metadata account
- Embedded metadata is only used when the metadata pointer points to the mint account itself, requiring no additional verification as the mint account is already verified

The metadata account is constrained in every case.

### Impact of change

With this change, most Token2022 tokens will be supported on Solana.

# Extension compatibility
### Metadata Pointer (Mint)

- Full support for metadata pointer extension
- Both embedded (self-referential) and external metadata pointers work
- For external accounts, metadata is assumed to be serialized in Metaplex standard format (the account doesn't need to be owned by Metaplex)

### CPI Guard

- This is enabled per token account. If the user disables the CPI guard on their account, they can transfer tokens. The Token Bridge will not enable CPI guard on its account. So transfers should work in theory.

### Default Account State (Mint)

- Allows configuring accounts to be frozen or not be default. No effect on the token bridge (obviously frozen tokens will not work)

### Interest Bearing Mint (Mint)/ ScaledUiAmount

- These extensions are UI changes and do not affect functionality

### Mint Close Authority (Mint)

- Shouldn't impact the bridge. If the token bridge has funds from the mint, the mint can't be closed

### Permanent Delegate

- A permanent delegate could revoke funds from the token bridge, so potentially the token could be bricked.
Otherwise should operate as normal

### Group Pointers (Mint)

- The Token Bridge will work only with tokens that use Metaplex for metadata

### Memo Transfer (Account)

- Requires an extra memo instruction to occur before each transfer, which the Token Bridge does not perform.
    - It be possible to deposit these in the token bridge and initiate a cross-chain transfer. If you try and bridge back to Solana the transfer out of the token bridge would fail.

### TransferFees/ConfidentialTransferFees

- The TransferFee extension requires a `mint` to be supplied. This is normally done with a `transfer_checked` instruction, but the Token Bridge uses the (deprecated) `transfer` function. So the transfer will revert.
    - https://github.com/solana-program/token-2022/blob/839b577fe6e11b309a5c53df1d66cd62f4a8167f/program/src/processor.rs#L385-L392
    - https://blog.offside.io/i/148944074/transfer-vs-transferchecked
- The accounting may not be correct after fees are applied, so the token bridge may not always be solvent for this token

### Transfer Hook

- Explicitly not supported

### Non-transferable

- Can't transfer to the token bridge

### Confidential Transfers

- The amount is hidden, so the token will be incompatible as the token bridge requires the ability to view the amount

### DIFF
--- a/whitepapers/0003_token_bridge.md
+++ b/whitepapers/0003_token_bridge.md
@@ -249,6 +249,10 @@ Since there is no way for a token bridge endpoint to know which other chain alre
 native asset on its chain, there may be transfers initiated for assets that don't have wrapped assets set up yet on the
 target chain. However, the transfer will become executable once the wrapped asset is set up (which can be done any time).
 
+The name and symbol fields of the Transfer payload are not guaranteed to be valid UTF8 strings.
+Implementations might truncate longer strings at the 32 byte mark, which may result in invalid UTF8 bytes at the end.
+Thus, any client wishing to present these as strings must validate them first, potentially dropping the garbage at the end.
+
 <!-- Local Variables: -->
 <!-- fill-column: 120 -->
 <!-- End: -->


nvsriram (approved):
LGTM!

* Verified the `spl-token` patch uses the correct Token 2022 address
* Verified `NEW_MAX_METADATA_LEN` is set correctly to 607 bytes
* Verified all (de)serializations and conversions between metadata formats work as intended
* Verified all supported metadata cases are handled as described in the comments

solana/Cargo.toml
```diff
@@ -11,6 +11,10 @@ members = [
     "solitaire/program",
     "solitaire/rocksalt",
 ]
+exclude = [
+    "modules/token_bridge/token_metadata_parser",
+]
 
 [patch.crates-io]
 memmap2 = { path = "bridge/memmap2-rs" }
+spl-token = { git = "https://github.com/wormholelabs-xyz/spl-token.git", rev = "7ae1b55" }
```
nonergodic: This would be more hygienic if it was under the wormhole-foundation org too (just like the metaplex-program-library is).

I assume it's in a standalone repo rather than part of the monorepo here because depending on the full weight of wormhole-foundation/wormhole.git would be extremely ugly when used elsewhere?

solana/modules/token_bridge/program/src/accounts.rs
```diff
@@ -106,39 +109,140 @@ impl<'b> Seeded<&SplTokenMetaDerivationData> for SplTokenMeta<'b> {
...
 }
 
+//New data length of spl token metadata account
+pub const NEW_MAX_METADATA_LEN: usize = 607;
+
+/// Converts Token-2022 metadata to Metaplex metadata format for compatibility
+fn convert_token2022_to_metaplex_metadata(
+    token_metadata: &token_metadata_parser::TokenMetadata,
+) -> spl_token_metadata::state::Metadata {
+    use spl_token_metadata::state::{
```
nonergodic, line 121: For this latest change, it's exceedingly painful/confusing that mpl-token-metadata is still aliased to spl-token-metadata. I think I would have eliminated the alias for greater clarity, even if it requires more code changes.

solana/modules/token_bridge/token_metadata_parser/src/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+            .ok_or(ParseError::UnexpectedEnd)?;
+        if end > mint_account_data.len() {
+            return Err(ParseError::UnexpectedEnd);
+        }
+        let val = &mint_account_data[offset..end];
+        offset = end;
+
+        if t == EXT_UNINITIALIZED {
+            break; // padding
+        }
```
nonergodic, line 136: This can never happen:

If a mint + extensions ends up having the same length as a multisig account, an [Unintialized](https://github.com/solana-program/token-2022/blob/de93095c0a32ad52d90b37d944fb17756ced4dc5/interface/src/extension/mod.rs#L1059) extension is [added at the end](https://github.com/solana-program/token-2022/blob/de93095c0a32ad52d90b37d944fb17756ced4dc5/interface/src/extension/mod.rs#L124-L130) to increase the length and avoid the clash. However, unlike all other extensions, `Uninitialized` is not followed by a length field.

At the same time, they always realloc the account when an extension is added or grows or shrinks. In other words, there can never be any useless padding at the end. (They also [top-up / return the lamports for such reallocs](https://github.com/solana-program/token-2022/blob/9cbf7a1e9bab57aabb71b4b84fc84e3670108573/program/src/extension/reallocate.rs#L76-L93))

So what will happen here is that if `t` were indeed `Uninitialized` then the `while` loop would have already skipped it, because `offset + TLV_TYPE_SIZE + TLV_LEN_SIZE` will be equal to `mint_account_data.len() + 2`.

So the code is safe as it is, but the check should be removed and the safety, even for the `Uninitialized` extension should be documented. The original implementation always parses stuff [in a two step manner](https://github.com/solana-program/token-2022/blob/de93095c0a32ad52d90b37d944fb17756ced4dc5/interface/src/extension/mod.rs#L168-L173), first reading the type, checking that it is not `Uninitialized` before reading the length.

solana/modules/token_bridge/token_metadata_parser/src/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+
+        match t {
+            EXT_METADATA_POINTER => {
+                if val.len() != 64 {
+                    return Err(ParseError::UnexpectedLength);
+                }
```
nonergodic, line 142: This feels overly defensive: If this check can be wrong, nothing can be trusted.

And likewise for all the following checks that, if they were to trigger, would mean that there would be a bug in how token 2022 serializes Metadata. Or I suppose a deeply breaking change, in which case the handling of `Uninitialized` should definitely be hardened to be in line with the defensiveness here.

solana/modules/token_bridge/token_metadata_parser/src/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+    None,
+    External(MetadataPointer),
+    Embedded(TokenMetadata),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MetadataPointer {
+    pub authority: Option<Pubkey>,
+    /// Where metadata lives. If equals the mint pubkey, metadata is embedded in the mint (TokenMetadata extension).
+    pub metadata_address: Pubkey,
```
nonergodic, line 34: [According to the spec](https://github.com/solana-program/token-2022/blob/de93095c0a32ad52d90b37d944fb17756ced4dc5/interface/src/extension/metadata_pointer/mod.rs#L21), both `authority` and `metadata_address` have optional type `OptionalNonZeroPubkey` :thinking:

solana/modules/token_bridge/program/src/accounts.rs
```diff
@@ -106,39 +109,140 @@ impl<'b> Seeded<&SplTokenMetaDerivationData> for SplTokenMeta<'b> {
...
+        Data,
+        Key,
+        Metadata,
+    };
+
+    let data = Data {
+        name: token_metadata.name.clone(),
+        symbol: token_metadata.symbol.clone(),
+        uri: token_metadata.uri.clone(),
```
nonergodic, line 130: Given that these can now all come from embedded token 2022 metadata, all of them can exceed the max length that was previously implicitly guaranteed by MPL metadata, since token-2022 metadata has no length limits for its fields.

This can now cause problems when creating attestations for tokens with names/symbols that use unicode and have more than 32 bytes in total, because [both the name and the symbol are simply truncated to 32 bytes without respecting unicode character bounds](https://github.com/wormhole-foundation/wormhole/blob/10eacbb06a0b75daff4b612d2112d0c733022a0b/solana/modules/token_bridge/program/src/messages.rs#L260-L268).

This was not a problem with MPL because the max length for names and symbols was [32](https://github.com/wormhole-foundation/metaplex-program-library/blob/1b71ebce639a0bf155074dcfeef509bbf1edd00a/token-metadata/program/src/state/metadata.rs#L11) and [10](https://github.com/wormhole-foundation/metaplex-program-library/blob/1b71ebce639a0bf155074dcfeef509bbf1edd00a/token-metadata/program/src/state/metadata.rs#L13) respectively, but that's now not the case any longer.

solana/modules/token_bridge/program/src/accounts.rs
```diff
@@ -106,39 +109,140 @@ impl<'b> Seeded<&SplTokenMetaDerivationData> for SplTokenMeta<'b> {
     }
 }
 
+//New data length of spl token metadata account
+pub const NEW_MAX_METADATA_LEN: usize = 607;
```
nonergodic, line 115: Might be worth linking to [here](https://developers.metaplex.com/token-metadata/guides/account-size-reduction) (or preferably a more stable link that is likely to persist over time, like a github diff) in the code, because "new" by itself doesn't really mean anything.

solana/modules/token_bridge/program/src/accounts.rs
kcsongor: addressed this by calling out this behaviour in the whitepaper. This was already the case from EVM chains, as the solidity contract truncates the first 32 bytes the same way

solana/modules/token_bridge/program/src/accounts.rs
kcsongor: added a link

solana/modules/token_bridge/token_metadata_parser/src/lib.rs
kcsongor: added a note

solana/Cargo.toml
kcsongor: moved over to the foundation

solana/modules/token_bridge/program/src/accounts.rs
```diff
@@ -106,39 +109,142 @@ impl<'b> Seeded<&SplTokenMetaDerivationData> for SplTokenMeta<'b> {
     }
 }
 
+/// New data length of spl token metadata account (post https://developers.metaplex.com/token-metadata/guides/account-size-reduction)
+/// The naming convention we adopt for future resizes is iterated NEW_* prefixes, e.g. NEW_NEW_NEW_NEW_NEW_MAX_METADATA_LEN.
+/// This will continue until morale improves.
```
nonergodic, line 114: :heart:

PR #4486, opened 2025-09-05
Aztec Fixes

svlachakis:


johnsaigle (changes requested):
Can you please explain why these changes are necessary? Also please get rid of the LLM-generated comments.

I don't understand why the JSON hashes are marked as optional also, that seems very unusual when fetching RPC info.

PR #4505, opened 2025-09-24
node: change aptos watcher to use a graphql indexer

panoel:


node/pkg/watchers/aptos/watcher.go
```diff
@@ -128,9 +139,12 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			logger.Info("Received obsv request", zap.Uint64("tx_hash", nativeSeq))
 
-			s := fmt.Sprintf(`%s?start=%d&limit=1`, eventsEndpoint, nativeSeq)
+			query := fmt.Sprintf(`query GetVersionBySequence { msg(where: {sequence_num: {_eq: %d}}) { version } }`, nativeSeq)
```
evan-gray, line 333: should we switch the reobservations to be based on the version directly?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -196,35 +223,47 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			// Walk through the array of events
+			for _, msg := range messages.Array() {
+				version := msg.Get("version")
+				sequenceNum := msg.Get("sequence_num")
+				if !version.Exists() || !sequenceNum.Exists() {
 					continue
 				}
 
-				eventSeq := eventSequence.Uint()
-				if nextSequence == 0 && eventSeq != 0 {
```
evan-gray, line 213: should this block stay? what prevents this new version from publishing the last event every startup?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -140,47 +154,60 @@ func (e *Watcher) Run(ctx context.Context) error {
...
 				// if nextSequence is 0, we look up the most recent event
-				s = fmt.Sprintf(`%s?limit=1`, eventsEndpoint)
+				// Get both version (for fetching tx) and sequence_num (for observeData)
+				query = "query GetLastEvent { msg(order_by: {sequence_num: desc}, limit: 1) { version sequence_num } }"
 			} else {
 				// otherwise just look up events starting at nextSequence.
 				// this will potentially return multiple events (whatever
 				// the default limit is per page), so we'll handle all of them.
-				s = fmt.Sprintf(`%s?start=%d`, eventsEndpoint, nextSequence)
+				query = fmt.Sprintf(`query GetNextEvents { msg(where: {sequence_num: {_gt: %d}}) { version sequence_num } }`, nextSequence)
```
evan-gray: does this need to be sorted ascending so we don't skip a page?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -128,9 +139,12 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			logger.Info("Received obsv request", zap.Uint64("tx_hash", nativeSeq))
 
-			s := fmt.Sprintf(`%s?start=%d&limit=1`, eventsEndpoint, nativeSeq)
+			query := fmt.Sprintf(`query GetVersionBySequence { msg(where: {sequence_num: {_eq: %d}}) { version } }`, nativeSeq)
 
-			body, err := e.retrievePayload(s)
+			// Add authorization header for indexer API calls
+			headers := make(map[string]string)
+			headers["Authorization"] = "Bearer " + e.aptosIndexerToken
```
evan-gray: should the token / auth be optional?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +321,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		return nil, err
+	}
+	return body, err
+}
+
+// processTransactionVersion fetches a transaction by version and extracts WormholeMessage events
+func (e *Watcher) processTransactionVersion(
+	logger *zap.Logger,
+	versionNum uint64,
+	sequenceForObserve uint64,
```
evan-gray: if a transaction contains multiple wormhole messages, this will end up hitting the endpoint multiple times. do we need / is it worthwhile to specify the sequence here?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +321,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	defer res.Body.Close()
+	body, err := common.SafeRead(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	return body, err
+}
+
+// processTransactionVersion fetches a transaction by version and extracts WormholeMessage events
+func (e *Watcher) processTransactionVersion(
```
evan-gray, line 586: can we add a unit or integration test for this?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +321,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
 	return body, err
 }
 
+//nolint:noctx // TODO: this function should use a context.
+func (e *Watcher) queryIndexer(query string, headers map[string]string) ([]byte, error) {
```
evan-gray, line 528: can we add a unit or integration test for this?

evan-gray (changes requested):
overall seems reasonable, though I'm not sure if tilt will pass. mostly nits. I could go either way about reobserving direcly by version or keeping it by event sequence. I think the obvious advantage of using the version is that reobservation would not require the indexer and that would probably be a good thing.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -82,16 +93,26 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			zap.String("rpc", e.aptosRPC),
+			zap.String("account", e.aptosAccount),
+			zap.String("handle", e.aptosHandle),
+		)
+	}
 
 	// SECURITY: the API guarantees that we only get the events from the right
 	// contract
-	var eventsEndpoint = fmt.Sprintf(`%s/v1/accounts/%s/events/%s/event`, e.aptosRPC, e.aptosAccount, e.aptosHandle)
+	var eventsEndpoint = e.aptosIndexerRPC
```
johnsaigle: nit: It looks like these two variables are just used for the log statement below, so you could consider in-lining them.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-	// sequence number to look up next.
+	// Important: There are 3 different sequence/version numbers in play:
+	// 1. Aptos Version: Transaction version number (like block number in EVM)
+	// 2. Aptos Sequence Number: Event-specific counter for WormholeMessage events
+	// 3. Wormhole Sequence: Protocol sequence inside the event data (goes into VAA)
+	//
+	// nextSequence tracks the Aptos Sequence Number (not version, not wormhole seq)
 	var nextSequence uint64 = 0
 
 	timer := time.NewTicker(time.Second * 1)
```
johnsaigle: This timer seems unused?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-	// (NOTE: this is not the same as the wormhole sequence id). The event
-	// endpoint is paginated, so we use this variable to keep track of which
-	// sequence number to look up next.
+	// Important: There are 3 different sequence/version numbers in play:
+	// 1. Aptos Version: Transaction version number (like block number in EVM)
+	// 2. Aptos Sequence Number: Event-specific counter for WormholeMessage events
+	// 3. Wormhole Sequence: Protocol sequence inside the event data (goes into VAA)
+	//
+	// nextSequence tracks the Aptos Sequence Number (not version, not wormhole seq)
 	var nextSequence uint64 = 0
```
johnsaigle: nit: This value is hard-coded to 0 and it's an argument to private methods. So it seems like it's not necessary to define here? The value could just be initialized within the methods themselves, and the parameter could be removed from the function signature.

node/cmd/guardiand/node.go
```diff
@@ -1658,12 +1678,18 @@ func runNode(cmd *cobra.Command, args []string) {
...
-			Rpc:       *aptosRPC,
-			Account:   *aptosAccount,
-			Handle:    *aptosHandle,
+			NetworkID:    "aptos",
+			ChainID:      vaa.ChainIDAptos,
+			Rpc:          *aptosRPC,
+			Account:      *aptosAccount,
+			Handle:       *aptosHandle,
+			IndexerRpc:   *aptosIndexerRPC,
+			IndexerToken: *aptosIndexerToken,
```
johnsaigle, line 1689: Should there be a check here that the indexer token is non-empty when indexer mode is selected?

johnsaigle, line 1689: Should there be a check here that the indexer token is non-empty when indexer mode is selected?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			}
+
+			if !gjson.Valid(string(eventsJson)) {
+				logger.Error("InvalidJson: " + string(eventsJson))
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+
+			events := gjson.ParseBytes(eventsJson)
+			for _, event := range events.Array() {
```
johnsaigle: Might be a good idea to do a length check and log on cases where the events.Array is empty. Otherwise the loop will be skipped silently

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +501,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	body, err := common.SafeRead(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	return body, err
```
johnsaigle:
```suggestion
	return body, nil
```
nit: Should be logically equivalent, but it's a bit more clear

node/pkg/watchers/aptos/watcher.go
```diff
@@ -140,47 +330,62 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			if !messages.Exists() || len(messages.Array()) == 0 {
+				logger.Error("No data.msg field in indexer response or empty array")
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
 
-				if newSeq.Uint() != nativeSeq {
-					logger.Error("newSeq != nativeSeq")
-					break
+			// Get the version from the first (and should be only) result
```
johnsaigle, line 360: Should we log if there are multiple results?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -196,37 +401,51 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			// Extract the msg array from data.msg
+			messages := response.Get("data.msg")
+			if !messages.Exists() {
+				logger.Error("No data.msg field in indexer response")
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+
+			// Walk through the array of events
+			for _, msg := range messages.Array() {
```
johnsaigle: Length check here? The loop will be skipped silently if the array is empty.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +501,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	}
+
+	// Parse the transaction data
+	txResult := gjson.ParseBytes(txData)
+
+	// Look for WormholeMessage in the events array
+	whEventType := fmt.Sprintf("%s::state::WormholeMessage", e.aptosAccount)
+	events := txResult.Get("events")
+	if events.Exists() {
+		for _, event := range events.Array() {
```
johnsaigle, line 615: Should this have a length check? The loop will be skipped silently if the array is empty.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -128,9 +313,14 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			logger.Info("Received obsv request", zap.Uint64("tx_hash", nativeSeq))
 
-			s := fmt.Sprintf(`%s?start=%d&limit=1`, eventsEndpoint, nativeSeq)
+			query := fmt.Sprintf(`query GetVersionBySequence { msg(where: {sequence_num: {_eq: %d}}) { version } }`, nativeSeq)
 
-			body, err := e.retrievePayload(s)
+			// Add authorization header for indexer API calls
+			headers := make(map[string]string)
```
johnsaigle: Nit: this field is static so it could be defined once outside of the ticker loop

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +501,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	// Set content type for GraphQL
+	req.Header.Set("Content-Type", "application/json")
+
+	// Add additional headers if provided
+	for key, value := range headers {
+		req.Header.Set(key, value)
+	}
+
+	client := &http.Client{}
+	res, err := client.Do(req)
```
johnsaigle, line 556: Could you add a comment here regarding the status code? I see in other places that the graphQL queries always return status code 200, but it would be good to make that explicit here.

johnsaigle: Similarly, is there an error field in the response that will get populated when something goes wrong? 
It seems like right now the error handling is implicit based on whether the JSON is parsed correctly or whether the message has the expected number of results. I think this could hide error cases within the GraphQL response (assuming that they all have 200 status codes and are valid JSON).

node/pkg/watchers/aptos/watcher.go
```diff
@@ -282,6 +501,87 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	// Look for WormholeMessage in the events array
+	whEventType := fmt.Sprintf("%s::state::WormholeMessage", e.aptosAccount)
+	events := txResult.Get("events")
+	if events.Exists() {
+		for _, event := range events.Array() {
+			eventType := event.Get("type")
+			if eventType.String() == whEventType {
+				// Extract the event data
+				data := event.Get("data")
+				if data.Exists() {
```
johnsaigle, line 628: is it an error to have Events that don't contain Data?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -140,47 +330,62 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			response := gjson.ParseBytes(body)
 
-			for _, chunk := range outcomes.Array() {
-				newSeq := chunk.Get("sequence_number")
-				if !newSeq.Exists() {
-					break
-				}
+			// Extract the msg array from data.msg
+			messages := response.Get("data.msg")
+			if !messages.Exists() || len(messages.Array()) == 0 {
```
johnsaigle, line 354: The bounds check is worth a loud comment here because the node will panic without it.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -140,47 +330,62 @@ func (e *Watcher) Run(ctx context.Context) error {
...
 				// otherwise just look up events starting at nextSequence.
 				// this will potentially return multiple events (whatever
 				// the default limit is per page), so we'll handle all of them.
-				s = fmt.Sprintf(`%s?start=%d`, eventsEndpoint, nextSequence)
+				query = fmt.Sprintf(`query GetNextEvents { msg(where: {sequence_num: {_gt: %d}}, order_by: {sequence_num: asc}) { version sequence_num } }`, nextSequence)
 			}
 
-			eventsJson, err := e.retrievePayload(s)
+			// Add authorization header for indexer API calls
+			headers := make(map[string]string)
```
johnsaigle: nit: this is static, so it could be built once outside of the ticker loop.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -196,37 +401,51 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-					continue
+				// Update nextSequence to track progress using the Aptos sequence_num
+				if aptosSeqNum > nextSequence {
+					nextSequence = aptosSeqNum
 				}
-				e.observeData(logger, data, eventSeq, false)
 			}
 
+			// Health check endpoint
+			aptosHealth := fmt.Sprintf(`%s/v1`, e.aptosRPC)
```
johnsaigle, line 471: nit: this is static, so it could be built once outside of the ticker loop.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+
+			logger.Debug(string(health) + string(eventsJson))
+
+			pHealth := gjson.ParseBytes(health)
+			blockHeight := pHealth.Get("block_height")
+
+			if blockHeight.Uint() > math.MaxInt64 {
```
johnsaigle, line 284: This type check looks wrong? (or maybe the log message should be more clear). Is the idea that the max block height for Aptos is MaxInt64?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+				}
+
+				// Update nextSequence to track progress
+				if sequenceNum > nextSequence {
+					nextSequence = sequenceNum
+				}
+			}
+
+			// Health check
+			aptosHealth := fmt.Sprintf(`%s/v1`, e.aptosRPC)
```
johnsaigle: nit: this is static, so it could be built once outside of the ticker loop.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -100,17 +121,181 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+			}
+
+			if !gjson.Valid(string(eventsJson)) {
+				logger.Error("InvalidJson: " + string(eventsJson))
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+
+			events := gjson.ParseBytes(eventsJson)
+			for _, event := range events.Array() {
```
johnsaigle: length check? otherwise the loop will be skipped silently

node/pkg/watchers/aptos/watcher.go
panoel: Removed them

node/pkg/watchers/aptos/watcher.go
panoel: Removed

node/cmd/guardiand/node.go
panoel: We are allowing for the possibility that future indexers may be run by the guardians themselves and then won't need a token.

panoel: We are allowing for the possibility that future indexers may be run by the guardians themselves and then won't need a token.

node/pkg/watchers/aptos/watcher.go
panoel: Added to indexer mode

node/pkg/watchers/aptos/watcher.go
panoel: Changed

node/pkg/watchers/aptos/watcher.go
panoel: Changed

node/pkg/watchers/aptos/watcher.go
panoel: Changed

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: It is def an error

node/pkg/watchers/aptos/watcher.go
panoel: Done

node/pkg/watchers/aptos/watcher.go
panoel: Changed

node/pkg/watchers/aptos/watcher.go
panoel: Changed

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: Honestly, I don't know.  I stole this from the legacy code.  Dirk added this into the legacy code in PR #4224.

node/pkg/watchers/aptos/watcher.go
panoel: Were you thinking that some error decoding be done in queryIndexer() or the calling function?

node/pkg/watchers/aptos/watcher.go
johnsaigle: Hmm OK the limitation seems to come from the Protobuf for Heartbeat messages: https://github.com/wormhole-foundation/wormhole/blob/565a8d1aff42fb5087dc6a6581b4dc01a8f1d9c2/node/pkg/proto/gossip/v1/gossip.pb.go#L1104-L1121

It's a little funky but not in-scope for this PR.

node/pkg/watchers/aptos/watcher.go
johnsaigle: Up to you. I'm not really familiar with the response format or contents. If it can be done in a general way, we may as well do it here. Otherwise, the calling function(s) might make more sense.

node/pkg/watchers/aptos/watcher.go
panoel: Fixed

node/pkg/watchers/aptos/watcher.go
panoel: Added

node/pkg/watchers/aptos/watcher.go
panoel: In theory, I agree.  I'm, also, scared to change more things because I think something in the reobservation cloud function may need to change.

evan-gray: keeping it as is for the time being should be reasonable then

node/cmd/guardiand/node.go
```diff
@@ -925,8 +929,24 @@ func runNode(cmd *cobra.Command, args []string) {
...
 	}
 
-	if !argsConsistent([]string{*aptosAccount, *aptosRPC, *aptosHandle}) {
-		logger.Fatal("Either --aptosAccount, --aptosRPC and --aptosHandle must all be set or all unset")
+	// Validate Aptos configuration - support both legacy and indexer modes
+	aptosBasic := []string{*aptosRPC, *aptosAccount, *aptosHandle}
+
+	if !argsConsistent(aptosBasic) {
+		logger.Fatal("Either --aptosRPC, --aptosAccount, and --aptosHandle must all be set together or all unset")
+	}
```
evan-gray, line 935: I don't think this is used anywhere else anymore, so can probably just return to the previous code for these 4 lines

node/cmd/guardiand/node.go
```diff
@@ -925,8 +929,24 @@ func runNode(cmd *cobra.Command, args []string) {
...
+
+	// If Aptos is enabled, validate indexer configuration
+	if *aptosRPC != "" {
+		// Only aptosIndexerRPC is required for indexer mode, token is optional
+		useIndexerMode := *aptosIndexerRPC != ""
+
+		// Log which mode will be used
```
evan-gray, line 942: nice, this will be critical to debugging. do you think this is worth a (really short) feature flag in the heartbeat as well?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -203,6 +230,12 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			// the endpoint returns an array of events, ordered by sequence
 			// id (ASC)
+			// Check if the array is empty
+
+			if !events.Exists() || !events.IsArray() || len(events.Array()) == 0 {
+				logger.Warn("No new events found")
```
evan-gray: should this really be a warning? I think we expect the array to be empty most of the time while there hasn't been a new event. is that not the case? it seems like maybe more a debug to me.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -203,6 +230,12 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+				continue
```
evan-gray: if we continue here, the health endpoint won't ever be hit. is that the intention? that would cause the block height reported by the guardian to never increase unless there is a new message, if I understand correctly.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -267,6 +300,216 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+
+			// Extract the msg array from data.msg
+			messages := response.Get("data.msg")
+			if !messages.Exists() {
+				logger.Error("No data.msg field in indexer response")
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+			// Check if the array is empty
+			if len(messages.Array()) == 0 {
```
evan-gray, line 430: same comments as above for this block

panoel: Fixed

node/cmd/guardiand/node.go
panoel: Changed back

node/pkg/watchers/aptos/watcher.go
```diff
@@ -279,7 +522,120 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+			}
+			return nil, fmt.Errorf("indexer error: %s", errors.String())
+		}
+	}
+
+	return body, nil
+}
+
+// processTransactionVersion fetches a transaction by version and extracts WormholeMessage events
+func (e *Watcher) processTransactionVersion(
```
johnsaigle, line 586: I think this function's logged errors should include a unique field that we can do lookups on in order to debug. Maybe `hash` listed here? I'm not sure if there's a better field though:
https://aptos.dev/rest-api/operations/get_transaction_by_version

Otherwise, we just get the error strings but we can't tell where they came from.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -279,7 +522,120 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		logger.Error("Invalid JSON in transaction response", zap.String("data", string(txData)))
+		p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+		return fmt.Errorf("invalid JSON in transaction response")
+	}
+
+	// Parse the transaction data
+	txResult := gjson.ParseBytes(txData)
+
+	// Look for WormholeMessage in the events array
+	whEventType := fmt.Sprintf("%s::state::WormholeMessage", e.aptosAccount)
```
johnsaigle, line 610: nit: This could be defined when the watcher is configured instead of inside of this function

node/pkg/watchers/aptos/watcher.go
```diff
@@ -279,7 +522,120 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	}
+	defer res.Body.Close()
+	body, err := common.SafeRead(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for GraphQL errors in the response
+	// The indexer may return errors with a 200 status code
+	if gjson.Valid(string(body)) {
```
johnsaigle, line 568: We always expect valid JSON from this query, right? So I think if this check fails, the function should return `nil, err`. Right now it will fallthrough to `body, nil`

node/pkg/watchers/aptos/watcher.go
```diff
@@ -279,7 +522,120 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		response := gjson.ParseBytes(body)
+		errors := response.Get("errors")
+		if errors.Exists() && errors.IsArray() && len(errors.Array()) > 0 {
```
johnsaigle, line 571: Similarly the bounds check on errors is worth a loud comment as it will cause a panic if removed

node/pkg/watchers/aptos/watcher.go
```diff
@@ -267,6 +300,216 @@ func (e *Watcher) Run(ctx context.Context) error {
...
+
+			blockHeight := pHealth.Get("block_height")
+
+			if blockHeight.Uint() > math.MaxInt64 {
+				logger.Error("Block height not a valid uint64: ", zap.Uint64("blockHeight", blockHeight.Uint()))
+				p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDAptos, 1)
+				continue
+			}
+
+			if blockHeight.Exists() {
```
johnsaigle, line 499: nit: It seems like this check should happen before the `Uint()` check above. It also seems like if the block height doesn't exist for whatever reason then we would never see a health check. Probably if this field does not exist or is 0 after `Uint()`, an error should be logged.

PR #4517, opened 2025-10-03
docs: Notary operating instructions

johnsaigle:
Adds operating instructions for the Notary for use by the Guardians.
It follows the same style as `docs/governor.md`.

Leaving in draft mode until #4474 is merged.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -126,6 +126,7 @@
 
 /docs @evan-gray @SEJeff
 /docs/transfer-verifier.md @djb15 @johnsaigle @mdulin2 @pleasew8t
+/docs/notary.md @djb15 @johnsaigle @mdulin2 @pleasew8t
 
 ## Audit Reports
 


.github/CODEOWNERS
```diff
@@ -116,6 +116,7 @@
 ## Documentation
 
 /docs @evan-gray @SEJeff
+/docs/notary.md @evan-gray @djb15 @johnsaigle @mdulin2 @pleasew8t
```
djb15: Defer to @evan-gray on whether he wants to be a codeowner of this doc

docs/notary.md
```diff
@@ -0,0 +1,145 @@
...
+
+The Notary feature is disabled by default. Guardians can enable it by passing the following flag to the `guardiand` command when starting it up:
+
+```bash
+--notaryEnabled=true
+```
+
+When enabled, the Notary will evaluate all incoming message publications and apply the appropriate verdict based on the message status and content.
+
+_NOTE: Currently, the Notary has no effect on message publication processing unless Transfer Verifier is also enabled._
```
djb15, line 15: Link to TV whitepaper?

.github/CODEOWNERS
johnsaigle: Makes sense, I was just extending the current rule for `docs/`.

docs/notary.md
mdulin2: I have always seen the transfer verifier as the *evaluator* and the notary as the *enforcer*. It may be worth adding this distinction into this doc.

docs/notary.md
```diff
@@ -0,0 +1,145 @@
...
+
+To move a delayed VAA to the blackholed list (permanently blocking it), Guardians can run the `notary-blackhole-delayed-message` admin command:
+
+```bash
+guardiand admin notary-blackhole-delayed-message "chain_id/emitter_address/sequence_number" --socket /path/to/admin.sock
+```
+
+This command moves the specified VAA from the Notary's delayed list to the blackholed list, preventing it from ever being published.
+
+**Warning:** *Blackholing a VAA should only be used when there is a confirmed hack that directly affects the security of the Wormhole network. This action permanently blocks the message from being processed.*
```
mdulin2, line 83: Possible to remove it from a below command? Doesn't feel 100% permanent as a result.

PR #4521, opened 2025-10-07
notary: Add prometheus analytics for Notary and EVM Transfer Verifier

johnsaigle:


node/pkg/watchers/evm/msg_verifier.go
```diff
@@ -49,9 +65,18 @@ func verify(
 		newState = state(ctx, localMsg, txHash, receipt, verifier)
 	}
 
+	// Track non-default verification states for analytics
+	if newState != common.NotVerified && newState != common.Valid {
```
djb15: Should you also be skipping `NotApplicable` here?

johnsaigle: Yes, thanks!

djb15 (approved):

node/pkg/notary/notary.go
```diff
@@ -80,6 +80,9 @@ const (
 	// The value should be long enough to allow for manual review and classification
 	// by the Guardians.
 	DelayFor = time.Hour * 24 * 4
+
+	// The ticker interval for the notary's periodic metrics update.
+	metricsUpdateInterval = time.Minute * 15
```
panoel: This seems like a long time?  How often are messages examined to see if they are blackholed or delayed?  I'm not sure what the user of the metrics wants, but I think the metrics are sent to grafana like every 15 - 30 seconds?

node/pkg/watchers/evm/msg_verifier.go
```diff
@@ -8,6 +8,22 @@ import (
...
 	eth_common "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	msgVerificationStates = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "wormhole_msg_verification_states_total",
```
panoel: Maybe 'unusual' can be in the name?  This is what is displayed on grafana.  Otherwise, it could be misinterpreted as any state.

panoel (dismissed):

node/pkg/notary/notary.go
johnsaigle: My strategy here was to only send metrics for unusual cases. We expect the Notary to flag messages essentially only the case where there's a major token bridge problem, so hopefully the bad messages are flagged almost never. For that reason, 15 minutes seemed fine.

What do you think about that approach? Should we also log in the normal case? That'll cause a lot more updates to occur and in that scenario I agree that this delay could be shortened by a lot.

node/pkg/watchers/evm/msg_verifier.go
johnsaigle: Thanks, agreed.

node/pkg/notary/notary.go
panoel: I believe these metrics are scraped and not pushed.  So, the errors would show up more timely with a shorter interval.

panoel (dismissed):

node/pkg/notary/notary.go
johnsaigle: Thanks, I updated this to be the interval that you suggested.

node/pkg/watchers/evm/msg_verifier.go
```diff
@@ -49,9 +65,18 @@ func verify(
 		newState = state(ctx, localMsg, txHash, receipt, verifier)
 	}
 
+	// Track unusual states for analytics
+	if newState != common.NotVerified && newState != common.Valid && newState != common.NotApplicable {
```
pleasew8t: Why not explicitly check if `Anomalous || Rejected || CouldNotVerify`?

node/pkg/watchers/evm/msg_verifier.go
```diff
@@ -8,6 +8,22 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	msgVerificationStates = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "wormhole_unusual_msg_verification_states_total",
+			Help: "Total number of message verification state changes to unusual values",
```
pleasew8t: Consider "Total number of messages assessed as unusual by transfer verifier"

node/pkg/notary/metrics.go
```diff
@@ -0,0 +1,39 @@
+package notary
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// Prometheus metrics for notary events
+var (
+	notaryMessagesReleased = promauto.NewCounter(
```
pleasew8t: I like the metric suffix, like `xxxGauge` or `xxxCounter`  maybe?

pleasew8t: `notaryReleasedMessageCounter`

node/pkg/notary/notary.go
```diff
@@ -82,6 +82,9 @@ const (
 	DefaultDelay = time.Hour * 24 * 4
 	MaxDelayDays = 30
 	MaxDelay     = time.Hour * 24 * MaxDelayDays
+
+	// The ticker interval for the notary's periodic metrics update.
+	metricsUpdateInterval = time.Second * 30
```
pleasew8t, line 88: I'm sure this was arbitrary, but I rate this could probably be in the order of minutes? Like 5 or 10?

node/pkg/notary/notary.go
pleasew8t: Oh no. Does this mean the notary was short-circuiting in the testnet deployment? I believe this would also short-circuit in devnet.

node/pkg/processor/processor.go
```diff
@@ -322,6 +322,10 @@ func (p *Processor) Run(ctx context.Context) error {
 				switch verdict {
 				case guardianNotary.Blackhole, guardianNotary.Delay:
 					p.logger.Error("notary evaluated message as threatening", k.ZapFields(zap.String("verdict", verdict.String()))...)
+
+					// Track messages that receive non-Approve verdicts
+					guardianNotary.NotaryTokenTransferNonApprove.WithLabelValues(verdict.String()).Inc()
```
pleasew8t: I don't know if this is done similarly elsewhere, but does it make sense for something outside of the package to trigger prometheus metrics updates?

node/pkg/notary/notary.go
johnsaigle: I had this at 30m originally and Paul asked me to reduce it to this range.
https://github.com/wormhole-foundation/wormhole/pull/4521#discussion_r2440589373

node/pkg/notary/notary.go
johnsaigle: Yep...

johnsaigle: We can refactor to use SupportedChains here instead? https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/txverifier/utils.go#L195-L205

johnsaigle: https://github.com/wormhole-foundation/wormhole/pull/4577

node/pkg/processor/processor.go
johnsaigle: Thanks, I moved it inside of the Notary

node/pkg/notary/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Prometheus metrics for notary events.
+// These are only registered when the Notary is actually enabled and Run() is called.
+var (
+	notaryReleasedMessagesCounter prometheus.Counter
+	notaryDelayedMessagesGauge    prometheus.Gauge
+	notaryBlackholedMessagesGauge prometheus.Gauge
+	notaryErrors                  *prometheus.CounterVec
+	NotaryTokenTransferNonApprove *prometheus.CounterVec
```
pleasew8t: Is there a reason for exporting `NotaryTokenTransferNonApprove`?

node/pkg/notary/notary.go
```diff
@@ -148,9 +156,52 @@ func (n *Notary) Run() error {
...
+	// Safety check: ensure metrics are initialized (should always be true after Run() is called)
+	if notaryDelayedMessagesGauge == nil || notaryBlackholedMessagesGauge == nil {
+		n.logger.Warn("Notary metrics not initialized, skipping gauges update")
+		return
+	}
+
+	n.mutex.RLock()
+	defer n.mutex.RUnlock()
+
+	notaryDelayedMessagesGauge.Set(float64(n.delayed.Len()))
```
pleasew8t, line 201: Not that it _should_ significantly increase over time, but is `delayed` and `blackholed` ever pruned for in-memory use? I noticed that `loadFromDB` always loads everything.

johnsaigle: `n.delayed` should be self-pruning based on the Pop operations for the queue. blackholed would never be pruned because that's how the Notary is able to tell the processor that a message shouldn't be processed.

djb15 (approved):

PR #4527, opened 2025-10-10
svm/wormhole-core-shims: support non-solana SVM deployments

kcsongor:
This PR modifies the shim programs (mostly through the definitions crate) to allow deploying to chains other than Solana. It does this by making the program id constants and chain id configurable via environment variables. This mode is made available by passing the`from-env` feature.

The change aims to maintain bakcwards compatibility.

### DIFF
--- a/svm/wormhole-core-shims/crates/definitions/src/lib.rs
+++ b/svm/wormhole-core-shims/crates/definitions/src/lib.rs
@@ -191,6 +191,9 @@ pub const fn make_anchor_discriminator(input: &[u8]) -> [u8; 8] {
 }
 
 /// Trait to encode and decode the SVM finality of a message.
+/// The byte encoding corresponds to the "commitment level" field in the VAA.
+/// See https://wormhole.com/docs/products/reference/consistency-levels/
+///
 /// Different SVM runtimes may have different finality options, so we just
 /// provide this conversion trait between u8s.
 /// We also define the "standard" finality options ([`standard_finality`]) which are
@@ -198,7 +201,7 @@ pub const fn make_anchor_discriminator(input: &[u8]) -> [u8; 8] {
 /// If you use this crate with a chain that support different finality modes,
 /// either just use `u8`, or define a a custom enum and an impl of this trait.
 pub trait EncodeFinality: Sized + Copy {
-    /// Encode SVM finality into a byte.
+/// Encode SVM finality into a byte.
     fn encode(&self) -> u8;
 
     /// Decode SVM finality from a byte.


svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -162,9 +204,155 @@ pub trait EncodeFinality: Sized + Copy {
...
+#[allow(dead_code)]
+/// A test to make sure the expected IDs are available.
+fn available_ids() {
+    let _ = crate::solana::mainnet::CORE_BRIDGE_PROGRAM_ID;
+    let _ = crate::solana::devnet::CORE_BRIDGE_PROGRAM_ID;
+    let _ = crate::solana::mainnet::CORE_BRIDGE_PROGRAM_ID;
+    // defaults to mainnet (for backwards compatibility)
+    let _ = crate::solana::CORE_BRIDGE_PROGRAM_ID;
+    let _ = crate::CORE_BRIDGE_PROGRAM_ID;
+}
```
bengtlofgren, line 334: Do you want to keep this?

svm/wormhole-core-shims/crates/definitions/src/env.rs
```diff
@@ -0,0 +1,34 @@
+pub const fn validate_bs58(s: &str) -> &str {
```
bengtlofgren: Maybe change the name of this function so it is clear that it only validates bs58s that are for pks?

Also, we're not checking that it is valid on the curve? But maybe that's intentional, just won't it panic on `const_crypto::bs58::decode_pubkey`? :S

svm/wormhole-core-shims/crates/definitions/src/lib.rs
kcsongor: yeah I wanted to keep this to test (+ illustrate) what the module hierarchy ends up looking like, given the conditional stuff and the code generation

svm/wormhole-core-shims/crates/definitions/src/env.rs
kcsongor: it doesn't have to be on the curve, PDAs are off the curve!

svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -162,9 +204,155 @@ pub trait EncodeFinality: Sized + Copy {
...
+            solana_program::pubkey::Pubkey::new_from_array(VERIFY_VAA_SHIM_PROGRAM_ID_ARRAY);
+    };
+}
+
+#[allow(dead_code)]
+/// A test to make sure the expected IDs are available.
+fn available_ids() {
+    let _ = crate::solana::mainnet::CORE_BRIDGE_PROGRAM_ID;
+    let _ = crate::solana::devnet::CORE_BRIDGE_PROGRAM_ID;
+    let _ = crate::solana::mainnet::CORE_BRIDGE_PROGRAM_ID;
```
johnsaigle:
```suggestion
    let _ = crate::solana::localnet::CORE_BRIDGE_PROGRAM_ID;
```
Mainnet is tested twice, so I think this line should be localnet.

svm/wormhole-core-shims/crates/definitions/README.md
```diff
@@ -4,31 +4,36 @@ Definitions relating to Wormhole SVM programs. These definitions include finding
...
+
+For Solana, we have a hardcoded set of addresses for `mainnet`, `devnet` and a
+local testing environment. If the `solana` feature is specified, it will default
+to the mainnet addresses. Otherwise, the `testnet` flag provides the addresses
+for Solana Devnet, and the `localnet` flag provides the addresses for the local
+testing environment.
+
+Alternatively, when using on another SVM chain (or on Solana, but against an
+independent Wormhole deployment) just specify the `from-env` feature flag (and
+don't specify Solana, even if the chain is Solana). In this case, the following
```
johnsaigle, line 20: Could you clarify why `from-env` and `solana` shouldn't be specified at the same time? Is it worth enforcing this as compile time to prevent mistakes?

svm/wormhole-core-shims/crates/definitions/src/env.rs
```diff
@@ -0,0 +1,34 @@
...
+    let b = s.as_bytes();
+    if b.len() != 44 {
+        panic!("invalid pubkey length (need to be 44 characters)");
+    }
```
nonergodic: Addresses can be 43 or 44 chars long.
[Random 43 character long address](https://explorer.solana.com/address/astraEJ2fEj8Xmy6KLG7B3VfbKfsHXhHrNdCQx7iGJK) (fished from a recent tx from the explorer)

svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -162,9 +204,155 @@ pub trait EncodeFinality: Sized + Copy {
...
+    fn encode(&self) -> u8 {
+        *self
+    }
+
+    fn decode(data: u8) -> Option<Self> {
+        Some(data)
+    }
+}
+
+pub mod std_finality {
```
nonergodic: Not a fan of the `std` prefix in the Rust context, given Rust's `std` which I found especially confusing in the Cargo.toml where any additional context is missing (though I admit that I struggle to think of anything decent that's also reasonably short). Edit: Maybe just spelling out "standard" would already be enough.

Also, the term consistency level should appear somewhere in the comments (or perhaps we want [a link to the docs](https://wormhole.com/docs/products/reference/consistency-levels/) outright) - if only to augment commitment level.

svm/wormhole-core-shims/crates/definitions/README.md
nonergodic: I'm seconding this. `solana` and `from-env` don't exactly scream "use one and only one". Perhaps adding a `compile_error!` - see comment below.

svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -3,16 +3,52 @@
...
+
+// We define the constants (chain id + addresses) here.
+// - For 'solana', we just re-export the definitions in the solana module.
+// - For 'from-env', we pick these up from environment variables and parse (+
+// validate) them into the right types via const functions.
+//
+// Consumers of this crate should mark 'from-env' as the default.
+mod defs {
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "solana")] {
```
nonergodic, line 21:
```suggestion
    cfg_if::cfg_if! {
        if #[cfg(all(feature = "solana", feature = "from-env"))] {
            compile_error!("Features `solana` and `from-env` cannot both be enabled.");
        } else if #[cfg(all(not(feature = "solana"), not(feature = "from-env")))] {
            compile_error!("You must enable exactly one of `solana` or `from-env`.");
        } else if #[cfg(feature = "solana")] {
```
See enforcement comment above.

svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -3,16 +3,52 @@
...
+            };
+
+            pub const CORE_BRIDGE_PROGRAM_ID_ARRAY: [u8; 32] =
+            // we use the same variable name as the core contracts
+                env_pubkey!("BRIDGE_ADDRESS");
+
+            pub const POST_MESSAGE_SHIM_PROGRAM_ID_ARRAY: [u8; 32] =
+                env_pubkey!("POST_MESSAGE_SHIM_PROGRAM_ID");
+
+            pub const VERIFY_VAA_SHIM_PROGRAM_ID_ARRAY: [u8;32] =
```
nonergodic:
```suggestion
            pub const VERIFY_VAA_SHIM_PROGRAM_ID_ARRAY: [u8; 32] =
```
*spasms in autistic seizure*

svm/wormhole-core-shims/crates/definitions/src/env.rs
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+
+    s
+}
+
+#[macro_export]
+macro_rules! env_pubkey {
+    ($name: literal) => {
+        const_crypto::bs58::decode_pubkey(crate::validate_bs58(env!($name)))
```
nonergodic: Given the length ambiguity, this is now not enough to ensure that the base58 string corresponds to a 32 byte value.
ChatGPT example:
base58 `4frqk3R2r2FJYF91BkKfU7CCVbAq5RczPpTGiP2eQqq` (43 chars) [decodes](https://appdevtools.com/base58-encoder-decoder) to 0xf0ae48b543c3f88b9efe866b614a9f76b0760aaceed9f0fc12a9b72c6bb7a4 which is only 31 bytes.

svm/wormhole-core-shims/crates/definitions/src/lib.rs
kcsongor: fixed

svm/wormhole-core-shims/crates/definitions/src/env.rs
kcsongor: Good catch. I ended up just removing this function altogether, as the const_crypto library does internal validation already, so it's redundant. I checked, and it accepts the 43 long address just fine.

svm/wormhole-core-shims/crates/definitions/README.md
kcsongor: done

svm/wormhole-core-shims/crates/definitions/src/lib.rs
kcsongor: good catch ,fixed

svm/wormhole-core-shims/crates/definitions/src/env.rs
kcsongor: resolved by removing the check entirely

svm/wormhole-core-shims/crates/definitions/src/lib.rs
kcsongor: enforced that they're mutually exclusive. it's OK to not set either

svm/wormhole-core-shims/crates/definitions/src/lib.rs
kcsongor: good idea, added a doc string for this

svm/wormhole-core-shims/crates/definitions/src/env.rs
kcsongor: removed this check

svm/wormhole-core-shims/crates/definitions/src/lib.rs
```diff
@@ -154,17 +191,173 @@ pub const fn make_anchor_discriminator(input: &[u8]) -> [u8; 8] {
...
+///
+/// Different SVM runtimes may have different finality options, so we just
+/// provide this conversion trait between u8s.
+/// We also define the "standard" finality options ([`standard_finality`]) which are
+/// the available ones on Solana.
+/// If you use this crate with a chain that support different finality modes,
+/// either just use `u8`, or define a a custom enum and an impl of this trait.
 pub trait EncodeFinality: Sized + Copy {
-    /// Encode SVM finality into a byte.
+/// Encode SVM finality into a byte.
```
nonergodic, line 204: I feel like you did this, just to spite me.

johnsaigle (approved):

svm/wormhole-core-shims/crates/definitions/src/lib.rs
johnsaigle: Do we not have cargo fmt enabled for this directory? 🤔

PR #4528, opened 2025-10-10
feat(evm): include contract verificiation in evm contract deployment

aapclark:
update readme and verification instructions

add template envs for testnet and mainnet deployments


ethereum/env/.env.template.mainnet
```diff
@@ -0,0 +1,45 @@
...
+# EVM chain ID
+INIT_EVM_CHAIN_ID=
+
+# =============================================================================
+# Bridge Contract Migrations -- Only required when WTT/TokenBridge is deployed
+# =============================================================================
+
+# These values are unlikely to change unless announced
+BRIDGE_INIT_GOV_CHAIN_ID=0x1
+BRIDGE_INIT_GOV_CONTRACT=0x0000000000000000000000000000000000000000000000000000000000000004
```
scnale: Maybe we could reuse the definitions?
```suggestion
BRIDGE_INIT_GOV_CHAIN_ID=$INIT_GOV_CHAIN_ID
BRIDGE_INIT_GOV_CONTRACT=$INIT_GOV_CONTRACT
```

And similar things could be done for other parameters like the Wormhole chain id too.

djb15 (dismissed):

ethereum/env/.env.template.mainnet
aapclark: Good idea. I'll make those changes.

PR #4530, opened 2025-10-13
SVM: Revamped Core and Bridge initialization scripts.

hernandp:


solana/scripts/initialize-core.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+  
+  for (const guardian of guardianSet) {
+    if (guardian.startsWith("0x")) {
+      console.error("GUARDIAN_SET must contain only non-0x prefixed hex addresses");
+      displayHelp();
+      process.exit(1);
+    }
+  }
+  const guardianSetBuffer = guardianSet.map((guardian) => Buffer.from(guardian, "hex"));
```
kcsongor: we should strengthen this test. currently if the guardian set string doesn't start with `0x` but is otherwise not a valid hex string, it will just be silently accepted. Or if it's not 20 bytes long, it will be accepted too. Maybe we remove the for loop and just do

```suggestion
  const guardianSetBuffer = guardianSet.map((guardian) => Buffer.from(guardian, "hex"));
  
  if (guardianSetBuffer.some((address) => address.length !== 32)) {
    ... error
   }
 
```

hernandp: i agree with your proposal.

PR #4537, opened 2025-10-17
ci: Enable tests and lints for svm/ directory; disable them for terra/

johnsaigle:
The `svm/` directory was added a while ago and contains load-bearing code given that the Shims are now widely used. It should follow the same code quality standards as other Rust projects.

Terra has also been deprecated: the Guardians no longer run Terra nodes. So, we don't need to continue testing and linting in every CI run.

### DIFF
--- a/svm/wormhole-core-shims/crates/definitions/src/lib.rs
+++ b/svm/wormhole-core-shims/crates/definitions/src/lib.rs
@@ -284,7 +284,7 @@ macro_rules! derive_consts {
 
         pub const CORE_BRIDGE_FEE_COLLECTOR_PDA: ([u8; 32], u8) =
             const_crypto::ed25519::derive_program_address(
-                &[crate::FEE_COLLECTOR_SEED],
+                &[$crate::FEE_COLLECTOR_SEED],
                 &CORE_BRIDGE_PROGRAM_ID_ARRAY,
             );
 
@@ -295,7 +295,7 @@ macro_rules! derive_consts {
 
         pub const CORE_BRIDGE_CONFIG_PDA: ([u8; 32], u8) =
             const_crypto::ed25519::derive_program_address(
-                &[crate::CORE_BRIDGE_CONFIG_SEED],
+                &[$crate::CORE_BRIDGE_CONFIG_SEED],
                 &CORE_BRIDGE_PROGRAM_ID_ARRAY,
             );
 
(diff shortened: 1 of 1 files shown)


johnsaigle:
I'll leave this in draft until #4527 is merged as it changes the feature flags and may influence the build steps.

.github/workflows/build.yml
```diff
@@ -216,23 +216,6 @@ jobs:
...
-
-  terra:
-    runs-on: ubuntu-24.04
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: "22"
-      - run: cd terra && make test
-  terra-2:
```
evan-gray, line 228: I'm not terra-2 should be removed. this may be a bit of a misnomer, but may actually test wormchain functionality as it is the same deployment for all cosmwasm chains.

.github/workflows/build.yml
```diff
@@ -344,27 +327,58 @@ jobs:
...
         manifest:
+          # CosmWasm
           - path: cosmwasm/Cargo.toml
             args: "--workspace --locked --exclude 'shutdown-*'"
+            test_args: ""
           - path: cosmwasm/Cargo.toml
             args: "-p 'shutdown-*' --no-default-features --locked"
-          - path: terra/Cargo.toml
-            args: "--workspace --locked"
+            test_args: ""
```
evan-gray, line 353: is this a functional change or we're just being explicit about test args?

.github/workflows/build.yml
johnsaigle: Thanks I'll revert this.

.github/workflows/build.yml
johnsaigle: Being explicit so that we can pass some optional, custom parameters to shims. Just a way to generalize across these different crates. My intention is that tests for other crates shouldn't change.

kcsongor:
> @kcsongor Can you review the latest lint failure? I'm not sure if this is a bug or one that we should allow.
> 
> 
> 
> ```
> 
> error: `crate` references the macro call's crate
> 
>    --> crates/definitions/src/lib.rs:287:19
> 
>     |
> 
> 287 |                 &[crate::FEE_COLLECTOR_SEED],
> 
>     |                   ^^^^^ help: to reference the macro definition's crate, use: `$crate`
> 
>     |
> 
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#crate_in_macro_def
> 
>     = note: `-D clippy::crate-in-macro-def` implied by `-D warnings`
> 
>     = help: to override `-D warnings` add `#[allow(clippy::crate_in_macro_def)]`
> 
> ```


in our case this doesn't matter because we're invoking the macro within the crate itself. but might be nice to fix, as the "proper" way doesn't hurt

kcsongor (approved):

PR #4542, opened 2025-10-20
node: add last observed vaa to hb

panoel:
Adding last observed VAA to heartbeat message for each supported chain.  The heartbeat will contain all chains supported by this guardian.  In the per-chain section there is a block height that tells us the guardian's node is functioning.  This PR adds a last observed VAA to indicate that this guardian is, also, observing VAAs.  This is not when this guardian saw another guardian's observation, but when it produced its own observation.


node/pkg/p2p/registry.go
```diff
@@ -18,12 +18,17 @@ type registry struct {
 	// Per-chain error counters
 	errorCounters  map[vaa.ChainID]uint64
 	errorCounterMu sync.Mutex
+
+	// Per-chain last VAA timestamp (Unix nanoseconds)
+	lastVaaTimestamps   map[vaa.ChainID]int64
+	lastVaaTimestampsMu sync.Mutex
```
johnsaigle:
```suggestion
	lastVaaTimestampsMu sync.RWMutex
```
You could consider using the RWMutex here instead. That way `GetLastVaaTimestamp` only needs to acquire a ReadLock, and `SetLastVaaTimestamp` could acquire a write lock. This should allow for concurrent reads of the LastVAATimestamp.

Not sure if this would have a meaningful impact on performance, but I do think it makes sense anyway.

node/pkg/p2p/registry.go
```diff
@@ -51,3 +56,18 @@ func (r *registry) GetErrorCount(chain vaa.ChainID) uint64 {
 	defer r.errorCounterMu.Unlock()
 	return r.errorCounters[chain]
 }
+
+// SetLastVaaTimestamp sets the timestamp (Unix nanoseconds) of the last VAA observed for a chain.
+func (r *registry) SetLastVaaTimestamp(chain vaa.ChainID, timestamp int64) {
+	r.lastVaaTimestampsMu.Lock()
+	defer r.lastVaaTimestampsMu.Unlock()
+	r.lastVaaTimestamps[chain] = timestamp
```
johnsaigle: Any thoughts about validating timestamp here? Maybe `timestamp` should be strictly greater than the value that's already stored?

node/pkg/p2p/registry.go
```diff
@@ -18,12 +18,17 @@ type registry struct {
 	// Per-chain error counters
 	errorCounters  map[vaa.ChainID]uint64
 	errorCounterMu sync.Mutex
+
+	// Per-chain last VAA timestamp (Unix nanoseconds)
+	lastVaaTimestamps   map[vaa.ChainID]int64
```
johnsaigle:
```suggestion
	lastVaaTimestamps   map[vaa.ChainID]timestamp
```
Can this type work as `timestamp`? Since that type is used in `setLastVaaTimestamp`'s signature, maybe it could be used everywhere?

node/pkg/p2p/registry_test.go
```diff
@@ -53,3 +53,52 @@ func TestGetErrorCount(t *testing.T) {
...
+
+func TestUpdateLastVaaTimestamp(t *testing.T) {
+	registry := NewRegistry()
+
+	// Set initial timestamp
+	timestamp1 := int64(1000000000000000000)
+	registry.SetLastVaaTimestamp(vaa.ChainIDEthereum, timestamp1)
+	assert.Equal(t, timestamp1, registry.GetLastVaaTimestamp(vaa.ChainIDEthereum))
+
+	// Update to a newer timestamp
```
johnsaigle, line 100: Could be good to test and prevent overwriting a new timestamp with an old timestamp, if the error handling is added that I described elsewhere.

node/pkg/p2p/registry.go
djb15: Agreed, we don't want reobservations to overwrite this value with an older block timestamp

node/pkg/p2p/registry.go
panoel: Very sensible

node/pkg/p2p/registry.go
panoel: Good catch

node/pkg/p2p/registry.go
panoel: int64 is used everywhere in this codebase to represent timestamps.  I'm not familiar with that `timestamp` type?  I hesitate to do something inconsistent with the rest of the codebase.

node/pkg/p2p/registry_test.go
panoel: Added tests

node/pkg/p2p/registry.go
johnsaigle: No worries, I think I misinterpreted the code anyway. Sorry about that.

johnsaigle (dismissed):

djb15 (dismissed):

johnsaigle:
@evan-gray Agreed with the above and I think those checks and that behaviours are present in the latest commit. (And so I approved the PR after my initial comment.)

>  the reobservation path might work but not direct observations which I feel is what monitoring would want to zero in on. 
If that's the goal, I agree with checking the IsReobservation flag on a message and dropping if true. We could consider renaming the metric to "most recent direct observation" to remove any ambiguity.

proto/gossip/v1/gossip.proto
```diff
@@ -50,9 +50,11 @@ message Heartbeat {
     // Connection error count
     uint64 error_count = 4;
     // Safe block height of the node, if supported.
-    int64 safe_height = 5;    
+    int64 safe_height = 5;
     // Finalized block height of the node, if supported.
     int64 finalized_height = 6;
+    // UNIX timestamp of the last time a VAA was observed for this chain.
+    int64 last_vaa_observed_at = 7;
```
evan-gray: if we're being pedantic, which we probably should, this actually represents the last time the guardian signed an observation for that network, not observed a VAA, since the valid VAA is not formed until quorum is reached. `last_observation_signed_at` would be a more accurate name.

evan-gray (changes requested):
just one nit please

proto/gossip/v1/gossip.proto
panoel: Changed

evan-gray (approved):

johnsaigle (approved):

PR #4545, opened 2025-10-21
node/pkg/watchers/evm: fix ccl bug

panoel:


node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -177,7 +177,7 @@ func (w *Watcher) cclHandleMessage(parentCtx context.Context, pe *pendingMessage
 			zap.Uint8("consistencyLevel", req.consistencyLevel),
 			zap.Uint16("additionalBlocks", req.additionalBlocks),
 		)
-		pe.message.ConsistencyLevel = req.consistencyLevel
```
mdulin2, line 180: On line 171, the `pe.message.ConsistencyLevel` is being set to finalized if the other `consistencyLevel` is not one of the main three. Can we edit this one as well? This could lead to the same bug otherwise I think.

mdulin2:
I couldn't find a good way to add a comment in the PR for this. Could we update this log message to be the effectiveCL and posted CL? 
https://github.com/wormhole-foundation/wormhole/blob/07d885b130811af8bb1ef15655f3239590d5ef45/node/pkg/watchers/evm/watcher.go#L867

mdulin2 (changes requested):
Can you check that all of the logging within the EVM watcher makes sense as well? When debugging TXs in the logs, we definitely want the appropriate data to show up in the logs given the context.

mdulin2 (approved):

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -153,7 +153,9 @@ func (w *Watcher) cclHandleMessage(parentCtx context.Context, pe *pendingMessage
 	r, err := w.cclReadAndParseConfig(parentCtx, emitterAddr)
 	if err != nil {
 		w.cclLogger.Error("failed to look up config for custom handling, treating as finalized", zap.String("msgId", pe.message.MessageIDString()), zap.Error(err))
-		pe.message.ConsistencyLevel = vaa.ConsistencyLevelFinalized
+		// If one guardian has an error reading the config, but others do not, they will produce different VAA hashes.
```
evan-gray, line 156: the consistency level is still modified on line 141 above. `pe.message.ConsistencyLevel` should never be overwritten. the message should have the same consistency level represented regardless of whether the feature is on or off

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -168,7 +170,7 @@ func (w *Watcher) cclHandleMessage(parentCtx context.Context, pe *pendingMessage
 				zap.Uint8("consistencyLevel", req.consistencyLevel),
 				zap.Uint16("additionalBlocks", req.additionalBlocks),
 			)
-			pe.message.ConsistencyLevel = vaa.ConsistencyLevelFinalized
+			pe.effectiveCL = vaa.ConsistencyLevelFinalized
```
evan-gray, line 173: ditto on line 165 above

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -177,10 +179,11 @@ func (w *Watcher) cclHandleMessage(parentCtx context.Context, pe *pendingMessage
 			zap.Uint8("consistencyLevel", req.consistencyLevel),
 			zap.Uint16("additionalBlocks", req.additionalBlocks),
 		)
-		pe.message.ConsistencyLevel = req.consistencyLevel
+		pe.effectiveCL = req.consistencyLevel
```
evan-gray, line 182: ditto on line 187 below

evan-gray (changes requested):

evan-gray (approved):

PR #4546, opened 2025-10-22
feat: squashed stacks implementation

jannik-stacks:
_This is the watcher and test code required to add the Stacks chain to Wormhole._

## Node Watcher Implementation

Added a polling-based watcher in `node/pkg/watchers/stacks/` that monitors the Stacks blockchain for Wormhole message publication events.

**Overview**

- Uses Bitcoin blocks as the anchor point for finality (based on node "stable" definition)
- Polls Stacks node RPC every n seconds for new blocks (defaults to `2s`)
- Reobserve implemented
- Uses only v2/v3 Stacks node RPC endpoints (however this needs a chain post-Nakamoto epoch)

## Smart Contracts

Contract files copied from other repos for integration testing (as well as some dep contracts).

## DevNet/Hacknet Infrastructure

Added Kubernetes configurations for local Stacks devnet in `devnet/` — Unfortunately this requires a lot of files for trivial testing.

- Bitcoin node
- Stacks node & Signer (single)
- Nakamoto Transaction broadcaster service (to keep chain alive it needs txs)
- Stacker services for PoX consensus

## Testing

Basic integration tests in `stacks/test/integration.test.ts` — launched in Tilt, but take a while (6mins) to run.

### DIFF
--- a/node/pkg/watchers/stacks/watcher.go
+++ b/node/pkg/watchers/stacks/watcher.go
@@ -386,7 +386,7 @@ func (w *Watcher) processStacksTransaction(_ context.Context, tx *StacksV3Tenure
 			zap.String("type", fmt.Sprintf("%T", clarityValue)))
 
 		// Process the core event
-		if err := w.processCoreEvent(clarityValue, replay.Timestamp); err == nil {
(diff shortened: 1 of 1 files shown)


jannik-stacks:
Feel free to change base branch, I don't have write access on this repo I believe.

PR #4553, opened 2025-10-28
ci: Add token metadata parser

johnsaigle:
Adds the new Token Metadata parser library to CI to ensure it gets linted and tested.
See also https://github.com/wormhole-foundation/wormhole/pull/4537

To get this to work in a bit more of a convenient way, the `rust-toolchain` file in `solana` was renamed so it has the `.toml` extension. Other changes were necessary to comply with `cargo fmt`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -361,7 +361,7 @@ jobs:
           - path: svm/wormhole-core-shims/anchor/Cargo.toml
             args: "--workspace"
             test_args: "--lib"  # Skip doctests; anchor interfaces lack solana_program dependency
-          # Token Mint parser for SVM token bridge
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -362,6 +362,10 @@ jobs:
           - path: svm/wormhole-core-shims/anchor/Cargo.toml
             args: "--workspace"
             test_args: "--lib"  # Skip doctests; anchor interfaces lack solana_program dependency
+          # Token Mint parser for SVM token bridge
```
djb15:
```suggestion
          # Token Metadata parser for SVM token bridge
```

djb15 (dismissed):

PR #4554, opened 2025-10-28
Script to help WTT deployments review

fergarrui:
Quick script to help reviewing WTT new EVM deployments.
Pushing just in case it can be useful to someone else too.


scripts/verify_wtt_deployment/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "verifytokenbridgedeployment",
```
kcsongor: it's better to use the `@wormhole-foundation` namespace here, even if the package never gets published, to avoid name squatters messing with package resolution

fergarrui: good point, updated

PR #4559, opened 2025-10-30
svm/wormhole-core-shims: allow anchor interfaces to be used from any anchor version

kcsongor:
We relax the anchor version bounds of the `*-interface` crates so they can be used from any anchor program. This works, because the interface crates rely on a very small subset of anchor features that are hopefully stable enough to not need to be tied to version bounds. Knock on wood.

Additionally, the definitions crate now has fine-grained control over which addresses to expose. This is only interesting in the `from-env` case, which rely on environment variables to be set. For example, if a user of this crate wants only the core bridge address (and its PDAs), it should not need to define the verify-vaa-shim program id.
This is in preparation for using this crate as the ultimate source of truth for all program ids and constants. As such, we also use hand-rolled u16 const parser, as the previously used version doesn't work on older compiler versions required by some anchor versions out there (looking at you, 0.29).

This is also a breaking change of the definitions crate, so bumped the major version to 1.0.0. Breaking, because the default flags now only provide core bridge ids, as opposed to all the ids.

### DIFF
--- a/svm/wormhole-core-shims/crates/definitions/Cargo.toml
+++ b/svm/wormhole-core-shims/crates/definitions/Cargo.toml
@@ -16,6 +16,7 @@ default = ["core", "standard-finality"]
 borsh = ["dep:borsh"]
 
 core = []
+chain-id = []
 token-bridge = []
 post-message-shim = []
 verify-vaa-shim = []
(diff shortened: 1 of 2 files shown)


kcsongor:
> > the interface crates rely on a very small subset of anchor features that are hopefully stable enough to not need to be tied to version bounds.
> 
> Any chance you could document the required features in the README? The idea here would be to make it easier for us to check this crate against Anchor's changelogs going forward to figure out if we're gonna get into trouble.

not sure I can explicitly enumerate these, but I think we will know if something breaks because it will fail to compile

johnsaigle (approved):

PR #4561, opened 2025-10-31
Add convenience script to run clippy on Rust crates locally

johnsaigle:
- Adds scripts/clippy.sh to run clippy on most Rust crates. This should make it easier for devs to keep clippy's lints up-to-date, as currently this requires doing a complete push-and-CI cycle for each lint violation. This is aggravated by CI failing early as soon as there is a lint violation instead of printing a full report of all clippy issues.
- Adds clippy to the Makefile's lint target.
- Adds me to the codeowners for the new script.

(There are some important exceptions that were skipped in this initial run as
  they require other artifacts or make assumptions about which directory the code is
  executed from.)
  
  This should help with tasks like #4560 which will arise anytime clippy lints are modified in future Rust releases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,10 @@ generate: dirs
 lint:
 # Lints spelling and Go
 	bash scripts/lint.sh lint
-# Runs clippy for most Rust crates
+
+.PHONY: lint-rust
+lint-rust:
+# Runs clippy for most Rust crates.
 	bash scripts/clippy.sh
 
 .PHONY: node


scripts/clippy.sh
```diff
@@ -0,0 +1,85 @@
...
+cd "$(dirname "$0")/.." || exit
+
+# This should be the latest available Rust version for the system in order to match
+# our CI configuration.
+echo "Running Rust linting with active toolchain: $(rustup show active-toolchain)"
+
+# Ensure clippy component is installed for current active toolchain
+# This will use the toolchain defined in rust-toolchain* files via rustup
+# echo "Ensuring clippy component is installed for toolchain: $(rustup show active-toolchain)"
+# rustup component add clippy --toolchain "$(rustup show active-toolchain | awk '{print $1}')"
```
djb15: Is this supposed to be commented out?

johnsaigle: Thanks, yeah it is. I didn't want to mess with the developer's set-up ultimately.

johnsaigle: I deleted it in the latest commit

scripts/clippy.sh
```diff
@@ -0,0 +1,75 @@
...
+# Skip:
+# - 3rd party crates
+# - Compiled code
+# - Version control information
+# - Deprecated chains (e.g. Terra)
+#
+# TODO: Near, CosmWasm/Wormchain, NFT Bridge, and memmap2-rs should probably be linted as well
+# but requires a little more work.
+find . -type f -name Cargo.toml \
+   -not -path "./target/*" \
```
djb15, line 32: Can you add `/ethereum/node_modules/rustbn.js` to this list? Or generalise to all `node_modules`?

Makefile
```diff
@@ -33,7 +33,10 @@ generate: dirs
 
 .PHONY: lint
 lint:
+# Lints spelling and Go
 	bash scripts/lint.sh lint
+# Runs clippy for most Rust crates
+	bash scripts/clippy.sh
```
evan-gray, line 42: I feel like this lint was intended for the guardian. can this instead be something specific for rust folks?

johnsaigle: Sure, here are two approaches from my view:

1. Keep this change, but also add `lint-go` and `lint-rust` that separates these actions
2. Revert this change and add a separate `lint-rust`

Any preference?

evan-gray: 2 as it maintains the status quo for existing contributors

djb15 (approved):

PR #4564, opened 2025-11-02
node: Remove references to deprecated chain Fantom

johnsaigle:
Fantom is deprecated and there are not enough Guardians running Fantom nodes to produce VAAs. As a result, we can remove references to it from the node and documentation. https://wormhole-foundation.github.io/wormhole-dashboard/#/?endpoint=Mainnet


sdk/vaa/chainid_generated.go
```diff
@@ -26,8 +26,6 @@ func (c ChainID) String() string {
 		return "avalanche"
 	case ChainIDAlgorand:
 		return "algorand"
-	case ChainIDFantom:
```
mdulin2, line 29: Did you manually modify these or regenerate them?

mdulin2 (dismissed):

sdk/vaa/chainid_generated.go
johnsaigle: I don't remember, but I just checked-out this PR, ran `make go-generate` from `sdk/vaa/`, and there was no change.

sdk/rust/supported-chains/src/lib.rs
```diff
@@ -75,7 +74,6 @@ impl From<u16> for Chain {
             7 => Chain::Oasis,
             8 => Chain::Algorand,
             9 => Chain::Aurora,
-            10 => Chain::Fantom,
```
djb15, line 78: Just checking whether it makes sense to remove here or keep around considering the other obsolete chains from [proto/publicrpc/v1/publicrpc.proto](https://github.com/wormhole-foundation/wormhole/pull/4564/files#diff-add5b16afbcd8c63c1a2c83765c5fa39d3989442f3080c807244ed11508f719e) are still here? It might make sense to comment out here too?

johnsaigle:
superseded by #4791

PR #4565, opened 2025-11-03
Stacks Implementation

jannik-stacks:
- Fork of [original PR](https://github.com/wormhole-foundation/wormhole/pull/4546), which had to be synced and the downstream org security rules didn't allow pushes anymore, Github closed the branch and can't recreate within the same PR.


johnsaigle:
@jannik-stacks it looks like this PR is opened on the wrong base, or that the branch contains some commits that are already merged in `main`.

PR #4572, opened 2025-11-04
sdk(rust): Update chain IDs

johnsaigle:
- Adds new chain IDs
- Removes old chain IDs with deprecation comment similar to Go SDK
- Bumps Cargo.toml to a new major version for related crates as this represents a breaking change for previous versions


johnsaigle:
#4573

PR #4574, opened 2025-11-05
governor: Add manual token entries

johnsaigle:


djb15 (approved):
Looks kinda scammy but sure 😄

PR #4577, opened 2025-11-06
notary: fix bug where non-Ethereum messages were always Approved

johnsaigle:
Reference: https://github.com/wormhole-foundation/wormhole/pull/4521#discussion_r2497995168

The expected behaviour is that all chains without a Transfer Verifier should get approved, but in this case all non-Ethereum chains were approved. Now, chains with transfer verifiers should get fully validated by the Notary.

This check was leftover from an earlier phase of development. The current test plan is to run the Notary feature in testnet for a while and to support both Ethereum and Sui messages.

The new approach ensures that the emitter has a Transfer Verifier implementation. If not, it always Approves messages as the Notary only supports transfer verifier messages for now.

### DIFF
--- a/node/pkg/notary/notary.go
+++ b/node/pkg/notary/notary.go
@@ -181,8 +181,8 @@ func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err erro
 		n.logger.Debug("skipping token bridge emitter check because environment is not mainnet or testnet")
 	}
 
-	// The devnet environment does not have token bridges configured for all the relevant chains that we want to test.
-	if n.env == common.MainNet || n.env == common.TestNet {
+	// Perform emitter checks when outside of unit tests or mock environments
+	if n.env == common.MainNet || n.env == common.TestNet || n.env == common.UnsafeDevNet {
 		if tokenBridge, ok := tbEmitters[msg.EmitterChain]; !ok {
 			// Return Unknown if the token bridge is not registered in the SDK.
 			n.logger.Error("notary: unknown token bridge emitter", msg.ZapFields()...)


node/pkg/notary/notary.go
```diff
@@ -154,30 +155,43 @@ func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err erro
...
-		if !bytes.Equal(msg.EmitterAddress.Bytes(), tokenBridge) {
-			n.logger.Debug("notary: automatically approving message publication because it is not from the token bridge", msg.ZapFields()...)
-			return Approve, nil
+	var tbEmitters = make(map[vaa.ChainID][]byte)
+	switch n.env {
+	case common.MainNet:
+		tbEmitters = sdk.KnownTokenbridgeEmitters
+	case common.TestNet:
+		tbEmitters = sdk.KnownTestnetTokenbridgeEmitters
+	case common.UnsafeDevNet, common.AccountantMock, common.GoTest:
```
pleasew8t: Why are we skipping devnet? Surely we'd want even devnet messages to be passed via the notary during tilt tests? The SDK has emitters for devnet: https://github.com/wormhole-foundation/wormhole/blob/main/sdk/devnet_consts.go

johnsaigle: Thanks, I was having an issue earlier while making the PR where my unit test was failing because there wasn't an emitter configured. I just now separated out Devnet from AccountMock and GoTest.

mdulin2 (dismissed):

node/pkg/notary/notary.go
```diff
@@ -154,30 +155,45 @@ func (n *Notary) ProcessMsg(msg *common.MessagePublication) (v Verdict, err erro
...
+		tbEmitters = sdk.KnownTestnetTokenbridgeEmitters
+	case common.UnsafeDevNet:
+		tbEmitters = sdk.KnownDevnetTokenbridgeEmitters
+	case common.AccountantMock, common.GoTest:
+	default:
+		n.logger.Debug("skipping token bridge emitter check because environment is not mainnet or testnet")
+	}
+
+	// The devnet environment does not have token bridges configured for all the relevant chains that we want to test.
+	if n.env == common.MainNet || n.env == common.TestNet {
```
djb15: Should we be auto approving all messages from devnet?

johnsaigle: Added a devnet conditional in the latest commit, and fixed the comment on 184.

mdulin2 (approved):

PR #4585, opened 2025-11-15
node: Add IsWTT() helper method for VAA and MessagePublication

johnsaigle:


sdk/token_bridge.go
```diff
@@ -0,0 +1,111 @@
+package sdk
+
+import (
+	"bytes"
+
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+)
+
+// Environment represents the Wormhole network environment
+type Environment uint8
```
pleasew8t, line 10: Is there a reason for not just importing this from `"github.com/certusone/wormhole/node/pkg/common"`?  Example use [here](https://github.com/johnsaigle/wormhole/blob/242b0483ba985fef727c1dfb19f97abce3351c7e/node/pkg/watchers/sui/watcher.go#L222).

node/pkg/common/chainlock.go
```diff
@@ -628,6 +629,49 @@ func (msg *MessagePublication) VAAHash() string {
...
+// 1. The payload is a transfer (payload type 1 or 3) via vaa.IsTransfer
+// 2. The emitter is a known token bridge emitter for the specified environment
+//
+// This method validates WTTs with respect to an environment's known token bridge emitters.
+// For a context-free check that only verifies the payload type, use vaa.IsTransfer instead.
+//
+// Returns false for test/mock environments (GoTest, AccountantMock) and if either check fails.
+//
+// Note: This method uses the same validation logic as VAA.IsWTT.
+func (msg *MessagePublication) IsWTT(env Environment) bool {
```
pleasew8t, line 643: Why not reuse `IsWTT()` from the sdk? Is it necessary to re-implement?

sdk/token_bridge.go
johnsaigle: The SDK works as a standalone package so it doesn't and shouldn't have a reference to anything under `node/`. Instead, `node/` should make use of the SDK. That's why the data structure is replicated in both places.

node/pkg/common/chainlock.go
johnsaigle: It's implemented once for each data structure. We can use this method when processing a MessagePublication _before_ it's fully validated and signed to become a VAA. In contrast, the SDK method exists so that users of the SDK can do analysis on complete VAAs, for example after they're broadcast.

PR #4592, opened 2025-11-21
Stacks Watcher Update (Full changes)

jannik-stacks:
## Stacks Watcher Update 
Base `main`

**Includes PRs**
- https://github.com/wormhole-foundation/wormhole/pull/4591
- https://github.com/wormhole-foundation/wormhole/pull/4583
- https://github.com/wormhole-foundation/wormhole/pull/4565
- https://github.com/wormhole-foundation/wormhole/pull/4546
- https://github.com/wormhole-foundation/wormhole/pull/4503

**Additional changes**
- Pin a node image version which is more common in the codebase, to increase likelihood of hitting image cache.

---

> Opening for CI runs and further fixes to be pushed in if CI still has failures.

### DIFF
--- a/node/pkg/governor/mainnet_chains.go
+++ b/node/pkg/governor/mainnet_chains.go
@@ -10,7 +10,7 @@ import (
 
 func ChainList() []ChainConfigEntry {
 	return []ChainConfigEntry{
-		{EmitterChainID: vaa.ChainIDSolana, DailyLimit: 20_000_000, BigTransactionSize: 5_000_000},
+		{EmitterChainID: vaa.ChainIDSolana, DailyLimit: 20_000_000, BigTransactionSize: 1_000_000},
 		{EmitterChainID: vaa.ChainIDEthereum, DailyLimit: 30_000_000, BigTransactionSize: 9_000_000},
 		{EmitterChainID: vaa.ChainIDBSC, DailyLimit: 1_000_000, BigTransactionSize: 100_000},
 		{EmitterChainID: vaa.ChainIDPolygon, DailyLimit: 3_000_000, BigTransactionSize: 300_000},


stacks/stacker/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:22.16-bullseye-slim@sha256:550b434f7edc3a1875860657a3e306752358029c957280809ae6395ab296faeb AS builder
+
+RUN apt-get update && apt-get install -y curl gettext-base jq
+
+WORKDIR /root
+COPY package.json package-lock.json /root/
+RUN npm i
```
johnsaigle:
```suggestion
RUN npm ci
```

node/pkg/watchers/stacks/fixture_stacks_simulate.json
```diff
@@ -0,0 +1,105 @@
+{
```
johnsaigle: Can you please remove these JSON files, or justify why they're here?

node/pkg/watchers/stacks/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Configuration
+
+The watcher requires the following configuration:
+
+- Stacks RPC URL: The URL of the Stacks blockchain API
+- Contract Address: The address of the Wormhole Core contract on Stacks
+
+## API Endpoints
+
+The watcher interacts with the Stacks Node RPC API endpoints:
```
johnsaigle: Could you add a link to official Stacks documentation here so that we can cross-reference?

stacks/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM rust@sha256:234b03300ecaae65919eb8cb13ca9b36588b8dff5b1c69581e361f9f25654db1 AS builder
+
+ARG STACKS_CORE_REPO=https://github.com/stacks-network/stacks-core.git
+ARG STACKS_CORE_BASE_BRANCH=develop
```
johnsaigle: Could you change this to work from a specific commit hash, ideally one tied to a release? I think we'll want to use a fixed version rather than tracking the latest changes on `develop`.

johnsaigle (changes requested):
I left some initial feedback. However, it looks like this PR doesn't yet contain the full changes: it's missing [@mdulin2 's fixes from the Stacks fork of Wormhole.](https://github.com/stx-labs/wormhole/pulls?q=is%3Apr+is%3Aclosed)

stacks/stacker/Dockerfile
jannik-stacks: Thx 🙏

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+		if err != nil {
+			logger.Error("Failed to decode raw value hex",
+				zap.String("tx_id", tx.TxId),
+				zap.Uint64("event_index", event.EventIndex),
+				zap.String("hex", event.ContractEvent.RawValue),
+				zap.Error(err))
+			continue
+		}
+
+		if len(hexBytes) > 1024*1024 {
```
johnsaigle: Could you move this length check above the hex-decoding step? I think it makes sense to first check for an unbounded length before doing hex-decoding. Note you'll probably have to double the size of the check because each hex-encoded byte will be two characters vs. the decoded bytes here.

Could you also move this value to a constant so it's easier to track? Finally, if you do so, it would be good to log the actual constant value rather than hardcode `(1MB)` in the log. This will help the log stay accurate if the upper bound needs to changes in the future.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+	}
+
+	// other runtime error
+	if tx.VmError != nil {
+		return 0, fmt.Errorf("transaction %s failed due to runtime error: %s", tx.TxId, *tx.VmError)
+	}
+
+	// success
+
+	wormholeEvents := uint32(0)
```
johnsaigle, line 402: This could maybe be a uint8 unless you're expecting a very large number of wormhole core events per transaction

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+}
+
+// Fetches and processes all transactions in a Stacks block
+func (w *Watcher) processStacksBlock(ctx context.Context, blockHash string, logger *zap.Logger) error {
+	replay, err := w.fetchStacksBlockReplay(ctx, blockHash)
+	if err != nil {
+		return fmt.Errorf("failed to fetch Stacks block replay: %w", err)
+	}
+
+	for _, tx := range replay.Transactions {
```
johnsaigle, line 369: Do you know whether the JSON `transactions` field has stable ordering? I would assume it does. It might be worth adding a comment here because this loop will process transactions according to whatever order the Stacks node yields them.

johnsaigle, line 369: Is it worth adding a length check here? Are blocks always expected to have transactions? If so, then we might want to return an error here.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+/// PROCESS
+
+// Processes all Stacks blocks anchored to the given Bitcoin (burn) block
+func (w *Watcher) processBitcoinBlock(ctx context.Context, tenureBlocks *StacksV3TenureBlocksResponse, logger *zap.Logger) {
+	logger.Info("Processing Bitcoin (burn) block",
+		zap.Uint64("bitcoin_block_height", tenureBlocks.BurnBlockHeight),
+		zap.String("bitcoin_block_hash", tenureBlocks.BurnBlockHash))
+
+	// Process each Stacks block anchored to this burn block
+	for _, block := range tenureBlocks.StacksBlocks {
```
johnsaigle, line 338: Worth adding a length check on `tenureBlocks.StacksBlocks`?

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+// Reobserve implements the interfaces.Reobserver interface.
+func (w *Watcher) Reobserve(ctx context.Context, chainID vaa.ChainID, txID []byte, customEndpoint string) (uint32, error) {
+	logger := supervisor.Logger(ctx)
+
+	// Verify this request is for our chain
+	if chainID != vaa.ChainIDStacks {
+		return 0, fmt.Errorf("unexpected chain ID: %v", chainID)
+	}
+
+	txIdString := hex.EncodeToString(txID)
```
johnsaigle, line 194: Can you add a length check for `txID` and fail if it's wrong? I assume it's a fixed value if we're querying a tx ID.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+
+			// We have a new stable Bitcoin (burn) block height
+			if nodeInfo.StableBurnBlockHeight > previousStableBitcoinHeight {
+				logger.Info("Found new stable Bitcoin (burn) block",
+					zap.Uint64("previous_stable_height", previousStableBitcoinHeight),
+					zap.Uint64("stable_height", nodeInfo.StableBurnBlockHeight))
+
+				w.stableBitcoinHeight.Store(nodeInfo.StableBurnBlockHeight)
+
+				// Convert StableBurnBlockHeight to int64 with overflow check
```
johnsaigle: It looks like the source value `json:"stable_burn_block_height"` is a uint64. Is the actual source block stored as uint64 or as int64? If the former, I don't think this check is needed. If the latter, then I think the JSON decoding should use int64 for representing the block.

node/pkg/watchers/stacks/fetch.go
```diff
@@ -0,0 +1,305 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/certusone/wormhole/node/pkg/common"
+)
+
+type (
```
johnsaigle, line 13: Can you link to the JSON documentation for each of these types? it would be helpful to have a link in case the responses change and these types need to be updated.

node/pkg/watchers/stacks/fetch.go
```diff
@@ -0,0 +1,305 @@
...
+		RewardPhaseBlockLength      uint64             `json:"reward_phase_block_length"`
+		Epochs                      []StacksV2PoxEpoch `json:"epochs"`
+	}
+
+	StacksV2InfoPoxAnchor struct {
+		AnchorBlockHash string `json:"anchor_block_hash"`
+		AnchorBlockTxid string `json:"anchor_block_txid"`
+	}
+
+	StacksV2InfoResponse struct {
```
johnsaigle, line 108: Can these types be more strict? Are all of these really uint64 or are smaller types more accurate for decoding these responses? As mentioned above, a link to documentation would be helpful here.

node/pkg/watchers/stacks/fetch.go
```diff
@@ -0,0 +1,305 @@
...
+func (w *Watcher) fetchStacksBlockReplay(ctx context.Context, blockId string) (*StacksV3TenureBlockReplayResponse, error) {
+	url := fmt.Sprintf("%s/v3/blocks/replay/%s", w.rpcURL, blockId)
+
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	if w.rpcAuthToken != "" {
+		req.Header.Set("Authorization", w.rpcAuthToken)
```
johnsaigle: Can you clarify which of these endpoints need authentication and why? 
It may be better to create an http client in the Watcher object itself and configure it with an API key rather than re-implementing a DefaultClient in each of these methods.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+	// Submit the message to the channel for processing
+	w.msgC <- msgPub
+
+	return nil
+}
+
+/// HELPERS
+
+// Extracts the event name from an event tuple
+func extractEventName(eventTuple *clarity.Tuple) (string, error) {
```
johnsaigle, line 591: Please add a nil check for this parameter.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+		eventName = strVal.Value
+	} else {
+		return "", fmt.Errorf("'event' field is not a string type: %T", eventNameVal)
+	}
+
+	return eventName, nil
+}
+
+// Extracts core message fields from an event tuple
+func extractMessageData(eventTuple *clarity.Tuple) (*MessageData, error) {
```
johnsaigle, line 615: Please add a nil check for this parameter.

node/pkg/watchers/stacks/watcher.go
```diff
@@ -0,0 +1,652 @@
...
+		return nil, fmt.Errorf("invalid 'consistency-level' field: %T", consistencyLevelVal)
+	}
+
+	payloadVal, ok := msgTuple.Values["payload"]
+	if !ok {
+		return nil, fmt.Errorf("missing 'payload' field in message")
+	}
+
+	payload, ok := payloadVal.(*clarity.Buffer)
+	if !ok || payload.Len() > 8192 {
```
johnsaigle: Could you move the value `8192` into a constant and document that this is related to the Clarity transaction size limit?

It may be helpful to create a `consts.go` where the constants are stored for the Stacks package.

node/pkg/watchers/stacks/watcher.go
jannik-stacks: I'd keep uint32 if it's not an issue, even if I believe there will be some other limit (tx size etc.) that would hinder a tx from ever having more than 255 events, just in case.
cc @jeff-stacks

node/pkg/watchers/stacks/fetch.go
jannik-stacks: Thanks, removed some and made stricter where possible. ✅ Most of them where uint64, or possibly uin64 depending on the system running (usize)

node/pkg/watchers/stacks/fetch.go
jannik-stacks: Linked to the openapi spec and the source code, some types aren't immediately clear, but checked for all used types.

node/pkg/watchers/stacks/watcher.go
jannik-stacks: Mainly added this for the linter wanting a check. The source does state `uint64` for this case.

node/pkg/watchers/stacks/watcher.go
jannik-stacks: Stable ordering is given, added a comment.

node/pkg/watchers/stacks/fetch.go
jannik-stacks: Added a comment, it's only this endpoint since for public nodes this endpoint and a few others can cause resource issues and may not be wanted to be exposed.

stacks/test/integration.test.ts
```diff
@@ -0,0 +1,545 @@
...
+    const rewriteClarity = (code: string) => {
+      return code
+        .replaceAll("SP2J933XB2CP2JQ1A4FGN8JA968BBG3NK3EKZ7Q9F", ADDRESS)
+        .replaceAll("SP1E0XBN9T4B10E9QMR7XMFJPMA19D77WY3KP2QKC", ADDRESS)
+        .replaceAll("SP102V8P0F7JX67ARQ77WEA3D3CFB5XW39REDT0AM", ADDRESS);
+    };
+
+    const dependencyFiles = [
+      "trait-sip-010.clar",
+      "proposal-trait.clar",
```
jannik-stacks: @jbencin can we get rid of some or all of these dep contracts?

mdulin2 (approved):

PR #4596, opened 2025-11-24
node: Add manual governed tokens for Monad

djb15:


johnsaigle (approved):

mdulin2 (approved):
Seems like a reasonable starting point. Thanks for doing some research!

PR #4602, opened 2025-11-26
fix: add NEAR address support and RPC override to info origin command

kinsyudev:
## Summary
- Fixes error when querying origin information for NEAR addresses by implementing proper NEAR address resolution
- Adds RPC override capability to the `info origin` command for more flexible querying

### DIFF
--- a/clients/js/src/cmds/info/origin.ts
+++ b/clients/js/src/cmds/info/origin.ts
@@ -8,42 +8,44 @@ import { toChain } from "@wormhole-foundation/sdk-base";
 export const command = "origin <chain> <address>";
 export const desc = `Print the origin chain and address of the asset that corresponds to the given chain and address.`;
 export const builder = (y: typeof yargs) =>
-  y
-    .positional("chain", {
-      describe:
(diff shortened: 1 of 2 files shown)


clients/js/src/cmds/info/origin.ts
```diff
@@ -8,42 +8,48 @@ import { toChain } from "@wormhole-foundation/sdk-base";
...
-  console.log({
-    ...res,
-    assetAddress: tryUint8ArrayToNative(res.assetAddress, toChain(res.chainId)),
-  });
+	const network = getNetwork(argv.network);
 
-  console.warn = consoleWarnTemp;
+	const res = await getOriginalAsset(
+		chainToChain(argv.chain),
+		network,
```
johnsaigle, line 41: `network` seems to always be default here, so it becomes mainnet. Is the idea that this should support other networks? If so it might be worth modifying the `command` text at the top to ensure that the network can be customized, and that users know that it'll default to mainnet.

djb15:
Stale

PR #4614, opened 2025-12-08
node(txverifier): skip logs with parsing errors instead of returning error

johnsaigle:
Modifies the design philosophy in the EVM Transfer Verifier with respect
to error-handling.

Previously, the code would return a "receipt error" when encountering an
EVM log that was not well-formed or else was extremely unusual. Instead,
that code now skips those logs and does not include them in the
invariant calculations.

Practically, this results in fewer cases where the watcher will give the
`CouldNotVerify` status to Message Publications that have unusual logs
in their receipts. This should limit the ability for a user to control the outcome of
querying the Transfer Verifier by passing in logs that are irrelevant
but malformed.

The logs that remain after this filtering will still be subject to the
invariant checks.

- Changes instances of log parsing to skip instead of return errors
- Adds more happy path unit tests to demonstrate that problematic logs
  are skipped
- Add additional comments and cleanup helper functions in unit tests

### DIFF
--- a/node/pkg/common/chainlock.go
+++ b/node/pkg/common/chainlock.go
@@ -131,7 +131,7 @@ const (
 	Valid
 	// Indicates that no verification is necessary.
 	NotApplicable
-	// The message could not complete the verification process. Indicates malformed input or network issues.
+	// The message could not complete the verification process.
 	CouldNotVerify
 )
 


node/pkg/txverifier/evm.go
```diff
@@ -527,6 +538,7 @@ func (tv *TransferVerifier[evmClient, connector]) parseReceipt(
 			transfers = append(transfers, transfer)
 
 		// Process LogMessagePublished events.
+		// Parsing errors that occur here should bubble up to the caller.
```
djb15: On this point, is it safer to do the same skipping action on the message published events? In the very unlikely event that the core bridge were exploited and upgrade authority seized from the Guardians, an attacker could insert a new event that fails some of the validation steps below.

They would then emit arbitrary transfer messages alongside this malformed event, also skipping the Transfer Verifier logic using the same approach. 

The counterargument would be that we might end up delaying legitimate transfers, but the logs **should** be of a consistent format guaranteed by the EVM and the smart contract

johnsaigle: IMO this is an argument for changing the `Notary` to delay messages that have `CouldNotVerify`, which is triggered by these errors.

If a single LogMessagePublished event log is broken in this way, I don't think it's a good idea to consider processing other logs from the same contract.

The threat model is different compared to what's patched here: users have a lot more control over Deposit/Transfer events based on their being derived from permissionless contracts. If the core bridge messages are malformed, even a single one, it seems like potentially an emergency situation (e.g. a hack as you indicated).

What do you think?

djb15: Yeah these are good points, I agree with that. If we only threw `CouldNotVerify` (or some equivalent) for invalid message publish events and then delayed all transfers from that transaction I think that makes sense given it should never be user controlled like you say. But I think you'd have to reorder the checks, specifically [this one](https://github.com/wormhole-foundation/wormhole/blob/82f8d98dca4916a716716c4189285b7b7497cc0b/node/pkg/txverifier/evm.go#L546-L550) should go first?

johnsaigle: I reordered the checks a little bit and added some comments. It's not that easy to do all the skip checks first, then all the parsing checks, but I did my best.

djb15: I think this is good now. I believe we will now skip events that aren't emitted by the core bridge and where the emitter (in the actual event) is not the token bridge. Any other errors in parsing will result in a receipt error which results in the transaction being delayed. This is fine as a parsing error from such events would indicate a core bridge or token bridge exploit

djb15 (approved):

node/pkg/txverifier/evm.go
johnsaigle: Yes that's my intention as well as my understanding of how the latest implementation operates.

djb15 (approved):

bemic (approved):

PR #4618, opened 2025-12-11
node(governor): Add defense in depth checks for computeValue

johnsaigle:
- Modify computeValue to validate its parameters. The callers are already safe in each instance, but this helps to facilitate unit tests and refactoring
- Add Price validation for mainnet tokens (must be non-nil and non-negative)
- Try to avoid unnecessary copies of tokens and chains in unit tests

### DIFF
--- a/node/pkg/governor/governor.go
+++ b/node/pkg/governor/governor.go
@@ -567,8 +567,8 @@ func (gov *ChainGovernor) processMsgForTime(msg *common.MessagePublication, now 
 		return false, err
 	}
 
-	// Compute the notional USD value of the transfers
-	value, err := computeValue(payload.Amount, token)
+	// Compute the notional USD value.
+	value, err := usdValue(payload.Amount, token)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


node/pkg/governor/governor.go
```diff
@@ -989,14 +990,26 @@ func (gov *ChainGovernor) checkPendingForTime(now time.Time) ([]*common.MessageP
 }
 
 func computeValue(amount *big.Int, token *tokenEntry) (uint64, error) {
+
+	// Ensure that the input is valid:
+	// - Price must be non-nil and greater than 0.
```
bemic: "Price greater than 0" check is missing here?

bemic: Or the comment is redundant

johnsaigle: Thanks, should be fixed now!

djb15 (dismissed):

djb15 (approved):

PR #4628, opened 2025-12-22
node: Add support for delegated guardian sets

nvsriram:
This PR adds support for delegated guardian sets:
* `WormholeDelegatedGuardians` EVM smart contract configures delegated guardians per chain id 
* `delegated-guardians-config` admin command to update delegated guardian set
* Guardian node stores `DelegatedGuardianChainConfig` locally (updated every 15 seconds by watcher) and references that to handle messages (see below for more details) from delegated chains
* New gossip message type, `DelegateObservation`, and its signed p2p wrapper, `SignedDelegateObservation`, that is emitted by delegated guardians via `delegatedAttestationPubsubTopic`
* Tilt setup to work with arbitrary delegated guardian set preset configuration
* End-to-end Go tests designed to run on Tilt

---

### Glossary of Terms

* **(Un)delegated Chain**:  A chain is undelegated  if there is no `DelegatedGuardianSet` configured for it in the <code>WormholeDelegatedGuardians</code> EVM smart contract. Conversely, a chain is delegated if a  `DelegatedGuardianSet` is configured for it.
* **Delegated Guardian**:  A guardian is delegated for a chain if the chain is delegated and the guardian is part of the  `DelegatedGuardianSet`  for the chain.
* **Canonical Guardian**:  A guardian is canonical for a chain if the chain is delegated and the guardian is not part of the  `DelegatedGuardianSet` for the chain.

> [!NOTE]
> **Delegated Guardian** and **Canonical Guardian** are scoped terms; by default they mean the delegated/canonical guardian for the message’s <code>EmitterChain</code>.</details>

### Updated Message Flow

This flow is applicable for delayed messages and reobservations as well.
The message flow is unchanged for undelegated chains. The following flow describes the updated flow for delegated chains:

#### For Delegated Guardian

1. Guardian runs watcher for delegated chain and observes event — processor receives `MessagePublication` via `msgC`
2. Guardian runs the message through the Notary — `continue` if unsuccessful
3. Guardian converts `MessagePublication` into `DelegateObservation`, populates missing fields, marshals it, and sends it to `gossipDelegatedAttestationSendC` channel:
   - In p2p, guardian adds the `signedDelegateObservationPrefix`, signs, coverts it to a `SignedDelegateObservation` gossip message and publishes it to the `delegatedAttestationPubsubTopic`. 
4. Guardian runs the message through the Governor, and/or the Accountant and then broadcasts the observation
5. Guardian ignores incoming `SignedDelegateObservation` for this chain

#### For Canonical Guardian

1. Guardian ignores any `MessagePublication` from local watcher (if exists)
2. In p2p, guardian receives `SignedDelegateObservation` gossip messages from `delegatedAttestationSubscription`:
   - Guardian processes (verifies prefix, signature, timestamp, etc.) the `SignedDelegateObservation` and sends it to the `delegateObsvC` channel
   - In the processor, guardian unmarshals the `DelegateObservation` from its signed wrapper and verifies that the observation is from a delegated guardian for this chain.
4. Guardian stores current delegated guardian's `DelegateObservation` in local `delegateObservationMap`
5. If delegate threshold is reached, then guardian converts `DelegateObservation` into `MessagePublication` and runs it through main message processing loop.

### DIFF
--- a/node/pkg/watchers/aptos/watcher.go
+++ b/node/pkg/watchers/aptos/watcher.go
@@ -126,6 +126,19 @@ func (e *Watcher) Run(ctx context.Context) error {
 				panic("invalid chain ID")
 			}
 
+			// Aptos's TxID is a uint64. Historically, all TxIDs used a fixed 32-byte hash type.
+			// This parsing is leftover from that time period. It should be possible to refactor
+			// this code such that the TxID received from p2p is exactly 8 bytes, which would
+			// obviate the need for the below bounds check and parsing.
+			//
+			// SECURITY: This acts as a bounds check for the BigEndian.Unint64 call below.
+			const AptosTxIDExpectedLen = 32
+			if len(r.TxHash) < AptosTxIDExpectedLen {
+				logger.Error("invalid TxID: too short")
+				p2p.DefaultRegistry.AddErrorCount(e.chainID, 1)
+				continue
+			}
+
 			// uint64 will read the *first* 8 bytes, but the sequence is stored in the *last* 8.
 			nativeSeq := binary.BigEndian.Uint64(r.TxHash[24:])
 
@@ -295,6 +308,7 @@ func (e *Watcher) observeData(logger *zap.Logger, data gjson.Result, nativeSeq u
 	emitter := make([]byte, 8)
 	binary.BigEndian.PutUint64(emitter, em.Uint())
 
+	// SECURITY: vaa.Address is guaranteed to be 32 bytes so copy's slice into `a` is safe.
 	var a vaa.Address
 	copy(a[24:], emitter)
 


node/pkg/p2p/p2p.go
```diff
@@ -87,9 +87,12 @@ var heartbeatMessagePrefix = []byte("heartbeat|")
 
 var signedObservationRequestPrefix = []byte("signed_observation_request|")
 
+var signedDelegateObservationPrefix = []byte("signed_delegate_observation|")
```
evan-gray: see whitepapers/0009_guardian_signer.md - please pad this to 35 bytes, e.g.
```suggestion
var signedDelegateObservationPrefix = []byte("signed_delegate_observation_000000|")
```

node/pkg/processor/delegated_guardian_test.go
```diff
@@ -86,11 +86,100 @@ func getNextConfigIndex(t *testing.T) uint64 {
 	return configIndex.Uint64()
 }
 
-func publishMessageToEthereum(t *testing.T, nonce uint32, payload []byte, consistencyLevel uint8) string {
-	client, err := ethclient.Dial(ethDevnet2RPC)
+func connectToEthereumRPC(t *testing.T, useDevnet bool) *ethclient.Client {
+	rpc := ethDevnetRPC
```
johnsaigle, line 125: It looks like both values for this bool are being set to some kind of devnet. Could you rename the variable or add a comment explaining this?

node/pkg/processor/delegated_guardian_test.go
```diff
@@ -505,7 +606,54 @@ func TestDelegateChainUndelegated(t *testing.T) {
...
+			Keys: []eth_common.Address{
+				eth_common.HexToAddress("0x" + guardian1),
+				eth_common.HexToAddress("0x" + guardian2),
+			},
+		},
+	}
+
+	createAndSubmitDelegatedGuardiansConfig(t, rollbackConfig)
+	t.Log("Waiting for guardian to pick up configuration...")
+	time.Sleep(20 * time.Second)
```
johnsaigle: Could you look into `t.Parallel()` for this and related tests? My understanding is that Go runs all tests in sequentially so each of these Sleeps will add up to much longer test runs.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -27,6 +27,12 @@ var (
 	)
 )
 
+// Chains like Ethereum and Solana should not be delegated due to their governance implications
+var nonDelegableChains = map[vaa.ChainID]struct{}{
+	vaa.ChainIDEthereum: {},
+	vaa.ChainIDSolana:   {},
```
johnsaigle: Could you please add Wormchain as well? This runs the Accountant so all Guardians need it.

node/pkg/processor/delegated_guardian_test.go
nvsriram: Since these tests run on Tilt and share the same delegated guardians smart contract address and guardian setup, it cannot be run in parallel. Secondly, this file could be thought of as one big test that mimics common delegated guardian config update scenarios.

node/pkg/processor/delegated_guardian_config.go
nvsriram: Done!

node/pkg/processor/delegated_guardian_test.go
johnsaigle: Hmm OK I didn't understand that this was being included via go:build as an integration test. I don't think we've done testing like that before. Do you think you could add some documentation to the top of this file giving that context?

On the `delegated-guardian-setup.yaml` also, I think you should check with @evan-gray on how that's set-up. We have existing Docker images that have foundry and so on installed so it may be better to leverage those images rather than do an ad-hoc installation.

wormchain/contracts/tools/deploy_wormchain.ts
```diff
@@ -402,6 +402,21 @@ async function main() {
         address: "wormhole1g25zz7gyuyh6chuejc3ppfemgfla4xpsm69lzq",
         name: "guardianAccountant1",
       }),
+      client.core.msgCreateAllowlistEntryRequest({
```
johnsaigle, line 405: What are the reasons for the modifications to Wormchain?

sdk/vaa/structs.go
```diff
@@ -81,6 +81,12 @@ type (
 		TargetChain   ChainID
 	}
 
+	// DelegatedGuardianConfig contains the threshold and keys for delegated guardians on a specific chain
+	DelegatedGuardianConfig struct {
+		Threshold uint8
+		Keys      []common.Address
```
johnsaigle, line 87: I know we're a bit late in the cycle for this, but any thoughts on converting this to a set (`map` with empty structs), as you did for the non-delegable chains?

This would take care of the edge-cases about duplicate Keys that we raised elsewhere. Arguably it's not worth the effort at this point for an extreme edge-case, but thought I'd raise it.

node/pkg/processor/delegated_guardian_test.go
nvsriram: > Hmm OK I didn't understand that this was being included via go:build as an integration test. I don't think we've done testing like that before. Do you think you could add some documentation to the top of this file giving that context?

You're right, the delegated guardian config tests needed to exist in the layer between Go unit tests and TypeScript client tests. Hence, they act as integration tests that run on Tilt CI and are gated using go:build tags. I've now documented the context.

wormchain/contracts/tools/deploy_wormchain.ts
nvsriram: I believe this was because we needed 3 guardians to be configured for the delegated guardian config integration tests.

sdk/vaa/structs.go
nvsriram: I think initially the idea was to mimic the regular GuardianSet and have `KeyIndex` return the index (which could be used for verification). But considering we have not come across any use case for it and the main use just seems to be to check if exists, I think the change to a set makes sense.

However, considering this could require changes in multiple places and would need to ensure any deterministic key ordering expectations are maintained, I will add that as a TODO for now and if there are free cycles, then can add it to this PR.

node/pkg/processor/delegated_guardian_test.go
fergarrui: Re `delegated-guardian-setup.yaml`, foundry image was lacking some of the dependencies needed for the setup guardians script. Iirc that image was kind of like a busybox and couldn't install utilities such as `jq`, etc.
Not sure what the best way of doing it in this CI would it be.

Will add some additional context to the Go e2e test.

sdk/vaa/structs.go
nvsriram: In order to handle the duplicate Keys edge case, I've added an explicit check against duplicate keys and have created unit tests as well. 

In the future, if we see that there is no reason to maintain deterministic key ordering that mimics the smart contract, we can always just port to the set.

Tiltfile
```diff
@@ -396,6 +497,37 @@ if num_guardians >= 2 and ci == False:
         trigger_mode = trigger_mode,
     )
 
+# delegated guardian sets
+if require_per_guardian_config:
+    script_content = str(read_file("scripts/delegated-guardian-set-preset.sh"))
```
evan-gray: this seems pretty convoluted. instead of using a debian image, build a docker image with the correct tools and files included

devnet/delegated-guardian-setup.yaml
```diff
@@ -0,0 +1,96 @@
...
+            - -c
+            - |
+              set -e
+              
+              echo "Installing dependencies..."
+              apt-get update -qq
+              apt-get install -y -qq curl jq ca-certificates git bsdmainutils
+              
+              echo "Installing kubectl..."
+              curl -LO "https://dl.k8s.io/release/v1.28.0/bin/linux/amd64/kubectl"
```
evan-gray: this will waste time on every run installing these dependencies. This should instead build/use a docker image with them already installed and simply run the script.

ethereum/forge-scripts/DeployWormholeDelegatedGuardians.s.sol
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  function run(address wormholeCore) public returns (address deployedDelegatedGuardians) {
+    vm.startBroadcast();
+    deployedDelegatedGuardians = _deploy(wormholeCore);
+    vm.stopBroadcast();
+  }
+
+  function _deploy(address wormholeCore) internal returns (address deployedDelegatedGuardians) {
+    WormholeDelegatedGuardians delegated = new WormholeDelegatedGuardians(wormholeCore);
```
evan-gray: since this is immutable, I recommend doing this with create2 instead.

node/cmd/guardiand/node.go
```diff
@@ -2055,6 +2063,18 @@ func checkEvmArgs(logger *zap.Logger, rpcURL string, contractAddr string, chainI
 	return contractAddr
 }
 
+// checkDelegatedGuardiansContract returns the delegated guardians contract address for the given chain.
+// If we are in devnet mode and the contract address is not specified, it returns the deterministic one for tilt.
+func checkDelegatedGuardiansContract(rpcURL string, contractAddr string, chainID vaa.ChainID) string {
```
evan-gray: I would recommend against this approach.
1. this is only used for Ethereum, for which rpcURL is required
2. you could simply specify the contract in the devnet config instead.

sdk/devnet_consts.go
```diff
@@ -52,3 +52,8 @@ var KnownDevnetWrappedNativeAddresses = map[vaa.ChainID]string{
 var KnownDevnetCoreContracts = map[vaa.ChainID]string{
 	vaa.ChainIDEthereum: "0xC89Ce4735882C9F0f0FE26686c53074E09B0D550",
 }
+
+// KnownDevnetDelegatedGuardiansContracts is a map of known delegated guardians contract addresses used during development.
```
evan-gray: with my other recommendation of solely relying on the guardiand parameter, this is not needed.

node/cmd/guardiand/node.go
```diff
@@ -1273,14 +1275,20 @@ func runNode(cmd *cobra.Command, args []string) {
...
-			Contract:               *ethContract,
-			GuardianSetUpdateChain: true,
-			CcqBackfillCache:       *ccqBackfillCache,
-			TxVerifierEnabled:      slices.Contains(txVerifierChains, vaa.ChainIDEthereum),
+			NetworkID:                     "eth",
+			ChainID:                       vaa.ChainIDEthereum,
+			Rpc:                           *ethRPC,
+			Contract:                      *ethContract,
+			GuardianSetUpdateChain:        true,
+			DelegatedGuardiansContract:    dgContract,
```
evan-gray: can the existence of the non-nil `!= ""` string be used instead of having two parameters? this doesn't require the same logic as `GuardianSetUpdateChain` since every watcher watches the core contract which requires the separation of only one watcher streaming updates. Here, if the `DelegatedGuardiansContract` is specified, it's only purpose is to sync delegate set updates.

node/pkg/node/options.go
```diff
@@ -58,7 +59,7 @@ func GuardianOptionP2P(
 ) *GuardianOption {
 	return &GuardianOption{
 		name:         "p2p",
-		dependencies: []string{"accountant", "alternate-publisher", "gateway-relayer", "governor", "query"},
+		dependencies: []string{"accountant", "alternate-publisher", "gateway-relayer", "governor", "processor", "query"},
```
evan-gray: I may have not gotten there yet, but what was the reason for the dependency change?

evan-gray: It looks like this might just be to add the feature flag which is very convoluted... see other comments

node/pkg/node/options.go
```diff
@@ -462,6 +467,15 @@ func GuardianOptionWatchers(watcherConfigs []watchers.WatcherConfig, ibcWatcherC
 					g.chainQueryReqC[wc.GetChainID()] = make(chan *query.PerChainQueryInternal, query.QueryRequestBufferSize)
 				}
 
+				// For EVM watchers, set the delegated guardian config channel
```
evan-gray, line 468: as far as I can tell, this is a very roundabout way just to set the contract address for the feature flag, when instead, it could have been passed in to `GuardianOptionP2P`'s `featureFlags` parameter in `node.go`. I recommend that much simpler approach instead.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+
+	return -1, false
+}
+
+type DelegatedGuardianConfig struct {
+	mu              sync.RWMutex
+	Chains          map[vaa.ChainID]*DelegatedGuardianChainConfig
+	contractAddress string
```
evan-gray: this and the functions that use it are not needed if the feature flag is populated by `node.go` in the first place.

node/pkg/p2p/p2p.go
```diff
@@ -682,6 +690,79 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+					)
+
+					if controlPubsubTopic == nil {
+						panic("controlPubsubTopic should not be nil when delegateObsvSendC is set")
+					}
+					logger.Debug("p2p: publishing delegate observation to control topic",
+						zap.Uint32("emitter_chain", msg.EmitterChain),
+						zap.Uint64("sequence", msg.Sequence),
+					)
+					err = controlPubsubTopic.Publish(ctx, b)
```
evan-gray: is the control topic appropriate for these messages? perhaps a new topic would be more appropriate. currently there's 
- control for heartbeats and configs
- attestation for observations
- vaa for signed vaas (which the guardians don't have to listen to)
maybe this warrants a new channel since it's possible that traffic will mirror a subset of the attestation channel

I would not want the control channel to become overwhelmed by these messages when heartbeats are critical for detecting guardian peers.

node/pkg/watchers/evm/watcher.go
```diff
@@ -188,6 +195,8 @@ func NewEthWatcher(
 	chainID vaa.ChainID,
 	msgC chan<- *common.MessagePublication,
 	setC chan<- *common.GuardianSet,
+	dgConfigC chan<- *processor.DelegatedGuardianConfig,
```
evan-gray, line 197: depending on prior comments, either this config having `contractAddress` or `dgContractAddr` is duplicative.

node/pkg/watchers/evm/connectors/delegated_guardians_connector.go
```diff
@@ -0,0 +1,66 @@
...
+	networkName string
+	address     ethCommon.Address
+	logger      *zap.Logger
+	client      *ethClient.Client
+	rawClient   *ethRpc.Client
+	caller      *dgAbi.DelegatedguardiansCaller
+}
+
+// NewDelegatedGuardiansConnector creates a new connector for the delegated guardians contract.
+func NewDelegatedGuardiansConnector(ctx context.Context, networkName, rawUrl string, address ethCommon.Address, logger *zap.Logger) (*DelegatedGuardiansConnector, error) {
```
evan-gray: I recommend against using another connector. Instead, reuse or augment `EthConn` akin to `fetchAndUpdateGuardianSet` because this code is not appropriately handling the client here and there's no need for a second websocket connection.

node/pkg/watchers/evm/watcher.go
```diff
@@ -357,6 +368,30 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 	}
 	defer messageSub.Unsubscribe()
 
+	// Initialize delegated guardians connector if configured
```
evan-gray: this block is not needed if reusing ethConn like `fetchAndUpdateGuardianSet`

devnet/delegated-guardian-setup.yaml
fergarrui: can you check [this commit](https://github.com/wormhole-foundation/wormhole/pull/4628/changes/27b9c1341bcebad1c8ff4eafe6ef3bc54040a4b5)? 

[the next one](https://github.com/wormhole-foundation/wormhole/pull/4628/changes/88c8ecb1aa8825f5935bbf50972bf71c0dc09c02) is just to pin the debian image

ethereum/forge-scripts/DeployWormholeDelegatedGuardians.s.sol
fergarrui: Done

node/pkg/processor/delegated_guardian_test.go
fergarrui: Actually resolved [here](https://github.com/wormhole-foundation/wormhole/pull/4628/changes/27b9c1341bcebad1c8ff4eafe6ef3bc54040a4b5)

devnet/delegated-guardian-setup/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Install foundry
+RUN curl -L https://foundry.paradigm.xyz | bash && \
+    /root/.foundry/bin/foundryup
+
+# Add foundry to PATH
+ENV PATH="/root/.foundry/bin:$PATH"
+
+# Create a simple health check
+HealthCheck CMD echo "setup-ready"
```
evan-gray: lint: Instructions should be written in uppercase letters

evan-gray: but also I'm not sure what the purpose of this is in this context.

devnet/delegated-guardian-setup/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
```
evan-gray, line 35: I would expect the script to get copied in here.

Tiltfile
```diff
@@ -396,6 +497,43 @@ if num_guardians >= 2 and ci == False:
...
+    
+    delegated_setup_yaml = read_yaml_stream("devnet/delegated-guardian-setup.yaml")
```
evan-gray: this should work with a simple `k8s_yaml_with_ns` without the need to make any modifications if using `build_args` below

Tiltfile
```diff
@@ -396,6 +497,43 @@ if num_guardians >= 2 and ci == False:
...
+
+    docker_build(
+        ref = "delegated-guardian-setup",
+        context = "./devnet/delegated-guardian-setup/",
+        dockerfile = "./devnet/delegated-guardian-setup/Dockerfile",
+    )
```
evan-gray, line 509: use `build_args` here. see above examples. there should be no need to do any of the below aside from `k8s_resource`

node/cmd/guardiand/node.go
```diff
@@ -553,6 +556,7 @@ func init() {
 	transferVerifierEnabledChainIDs = NodeCmd.Flags().UintSlice("transferVerifierEnabledChainIDs", make([]uint, 0), "Transfer Verifier will be enabled for these chain IDs (comma-separated)")
 
 	notaryEnabled = NodeCmd.Flags().Bool("notaryEnabled", false, "Run the notary")
+	delegatedGuardiansEnabled = NodeCmd.Flags().Bool("delegatedGuardiansEnabled", false, "Enable the delegated guardians protocol")
```
evan-gray: what happens if `delegatedGuardiansEnabled` is set but `ethDelegatedGuardiansContract` is not? should these be enforced to both be set? is there value in having two parameters for this if one is a boolean and both are required - e.g. could the non-empty string check be used instead?

evan-gray: furthermore, I expect if this was done the call to processor wouldn't need an additional flag as the delegated guardian config would simply never be set. would be curious to get other's opinions.

node/pkg/node/options.go
```diff
@@ -462,6 +465,11 @@ func GuardianOptionWatchers(watcherConfigs []watchers.WatcherConfig, ibcWatcherC
...
+				if evmWc, ok := wc.(*evm.WatcherConfig); ok {
```
evan-gray, line 469: just a nit, but it is a bit odd that this is set for all evm watchers instead of just Ethereum / ones that the delegate set contract is != "" for. It is also odd that this mechanism works different than `g.setC.writeC` which is always passed to create, but I gather that was done to avoid changing the interface. feel free to resolve this though, as I don't think the check is too big an issue as currently written.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,180 @@
...
+// SetContractAddress sets the delegated guardians contract address being monitored.
+func (d *DelegatedGuardianConfig) SetContractAddress(addr string) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.contractAddress = addr
+}
+
+// GetFeatures returns the delegated guardian feature string for heartbeat messages.
+// Returns "dgset:<contract_address>" if a contract is configured, or "" otherwise.
+func (d *DelegatedGuardianConfig) GetFeatures() string {
```
evan-gray: this function is not used.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,180 @@
...
+	ret := make(map[vaa.ChainID]*DelegatedGuardianChainConfig, len(d.Chains))
+	for chain, cfg := range d.Chains {
+		ret[chain] = cfg
+	}
+
+	return ret
+}
+
+// SetContractAddress sets the delegated guardians contract address being monitored.
+func (d *DelegatedGuardianConfig) SetContractAddress(addr string) {
```
evan-gray: this function is not used.

node/pkg/watchers/evm/watcher.go
```diff
@@ -730,6 +767,83 @@ func fetchCurrentGuardianSet(ctx context.Context, ethConn connectors.Connector)
...
+	logger *zap.Logger,
+	ctx context.Context,
+) error {
+	// defensive check in case this function is called from another path in the future.
+	if w.dgContractAddr == "" || w.dgConfigC == nil {
+		return nil
+	}
+
+	msm := time.Now()
+	logger.Info("fetching delegated guardian config from contract")
```
evan-gray: this feels like a debug

node/pkg/watchers/evm/watcher.go
```diff
@@ -730,6 +767,83 @@ func fetchCurrentGuardianSet(ctx context.Context, ethConn connectors.Connector)
...
+
+	configs, err := w.ethConn.GetDelegatedGuardianConfig(timeout)
+	if err != nil {
+		ethConnectionErrors.WithLabelValues(w.networkName, "delegated_guardian_config_fetch_error").Inc()
+		p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
+		logger.Error("failed to fetch delegated guardian config", zap.Error(err))
+		return err
+	}
+
+	logger.Info("successfully fetched delegated guardian config",
```
evan-gray: this feels like a debug - the guardian set one only logs when it changes, but this seems to log every time

node/pkg/watchers/evm/watcher.go
```diff
@@ -137,6 +138,11 @@ type (
 		latestSafeBlockNumber      uint64
 		latestFinalizedBlockNumber uint64
 
+		// Delegated guardians config
+		dgConfigC                               chan<- *processor.DelegatedGuardianConfig
+		dgContractAddr                          string
```
evan-gray: should this have been `eth_common.Address` like `contract` above?

node/pkg/watchers/evm/watcher.go
```diff
@@ -1042,7 +1156,13 @@ func (w *Watcher) createConnector(ctx context.Context, url string) (ethConn conn
 		return
 	}
 
-	baseConnector, err := connectors.NewEthereumBaseConnector(ctx, w.networkName, url, w.contract, w.logger)
+	var delegatedGuardiansAddr *eth_common.Address
+	if w.dgContractAddr != "" && w.dgConfigC != nil {
```
evan-gray: wonder if this could be avoided if this was passed in as an address in the first place, like is done with `contract` in the `Create` method of config.go

node/pkg/p2p/p2p.go
```diff
@@ -416,7 +425,7 @@ func Run(params *RunParams) func(ctx context.Context) error {
 				}
 			}()
 
-			if params.obsvReqRecvC != nil || params.signedGovCfgRecvC != nil || params.signedGovStatusRecvC != nil || params.gst.IsSubscribedToHeartbeats() {
+			if params.obsvReqRecvC != nil || params.signedGovCfgRecvC != nil || params.signedGovStatusRecvC != nil || params.delegateObsvRecvC != nil || params.gst.IsSubscribedToHeartbeats() {
```
evan-gray: `delegateObsvRecvC` is missing from the above if, so this won't be hit if only `delegateObsvRecvC` is set. it should be added to the outer if as well.

evan-gray (commented):
some nits

node/pkg/p2p/p2p.go
nvsriram: I think `delegateObsvRecvC` should not be part of the inner if actually since it has a dedicated channel. I've now removed it from the control channel setup.

node/cmd/guardiand/node.go
fergarrui: That was my question too, it was suggested by @johnsaigle 

Image

fergarrui: Not sure if I misunderstood something

node/pkg/watchers/evm/watcher.go
fergarrui: To compare it to `!= ""` but we can use `*eth_common.Address` and `!= nil` instead
Will do that

node/cmd/guardiand/node.go
evan-gray: yea, I agree I'd like to remove the opportunity for "weird configuration"

johnsaigle: I believe cobra.Command has a built-in method to state "these flags must be passed in together". It might be `MarkFlagsRequiredTogether`?

evan-gray: I'd still lean towards just using `ethDelegatedGuardiansContract != ""` in place of `delegatedGuardiansEnabled`

fergarrui: done, resolving

evan-gray (dismissed):

node/pkg/adminrpc/adminserver.go
```diff
@@ -702,6 +702,51 @@ func coreBridgeSetMessageFeeToVaa(req *nodev1.CoreBridgeSetMessageFee, timestamp
...
+		Keys      []string `json:"keys"`
+		Threshold uint8    `json:"threshold"`
+	}
+	if err := json.Unmarshal([]byte(req.Config), &rawConfig); err != nil {
+		return nil, fmt.Errorf("failed to parse JSON: %w", err)
+	}
+
+	configs := make(map[vaa.ChainID]vaa.DelegatedGuardianConfig)
+	for chainIDStr, cfg := range rawConfig {
+		chainID, err := strconv.ParseUint(chainIDStr, 10, 16)
```
johnsaigle: Please change this to use the SDK's KnownChainIDFromString.
This prevents two invalid cases:
* Processing a config for an invalid chain, e.g. one that does not exist or has been deprecated
* Handling the ChainID == 0 case.

node/pkg/p2p/p2p.go
```diff
@@ -948,6 +1016,96 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+								break
+							}
+
+							// SECURITY: The p2p signature on SignedDelegateObservation is verified in processSignedDelegateObservation. The processor
+							// assumes this verification has already occurred. Do not bypass this check, or untrusted data may reach the processor.
+							_, err := processSignedDelegateObservation(d, gs)
+							if err != nil {
+								p2pMessagesReceived.WithLabelValues("invalid_signed_delegate_observation").Inc()
+								if logger.Level().Enabled(zapcore.DebugLevel) {
+									logger.Debug("invalid signed delegate observation received",
```
johnsaigle:
```suggestion
									logger.Error("invalid signed delegate observation received",
```
This seems like something we want to be loud about. It shows that the p2p is seriously misbehaving or that someone is trying to forge a message, right?

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		[]string{"chain"},
+	)
+)
+
+// Chains like Ethereum, Solana, and Wormchain should not be delegated due to their governance implications
+var nonDelegableChains = map[vaa.ChainID]struct{}{
+	vaa.ChainIDEthereum:  {},
+	vaa.ChainIDSolana:    {},
+	vaa.ChainIDWormchain: {},
```
johnsaigle, line 35: Could you please add `vaa.ChainIDUnset`?

node/pkg/processor/processor.go
```diff
@@ -290,91 +362,203 @@ func (p *Processor) Run(ctx context.Context) error {
...
-					if verdict == guardianNotary.Blackhole {
-						// Black-holed messages should not be processed.
-						p.logger.Error("message will not be processed", k.ZapFields(zap.String("verdict", verdict.String()))...)
-					} else {
-						// Delayed messages are added to a separate queue and processed elsewhere.
-						p.logger.Error("message will be delayed", k.ZapFields(zap.String("verdict", verdict.String()))...)
-					}
-					// We're done processing the message.
+			// Log details for new/updated chain configs
+			for chain, cfg := range chains {
```
johnsaigle: Iterating over maps is non-deterministic. Could you add a comment like: "NOTE: Iteration is non-deterministic. Ensure that the code in this loop does not modify control flow." ?

Otherwise, consider sorting the keys of `chains` and iterating deterministically.

node/pkg/processor/processor.go
```diff
@@ -290,91 +362,203 @@ func (p *Processor) Run(ctx context.Context) error {
...
+				if _, isNonDelegable := nonDelegableChains[chain]; isNonDelegable {
+					p.logger.Error("attempted to set config for non-delegable chain; ignoring",
+						zap.Stringer("chainID", chain),
+					)
+					delete(chains, chain)
```
johnsaigle, line 452: Worth adding a comment here that `delete` on a `map` while iterating over it _should_ be safe in Go so long as this is not done concurrently.

Since we already have the a mutex for `DelegatedGuardianConfig` I think it would be worth adding a RLock() for this loop. I doubt there's any issue with the code as written now but it'll help us maintain the code safely going forward.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type DelegatedGuardianChainConfig struct {
+	// TODO: Use map[common.Address]struct{} instead
+	// Guardian's public key hashes truncated by the ETH standard hashing mechanism (20 bytes).
+	Keys []common.Address
+
+	// quorum value for this set of keys
+	quorum int
```
johnsaigle, line 45: Worth adding a note that this must be a positive integer in practice and that this should be guaranteed via the EVM ABI bindings in the EVM watcher.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,159 @@
...
+		dgQuorum.WithLabelValues(label).Set(float64(cfg.Quorum()))
+	}
+	for chain := range d.Chains {
+		if _, ok := chains[chain]; !ok {
+			label := chain.String()
+			dgSigners.DeleteLabelValues(label)
+			dgQuorum.DeleteLabelValues(label)
+		}
+	}
+	defer d.mu.Unlock()
```
johnsaigle: Please move this to just below L122. We always want to Unlock when this function returns. In Go generally the deferred Unlock always goes right after the lock. If "something bad happens" between a Lock and Unlock, and the function were to return, then we would never Unlock. 

This is scenario should be impossible because there should be no control flow change in these for loops, but there's only downside to have the deferred unlock far away from the Lock.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,159 @@
...
+// NewDelegatedGuardianConfig returns a new DelegatedGuardianConfig.
+func NewDelegatedGuardianConfig() *DelegatedGuardianConfig {
+	return &DelegatedGuardianConfig{
+		Chains: map[vaa.ChainID]*DelegatedGuardianChainConfig{},
+	}
+}
+
+// Set sets the chains map
+func (d *DelegatedGuardianConfig) Set(chains map[vaa.ChainID]*DelegatedGuardianChainConfig) {
+	d.mu.Lock()
```
johnsaigle, line 150: Could you add `nil` checks for the `chains` values here? The function would end up looking like

```go
d.mu.Lock()
defer d.mu.Unlock()
// new nil loop
// the two loops you already have
```

johnsaigle: My understanding is that this is the only writer to `d` so it's crucial that we make it impossible for it to write a `nil` value for `cfg`. (Although we'd get a nil-deref for the `cfg.Keys` and `cfg.Quorum` calls below first).

node/pkg/adminrpc/adminserver.go
nvsriram: Done!

node/pkg/p2p/p2p.go
nvsriram: Yep, agreed.

node/pkg/processor/delegated_guardian_config.go
nvsriram: Done!

node/pkg/processor/delegated_guardian_config.go
nvsriram: Done!

node/pkg/processor/delegated_guardian_config.go
nvsriram: Done!

node/pkg/processor/delegated_guardian_config.go
nvsriram: Done!

node/pkg/processor/processor.go
nvsriram: Done — it's now sorted so iteration should be deterministic.

node/pkg/processor/processor.go
nvsriram: Done — I've added the `dgConfig.mu.Lock()`/`defer dgConfig.mu.Unlock()` so `chains` is guarded even if the code changes to becoming concurrent.

node/pkg/proto/gossip/v1/gossip.pb.go
```diff
@@ -1101,6 +1115,230 @@ func (x *Observation) GetMessageId() string {
...
+	// The address on the source chain that emitted this message.
+	EmitterAddress []byte `protobuf:"bytes,4,opt,name=emitter_address,json=emitterAddress,proto3" json:"emitter_address,omitempty"`
+	// The sequence number of this observation.
+	Sequence uint64 `protobuf:"varint,5,opt,name=sequence,proto3" json:"sequence,omitempty"`
+	// The consistency level requested by the emitter.
+	ConsistencyLevel uint32 `protobuf:"varint,6,opt,name=consistency_level,json=consistencyLevel,proto3" json:"consistency_level,omitempty"`
+	// The serialized payload.
+	Payload []byte `protobuf:"bytes,7,opt,name=payload,proto3" json:"payload,omitempty"`
+	// Transaction hash on the emitter chain in which the transfer was performed.
+	// Optional, included for observability.
```
johnsaigle: Is this optional? Now that we have the serialization checks when moving from MessagePublication <-> DelegateObservation, I think this field needs to be present or else it'll fail the length checks.

Also, I'm wondering if this ought to be `TxID` to match the MessagePublication's field? Basically, the MessagePublication used to cram every "transaction ID" into a 32 byte array to match what EVM does. However, this field is now runtime-dependent and they can put more or less arbitrary bytes inside of it.

node/pkg/processor/delegated_guardian_config.go
```diff
@@ -0,0 +1,168 @@
...
+			dgQuorum.DeleteLabelValues(label)
+		}
+	}
+	d.Chains = chains
+
+	return nil
+}
+
+// GetChainConfig returns the delegated guardian chain config for a specific chain, or nil if none.
+func (d *DelegatedGuardianConfig) GetChainConfig(chain vaa.ChainID) *DelegatedGuardianChainConfig {
```
johnsaigle: I believe that both of the `Get*` methods are returning only shallow copies of the `DelegatedGuardianChainConfig`, meaning that a caller who gets the return value can potentially modify the underlying data. 

I think the intention here is a getters/setters pattern and the `GetAll()` seems like it wants to make a deep copy rather than directly share the pointer to the underlying struct. If so, these methods can be improved by changing them to make deep copies of the underlying data.

If that's the case, you might also want to rename them to `Read*` instead of `Get*` to provide a little more clarity on the mutability of the data. If you choose not to, it might be worth a code comment explaining it at least.

johnsaigle: Similarly, you might want to make `Chains` private by renaming to `chains`, and then adding a method called e.g. `Chains()` or `ReadChains()` that provides a deep copy. This would be toward the same goal providing an API that limits mutability by the callers.

johnsaigle: And if you make any of the choices above, it might be worth adding unit tests to prove the mutability. (Sorry for all of the suggestions 😅 )

I just implemented a similar pattern in this PR in case that's helpful: https://github.com/wormhole-foundation/wormhole/pull/4657

node/pkg/processor/observation.go
```diff
@@ -72,8 +95,147 @@ func signaturesToVaaFormat(signatures map[common.Address][]byte, gsKeys []common
...
+// This is the main message processing loop. It is responsible for handling messages that are
+// received on the message channel. Depending on the configuration, a message may be processed
+// by the Notary, the Governor, and/or the Accountant.
+// This loop effectively causes each of these components to process messages in a modular
+// manner. The Notary, Governor, and Accountant can be enabled or disabled independently.
+// As a consequence of this loop, each of these components updates its internal state, tracking
+// whether a message is ready to be processed from its perspective. This state is used by the
+// processor to determine whether a message should be processed or not. This occurs elsewhere
+// in the processor code.
+// Returns error if accountant fails to process message
```
johnsaigle: Could you add a comment here similar to this one? https://github.com/wormhole-foundation/wormhole/blob/a6ac0eefd2d0f58fcdb4dcad7d9c3b7917857f50/node/pkg/governor/governor.go#L805-L806

Basically now that the processor has been refactored a little bit, we should warn that this function can have scary consequences.

node/pkg/processor/observation.go
```diff
@@ -360,3 +549,280 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(m *gossipv1.SignedVAAWithQu
...
+			zap.Uint32("emitter_chain", d.EmitterChain),
+			zap.String("emitter_address", hex.EncodeToString(d.EmitterAddress)),
+			zap.Uint64("sequence", d.Sequence),
+			zap.String("guardian_addr", addr.Hex()),
+		)
+		delegateObservationsFailedTotal.WithLabelValues("no_delegate_chain_config").Inc()
+		return nil
+	}
+
+	_, ok := cfg.KeyIndex(p.ourAddr)
```
johnsaigle, line 684: I think it's worth adding a `// SECURITY:` comment here. This bit of code sort of fades into the background while reading the function, but it's actually crucial that if we're a Delegated Guardian that we do not process observations from other Guardians.

node/pkg/processor/observation.go
```diff
@@ -360,3 +549,280 @@ func (p *Processor) handleInboundSignedVAAWithQuorum(m *gossipv1.SignedVAAWithQu
 
 	p.storeSignedVAA(v)
 }
+
+// handleSignedDelegateObservation processes a signed delegate observation
+//
+// SECURITY: This function assumes the p2p layer has already verified the SignedDelegateObservation signature and hence does not re-verify it.
```
johnsaigle, line 558: Optional extra comment here, something to the effect of "This method must only proceed to process observations if the current Guardian is a canonical Guardian for this chain. Otherwise, the Guardian must process messages using its own watcher."

node/pkg/watchers/evm/watcher.go
```diff
@@ -730,6 +767,83 @@ func fetchCurrentGuardianSet(ctx context.Context, ethConn connectors.Connector)
 	return currentIndex, &gs, nil
 }
 
+func (w *Watcher) fetchAndUpdateDelegatedGuardianConfig(
+	logger *zap.Logger,
```
johnsaigle: `w.logger` is available, so this argument can be deleted. (Not a big deal though.)

node/pkg/proto/gossip/v1/gossip.pb.go
evan-gray: Just clarifying this file is generated and comment should apply to the protobuf file

nvsriram: I've removed the `optional` comment.

Regarding the name, `tx_hash` was adopted from [the Observation gossip message](https://github.com/wormhole-foundation/wormhole/blob/main/proto/gossip/v1/gossip.proto#L212). But if you feel it's better to rename it to `tx_id`, I can make that change.

johnsaigle: Ah OK. I think that one is probably out of date too.

nvsriram: Yeah, I think all gossip messages ([`ObservationRequest`](https://github.com/wormhole-foundation/wormhole/blob/main/proto/gossip/v1/gossip.proto#L99), [`ChainGovernorStatus`](https://github.com/wormhole-foundation/wormhole/blob/main/proto/gossip/v1/gossip.proto#L150), and `Observation`) all use the outdated `tx_hash` as the name. I think it would be better to just rename all of those in a separate PR as I wouldn't want to touch the other message types.

node/pkg/processor/observation.go
nvsriram: Done!

node/pkg/processor/observation.go
nvsriram: Done!

node/pkg/processor/observation.go
nvsriram: Done!

node/pkg/proto/gossip/v1/gossip.pb.go
johnsaigle: Sounds good!

node/pkg/processor/delegated_guardian_config.go
nvsriram: The `GetChainConfig` and `GetAll` methods of the `DelegatedGuardianConfig` were made to match how the `GuardianSet` currently does it via its own [`Get`](https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/common/guardianset.go#L142) and [`GetAll`](https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/common/guardianset.go#L186) methods that return a pointer to the struct or create a nested map of pointers respectively.

However, I agree that this forces the caller to abide by an unsaid "Do not modify after Get" rule so will rename to `Read*` and make the changes.
Having said that, it sounds like a good idea to update `GuardianSet` to also follow the same pattern.

nvsriram: Done, modified the `Get*` helpers to `Read*` and updated the unit tests for the same.
Regarding making `Chains` private, I've added a TODO since that is not as straightforward a change.

node/pkg/p2p/p2p.go
```diff
@@ -948,6 +1016,96 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+								break
+							}
+
+							// SECURITY: The p2p signature on SignedDelegateObservation is verified in processSignedDelegateObservation. The processor
+							// assumes this verification has already occurred. Do not bypass this check, or untrusted data may reach the processor.
+							_, err := processSignedDelegateObservation(d, gs)
+							if err != nil {
+								p2pMessagesReceived.WithLabelValues("invalid_signed_delegate_observation").Inc()
+								if logger.Level().Enabled(zapcore.DebugLevel) {
+									logger.Error("invalid signed delegate observation received",
```
evan-gray: just confirming that the intention here is to only log an error if the log level is debug?

johnsaigle (approved):
re-approve after log change

ethereum/contracts/delegated_guardians/WormholeDelegatedGuardians.sol
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  function getHistoricalConfigLength(uint16 _chainId) public view returns (uint256) {
+    return delegatedGuardianSets[_chainId].length;
+  }
+
+  function getHistoricalConfig(uint16 _chainId, uint256 _index) public view returns (DelegatedGuardianSet memory) {
+    return delegatedGuardianSets[_chainId][_index];
+  }
+
```
kcsongor, line 115: a small wrinkle here is that, since we don't guarantee/require that each chain gets a new config entry in each config bump, the number of elements between chain configs will differ. This also means that the "index" here local to this particular chain's configs, and the event we emit when the config is updated includes a different index (the "global" one).
Since these functions are only used in tests (and in particular not the watcher), this does not seem to be an issue on any critical path, just wanted to call out

fergarrui: Yeah, historical as you mentioned, is not used. We always append to `delegatedGuardianSets` and retrieve the latest entry in `getConfig`. The index in `getHistoricalConfig` is supposed to be used with `getHistoricalConfigLength(_chainId)` in order to loop over previous updates if needed, but it is unrelated to the global `nextConfigIndex`.

fergarrui: I'm resolving so we can merge! Thanks!

nonergodic (approved):
rubberstamp

PR #4635, opened 2026-01-11
feat(node): Manager Service and Dogecoin signing

evan-gray:
This PR add a new service as described in whitepapers/0016_manager_service.md. This broadly achieves the following.
- uses a hardcoded 5-7 manager signer set index 0 for devnet
- uses hardcoded emitters to forward to the signer
- wires up a new "manager" service in the guardian
- that manager service subscribes to VAAs (either only those generated by this guardian or those from any guardian, if --subscribeToVAAs is enabled), parses them, and signs each input according to the spec
- it then gossips the signatures to other manager service enabled guardians
- it aggregates all of the signatures and makes them available at a new REST endpoint (currently it just stores them in an in-memory map, but one step at a time)

This PR adds two new guardian parameters:
- `--managerServiceEnabled` indicates that the manager service should be run and this guardian should subscribe to the manager gossip channel. This defaults to false.
- `--dogecoinManagerSignerUri` works identically to `guardianSignerUri`. Use like `file:///tmp/dogecoin.key`.

It also adds governance generation for `DelegatedManagerSetUpdate` to the guardian and CLI. Usage examples:
```bash
# Using component method (for secp256k1 multisig)
guardiand template delegated-manager-set-update \
  --manager-chain-id 65 \
  --manager-set-index 1 \
  --threshold 2 \
  --num-keys 3 \
  --public-keys "02...,03...,02..."
# Using raw manager set bytes
guardiand template delegated-manager-set-update \
  --manager-chain-id 65 \
  --manager-set-index 1 \
  --manager-set "010203..."
```

The generated prototxt can then be used with `guardiand admin governance-vaa-inject --socket /path/to/admin.sock governance.prototxt`.

Finally, it includes contracts to verify and store the `DelegatedManagerSetUpdate` governances. It has been deployed to Sepolia [here](https://sepolia.etherscan.io/address/0x086a699900262d829512299abe07648870000dd1#readContract).

I was able to test [this](https://doge-testnet-explorer.qed.me/tx/527e5225b9751111af110f21c104f7dbec1308629e937b43e2ca96e7323fa836) dogecoin testnet deposit, spin up tilt with `tilt up -- --num=7 --manager_service`, emit a VAA according to the spec and viola! 7 signatures which, along with the VAA, I can construct into a dogecoin transaction and unlock the funds like [this](https://doge-testnet-explorer.qed.me/tx/28d04cb6abc684f796a4c4abe8fc41d923349e083ae830a477bdebed1c77755a)!

Future steps:
- It should be considered if a guardian could configure multiple signing URIs because it is possible that a delegated manager set may transition key types and both keys would need to be supported for a period of time while integrators migrate, such as multisig switching to threshold. I expect this could be done by passing a comma separated list of URIs to the same argument, keeping it backwards compatible.

Reviewer notes:
- There are two different protobuf responses because I got the following error in CI
```
#15 0.946 proto/publicrpc/v1/publicrpc.proto:142:3:"publicrpc.v1.GetSignedManagerTransactionResponse" is used as the request or response type for multiple RPCs.
#15 0.946 proto/publicrpc/v1/publicrpc.proto:147:3:"publicrpc.v1.GetSignedManagerTransactionResponse" is used as the request or response type for multiple RPCs.
#15 0.946 proto/publicrpc/v1/publicrpc.proto:147:92:RPC response type "GetSignedManagerTransactionResponse" should be named "GetSignedManagerTransactionByHashResponse" or "PublicRPCServiceGetSignedManagerTransactionByHashResponse".
```
- Please carefully review the payloads, nosec comments, service start up, channel usage, signer usage, and public API.
- The scripts I used for testing are in #4645 
- If it's preferred, I am willing to split this PR into separate PRs per code area, but I felt it was easiest to have the full context of the required components.

### DIFF
--- a/sdk/vaa/payloads_test.go
+++ b/sdk/vaa/payloads_test.go
@@ -1053,11 +1053,11 @@ func TestDeserializeUTXOUnlockPayloadErrors(t *testing.T) {
 func TestDeserializeUTXOUnlockPayload_OversizedInputCount(t *testing.T) {
 	// Build a minimal valid header: prefix (4) + chain (2) + manager_set (4) + lenInput (4) + lenOutput (4)
 	buf := make([]byte, 18)
-	copy(buf[0:4], UTXOPayloadPrefix[:])                // prefix
-	binary.BigEndian.PutUint16(buf[4:6], uint16(65))    // destination chain
-	binary.BigEndian.PutUint32(buf[6:10], 1)            // manager set index
-	binary.BigEndian.PutUint32(buf[10:14], 0xFFFFFFFF)  // lenInput = max uint32
-	binary.BigEndian.PutUint32(buf[14:18], 0)           // lenOutput = 0
+	copy(buf[0:4], UTXOPayloadPrefix[:])               // prefix
+	binary.BigEndian.PutUint16(buf[4:6], uint16(65))   // destination chain
+	binary.BigEndian.PutUint32(buf[6:10], 1)           // manager set index
+	binary.BigEndian.PutUint32(buf[10:14], 0xFFFFFFFF) // lenInput = max uint32
+	binary.BigEndian.PutUint32(buf[14:18], 0)          // lenOutput = 0
 
 	_, err := DeserializeUTXOUnlockPayload(buf)
 	require.ErrorContains(t, err, "UTXO input count")


node/pkg/manager/manager.go
```diff
@@ -0,0 +1,688 @@
...
+	DestinationChain vaa.ChainID
+	// ManagerSetIndex is the delegated manager set index from the payload.
+	ManagerSetIndex uint32
+	// Required is the M value (number of signatures needed).
+	Required uint8
+	// Total is the N value (total number of possible signers).
+	Total uint8
+	// Signatures maps signer index to their signatures.
+	// Each entry contains the per-input signatures from that signer.
+	Signatures map[uint8][][]byte
```
kcsongor: we might want to think a bit more about this, since the signature-per-input stuff seems utxo specific. I guess for non-utxo use cases we can just treat the tx as having a single input, but I'm worried some other chain will come along that requires another schema. so maybe `Signatures` is just `[]bytes` with a chain-specific serialisation format?

node/pkg/manager/manager.go
```diff
@@ -0,0 +1,688 @@
...
+		}
+	}
+}
+
+// handleVAA processes an incoming signed VAA.
+func (c *ManagerService) handleVAA(v *vaa.VAA) {
+	// SECURITY: this channel should only be pushed to by a process that has verified the signatures on the VAA to belong to the current guardian set
+
+	// Check if this VAA is from a known manager emitter
+	if !c.isKnownEmitter(v.EmitterChain, v.EmitterAddress) {
```
kcsongor, line 208: IIUIC this is a "soft" registration check. what exactly is its purpose?

- there is no safety impact for doge, because the utxo redeem scripts commit to the emitter chain/address, so a spoofed transfer will simply attempt to spend a non-existent script hash
- for other chains that don't have such commitment mechanism, this check is not sufficient, because any known emitter could then control any managed address

so is this just a defensive sanity check? I think it will get in the way if we want to support dynamic chain topologies such as NTT deployments. In that case, we will likely need a way to recover the registration information on the fly

node/pkg/manager/manager.go
```diff
@@ -0,0 +1,688 @@
...
+	sig := &ManagerSignature{
+		VAAHash:          tx.VaaHash,
+		VAAID:            tx.VaaId,
+		DestinationChain: destChain,
+		ManagerSetIndex:  tx.ManagerSetIndex,
+		SignerIndex:      uint8(tx.SignerIndex), // #nosec G115 -- validated above: tx.SignerIndex < managerSet.N (uint8)
+		InputSignatures:  tx.Signatures,
+	}
+
+	c.storeSignature(sig)
```
kcsongor, line 349: we should probably do some validation here before accepting the signature

node/pkg/manager/manager.go
```diff
@@ -0,0 +1,688 @@
...
+
+	// VAA is from a known manager emitter, process it
+	c.logger.Info("received VAA from manager emitter",
+		zap.String("message_id", v.MessageID()),
+		zap.Stringer("emitter_chain", v.EmitterChain),
+		zap.Uint64("sequence", v.Sequence),
+	)
+
+	// Parse the UTXO unlock payload
+	payload, err := vaa.DeserializeUTXOUnlockPayload(v.Payload)
```
kcsongor, line 227: I think in order to generalise this module (so it works with non-utxo chains etc), we should only do partial parsing of the payload here. maybe we say every managed payload is 4 bytes prefix + destination chain id + delegated manager set index + payload? Then we just parse the known prefix here to route it to the correct place (so `signTransaction` actually does the chain-specific parsing on top of the chain-specific signature generation it already performs)

node/pkg/manager/manager.go
evan-gray: I'm on the fence here as, while I don't know all the future requirements, assuming they are arrays of signatures allows us to pack any chain specific information in bytes in already. This just simplifies the case when they are arrays of signatures and eases any case that necessitates one or more series of information.

node/pkg/manager/manager.go
evan-gray: Concur, there is no safety impact to the Dogecoin case given the emitter chain and address requirements in the redeem script. This was done as a defense-in-depth / DoS prevention to intentionally limit the scope of unknown emitters incidentally triggering the signing logic. In the future, this could be moved to a permissionless on-chain registration, but I figured that falls outside the scope of this already large feature and is not needed for initial production use cases.

I will add a note to that effect.

node/pkg/manager/manager.go
evan-gray: yes, for the MVP (in a future commit to this PR) I plan to go back and have the gossip message signed by the guardian key and have that be verified before pushing to this channel.

node/pkg/manager/manager.go
evan-gray: concur, if I get to it I will refactor here, otherwise we can definitely achieve that in a follow up PR.

evan-gray: left a comment for whoever implements the next chain

PR #4644, opened 2026-01-21
relayer: rm

evan-gray:
This PR removes all deprecated generic relayer related code that is related to development and CI but not guardian or CLI.

Note: with the relayer contracts no longer deployed, the NTT accountant test for relayer transfers is now skipped.


wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
```diff
@@ -1475,7 +1475,8 @@ describe("NTT Global Accountant Tests", () => {
         ).rejects.toThrow();
       }
     });
-    test("k. Relayed message gets accounted", async () => {
+    // NOTE: the relayer contract is no longer deployed in devnet
+    test.skip("k. Relayed message gets accounted", async () => {
```
panoel, line 1479: Should we just remove this test, then?

panoel (dismissed):

nonergodic (approved):
lgtm

johnsaigle (dismissed):

wormchain/contracts/tools/__tests__/test_ntt_accountant.ts
evan-gray: I was on the fence about removal since we could technically re-add support if folks were concerned about the coverage. Would remove in a follow-up PR if that's the consensus.

devnet/eth-devnet.yaml
```diff
@@ -57,12 +57,11 @@ spec:
           command:
             - /bin/sh
             - -c
-            - "cd ../../ethereum && CHAIN_ID=2 NETWORK=devnet ./sh/devnetInitialization.sh && cd ../relayer/ethereum && npm run deploy-relayers-evm1 && nc -lkn 2000"
```
johnsaigle, line 60: Is the `cd` command not necessary here and in eth-devnet2 anymore? Seems like the working directory is different here.

johnsaigle (approved):

panoel (approved):

devnet/eth-devnet.yaml
djb15: I also don't understand why this would work now without the directory traversal?

evan-gray: correct, the WORKDIR in `ethereum/Dockerfile` is now the ethereum folder instead of the relayer one

djb15 (approved):

kcsongor (approved):

gator-boi (approved):

PR #4652, opened 2026-01-30
meta: Add guidelines for writing safe code

johnsaigle:
Specific discussion points:
- Should this be longer or shorter?
- Any points of disagreement?
- Should we include code examples or leave this to be more general across different programming languages?


SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,244 @@
...
+
+These principles apply to all Wormhole code, with special emphasis on Guardian nodes and core protocol implementation. They favor **boredom over beauty**, **predictability over cleverness**, and **testability over conciseness**.
+
+---
+
+## Core Principles
+
+### Parse, Don't Verify
+
+Transform untrusted data into constrained types rather than checking and discarding the proof. Once parsed, the type system enforces your constraints automatically. Invalid states become unrepresentable. Use the compiler as your ally—let it prevent misuse at compile time rather than catching it at runtime.
```
djb15:
```suggestion
Transform untrusted data into constrained types rather than checking and discarding the proof. Once parsed, the type system enforces your constraints automatically. Invalid states become unrepresentable. Use the compiler as your ally — let it prevent misuse at compile time rather than catching it at runtime.
```

SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Simplify Conditionals
+
+Split compound boolean expressions into simple, nested conditions. Avoid complex `if` conditions with multiple `&&` or `||` operators. Each branch should be obvious. Each case should be explicitly handled.
+
+State invariants positively rather than negatively. `if index < length` is clearer than `if !(index >= length)`.
+
+### Centralize Control Flow
+
+When splitting large functions, keep all branching logic (if/switch/match) in the parent function. Push child functions toward pure computation without branching. This separates decision-making from execution and makes both easier to understand and test.
```
djb15:
```suggestion
When splitting large functions, keep all branching logic (`if/switch/match`) in the parent function. Push child functions toward pure computation without branching. This separates decision-making from execution and makes both easier to understand and test.
```

djb15 (commented):
This is really detailed already which I think is great, thanks for putting this together. Couple of more concrete comments:

- The risk/downside of the verbosity is that it's easy for contributors to skip/compromise on some of these or simply forget a large chunk of the list. Consider making a couple of these "non-negotiables" or add more emphasis to a few? Doing 3/4 of these things would be better than 0.
- Have you considered adding some short examples for those principles that are relevant? I would suggest using Golang examples given this guide is especially relevant for Guardian contributions

johnsaigle:
@djb15 I made some changes to try to keep only the more important things. Are there specific points you think are of most importance here? Happy to rearrange but I already thought this was the concise version of what I wanted to say 😆 😭 

In terms of code examples, I was trying to write something pretty general that code apply beyond just Go in the Guardian code. With that in mind, are you in favour of adding only Go-specific examples or should we omit them?

bemic (commented):
It is well written and the points are easy to understand. I can imagine that code examples in the style of "Do this:", "Not this:" could clarify it even more, but I don't think it's necessary. It's easy to get carried away and make it twice as long with all the good coding practices we know. This is a good balance of readability and information 👍

djb15:
> @djb15 I made some changes to try to keep only the more important things. Are there specific points you think are of most importance here? Happy to rearrange but I already thought this was the concise version of what I wanted to say 😆 😭
> 
> In terms of code examples, I was trying to write something pretty general that code apply beyond just Go in the Guardian code. With that in mind, are you in favour of adding only Go-specific examples or should we omit them?

Thanks John! I would add `Pin Exact Versions` back but appreciate you trying to cull a little bit. It's hard because I agree these are all important, just want to avoid information overload for readers (although maybe it's just LLMs we're training anyway 😄 ).

Fair point from both you @bemic, let's keep the examples out for brevity.

SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,217 @@
+# Safety-Critical Code Standards
+
+This document outlines coding principles for Wormhole's mission-critical infrastructure. These are aspirational standards representing where we're heading through continuous improvement and preventing regression.
```
SEJeff:
```suggestion
This document outlines coding principles for Wormhole's mission-critical infrastructure. These are aspirational standards representing where we're heading through continuous improvement and preventing regressions.
```

SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Make Invalid States Unrepresentable
+
+Design data structures so that illegal states cannot be constructed. Use sum types (enums/unions) instead of boolean flags. Use bounded types instead of magic sentinel values. If a value can only exist in valid states, you cannot accidentally create an invalid one. The compiler prevents entire classes of bugs.
+
+### Document Assumptions and Invariants
+
+Every function must explicitly document its preconditions, postconditions, and invariants. Assert all preconditions at function entry. Assert all postconditions before return. Assert invariants at boundaries where they matter. Comments should explain *why*, not *what*—the code already shows what.
+
+Use `REQUIRES:` and `ENSURES:` in function documentation to specify preconditions and postconditions. Add inline comments explaining non-obvious design decisions.
```
SEJeff: Does it make sense to link to an example in real code for this? Might be a bit nice to see?

SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,217 @@
...
+
+**Collections must not grow unbounded.** Every map, slice, array, or buffer must have:
+- A maximum capacity enforced at runtime
+- Explicit eviction or cleanup policies when limits are reached
+- Alerting when approaching capacity thresholds
+- Clear behavior when full (reject, evict oldest, fail explicitly)
+
+### Minimize Variable Scope
+
+Declare variables at the smallest possible scope. Fewer variables in scope means fewer ways to misuse them. Calculate values when needed, not before. Minimize the gap between where data is validated and where it's used (place-of-check to place-of-use distance).
```
SEJeff, line 99: Does it make sense to mention immutable types e.g. constants here?

SAFETY_CRITICAL_CODE.md
```diff
@@ -0,0 +1,217 @@
...
+- Can we implement this in under 200 lines?
+- Is this dependency actively maintained?
+- Has it been security audited?
+- How many transitive dependencies does it pull in?
+
+---
+
+## Security Practices
+
+**Defense in Depth:** Multiple independent layers of validation are better than one. Validate early and often. Check at boundaries. Trust nothing from external sources.
```
SEJeff:
```suggestion
**Defense in Depth:** Multiple independent layers of validation are better than one. Validate early and often. Check at both sides of boundaries. Trust nothing from external sources.
```

maybe?

SEJeff (dismissed):
Some tiny nits, but excellent work here! This documentation should be pointed at as table stakes for building serious web3 systems.

djb15 (dismissed):

djb15 (approved):

bemic (approved):

PR #4656, opened 2026-02-02
node(governor): Add governed tokens

johnsaigle:


bemic (approved):
Thanks for taking care of this!

node/pkg/governor/manual_tokens.go
```diff
@@ -12,6 +12,8 @@ func manualTokenList() []TokenConfigEntry {
 		{Chain: 2, Addr: "000000000000000000000000cbB7C0000aB88B473b1f5aFd9ef808440eed33Bf", Symbol: "CBBTC", CoinGeckoId: "coinbase-wrapped-btc", Decimals: 8, Price: 90_313.28},
 		{Chain: 2, Addr: "000000000000000000000000320623b8E4fF03373931769A31Fc52A4E78B5d70", Symbol: "RSR", CoinGeckoId: "reserve-rights-token", Decimals: 18, Price: 0.002880},                   // https://www.coingecko.com/en/coins/reserve-rights-token
 		{Chain: 2, Addr: "0000000000000000000000002798b1cC5A993085E8A9D46e80499F1B63f42204", Symbol: "GWEI", CoinGeckoId: "ethgas-2", Decimals: 18, Price: 0.0265},                                // https://www.coingecko.com/en/coins/ethgas-2
+		{Chain: 2, Addr: "000000000000000000000000939069722D568B5498CcBA4356E800EaefEFD2a5", Symbol: "FNXAI", CoinGeckoId: "finanx-ai", Decimals: 18, Price: 0.1692},                              // https://www.coingecko.com/en/coins/finanx-ai
```
mdulin2, line 15: Do we ever check whether these manual token prices are out of date? I feel like we mostly set and forget these.

johnsaigle: That's correct.

The Guardian should update their prices with the live amount every 15m though and choose the greater of the two prices. I think the only risk there is that the Governor's price might be too high relative to CG, which congests the Governor's usage. The chains haven't been congested for a while though so I don't think there's much of an impact.

PR #4657, opened 2026-02-04
node(svm-watcher): Add type to validate message account data

johnsaigle:
- Add new MessageAcccountData type to validate message account data is the right size and has a valid prefix.
- Modify relevant function signatures to accept the new type instead of []byte.
- Modify the recent changes to the SVM watcher's unit test by using the new safe struct instead of manually building the message account data.


node/pkg/watchers/solana/client.go
```diff
@@ -1122,10 +1171,15 @@ func (s *SolanaWatcher) getLatestFinalizedBlockNumber() uint64 {
...
+	if messageAccountData == nil {
+		return nil, errors.New("message account data is nil")
+	}
+
 	prop := &MessagePublicationAccount{}
-	// Skip the b"msg" prefix
-	if err := borsh.Deserialize(prop, data[3:]); err != nil {
+	// Skip the prefix and deserialize the rest of the data.
+	// SECURITY: The MessageAccountData type ensures that the data is at least 3 bytes long.
+	if err := borsh.Deserialize(prop, messageAccountData.data[MessageAccountDataMinLength:]); err != nil {
```
djb15: I'm wondering whether we should stick with `3` or some other specified constant of value 3 here instead of `MessageAccountDataMinLength`. As the docstring for the variable declaration suggests, it's possible that we could increase `MessageAccountDataMinLength` to be more restrictive, but here we already want to just skip the first 3 bytes

node/pkg/watchers/solana/client_test.go
```diff
@@ -107,3 +109,116 @@ func TestIsPossibleWormholeMessageFail(t *testing.T) {
...
+		},
+		{
+			name:               "failure -- nil argument",
+			messageAccountData: nil,
+			want:               &MessagePublicationAccount{},
+			wantErr:            true,
+		},
+		{
+			name:               "failure -- no data following prefix (msg)",
+			messageAccountData: &MessageAccountData{[]byte("msg")},
```
djb15: What are the errors we're expecting? Can you expand the test to check for the specific error strings?

The new checks are for `length < 3` which these aren't, so it must be failing somewhere else in the existing logic?

node/pkg/watchers/solana/client.go
johnsaigle: Good point. In this case it makes sense to decouple the constants. I'll make another one that's called "PrefixLength"

node/pkg/watchers/solana/client_test.go
johnsaigle: Borsh fails to deserialize the data into the struct and give an error message to that effect. I could include that error string here if you want, though we'll need to update the tests if the `borsh-go` library changes.

djb15: No that's fine, maybe also just add another test case with <3 bytes but not nil

johnsaigle: I added detection for the error strings before you replied 😅 . I also added the unit test case you described.

djb15 (dismissed):

node/pkg/watchers/solana/client.go
```diff
@@ -205,6 +244,26 @@ const (
...
+// NewMessageAccountData creates a new MessageAccountData from the given data.
+// SECURITY:
+// 1. Data must be non-empty.
+// 2. Data must begin with b"msg" for regular message accounts or b"msu" for unreliable accounts (used by the shim contract).
+// The program must never parse account data as a message account if it is does not meet these requirements.
+func NewMessageAccountData(data []byte) *MessageAccountData {
+	if len(data) < MessageAccountDataMinLength {
+		return nil
+	}
+	if !bytes.HasPrefix(data, []byte(accountPrefixReliable)) && !bytes.HasPrefix(data, []byte(accountPrefixUnreliable)) {
```
mdulin2, line 258: I like the explicit parsing of the prefixes.

node/pkg/watchers/solana/client.go
```diff
@@ -968,38 +1029,42 @@ func (s *SolanaWatcher) processAccountSubscriptionData(_ context.Context, logger
...
 	}
 
-	data, err = base64.StdEncoding.DecodeString(value.Account.Data[0])
+	// SECURITY: Bounds check the Data slice to ensure it's not empty.
+	var accountDataBase64 string
+	if len(value.Account.Data) == 0 {
+		return nil
+	} else {
+		// The data is received as an array of base64 strings, but we only care about the first one.
+		accountDataBase64 = value.Account.Data[0]
```
mdulin2, line 1139: Why does this have more than one element?

mdulin2 (dismissed):

node/pkg/watchers/solana/client.go
johnsaigle: We're parsing the JSON as a slice of strings: https://github.com/wormhole-foundation/wormhole/blob/601307f295513dcff35de20446bad02029443664/node/pkg/watchers/solana/client.go#L123

I'm not totally sure if that's correct. Maybe it's possible that it's really just a single string and the watcher is not representing it correctly.

djb15 (dismissed):

node/pkg/watchers/solana/client.go
```diff
@@ -156,6 +170,31 @@ const (
...
+	//   consistency_level      u8        =  1
+	//   vaa_time               u32       =  4
+	//   vaa_signature_account  Pubkey    = 32
+	//   submission_time        u32       =  4
+	//   nonce                  u32       =  4
+	//   sequence               u64       =  8
+	//   emitter_chain          u16       =  2
+	//   emitter_address        [u8; 32]  = 32
+	//   payload_len            u32       =  4  (Borsh Vec length prefix)
+	//   payload                [u8; N]   =  N  (variable, 0 minimum)
```
pleasew8t: Wouldn't this more accurately be `[u8; payload_len]`?

node/pkg/watchers/solana/client.go
pleasew8t: According to https://solana.com/docs/rpc/websocket/programsubscribe it looks like Data is a 2-element array containing:

`[ "encoded data", "encoding type" ]`

The example in the docs:

```
"data": [
    "11116bv5nS2h3y12kD1yUKeMZvGcKLSjQgX6BeV7u1FrjeJcKfsHPXHRDEHrBesJhZyqnnq9qJeUuF7WHxiuLuL5twc38w2TXNLxnDbjmuR",
    "base58"
],
```

pleasew8t (dismissed):

djb15 (dismissed):

djb15 (dismissed):

johnsaigle:
@djb15 @pleasew8t @mdulin2  could you please re-review? the only changes since your previous approvals are the ones in the unit test file

node/pkg/watchers/solana/client_test.go
```diff
@@ -265,56 +269,64 @@ func TestParseMessagePublicationAccount(t *testing.T) {
...
+			errStr: errStringPrefix,
 		},
 		{
-			name:               "failure -- data too short",
-			messageAccountData: MessageAccountData{[]byte("ms")},
-			want:               &MessagePublicationAccount{},
-			errStr:             errStringPrefix,
+			name: "failure -- data too short",
+			messageAccountData: func(t *testing.T) MessageAccountData {
+				return MessageAccountData{[]byte("ms")}
```
djb15, line 287: Why not use `mustNewMessageAccountData` here too?

johnsaigle: I added some clarification in the latest commit

djb15 (dismissed):

mdulin2 (approved):

djb15 (approved):

PR #4676, opened 2026-02-17
EVM Watcher Security Improvements

mdulin2:
We support a large number of EVM chains. There are several invariants that we currently rely on that we should probably explicitly check for that we were not. Specifically, we assume that the EVM subscription will return only logs matching the filter. We assume that a failed Tx will not have any logs. New checks have been added below. 

- Tx Status. 
- Contract emitter. 
- Event type 
- LogRemoved is false.

### DIFF
--- a/node/pkg/txverifier/evm.go
+++ b/node/pkg/txverifier/evm.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	geth "github.com/ethereum/go-ethereum/core/types"
-	gethTypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 	"go.uber.org/zap"
 )
@@ -433,7 +432,7 @@ func (tv *TransferVerifier[evmClient, connector]) parseReceipt(
 		zap.String("txHash", receipt.TxHash.String()),
 	)
 
-	if receipt.Status != gethTypes.ReceiptStatusSuccessful {
+	if receipt.Status != geth.ReceiptStatusSuccessful {
 		return nil, errors.Join(
 			ErrInvalidReceiptArgument,
 			errors.New("non-success transaction status"),
--- a/node/pkg/txverifier/evm.go
+++ b/node/pkg/txverifier/evm.go
@@ -11,8 +11,8 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	geth "github.com/ethereum/go-ethereum/core/types"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
-
 	"go.uber.org/zap"
 )
 
@@ -433,7 +433,7 @@ func (tv *TransferVerifier[evmClient, connector]) parseReceipt(
 		zap.String("txHash", receipt.TxHash.String()),
 	)
 
-	if receipt.Status != 1 {
+	if receipt.Status != gethTypes.ReceiptStatusSuccessful {
 		return nil, errors.Join(
 			ErrInvalidReceiptArgument,
 			errors.New("non-success transaction status"),
--- a/node/pkg/watchers/evm/by_transaction.go
+++ b/node/pkg/watchers/evm/by_transaction.go
@@ -20,6 +20,27 @@ var (
 	LogMessagePublishedTopic = eth_common.HexToHash("0x6eb224fb001ed210e379b335e35efe88672a8ce935d981a6896b27ffdf52a3b2")
 )
 
+// isLogValid checks that a log entry was emitted by the expected contract,
+// has the expected LogMessagePublished event topic, and has not been removed
+// due to a chain reorganization. This is called from both the real-time
+// subscription path (postMessage) and the reobservation path
+// (MessageEventsForTransaction) to ensure consistent validation.
+func isLogValid(l types.Log, contract eth_common.Address) bool {
+	// SECURITY: Reject logs that have been flagged as removed due to a chain reorg.
+	if l.Removed {
+		return false
+	}
+	// SECURITY: Verify the log was produced by our contract.
+	if l.Address != contract {
+		return false
+	}
+	// SECURITY: Verify the event is LogMessagePublished.
+	if len(l.Topics) == 0 || l.Topics[0] != LogMessagePublishedTopic {
+		return false
+	}
+	return true
+}
+
 // MessageEventsForTransaction returns the lockup events for a given transaction.
 // Returns the block number and a list of MessagePublication events.
 func MessageEventsForTransaction(
@@ -35,6 +56,7 @@ func MessageEventsForTransaction(
 		return nil, 0, nil, fmt.Errorf("failed to get transaction receipt: %w", err)
 	}
 
+	// SECURITY
 	// Bail early when the transaction receipt status is anything other than
 	// 1 (success). In theory, this check isn't strictly necessary - a failed
 	// transaction cannot emit logs and will trigger neither subscription
@@ -61,12 +83,7 @@ func MessageEventsForTransaction(
 			continue
 		}
 
-		// SECURITY: Skip logs not produced by our contract.
-		if l.Address != contract {
-			continue
-		}
-
-		if l.Topics[0] != LogMessagePublishedTopic {
+		if !isLogValid(*l, contract) {
 			continue
 		}
 
@@ -80,7 +97,7 @@ func MessageEventsForTransaction(
 			Timestamp:        time.Unix(int64(blockTime), 0), // #nosec G115 -- This conversion is safe indefinitely
 			Nonce:            ev.Nonce,
 			Sequence:         ev.Sequence,
-			EmitterChain:     chainId,
+			EmitterChain:     chainId, // SECURITY: Hardcoded to watcher chain id
 			EmitterAddress:   PadAddress(ev.Sender),
 			Payload:          ev.Payload,
 			ConsistencyLevel: ev.ConsistencyLevel,
--- a/node/pkg/watchers/evm/by_transaction.go
+++ b/node/pkg/watchers/evm/by_transaction.go
@@ -10,6 +10,7 @@ import (
 	"github.com/certusone/wormhole/node/pkg/common"
 	eth_common "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 )
 
@@ -65,7 +66,7 @@ func MessageEventsForTransaction(
 	// However, relying on that invariant is brittle - we connect to a lot of
 	// EVM-compatible chains which might accidentally break this API contract
 	// and return logs for failed transactions. Check explicitly instead.
-	if receipt.Status != 1 {
+	if receipt.Status != gethTypes.ReceiptStatusSuccessful {
 		return nil, 0, nil, fmt.Errorf("non-success transaction status: %d", receipt.Status)
 	}
 
--- a/node/pkg/watchers/evm/by_transaction.go
+++ b/node/pkg/watchers/evm/by_transaction.go
@@ -30,7 +30,7 @@ func isLogValid(l types.Log, contract eth_common.Address) bool {
 	if l.Removed {
 		return false
 	}
-	// SECURITY: Verify the log was produced by our contract.
+	// SECURITY: Verify the log was produced by the supplied contract.
 	if l.Address != contract {
 		return false
 	}
--- a/node/pkg/watchers/evm/by_transaction.go
+++ b/node/pkg/watchers/evm/by_transaction.go
@@ -21,12 +21,12 @@ var (
 	LogMessagePublishedTopic = eth_common.HexToHash("0x6eb224fb001ed210e379b335e35efe88672a8ce935d981a6896b27ffdf52a3b2")
 )
 
-// isLogValid checks that a log entry was emitted by the expected contract,
+// isValidCoreBridgeMessagePublicationLog checks that a log entry was emitted by the expected contract,
 // has the expected LogMessagePublished event topic, and has not been removed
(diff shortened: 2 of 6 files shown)


node/pkg/watchers/evm/watcher.go
```diff
@@ -635,6 +639,46 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
+						logger.Error("log index is before the first log in the receipt",
+							zap.String("msgId", pLock.message.MessageIDString()),
+							zap.String("txHash", pLock.message.TxIDString()),
+						)
+						delete(w.pending, key)
+						continue
+					}
+
+					// O(1) lookup by using relative indexing into the logs
+					pos := pLock.logIndex - firstLogIndex
```
johnsaigle: Is `pLock.logIndex` guaranteed to be >= `firstLogIndex`? This could underflow otherwise.

node/pkg/watchers/evm/watcher.go
```diff
@@ -829,6 +885,31 @@ func (w *Watcher) postMessage(
...
+		queryLatency.WithLabelValues(w.networkName, "transaction_receipt").Observe(time.Since(msm).Seconds())
+		if err != nil {
+			w.logger.Error("failed to get transaction receipt for instant message",
+				zap.String("msgId", msg.MessageIDString()),
+				zap.String("txHash", msg.TxIDString()),
+				zap.Error(err),
+			)
+			return
+		}
+		if receipt.Status != 1 {
```
johnsaigle: Small point but is there a constant that we can read instead of `1`? e.g. is there like `TransactionSuccess` value somewhere?

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -20,6 +20,27 @@ var (
...
+// has the expected LogMessagePublished event topic, and has not been removed
+// due to a chain reorganization. This is called from both the real-time
+// subscription path (postMessage) and the reobservation path
+// (MessageEventsForTransaction) to ensure consistent validation.
+func isLogValid(l types.Log, contract eth_common.Address) bool {
+	// SECURITY: Reject logs that have been flagged as removed due to a chain reorg.
+	if l.Removed {
+		return false
+	}
+	// SECURITY: Verify the log was produced by our contract.
```
pleasew8t: Suggestion: `Verify the log was produced by the supplied contract.`

node/pkg/watchers/evm/connectors/batch_poller.go
```diff
@@ -174,6 +174,8 @@ func (b *BatchPollConnector) pollBlocks(ctx context.Context, sink chan<- *NewBlo
 	}
 
 	for idx, newBlock := range newBlocks {
+
+		// SECURITY: Assumptions about blocks finalization. If block X is CL, and X+2 is a CL, assumes that X+1 is a CL.
```
pleasew8t: Should this be "If block X is _a_ CL"?

pleasew8t: I'm not entirely sure what this is saying. Is CL consistency level in this context?

node/pkg/watchers/evm/watcher.go
pleasew8t: I think the check above this is fine to ensure there is no underflow, no?

node/pkg/watchers/evm/watcher.go
```diff
@@ -829,6 +885,31 @@ func (w *Watcher) postMessage(
 			zap.Uint8("ConsistencyLevel", ev.ConsistencyLevel),
 		)
 
+		// SECURITY: Defense-in-depth check of transaction status.
+		// The EVM invariant is that failed transactions cannot emit logs, but we verify
+		// explicitly in case an EVM-compatible chain breaks this contract.
```
pleasew8t: Suggestion: "breaks this _invariant_"

node/pkg/watchers/evm/connectors/batch_poller.go
johnsaigle: I'd also prefer if the comment was more verbose. Clarity is more important than brevity here.

node/pkg/watchers/evm/watcher.go
mdulin2: Yeah, exactly. There's already a check for that case directly above.

node/pkg/watchers/evm/watcher.go
mdulin2: Found the constant `ReceiptStatusSuccessful` and used this instead.

node/pkg/watchers/evm/watcher.go
```diff
@@ -941,6 +1007,31 @@ func (w *Watcher) postMessage(
...
+		receipt, err := w.ethConn.TransactionReceipt(receiptCtx, ev.Raw.TxHash)
+		receiptCancel()
+		queryLatency.WithLabelValues(w.networkName, "transaction_receipt").Observe(time.Since(msm).Seconds())
+		if err != nil {
+			w.logger.Error("failed to get transaction receipt for instant message",
+				zap.String("msgId", msg.MessageIDString()),
+				zap.String("txHash", msg.TxIDString()),
+				zap.Error(err),
+			)
+			return
```
djb15, line 981: In this case a reobservation request would be required right?

djb15 (dismissed):
I think the changes are fine, but it's worth running this on a Guardian in testnet first before wider rollout given the code changes are on the critical path

node/pkg/watchers/evm/watcher.go
```diff
@@ -672,6 +676,56 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 						continue
 					}
 
+					// SECURITY: Defense-in-depth validation of the finalized receipt log.
```
mdulin2: I've given this some more thought. 

Given that the blockhash is checked above, this doesn't feel necessary. If the blockhash changed then the events must have also changed. So, this would need to be reobserved anyway. Given the extra complexity and minimal security gain, I'm considering removing this part of the PR.

node/pkg/watchers/evm/watcher.go
mdulin2: No, this is the standard path provided by the subscription.

mdulin2: Note that this requires an additional RPC call to retrieve the receipt status. This was already being done in the non-instant path.

node/pkg/watchers/evm/watcher.go
mdulin2: I removed the code and just added a security comment indicating why it's okay.

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -20,6 +21,27 @@ var (
 	LogMessagePublishedTopic = eth_common.HexToHash("0x6eb224fb001ed210e379b335e35efe88672a8ce935d981a6896b27ffdf52a3b2")
 )
 
+// isLogValid checks that a log entry was emitted by the expected contract,
+// has the expected LogMessagePublished event topic, and has not been removed
+// due to a chain reorganization. This is called from both the real-time
+// subscription path (postMessage) and the reobservation path
+// (MessageEventsForTransaction) to ensure consistent validation.
+func isLogValid(l types.Log, contract eth_common.Address) bool {
```
johnsaigle:
```suggestion
func isValidCoreBridgeMessagePublicationLog(l types.Log, contract eth_common.Address) bool {
```

A more targeted name might be better here given that this method isn't doing general log validation. Instead, it's really specifically checking a particular contract for a particular event.

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -20,6 +21,27 @@ var (
...
+	// SECURITY: Reject logs that have been flagged as removed due to a chain reorg.
+	if l.Removed {
+		return false
+	}
+	// SECURITY: Verify the log was produced by the supplied contract.
+	if l.Address != contract {
```
johnsaigle, line 35: IMO it would be best to restructure both the function and `MessageEventsForTransaction` as methods on the watcher instead of standalone functions. I get that this is scope-creep though, so we don't have to do it now.

The motivation for that is:
- We could take advantage of `env` to validate the `contract` against the SDK's known core bridge emitters field
- We could use `watcher.contract` instead of a function argument here
- Both of these methods require watcher-related state information anyway

The eth hack script using `MessageEventsForTransaction` would need to be restructured a bit too though.

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -80,7 +98,7 @@ func MessageEventsForTransaction(
 			Timestamp:        time.Unix(int64(blockTime), 0), // #nosec G115 -- This conversion is safe indefinitely
 			Nonce:            ev.Nonce,
 			Sequence:         ev.Sequence,
-			EmitterChain:     chainId,
+			EmitterChain:     chainId, // SECURITY: Hardcoded to watcher chain id
```
johnsaigle, line 101: This would be another advantage of making this a method on the watcher and leveraging its `chainId` instance directly.

node/pkg/watchers/evm/custom_consistency_level.go
```diff
@@ -138,6 +138,7 @@ func (w *Watcher) cclEnable(ctx context.Context) error {
 
 // cclHandleMessage is called for new observations that have the consistency level set to custom handling.
 // It reads the configuration for the emitter and updates the `pendingMessage` object for custom handling.
+// SECURITY: This function SHOULD NOT modify the actual data signed in the VAA. Otherwise, replay protection is broken.
```
johnsaigle:
```suggestion
// SECURITY: This function MUST NOT modify the actual data signed in the VAA. Otherwise, replay protection is broken.
```

May be nice to follow RFC-related language here for extra clarity.
https://www.rfc-editor.org/rfc/rfc2119

> [4](https://www.rfc-editor.org/rfc/rfc2119#section-4). SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

> [1](https://www.rfc-editor.org/rfc/rfc2119#section-1). MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.

node/pkg/watchers/evm/watcher.go
```diff
@@ -568,6 +568,7 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 					}
 
 					// Don't process the observation if we haven't reached the desired block height yet.
+					// CCL 'additionalBlocks' after consistency Level handling.
```
johnsaigle: Can you turn this into a full sentence? I'm not really clear on what this is communicating.

node/pkg/watchers/evm/watcher.go
```diff
@@ -1012,6 +1057,7 @@ func canRetryGetBlockTime(err error) bool {
 // Modifies the verificationState field of the message as a side-effect.
 // Even if an invalid Transfer is detected, the message will still be published. It is the responsibility of the calling code to handle
 // a status of Rejected.
+// SECURITY: should be the only location where the watcher's msgC channel is written to.
```
johnsaigle:
```suggestion
// SECURITY: must be the only location where the watcher's msgC channel is written to.
```

node/pkg/watchers/evm/watcher.go
johnsaigle: > No, this is the standard path provided by the subscription.

I don't understand this part. To me it does look like that if the instant message lookup fails, we would need to manually reobserve it.

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -20,6 +21,27 @@ var (
 	LogMessagePublishedTopic = eth_common.HexToHash("0x6eb224fb001ed210e379b335e35efe88672a8ce935d981a6896b27ffdf52a3b2")
 )
 
+// isLogValid checks that a log entry was emitted by the expected contract,
+// has the expected LogMessagePublished event topic, and has not been removed
+// due to a chain reorganization. This is called from both the real-time
+// subscription path (postMessage) and the reobservation path
+// (MessageEventsForTransaction) to ensure consistent validation.
```
johnsaigle: Are the parentheses here meant to be links? If so, they need to be square brackets instead. This lets them render in HTML or else editors can use the links to navigate around the source code. https://go.dev/doc/comment#links


```suggestion
// [MessageEventsForTransaction] to ensure consistent validation.
```

If you don't care about the links, then the parentheses can be deleted.

johnsaigle (commented):
Left some suggestions. Overall the changes look good 👍🏻

node/pkg/txverifier/evm.go
```diff
@@ -11,8 +11,8 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	geth "github.com/ethereum/go-ethereum/core/types"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
```
pleasew8t: Why is `"github.com/ethereum/go-ethereum/core/types"` being imported twice? I see this in `node/pkg/watchers/evm/by_transaction.go` as well.

node/pkg/watchers/evm/by_transaction.go
mdulin2: I think this is scope creep and I'd like to leave this for another PR.

node/pkg/watchers/evm/by_transaction.go
mdulin2: I think it was more a break in thought than anything else. But, I'll remove the parens altogether then.

node/pkg/txverifier/evm.go
mdulin2: Yeah, good callout. I think I copied over some code from a different file, didn't have the import with the exact name so I reimported it. I'll fix that.

node/pkg/watchers/evm/watcher.go
mdulin2: I think I misunderstood Dirk's point. If there was an error in getting the receipt, we would need to reobserve it later.

johnsaigle (approved):

djb15 (approved):

PR #4701, opened 2026-03-06
ci: Upgrade golangci-lint version to latest version

johnsaigle:
Changes:

* channelcheck isn't a configured linter, so this modifies these instances into comments
* Adds exceptions for what look like attacker-controlled URLs, hard-coded API keys, etc.
* Fixes variable shadowing
 * Workarounds for context cancellation functions that appeared to not get called (e.g. via `defer`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.4
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.0
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.1
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.2
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.2
+        run: curl -sSfL https://golangci-lint.run/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.2
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.0
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.1
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.3
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.4
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.1
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.3
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.0
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.11.1
       - name: Run linters
         run: make generate && golangci-lint --version && ./scripts/lint.sh -g lint
       - name: Ensure generated proto matches
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -423,7 +423,9 @@ func runInjectGovernanceVAA(cmd *cobra.Command, args []string) {
 }
 
 func runFindMissingMessages(cmd *cobra.Command, args []string) {
-	chainId, err := vaa.StringToKnownChainID(args[0])
+	// Effectively allow any uint16 value for ChainID here. A Guardian may want to find missing messages
+	// for deprecated chains, so we don't use [`vaa.StringToKnownChainID`] here.
+	chainId, err := vaa.ChainIDFromString(args[0])
 	if err != nil {
 		log.Fatalf("invalid chain ID: %v", err)
 	}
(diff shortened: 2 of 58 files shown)


node/pkg/notary/admincommands.go
```diff
@@ -285,10 +286,10 @@ func encodePayloadBytes(payload *vaa.TransferPayloadHdr) []byte {
 	copy(bz[1+offset:33], amtBytes)
 
 	copy(bz[33:], payload.OriginAddress[:])
-	bz[65] = byte(payload.OriginChain)
+	binary.BigEndian.PutUint16(bz[65:67], uint16(payload.OriginChain))
```
djb15, line 289: Good catch!

djb15 (dismissed):

node/cmd/guardiand/adminnodes.go
```diff
@@ -179,15 +178,16 @@ func runListNodes(cmd *cobra.Command, args []string) {
 		truncAddrs := make(map[vaa.ChainID]string)
 		errors := map[vaa.ChainID]uint64{}
 		for _, n := range h.RawHeartbeat.Networks {
-			if n.Id > math.MaxUint16 {
-				log.Fatalf("heartbeat chain id is greater than MaxUint16: %v", n.Id)
+			chainId, err := vaa.KnownChainIDFromNumber[uint32](n.Id)
+			if err != nil {
+				log.Fatalf("invalid chain id in heartbeat: %v", err)
```
bemic, line 186: Can we preserve `n.Id` in the log? In other places where we use `KnownChainIDFromNumber` we do that as well.

node/cmd/ccq/query_server.go
```diff
@@ -285,7 +285,7 @@ func runQueryServer(cmd *cobra.Command, args []string) {
 	if permWatcherErr != nil {
 		// Cleanup p2p connection.
 		cancel()
-		logger.Fatal("Could not start permissions file watcher", zap.Error(err))
+		logger.Fatal("Could not start permissions file watcher", zap.Error(permWatcherErr))
```
bemic, line 288: Nice. I see it was logging some random old error before 👍

node/cmd/guardiand/adminnodes.go
johnsaigle: Hm it's in the error string already: https://github.com/wormhole-foundation/wormhole/blob/467b7512644aaeb4dc2b10ae0cfb8ad3266866dd/sdk/vaa/structs.go#L189

Do you think there's value in keeping it as a separate key in the log?

bemic: I didn't realize it is in the error. I'm fine with it as it is.

johnsaigle:
Converting this PR to draft mode until we figure out the sequencing order with #4841: this PR strips channelcheck references because the linter is not active, but that PR relies on adding new ones. That PR also assumes that golangci-lint v2.8.0 is in use, whereas this PR changes the version.

Because there is likely a risk of conflicts here, we should figure out a way to properly knit these changes together. cc @mdulin2

node/hack/evm_test/wstest.go
```diff
@@ -77,7 +77,7 @@ func main() {
 				case <-ctx.Done():
 					return
 				case err := <-headerSubscription.Err():
-					errC <- fmt.Errorf("block subscription failed: %w", err) //nolint:channelcheck // The watcher will exit anyway
+					errC <- fmt.Errorf("block subscription failed: %w", err) // Note on channel capacity: The watcher will exit anyway
```
mdulin2, line 80: Thanks for keeping these comments in some capacity. It'll make it easier for the channel check linter later.

node/hack/repair_solana/repair.go
```diff
@@ -220,6 +220,7 @@ func main() {
 						if err != nil {
 							panic(err)
 						}
+						// #nosec G704 -- Maintenance script calling public Wormhole RPC endpoint
```
mdulin2, line 223: This lint is really aggressive. It seems like every `.Do()` is considered an SSRF bug. I'd consider disabling for the amount of noise it's causing.

node/pkg/common/symmetric.go
mdulin2, line 10: Is this code used by anything? I'd consider deleting it if not.

node/pkg/manager/xrpl/transaction.go
```diff
@@ -229,17 +221,25 @@ func EncodeDERSignature(r, s []byte) []byte {
...
-	sig = append(sig, byte(sLen))                         // s length
-	sig = append(sig, s...)                               // s value
+	totalLen := der.FixedOverhead + rLen + sLen // 0x02 + rLen + r + 0x02 + sLen + s
+
+	// DER X.690 §8.1.3 — https://www.itu.int/rec/T-REC-X.690
+	// Short-form length encoding uses a single byte for lengths 0–127.
+	// The constants below reflect the secp256k1 physical maximums; this
+	// ensures that every length can be encoded in a single byte.
+	if totalLen > der.MaxTotalLen || rLen > der.Secp256k1MaxIntEncodedLen || sLen > der.Secp256k1MaxIntEncodedLen {
+		return nil
```
mdulin2, line 231: Will this error be handled correctly? There's `nil` case that I can see in the handlers.

mdulin2 (dismissed):

node/hack/repair_solana/repair.go
johnsaigle: I'm in favour of disabling for the hack scripts for sure.

node/pkg/common/symmetric.go
johnsaigle: I think the `symmetric` stuff is only used by `node/hack/encrypt/encrypt.go` and I don't have a good signal on who's using it. I agree we could likely delete it, but my preference would be to do it in a separate PR.

node/pkg/manager/xrpl/transaction.go
johnsaigle: I'm a bit confused by the second sentence. This PR also modifies the callers in `node/pkg/manager/manager.go`. Are there specific callsites that you're worried about?

pleasew8t (dismissed):

node/pkg/common/symmetric.go
johnsaigle: Deleted here: https://github.com/wormhole-foundation/wormhole/pull/4861

.github/workflows/build.yml
```diff
@@ -307,7 +307,7 @@ jobs:
       - name: Formatting checks
         run: ./scripts/lint.sh -l -g format
       - name: Install linters
-        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.0
+        run: curl -sSfL https://golangci-lint.run/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.12.2
```
evan-gray, line 310: fwiw: https://golangci-lint.run/docs/welcome/install/ci/#github-actions

but I did confirm this change matches https://golangci-lint.run/docs/welcome/install/ci/#github-actions

node/cmd/guardiand/adminclient.go
```diff
@@ -470,13 +466,10 @@ func runDumpVAAByMessageID(cmd *cobra.Command, args []string) {
 	if len(parts) != 3 {
 		log.Fatalf("invalid message ID: %s", args[0])
 	}
-	chainID, err := strconv.ParseUint(parts[0], 10, 32)
+	chainId, err := vaa.StringToKnownChainID(parts[0])
```
evan-gray: these changes to "known chain id" will be a functional change, no? what if they want to dump messages from a deprecated chain?

node/cmd/guardiand/adminnodes.go
```diff
@@ -179,15 +178,16 @@ func runListNodes(cmd *cobra.Command, args []string) {
 		truncAddrs := make(map[vaa.ChainID]string)
 		errors := map[vaa.ChainID]uint64{}
 		for _, n := range h.RawHeartbeat.Networks {
-			if n.Id > math.MaxUint16 {
-				log.Fatalf("heartbeat chain id is greater than MaxUint16: %v", n.Id)
+			chainId, err := vaa.KnownChainIDFromNumber[uint32](n.Id)
```
evan-gray: I'm not sure this is advisable. other guardians may be on different versions. I don't see a reason to not accept this as a u16.

node/pkg/common/symmetric.go
evan-gray, line 1: if this file is unused, perhaps it can be removed.

node/pkg/governor/governor_monitoring.go
```diff
@@ -515,13 +515,12 @@ func (gov *ChainGovernor) CollectMetrics(ctx context.Context, hb *gossipv1.Heart
 			continue
 		}
 
-		if n.Id > math.MaxUint16 {
-			gov.logger.Error("CollectMetrics: chain id is not a valid uint16", zap.Uint32("chain_id", n.Id))
+		chain, err := vaa.KnownChainIDFromNumber[uint32](n.Id)
```
evan-gray: is it possible for there to be pending messages loaded from a prior version of the guardian which may have contained a chain that has since been removed (or a newer version that was rolled back)? what is the importance of this being a "known" chain?

node/pkg/watchers/near/nearapi/mock/mock_server.go
```diff
@@ -157,9 +160,10 @@ func (s *ForwardingCachingServer) ServeHTTP(w http.ResponseWriter, req *http.Req
 		}
 
 		httpClient := http.Client{}
+		// #nosec G704 -- Test mock server making requests to configured upstream
 		resp, reqErr := httpClient.Do(proxyReq)
 		if reqErr != nil {
-			http.Error(w, reqErr.Error(), http.StatusBadGateway)
+			http.Error(w, err.Error(), http.StatusBadGateway)
```
evan-gray: why did this line change?

node/pkg/node/reobserve.go
```diff
@@ -63,15 +62,17 @@ func handleReobservationRequests(
 				}
 			}
 		case req := <-obsvReqC:
-			if req.ChainId > math.MaxUint16 {
-				logger.Error("chain id is larger than MaxUint16",
+			chainId, err := vaa.KnownChainIDFromNumber[uint32](req.ChainId)
```
evan-gray, line 65: I think this one should be okay, along with the watcher specific checks, because I think all the watchers in node.go are set up with known chain IDs (e.g. `vaa.chainIDEthereum`)

evan-gray (commented):
I tried to only leave comments on the `KnownChainIDFromNumber` that I felt could potentially be problematic, though in general I would lean against needlessly restricting `u16`s into "known" chains when they could otherwise / just as safely be treated as generic `u16`s per the spec.

.github/workflows/build.yml
johnsaigle: Thanks, happy to investigate using a more direct GitHub Action approach in a future PR

node/pkg/common/symmetric.go
johnsaigle: tracking here: https://github.com/wormhole-foundation/wormhole/pull/4861

node/pkg/watchers/near/nearapi/mock/mock_server.go
johnsaigle: I'll take a look, might've been a rebasing issue

node/cmd/guardiand/adminclient.go
johnsaigle: I wasn't aware that was a use case, fair point. We can use `ChainIDFromNumber` (same thing but without the "known" check) in this case.

node/cmd/guardiand/adminnodes.go
johnsaigle: Fair, I'll change this to use ChainIDFromNumber

node/pkg/governor/governor_monitoring.go
johnsaigle: I'm not sure that it's meaningful for the Governor to have pending data for a deprecated chain generally, since it needs to be tracked in `node/pkg/governor/mainnet_chains.go` to have any effect. In this particular case, if the chain is truly deprecated (in the sense of having its ChainID commented out in the SDK), then the `if exists {` block shouldn't execute, which I think would produce a heartbeat with info where the chain ID would appear as `unset`.

That said this case is pretty minor so if you want I can change this check to not use the `Known` check.

node/pkg/watchers/near/nearapi/mock/mock_server.go
johnsaigle: Reverted in latest commit 👍🏻

node/pkg/governor/governor_monitoring.go
johnsaigle: Changed to skip Known check in latest commit.

node/cmd/guardiand/adminnodes.go
johnsaigle: Changed to skip Known check in latest commit.

node/cmd/guardiand/adminclient.go
johnsaigle: Changed to skip Known check in latest commit.

evan-gray (dismissed):

mdulin2 (dismissed):

evan-gray (dismissed):

djb15 (approved):

evan-gray (approved):

PR #4705, opened 2026-03-11
XRPL Watcher and Manager Service support

evan-gray:
draft awaiting further changes in https://github.com/wormholelabs-xyz/wormhole/pull/16


node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+	}
+
+	return &common.MessagePublication{
+		TxID:             txHash,
+		Timestamp:        tx.Timestamp,
+		Nonce:            coreMemo.nonce,
+		Sequence:         sequence,
+		EmitterChain:     vaa.ChainIDXRPL,
+		EmitterAddress:   sender,
+		Payload:          coreMemo.payload,
```
mdulin2, line 401: Could we add a `prefix` to this, like `Core`? I think this just makes future attacks harder to pull off. The audit also pointed this out.

mdulin2 (dismissed):

bemic (commented):
Once we have the changes from the other PR we are good to go

node/pkg/watchers/xrpl/parse.go
evan-gray: I don't think so, that would modify the payload from whatever an integrator emitted on chain, which goes against [the spec](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0001_generic_message_passing.md#detailed-design) and every other implementation. what is the concern you want to mitigate? If it is the other automated messages, those should instead use a calculated emitter like NTT in `calculateEmitterAddress`.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+	common.RunWithScissors(ctx, errC, "xrpl_fetch_obvs_req", func(ctx context.Context) error {
+		for {
+			select {
+			case err := <-errC:
+				logger.Error("xrpl_fetch_obvs_req died", zap.Error(err))
+				return fmt.Errorf("xrpl_fetch_obvs_req died: %w", err)
+			case <-ctx.Done():
+				return ctx.Err()
+			case req := <-w.obsvReqC:
+				if req.ChainId > math.MaxUint16 {
```
johnsaigle, line 222: The validation here could be done with the SDK's Known chain ID method instead.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+						zap.Uint32("chainID", req.ChainId),
+						zap.String("txID", hex.EncodeToString(req.TxHash)),
+					)
+					continue
+				}
+				if vaa.ChainID(req.ChainId) != vaa.ChainIDXRPL {
+					panic("invalid chain ID")
+				}
+
+				txHash := req.TxHash
```
johnsaigle, line 233: Is there a known minimum length for this value? We could avoid doing the websocket request for bad lengths here if so.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+			Help: "Current XRPL validated ledger index",
+		})
+	xrplTxDropped = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Name: "wormhole_xrpl_transactions_dropped_total",
+			Help: "Total number of XRPL transactions dropped due to full channel",
+		})
+)
+
+type Watcher struct {
```
johnsaigle, line 68: Is there a reason that this watcher doesn't have a logger? I think all of the other ones do. Here it looks like we use the supervisor's logger all the time.

If there's only one logger in use by the watcher, the methods could be simplified so that they don't need to take a pointer to a logger. Instead, they could just use `w.logger` for example

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		}
+	}
+
+	if len(ticketSequences) == 0 {
+		return nil, fmt.Errorf("TicketCreate transaction has no created Ticket entries")
+	}
+
+	// Find the minimum ticket sequence (ticket_start) and count
+	ticketStart := ticketSequences[0]
+	for _, seq := range ticketSequences[1:] {
```
johnsaigle: It looks like the `ticketSequences` slice needs to be at least length 2 here. Otherwise this could panic with an out of bounds slice index.

Maybe the check on L531 could assert that length is 2? (Or even that it's > 0 and an even number? assuming that theses sequences are always ranges of values. I'm not familiar with the underlying chain)

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+	return &common.MessagePublication{
+		TxID:             txHash,
+		Timestamp:        tx.Timestamp,
+		Nonce:            0,
+		Sequence:         sequence,
+		EmitterChain:     vaa.ChainIDXRPL,
+		EmitterAddress:   emitterAddress,
+		Payload:          payload,
+		ConsistencyLevel: 0,
+		IsReobservation:  false,
```
johnsaigle, line 581: What about Unreliable? I think the linter should flag this automatically but I guess not.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+
+// MPTAssetScaleFetcher is a function type for fetching MPT AssetScale from the ledger.
+// This allows the parsing logic to be tested without network calls.
+type MPTAssetScaleFetcher func(mptIssuanceID string) (uint8, error)
+
+// Parser handles parsing of XRPL NTT transactions.
+// It is stateless except for the MPT asset scale fetcher, which can be injected for testing.
+type Parser struct {
+	coreAccount        string          // Core Wormhole manager account — payments to this account are not NTT
+	managedAccounts    map[string]bool // Managed accounts (NTT accounts) — TicketCreate on these emits XTCF
```
johnsaigle: This is probably better written as a set (e.g. `managedAccounts map[string]struct{}`) unless there is a valid `false` case for a managed account.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		scaledAmount,
+		tokenInfo.sourceToken,
+		memo.recipientAddress,
+		memo.recipientChain,
+	)
+
+	return &common.MessagePublication{
+		TxID:      txHash,
+		Timestamp: tx.Timestamp,
+		Nonce:     0, // NTT payloads do not include a nonce
```
johnsaigle, line 299: What about other messages types? They also seem to always get 0.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		seqRaw, ok := node.CreatedNode.NewFields["TicketSequence"]
+		if !ok {
+			continue
+		}
+		// JSON numbers deserialize as float64
+		switch v := seqRaw.(type) {
+		case float64:
+			ticketSequences = append(ticketSequences, uint64(v))
+		case json.Number:
+			n, err := v.Int64()
```
johnsaigle: FYI json.Int64() just wraps `strconv.ParseInt`:

```
// Int64 returns the number as an int64.
func (n Number) Int64() (int64, error) {
	return strconv.ParseInt(string(n), 10, 64)
}
```

So this section and others like it could be simplified by writing

```
n, err := strconv.ParseUint(string(v), 10, 64))
```

This would handle catch the negative case and allow us to avoid the casting and nosec comments.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		EmitterChain:     vaa.ChainIDXRPL,
+		EmitterAddress:   emitterAddress,
+		Payload:          payload,
+		ConsistencyLevel: 0, // XRPL validated ledgers are final
+		IsReobservation:  false,
+	}, nil
+}
+
+// parseTransaction dispatches to parseTicketCreateTransaction, parseCoreTransaction, and parseNttTransaction.
+// TicketCreate is checked first because it has no Destination field and would fail in the other parsers.
```
johnsaigle: Can you expand on this comment and make the ordering clear? The comment in `parseTicketCreateTransaction` specifically says `	// Failed TicketCreate falls through to parseXACKTransaction.`, so it seems like it relies on a strict ordering of processing in its callers.
It would be good to note this here in case someone tries to reorder these calls for whatever reason.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		return nil, fmt.Errorf("failed to decode MemoData: %w", err)
+	}
+
+	// Verify length
+	if len(data) != memoDataLength {
+		return nil, fmt.Errorf("invalid memo data length: got %d, want %d", len(data), memoDataLength)
+	}
+
+	// Verify NTT prefix
+	if data[0] != nttPrefix[0] || data[1] != nttPrefix[1] ||
```
johnsaigle: Using `bytes.Equal` is a little nicer

```
if !bytes.Equal(data[:4], nttPrefix[:]) {
    return nil, nil
}
```

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+	// This avoids overflow from parsing at high precision first
+	multiplier := new(big.Float).SetInt(new(big.Int).Exp(big.NewInt(10), big.NewInt(int64(targetDecimals)), nil))
+	result := new(big.Float).Mul(f, multiplier)
+
+	// Truncate to integer (floor) - any fractional part beyond targetDecimals is dust
+	intVal, _ := result.Int(nil)
+	// COVERAGE: This branch is mathematically unreachable. We check for negative input above (line 545),
+	// and multiplying a non-negative number by a positive power of 10 cannot produce a negative result.
+	// This check exists as defensive programming against potential floating-point edge cases.
+	if intVal.Sign() < 0 {
```
johnsaigle: This is included in the IsUint64() check, but this extra check could be kept if having the unique error message is important.

node/pkg/watchers/xrpl/parse.go
johnsaigle: Also it could be nice to wrap this parsing in a function since it's repeated in a few places. e.g. something like `jsonNumberToUint64`

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+}
+
+// =============================================================================
+// Token identification
+// =============================================================================
+
+// normalizeCurrency converts a currency code to its canonical 20-byte internal format.
+// Standard codes: [0x00][ASCII bytes][trailing zeros]
+// Non-standard codes: [raw 160-bit value] (40-character hex string)
+func (p *Parser) normalizeCurrency(currency string) ([20]byte, error) {
```
johnsaigle: It would be nice to have this function live in the same package beside `encodeCurrency`. Currently they're split between two different packages even though they're inverses of each other. Ideally they'd both be `normalize/denormalize` or else `encode/decode` to make this more clear. My suggestion would be `encode/decode` since the normalize/denormalize for Wormhole usually refers to decimal-based conversions for NTT and WTT.

Also, it would be great to have a round-trip test to make sure they're paired correctly.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+		decoded, err := hex.DecodeString(currency)
+		if err != nil {
+			return result, fmt.Errorf("failed to decode hex currency: %w", err)
+		}
+		copy(result[:], decoded)
+		return result, nil
+	}
+
+	// Standard currency code (3-character ASCII)
+	if len(currency) < 1 || len(currency) > 3 {
```
johnsaigle:
```suggestion
	if len(currency) == 0 || len(currency) > 3 {
```

Sorry this is super nitty but I do think it's easier to read.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	// Send to processor
+	w.msgChan <- msg
+	xrplMessagesConfirmed.Inc()
+
+	logger.Info("Message observed",
+		zap.String("txHash", hex.EncodeToString(msg.TxID)),
+		zap.Uint64("sequence", msg.Sequence),
+		zap.Uint32("nonce", msg.Nonce),
```
johnsaigle: Is this always 0 for this chain? Maybe they're something more useful to log here. Consider also using `ZapFields` https://github.com/wormhole-foundation/wormhole/blob/559162904fe0733fbd653d785a50124022bd707f/node/pkg/common/chainlock.go#L613-L628

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+	// States are here: https://xrpl.org/docs/references/http-websocket-apis/api-conventions/rippled-server-states
+	if state != "full" && state != "proposing" && state != "validating" {
+		logger.Warn("XRPL node not fully synced", zap.String("state", state))
+	}
+
+	seq := infoResp.Info.ValidatedLedger.Seq
+	if seq > math.MaxInt64 {
+		return 0, fmt.Errorf("ledger sequence %d exceeds max int64", seq)
+	}
+	return int64(seq), nil
```
johnsaigle, line 426: What's the reason for this type? If it should really be a int64 instead of a `uint` maybe we can open an issue in their SDK?

If we find a way to thread an encoding of uint64 through this codebase then we can save ourselves the hassle of doing some of the negative/overflow checks elsewhere in the code.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+// mptLedgerEntryRequest is a custom request type for fetching MPT issuance details.
+// The xrpl-go library doesn't have built-in support for MPT ledger entries.
+type mptLedgerEntryRequest struct {
+	MPTIssuance string `json:"mpt_issuance"`
+	LedgerIndex string `json:"ledger_index"`
```
johnsaigle, line 433:
```suggestion
	LedgerIndex int64 `json:"ledger_index"`
```

I think this change makes it so that we don't need some of `getValidatedLedgerIndex`. Also, it may be the case that this should be `uint64`, as I commented elsewhere.

sdk/vaa/payloads.go
```diff
@@ -806,9 +806,306 @@ func LeftPadBytes(payload string, length int) (*bytes.Buffer, error) {
 	return buf, nil
 }
 
+// GetPayloadPrefix extracts the 4-byte prefix from a payload.
+// This is used to dispatch on payload type (UTX0 vs XREL, etc.).
+func GetPayloadPrefix(bz []byte) [4]byte {
```
johnsaigle, line 860: Could this be `GetUTXOPayloadPrefix`? Or is it relevant for other types of payloads?

sdk/vaa/payloads.go
```diff
@@ -1061,3 +1358,128 @@ func (p *UTXOUnlockPayload) Serialize() ([]byte, error) {
...
 }
+
+// XRPLTicketRefillPayload represents a VAA payload for refilling tickets on the XRP Ledger.
+// This payload is emitted by the XRPL sequencer when tickets are running low.
+type XRPLTicketRefillPayload struct {
+	// Account is the 20-byte XRPL account ID that will create the tickets
+	Account [20]byte
+	// UseTicket is the ticket sequence to consume for the TicketCreate transaction
+	UseTicket uint64
+	// RequestCount is the number of tickets to create (1-250)
```
johnsaigle: Should this be a uint8 then? Or if you want this to match the serializatoin format, then the Deserialize should give an error if the RequestCount > 8

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+	}
+
+	sequence := (uint64(tx.LedgerIndex.Uint32()) << 32) | tx.MetaTransactionIndex
+	return txHash, sequence, nil
+}
+
+// validateTransactionResult checks that the transaction result is tesSUCCESS.
+// Returns nil if valid, or an error describing why the transaction should be skipped.
+// This function is strict - if the result cannot be determined, it returns an error.
+func (p *Parser) validateTransactionResult(tx GenericTx) error {
```
johnsaigle: This function doesn't access `p` so this may be better written (and easier to test) as a standalone function.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -0,0 +1,1290 @@
...
+	if tx.MetaTransactionResult != tesSUCCESS {
+		return fmt.Errorf("transaction result is %s, not %s", tx.MetaTransactionResult, tesSUCCESS)
+	}
+
+	return nil
+}
+
+// validateTransactionType checks that the transaction type is "Payment".
+// Returns nil if valid, or an error if the transaction type doesn't match.
+func (p *Parser) validateTransactionType(tx transaction.FlatTransaction) error {
```
johnsaigle: Same as above: this function doesn't access `p` so this may be better written (and easier to test) as a standalone function.

johnsaigle (commented):
Left several suggestions below, mostly for simplifying the code itself or making it a bit easier to reason about.
None of these need to be blockers (except the slice access check) so it's OK if you need to resolve them instead of refactoring. That said if you need to restructure things a bit to pass CI anyway, I'd appreciate if you could consider working some of these in.

node/pkg/watchers/xrpl/watcher.go
evan-gray: the existing code matches more closely with the other watchers

node/pkg/watchers/xrpl/parse.go
evan-gray: will address https://github.com/wormhole-foundation/wormhole/issues/4808 in a follow up

node/pkg/watchers/xrpl/parse.go
evan-gray: this was added in linting rules after this PR was opened. updated.

node/pkg/watchers/xrpl/parse.go
evan-gray: outdated, I think.

node/pkg/watchers/xrpl/parse.go
evan-gray: I don't think it panics, but I'll add an `if` anyway

```go
package main

import "fmt"

func main() {
	s := []uint64{42}
	fmt.Printf("len=%d\n", len(s))
	fmt.Printf("s[0]=%d\n", s[0])
	tail := s[1:]
	fmt.Printf("s[1:] len=%d, value=%v\n", len(tail), tail)
	for _, v := range tail {
		fmt.Printf("iter: %d\n", v)
	}
	fmt.Println("no panic")
}
```

node/pkg/watchers/xrpl/watcher.go
evan-gray: please see that `fetchMPTAssetScale` uses the string `validated`.

node/pkg/watchers/xrpl/watcher.go
evan-gray: I think it's exactly 32, added a check

node/pkg/watchers/xrpl/watcher.go
evan-gray: don't think so, it was probably based on the aptos one which doesn't either. fixed, I think. please check

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
evan-gray: added comments for the others, but they are generated so they don't get a nonce, though the core message does

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
evan-gray: agreed, updated

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/watcher.go
evan-gray: it's not 0 for core messages, switched to zap fields.

node/pkg/watchers/xrpl/watcher.go
evan-gray: pretty sure this return is just to match the proto for Heartbeat_Network.Height. that said, idk why their SDK doesn't match the u32 that appears in the [C++ code](https://github.com/XRPLF/rippled/blob/028f0cb5da67175cb7ae9cf632502fe27d94525e/include/xrpl/protocol/Protocol.h#L255)

sdk/vaa/payloads.go
evan-gray: it's used to switch on all the payload types - currently UTX0, XREL, XRFL, XBRN

sdk/vaa/payloads.go
evan-gray: this is an XRPL arbitrary limit, not our spec. we want to remain flexible to the possibility that they update the limit. I've updated the comment to this effect.

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
evan-gray: updated

node/pkg/watchers/xrpl/parse.go
johnsaigle: Thanks for writing the test. TIL. I misunderstood this bit of code so apologies for the noise.

mdulin2 (dismissed):

node/pkg/watchers/xrpl/watcher.go
johnsaigle: Hm this check seems restrictive then if their SDK uses `uint`, though if the C++ code only goes up to uint32 then max.Int64 should be safe, I guess. 

It seems like it's not a problem in practice but it might be worth a code comment explaining that the int64 restriction is on our side and isn't invalid ledger sequence... unless it is, in which case their SDK has a bug.

mdulin2 (approved):

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if isStandard {
+		for i := 15; i < 20; i++ {
+			if currency[i] != 0 {
+				isStandard = false
+				break
+			}
+		}
+	}
+	if isStandard && currency[12] != 0 {
```
johnsaigle, line 75: Can you explain this `currency[12] != 0` bit? I don't understand why only this byte would be checked and not the others.

Maybe it's handling the all-zeroes case? but that leaves an edge-case where if bytes 13-14 are non-zero then an invalid three character currency ID

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Decode converts an XRPL currency string to its canonical 20-byte form.
+// Standard codes: [0x00][ASCII bytes (3)][trailing zeros]
+// Non-standard codes: [raw 160-bit value] (40-character hex string)
+// "XRP" is rejected because it is not a valid trust-line currency.
+func Decode(currency string) ([NormalizedLen]byte, error) {
+	var result [NormalizedLen]byte
+
+	// XRP is disallowed as a currency code
+	if strings.ToUpper(currency) == "XRP" {
```
johnsaigle, line 28: This isn't enforced on the encode side. Is that intentional?

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
```diff
@@ -0,0 +1,81 @@
+// Package currencycodec encodes and decodes XRPL currency codes between their
+// JSON string representation and the canonical 20-byte internal form used in
+// NTT source token derivation.
```
johnsaigle, line 3: Is the intention for this file to just pick the encoding format (hex vs. standard 3 charcter encoding), or is it also doing semantic validation? It would be good to list goals/non-goals here.

In the former case, I think it would be more consistent to strip the check for XRP in the Decode case. I also think it makes sense to drop the zero-check on byte 12 in the Encode case.

If the latter, then this function should go further to enforce the rules including:

Character allow-listing:
> Currency codes must be exactly 3 ASCII characters in length. The following characters are permitted: all uppercase and lowercase letters, digits, as well as the symbols ?, !, @, #, $, %, ^, &, *, <, >, (, ), {, }, [, ], and |. ([link](https://xrpl.org/docs/references/protocol/data-types/currency-formats#standard-currency-codes))

Handling the all-zero case:
> The currency code 0x0000000000000000000000000000000000000000 (all zeroes) is generally disallowed. Usually, XRP amounts are not specified with currency codes. However, this code is used to indicate XRP in rare cases where a field must specify a currency code for XRP. ([link](https://xrpl.org/docs/references/protocol/binary-format#currency-codes))

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: IIRC that was to guard the all-zeroes case from returning an invalid `"\x00\x00\x00"` which is not valid, so instead it falls back to hex. Regarding the edge case, if this is zero, it falls back to hex, if it's not, it assumes a valid 3 characters, unless you mean to say what if bytes 13 and 14 are zero? Will include this in the issue, but for the time being this code always stems from on-chain input which is assumed to be valid (or would be rejected on-chain if not).

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: before being moved, encode was only used by the manager where that shouldn't have been feasible. will add this to the issue, as further validation would have to be done to reject if it would be invalid on-chain. that said, I think we can scope the purpose of this to be for conversion purposes only, not to replicate the XRPL semantics.

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: yes, the intention was just to pick the correct encoding format based on the data that originally came from on-chain (even circularly via XRPL -> remote chain -> XRPL Sequencer). will cover this in an issue

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: #4813

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: #4813

node/pkg/watchers/xrpl/currencycodec/currencycodec.go
evan-gray: #4813

node/pkg/watchers/xrpl/watcher.go
evan-gray: my understanding is that their SDK types this incorrectly

bemic (approved):

djb15 (approved):
Approving Tiltfile change

panoel (approved):

PR #4713, opened 2026-03-11
node: Update mutex handling in processor

djb15:
Resolves https://github.com/asymmetric-research/wormhole-llm-issues/issues/68.

The previous `defer` wasn't actually releasing the mutex during the loop.


johnsaigle (approved):
Makes sense as a change. I think this could break if the control flow is ever altered to continue/return early but hopefully future LLM scans will catch this. 😅 
Anyway the current change is an improvement.

mdulin2 (approved):
How did this code even work before? Good catch!

PR #4716, opened 2026-03-12
node: Don't hold mutex on subscription channels in Spy

djb15:
Resolves https://github.com/asymmetric-research/wormhole-llm-issues/issues/184.

### DIFF
--- a/node/cmd/spy/spy.go
+++ b/node/cmd/spy/spy.go
@@ -110,8 +110,12 @@ func subscriptionId() string {
 }
 
 func (s *spyServer) PublishSignedVAA(vaaBytes []byte) error {
+	// Collect matching subscribers under the lock, then send outside the lock.
+	// This prevents a slow subscriber from blocking the mutex and stalling all
+	// other publishers and subscriber cleanup.
+	var targets []*subscriptionSignedVaa
+
 	s.subsSignedVaaMu.Lock()
-	defer s.subsSignedVaaMu.Unlock()
 
 	var v *vaa.VAA
 	var err error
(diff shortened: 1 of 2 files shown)


node/cmd/spy/spy.go
```diff
@@ -123,16 +127,18 @@ func (s *spyServer) PublishSignedVAA(vaaBytes []byte) error {
...
+					s.subsSignedVaaMu.Unlock()
 					return err
 				}
 			}
-			sub.ch <- message{vaaBytes: vaaBytes} //nolint:channelcheck // Don't want to drop incoming VAAs
+			targets = append(targets, sub)
 			continue
 		}
 
 		if v == nil {
```
pleasew8t, line 138: While we are here, does it make sense to move this up before the for-loop starts? Seems like a quick early-exit condition.

djb15: I don't think we can do this given the assignment to `v` on line 128? It is weirdly written though and could probably do with a refactor

node/cmd/spy/spy.go
```diff
@@ -143,15 +149,22 @@ func (s *spyServer) PublishSignedVAA(vaaBytes []byte) error {
...
+				targets = append(targets, sub)
 			}
 		}
 
 	}
 
+	s.subsSignedVaaMu.Unlock()
+
+	for _, sub := range targets {
+		sub.ch <- message{vaaBytes: vaaBytes} //nolint:channelcheck // Don't want to drop incoming VAAs
```
johnsaigle, line 168:
```suggestion
		sub.ch <- message{vaaBytes: vaaBytes} // Don't want to drop incoming VAAs
```

channelcheck didn't work out, and future versions of golangci-lint will mark linters that don't actually run. #4701

johnsaigle (commented):
Good to approve after the comment is updated

johnsaigle (dismissed):

pleasew8t (approved):

node/cmd/spy/spy.go
bemic, line 245: I guess this comment is not true anymore.

bemic (approved):
Only one in-code comment is outdated. Otherwise it looks fine

PR #4718, opened 2026-03-13
ci: enable "Poorly Chosen Identifer" lint

johnsaigle:


node/cmd/ccq/p2p.go
```diff
@@ -46,7 +46,7 @@ func runP2P(
 	port uint,
 	networkID string,
 	bootstrapPeers string,
-	ethRpcUrl string,
+	ethRPCURL string,
```
djb15: Imo this is harder to read now?

djb15 (commented):
LGTM otherwise

node/cmd/ccq/p2p.go
johnsaigle: Fair, I'll look into renaming this or adding an exception.

johnsaigle: renamed to `rpcURL`

djb15 (approved):

bemic (approved):

PR #4720, opened 2026-03-16
node: add exhaustruct lint for MessagePublication

johnsaigle:
https://golangci-lint.run/docs/linters/configuration/#exhaustruct (Checks each field is initialized)

The syntax for the `include` rule was broken probably since forever, my bad.
This fixes the detection of MessagePublication and the violations. In each case this preserves existing behaviour, as the default value for `bool` when not specified is false.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -118,9 +118,8 @@ linters:
         # Regular expressions must match complete canonical struct package/name/structname.
         # If this list is empty, all structs are tested.
         # Default: []
-        - ./+governor.tokenConfigEntry$
-        # TODO: MessagePublications _should_ be exhaustive but this will require many changes.
-        - ./_common.MessagePublication$
+        - .+/MessagePublication$
+        - .+/common\.MessagePublication$
       exclude:
         - .+/cobra\.Command$
         - .+/http\.Client$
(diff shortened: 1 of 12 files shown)


.golangci.yml
```diff
@@ -118,9 +118,8 @@ linters:
         # Regular expressions must match complete canonical struct package/name/structname.
         # If this list is empty, all structs are tested.
         # Default: []
-        - ./+governor.tokenConfigEntry$
-        # TODO: MessagePublications _should_ be exhaustive but this will require many changes.
-        - ./_common.MessagePublication$
+        - .+/MessagePublication$
+        - .+/common\.MessagePublication$
       exclude:
         - .+/cobra\.Command$
```
mdulin2, line 124: Is this redundant then? I assume that if you use `include` then these are the only struct types that would be reviewed.

mdulin2 (approved):

.golangci.yml
johnsaigle: Yeah i think it's still helpful going forward though. It may be the case that we eventually want to include all structs except for a few, and it's nice to have the deny-list ready to go. No harm keeping it IMO.

mdulin2: I find it misleading... but, if you think it's fine, then I'm happy to let it go.

pleasew8t (approved):

djb15 (approved):

PR #4722, opened 2026-03-17
ci: Add govet shadow detection

johnsaigle:
Augment our linting set-up with shadow detection. In practice, most of our variable shadowing involves reusing `err` within the same scope. Even though this is pretty common in Go, we've had a few small bugs where the wrong error was logged, or where this pattern ended up returning a nil error.

To address this, shadowing is now blocked generally in the Go code.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -144,6 +144,11 @@ linters:
         - sloppyLen
         - underef
         - unslice
+    # TODO: Add this later.
+    # gosec:
+    #   config:
+    #     global:
+    #       audit: true
     govet:
       # The following list includes all the linters that are disabled by default.
       # Uncomment any of these to enable them:
@@ -159,7 +164,7 @@ linters:
         - lostcancel           # Check that context cancel functions are called
         - nilness              # Check for redundant or impossible nil comparisons
         - reflectvaluecompare  # Check for comparing reflect.Value values with == or reflect.DeepEqual
-        # - shadow               # Check for possible unintended shadowing of variables
+        - shadow               # Check for possible unintended shadowing of variables
         - sigchanyzer          # Check for unbuffered channels with signal.Notify
         - slog                 # Check for invalid structured logging calls
         - sortslice            # Check the argument type of sort.Slice
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -645,9 +645,9 @@ func runNode(cmd *cobra.Command, args []string) {
 	if env == common.UnsafeDevNet {
 		// Use the hostname as nodeName. For production, we don't want to do this to
 		// prevent accidentally leaking sensitive hostnames.
-		hostname, err := os.Hostname()
-		if err != nil {
-			panic(err)
+		hostname, hostErr := os.Hostname()
+		if hostErr != nil {
+			panic(hostErr)
 		}
 		*nodeName = hostname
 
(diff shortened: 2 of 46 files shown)


.golangci.yml
```diff
@@ -117,9 +117,7 @@ linters:
         # Regular expressions must match complete canonical struct package/name/structname.
         # If this list is empty, all structs are tested.
         # Default: []
-        - ./+governor.tokenConfigEntry$
-        # TODO: MessagePublications _should_ be exhaustive but this will require many changes.
-        - ./_common.MessagePublication$
```
djb15: This change conflicts with https://github.com/wormhole-foundation/wormhole/pull/4721/changes...is that deliberate?

johnsaigle: They're both based off of `main`, but #4720 should be the one that ultimately decides the MessagePublication behaviour.

mdulin2 (dismissed):

bemic (dismissed):
The code is better to read with this one. I like that

node/cmd/guardiand/node.go
```diff
@@ -780,8 +780,8 @@ func runNode(cmd *cobra.Command, args []string) {
 	if env == common.UnsafeDevNet {
 		// Only if the signer is file-based should we generate the deterministic key and write it to disk
 		if st, _, _ := guardiansigner.ParseSignerUri(*guardianSignerUri); st == guardiansigner.FileSignerType {
-			err := devnet.GenerateAndStoreDevnetGuardianKey(*guardianKeyPath)
-			if err != nil {
+			genErr := devnet.GenerateAndStoreDevnetGuardianKey(*guardianKeyPath)
+			if genErr != nil {
 				logger.Fatal("failed to generate devnet guardian key", zap.Error(err))
```
pleasew8t: Should this be `zap.Error(genErr)`?

node/cmd/guardiand/node.go
```diff
@@ -803,22 +803,22 @@ func runNode(cmd *cobra.Command, args []string) {
...
+			firstGuardianName, lookupErr := devnet.GetFirstGuardianNameFromBootstrapPeers(*p2pBootstrap)
+			if lookupErr != nil {
+				logger.Fatal("failed to get first guardian name from bootstrap peers", zap.String("bootstrapPeers", *p2pBootstrap), zap.Error(lookupErr))
 			}
 			// try to connect to guardian-0
 			for {
 				//nolint:noctx // TODO: this should be refactored to use context.
-				_, err := net.LookupIP(firstGuardianName)
-				if err == nil {
+				_, resolveErr := net.LookupIP(firstGuardianName)
```
pleasew8t, line 794: I guess this isn't meant to change the original behaviour. But would it be useful logging the error, while we are here?

node/pkg/txverifier/evm.go
```diff
@@ -835,7 +835,7 @@ func (tv *TransferVerifier[evmClient, connector]) validateReceipt(
 	// To avoid processing these messages individually in the future, we mark them as invalid here.
 	if len(insolventAssets) > 0 {
 		for _, msgPub := range *receipt.MessagePublications {
-			msgID, err := tv.MsgID(msgPub)
+			msgID2, err := tv.MsgID(msgPub)
```
pleasew8t: I'm not sure I understand some of these changes. Why rename this to `msgID2`? It's scoped in its own block, similar to `msgID` and `msgID2` (another rename) earlier?

djb15 (dismissed):
I was concerned about the possibility of some of these semantically changing the code behaviour but I've looked at those in more detail + got claude to test some ideas for me. I think the changes are net-the-same besides the couple of deliberate fixes

pleasew8t (dismissed):

mdulin2:
There currently a build error. Worth fixing and then probably good to merge.

djb15 (approved):

pleasew8t (approved):

evan-gray (approved):

PR #4736, opened 2026-03-31
Add delegated guardian set 1, 2 and 3 

aadam-10:
Adds the governance files and the VAAs corresponding to the phased initial delegated guardian set rollout


SEJeff (approved):

SEJeff (dismissed):

SEJeff (commented):
Putting them in the same directory is slightly confusing. What about:

guardianset/mainnetv2/delegated/*.prototxt and leave the existing bits where they are?

Then you keep the v1.prototxt, v2.prototxt, v3,prototxt format in both.

aadam-10:
good point, organized them into folders

djb15 (dismissed):

guardianset/README.md
```diff
@@ -1,5 +1,7 @@
-These files document historic guardian set changes. 
+The files named "v1" to "vX" document historic guardian set changes.
 
-Wormhole Core contract deployments that are set to a particular guardian set index only accept VAAs that are signed by a quorum of entities in that set.
+Wormhole Core contract deployments that are set to a particular guardian set index only accept VAAs that are signed by a quorum of entities in that set. The VAAs that guardian sets had signed to approve the next guardian set are stored in deployments/mainnet/guardianSetVAAs.csv
+----
```
evan-gray, line 5: I think this makes the markdown yelling for some reason (see https://github.com/aadam-10/wormhole/blob/7f3ce54a53fbe8fc5b2d33643a6c1658555bd34a/guardianset/README.md)

---

maybe you need a newline above and below and three `---`? that seems to work in this comment here

evan-gray (dismissed):

evan-gray (approved):

djb15 (approved):

PR #4744, opened 2026-04-09
feat(node): guardian signed delegate signatures broadcast

evan-gray:
This returns the gossip message time restriction pre #4743 in favor of a canonical guardian only approach. This allows any guardian to collect a quorum of delegated guardian set signatures for a message and broadcast them in a single message.

The goal is that this should maintain the same safeguards and security of the original delegated guardian set messages but provide the same security / DoS prevention as observation requests. 

Tested with https://api.wormholescan.io/api/v1/observations/delegate/50/00000000000000000000000062deeafee06c7442a21c93ededc79a0cb5791c83/1750

This required a change in Wormholescan to store the full data from the gossip message, so some earlier messages may not contain all the relevant fields. I expect that if a prior message was needed for some reason, it could either be re-observed or the data could be backfilled.

Reviewers: the most critical code is in p2p.go 👀 

  ---                                                                                                                                                                                                           
  Summary                                                                                                                                                                                                       
                                                                                                                                                                                                                
  - Add broadcast-delegate-signatures admin command that fetches delegate observations from the wormholescan API for a given VAA ID and rebroadcasts them on the delegated_attestation p2p topic                
  - Introduces a new DelegateSignaturesBroadcast gossip message type that carries observation data once with a compact array of per-guardian (addr, sent_timestamp, signature) tuples                           
  - The CLI client fetches from wormholescan, groups by hash, verifies each delegate signature by reconstructing the original DelegateObservation bytes, and sends only verified signatures to the guardian node
  - The admin server is a thin passthrough that forwards the pre-built broadcast to p2p for signing and publication                                                                                             
  - Receiving guardians verify both the broadcast envelope (canonical guardian signature) and each inner delegate signature, then feed them into the existing delegate observation processor pipeline


johnsaigle:
@claude

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,254 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	}
+
+	if bestGroup == nil || len(bestGroup.observations) == 0 {
+		return nil, fmt.Errorf("no valid observation groups found")
+	}
+
+	// Build the compact broadcast from the best hash group.
+	ref := bestGroup.observations[0]
+
+	chainIDNum, err := strconv.Atoi(chainID)
```
johnsaigle: Probably better to use  `StringToKnownChainID` so that we know it's a valid chain. https://github.com/wormhole-foundation/wormhole/blob/5ea4844fc26de12902162cc4d4d6ef18c10d4aa9/sdk/vaa/structs.go#L195

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,254 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+
+	// Build and verify each delegate signature by reconstructing the DelegateObservation
+	// and checking the signature matches.
+	signedDelegateObservationPrefix := []byte("signed_delegate_observation_000000|")
+
+	signatures := make([]*gossipv1.DelegateSignature, 0, len(bestGroup.observations))
+	for _, obs := range bestGroup.observations {
+		guardianAddrHex := strings.TrimPrefix(obs.DelegatedGuardianAddr, "0x")
+		guardianAddrBytes, err := hex.DecodeString(guardianAddrHex)
+		if err != nil {
```
johnsaigle, line 781: This might be worth logging? Seems weird to get an invalid Guardian address in this way

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,254 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+
+func runBroadcastDelegateSignatures(cmd *cobra.Command, args []string) {
+	vaaID := args[0]
+
+	// Parse VAA ID: chain/emitter/sequence
+	parts := strings.Split(vaaID, "/")
+	if len(parts) != 3 {
+		log.Fatalf("vaa_id must be in format chain/emitter/sequence")
+	}
+	chainID := parts[0]
```
johnsaigle, line 904: A check for known chain ID here might be nice too. It's minor, and will give a 404 for bad chain IDs, but in the future this could end up giving results for chains that are deprecated yet have historical delegate observations

node/pkg/p2p/p2p.go
```diff
@@ -1175,6 +1232,44 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+							if gs == nil {
+								if logger.Level().Enabled(zapcore.DebugLevel) {
+									logger.Debug("dropping SignedDelegateSignaturesBroadcast - no guardian set",
+										zap.String("from", envelope.GetFrom().String()),
+									)
+								}
+								break
+							}
+
+							// SECURITY: Verify the broadcast was signed by a canonical guardian.
```
johnsaigle:
```suggestion
							// SECURITY: Verify the broadcast was signed by a Guardian in the current Guardian Set.
```
My understanding is that the term Canonical Guardian refers to any Guardian which is _not_ a member of the Delegated Guardian Set for a chain configured to use the delegated Guardian Set feature.

In the current implementation (where all Canonical and Delegated Guardians are also in the current, main Guardian Set), this wording to me suggests that broadcasts can only be signed by non-Delegated Guardians, whereas the check actually is against the current Guardian Set.

Sorry if this feels nit-picky but I do think that being strict about the wording will help other people understand the intended checks.

proto/gossip/v1/gossip.proto
```diff
@@ -295,6 +296,57 @@ message ManagerTransaction {
   int64 sent_timestamp = 7;
 }
 
+// DelegateSignaturesBroadcast is a batch of delegate observations rebroadcast by a canonical guardian.
```
johnsaigle:
```suggestion
// DelegateSignaturesBroadcast is a batch of delegate observations rebroadcast by a Guardian in the current Guardian Set.
```

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,254 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	ConsistencyLevel      uint32 `json:"consistencyLevel"`
+	Timestamp             string `json:"timestamp"`
+	SentTimestamp         string `json:"sentTimestamp"`
+	Unreliable            bool   `json:"unreliable"`
+	IsReobservation       bool   `json:"isReobservation"`
+	VerificationState     uint32 `json:"verificationState"`
+}
+
+// buildDelegateSignaturesBroadcast takes parsed wormholescan API observations and a VAA ID,
+// groups by hash, checks quorum, verifies each delegate signature, and returns the broadcast.
```
johnsaigle: > checks quorum

There's no quorum check in this function. I don't think we want one though, given that the signatures for observations are stored in-memory in the processor and the goal here is to replay signatures that might've been missed.

What do you think?

node/pkg/p2p/p2p.go
```diff
@@ -1175,6 +1232,44 @@ func Run(params *RunParams) func(ctx context.Context) error {
 								}
 							}
 						}
+					case *gossipv1.GossipMessage_SignedDelegateSignaturesBroadcast:
```
johnsaigle, line 1236: Do we want to do a Peer node ID check here? Right now, anyone can re-publish the Broadcast message within the 15m window (as long as it was signed by a current Guardian). Maybe only Guardians should be able to re-publish?

node/pkg/p2p/p2p.go
```diff
@@ -1551,6 +1646,108 @@ func processSignedDelegateObservation(d *gossipv1.SignedDelegateObservation, gs
...
+		if err != nil {
+			logger.Warn("failed to marshal reconstructed delegate observation", zap.Error(err))
+			continue
+		}
+
+		// Verify the delegate signature over the reconstructed bytes.
+		claimedAddr := eth_common.BytesToAddress(sig.GuardianAddr)
+
+		if gs != nil {
+			if _, ok := gs.KeyIndex(claimedAddr); !ok {
```
johnsaigle, line 1701: I'm not sure that this check is exactly right. My understanding is that this check is saying "Did a member of the current Guardian Set sign the Delegate Observation?" but conceptually these signatures should come from a chain's _Delegated_ Guardian Set, not the overall Guardian Set.

This check allows for a signature of a Canonical Guardian (i.e. one not necessarily running a full node) to have a valid signature for a _Delegate_ Observation, so long as it is in the Guardian Set. 
In theory, it would also exclude a valid signature from a Guardian that is in the Delegated Guardian Set but not in the current Guardian Set. I guess this is probably unlikely, but the main Guardian Set and Delegated Guardian Set(s) may not overlap in certain periods.

johnsaigle, line 1701: I'm not sure that this check is exactly right. My understanding is that this check is saying "Did a member of the current Guardian Set sign the Delegate Observation?" but conceptually these signatures should come from a chain's Delegated Guardian Set, not the overall Guardian Set.

This check allows for a signature of a Canonical Guardian (i.e. one not necessarily running a full node) to have a valid signature for a Delegate Observation, so long as it is in the Guardian Set.
In theory, it would also exclude a valid signature from a Guardian that is in the Delegated Guardian Set but not in the current Guardian Set. I guess this is probably unlikely, but the main Guardian Set and Delegated Guardian Set(s) may not overlap in certain periods.

johnsaigle (commented):
Given that we're doing checks against the current Guardian Set on one level, and the delegated/canonical sets on another level, do we need to worry about the boundary cases like:

* Changing a Guardian Set?
* Changing the Delegated Guardian Set for a chain?

node/cmd/guardiand/adminclient.go
evan-gray: yes! I originally tried to have a quorum check but then realized I would need the delegate set and as you said we probably want to just send what we have regardless

node/pkg/p2p/p2p.go
evan-gray: and here I thought a canonical guardian was one that was in the core contract 🤔

node/pkg/p2p/p2p.go
evan-gray: so, this is the identical check of the existing `processSignedDelegateObservation` which has this limitation that the delegate must be in the guardian set for the time being.

evan-gray: dup of the above

node/pkg/p2p/p2p.go
```diff
@@ -1551,6 +1665,108 @@ func processSignedDelegateObservation(d *gossipv1.SignedDelegateObservation, gs
...
+func processSignedDelegateSignaturesBroadcast(s *gossipv1.SignedDelegateSignaturesBroadcast, gs *common.GuardianSet) (*gossipv1.DelegateSignaturesBroadcast, error) {
+	envelopeAddr := eth_common.BytesToAddress(s.GuardianAddr)
+	idx, ok := gs.KeyIndex(envelopeAddr)
+	if !ok {
+		return nil, fmt.Errorf("invalid message: %s not in guardian set", envelopeAddr)
+	}
+	pk := gs.Keys[idx]
+
+	// SECURITY: see whitepapers/0009_guardian_signer.md
+	if len(signedDelegateSignaturesBroadcastPrefix)+len(s.Broadcast) < 34 {
```
johnsaigle, line 1741: It looks like this length calculation is wrong, same with other parts of the code unfortunately:

| Prefix Variable | Actual Prefix | Line | Length | Check Status |
| --- | --- | ---: | ---: | --- |
| `signedDelegateSignaturesBroadcastPrefix` | `signed_delegate_signatures_bc_0000\|` | 1741 | 35 bytes | No-op |
| `signedDelegateObservationPrefix` | `signed_delegate_observation_000000\|` | 1635 | 35 bytes | No-op |
| `signedManagerTransactionPrefix` | `signed_manager_transaction_0000000\|` | 1781 | 35 bytes | No-op |
| `signedObservationRequestPrefix` | `signed_observation_request\|` | 1594 | 27 bytes | Valid |
| `heartbeatMessagePrefix` | `heartbeat\|` | 1488 | 10 bytes | Valid |

idk where `34` comes from originally... probably `0x` + 32 byte addresses? I guess it worked fine as a sanity check for the heartbeats but not for the newer and longer prefixes.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,261 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	if len(apiObservations) == 0 {
+		return nil, fmt.Errorf("no delegate observations provided")
+	}
+
+	// Group observations by hash — only broadcast the group with the most signatures.
+	type hashGroup struct {
+		observations []wormholescanDelegateObservation
+	}
+	groups := make(map[string]*hashGroup)
+	for _, obs := range apiObservations {
```
johnsaigle: Bear with me here because there are several moving parts. I also wrote a really rambly version of this and had an LLM clean it up, so if the tone is weird, that's why.

--- 

I think we need some extra validation before batching here.
My understanding is that `obs.Hash` corresponds to the normal VAA signing digest, i.e. the VAA signing body in sdk/vaa/structs.go#L732-L743 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/sdk/vaa/structs.go#L732-L743). 

But a delegated observation signs more than just that body. The signed `DelegateObservation` also includes fields like tx_hash, unreliable, is_reobservation, verification_state, guardian_addr, and sent_timestamp. 

* The MessagePubliccation gets converted to an Observation node/pkg/processor/delegate_obs.go#L89-L126 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/node/pkg/processor/delegate_obs.go#L89-L126)
* The Observation is observed on the channel and the (Delegated) Guardian signs the Observation, which includes these extra fields. node/pkg/p2p/p2p.go#L726-L758 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/node/pkg/p2p/p2p.go#L726-L758)

That matters because this code groups observations only by `obs.Hash`. If two observations share the same VAA digest but differ in one of those extra fields, they will still end up in the same hash group.

For example, imagine two observations with the same hash, but one has `isReobservation=false` and the other has `isReobservation=true`. Both signatures can verify individually here, because we reconstruct each `DelegateObservation` using that row’s own fields.

The problem comes later in `expandBroadcastToSignedDelegateObservations()`. That function reconstructs each observation using the top-level broadcast fields for tx_hash, unreliable, is_reobservation, and verification_state rather than preserving them per observation: node/pkg/p2p/p2p.go#L1684-L1687 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/node/pkg/p2p/p2p.go#L1684-L1687)

Then on the receiving side, the reconstructed bytes may no longer match what a given delegated guardian originally signed, so that signature gets dropped during verification.

Separately, the processor’s delegate aggregation path uses a stronger identity than the normal VAA digest. It converts the delegated observation back into a MessagePublication and hashes that representation in node/pkg/processor/observation.go#L737-L760 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/node/pkg/processor/observation.go#L737-L760)

And `MessagePublication.MarshalBinary()` includes TxID, IsReobservation, Unreliable, and verificationState in node/pkg/common/chainlock.go#L260-L354 (https://github.com/wormhole-foundation/wormhole/blob/0d5b3f9e6fb88d24ca999f1a158eaf27b05d40dc/node/pkg/common/chainlock.go#L260-L354).

So I think grouping by `obs.Hash` alone is too weak for this compact broadcast format. 

We either need to:
1. group on a stronger key / explicitly validate equality of all fields being compressed into the top-level DelegateSignaturesBroadcast, or
2. move those fields into the per-signature portion of the broadcast.

evan-gray: 1. pretty sure that hash _is_ the one from `SignedDelegateObservation` and
2. the signatures are reverified on lines 794-803 against the `DelegateObservation`

does that address the concerns? the grouping by hash is just a pre-optimization, but the result should never have mismatches due to point 2.

evan-gray: proved 1 in 335ee21 and used VAA hash for the grouping in 
966587a but sends multiple groups based on the `MessagePublicationHash` since, as you noted, the non-VAA fields may be different.

node/pkg/p2p/p2p.go
evan-gray: IIRC it was done to prevent possible hash confusion with prior message types, see [whitepaper](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0009_guardian_signer.md),  #1958, #1992. I'd prefer to keep it consistent for now even though it's a no-op as an added defense-in-depth in addition to someone copy-pasting, not reading the whitepaper, and not padding the hash. can always handle in a follow up.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
 	fmt.Println(resp.GetResponse())
 }
 
+// wormholescanDelegateObservation represents a delegate observation from the wormholescan API.
+type wormholescanDelegateObservation struct {
+	Sequence               uint64 `json:"sequence"`
+	EmitterChain           uint32 `json:"emitterChain"`
```
johnsaigle:
```suggestion
	EmitterChain           uint16 `json:"emitterChain"`
```

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	Payload                string `json:"payload"`
+	DelegatedGuardianAddr  string `json:"delegatedGuardianAddr"`
+	Signature              string `json:"signature"`
+	Nonce                  uint32 `json:"nonce"`
+	ConsistencyLevel       uint32 `json:"consistencyLevel"`
+	Timestamp              string `json:"timestamp"`
+	SentTimestamp          string `json:"sentTimestamp"`
+	Unreliable             bool   `json:"unreliable"`
+	IsReobservation        bool   `json:"isReobservation"`
+	VerificationState      uint32 `json:"verificationState"`
```
johnsaigle:
```suggestion
	VerificationState      uint8 `json:"verificationState"`
```

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+}
+
+// delegateObservationVAAHash computes the VAA body hash (double Keccak256) from a delegate observation's fields.
+func delegateObservationVAAHash(obs *wormholescanDelegateObservation) (string, error) {
```
johnsaigle, line 664: It might be better to leverage the `VAA` type from the SDK and use `serializeBody()` rather than re-implement here.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+			mpGroups[h] = g
+		}
+		g.observations = append(g.observations, obs)
+	}
+
+	chainIDParsed, err := vaa.StringToKnownChainID(chainID)
+	if err != nil {
+		return nil, fmt.Errorf("invalid chain ID: %v", err)
+	}
+	chainIDNum := int(chainIDParsed)
```
johnsaigle: I think this can be deleted since you cast the `chainID` value to uint32 below.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+		for _, obs := range mg.observations {
+			guardianAddrHex := strings.TrimPrefix(obs.DelegatedGuardianAddr, "0x")
+			guardianAddrBytes, err := hex.DecodeString(guardianAddrHex)
+			if err != nil {
+				log.Printf("Warning: skipping observation with invalid guardian address %q: %v", obs.DelegatedGuardianAddr, err)
+				continue
+			}
+
+			var sigBytes []byte
+			if obs.Signature != "" {
```
johnsaigle: Should we log and return early if `obs.Signature` is empty?

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +642,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+				GuardianAddr:      guardianAddrBytes,
+				SentTimestamp:     sentTimestamp,
+			}
+			b, err := proto.Marshal(d)
+			if err != nil {
+				continue
+			}
+
+			// Verify the signature over the reconstructed bytes.
+			if len(sigBytes) == 0 {
```
johnsaigle: Should be fine either way, but we can specify the exact length here instead of only doing the emptiness check

```suggestion
			if len(sigBytes) == 65 {
```

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,300 @@
...
+		TxID:             txHash,
+		Timestamp:        ts,
+		Nonce:            obs.Nonce,
+		Sequence:         obs.Sequence,
+		ConsistencyLevel: uint8(obs.ConsistencyLevel),   // #nosec G115 -- test data is a known constant
+		EmitterChain:     vaa.ChainID(obs.EmitterChain), // #nosec G115 -- test data is a known constant
+		EmitterAddress:   emitterAddr,
+		Payload:          payload,
+		IsReobservation:  obs.IsReobservation,
+		Unreliable:       obs.Unreliable,
```
johnsaigle, line 158: Could you add verification state to the test? It may need to be done via `SetVerificationState` https://github.com/wormhole-foundation/wormhole/blob/966587a13b76d40b1f365557274ce9acd857b41c/node/pkg/common/chainlock.go#L220

Side-note: I'm trying to land a lint for MessagePublication to make sure that all fields are specified. This struct in particular has recurring problems because these fields are easy to miss. https://github.com/wormhole-foundation/wormhole/pull/4720

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,300 @@
...
+    "emitterChain": 50,
+    "emitterAddr": "00000000000000000000000062deeafee06c7442a21c93ededc79a0cb5791c83",
+    "hash": "UqjHt/0rrOhV1iVtFII4F3cuCpRKkagyWV+b4RF1oQo=",
+    "txHash": "jBQRYi+ZJKFAGdfKnMOBA68LwXfMprBnNPhDWnSYN7w=",
+    "payload": "mUX/EAAAAAAAAAAAAAAAAH77OGZ111KA05quQpZKZ3beDuC9AAAAAAAAAAAAAAAAPrQYvb6VtLnPRl7PvYQkaFrNG8EAkQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABoPAAAAAAAAAAAAAAAASEtVk7u5A4P5T7KZRw8JQnz2z+IAT5lOVFQIAAAAACDebcAAAAAAAAAAAAAAAADdRood3Dktzb7222406JqjOPnxhgAAAAAAAAAAAAAAABvct/SlP0APQnP1bTAtUzDzrKSpAB4AAA==",
+    "delegatedGuardianAddr": "0x938f104aeb5581293216ce97d771e0cb721221b1",
+    "signature": "z/fhKpSdTeE+hsSQs5FiYZM7azrC16NkxlvLKRZuCkso3741gsoJFydNhmEm7STzs3lrsLGDPyhAaoIK/XInjgA=",
+    "nonce": 0,
+    "consistencyLevel": 202,
+    "unreliable": false,
```
johnsaigle, line 116: I see the comment that this is taken from a live endpoint, but could you include test data that has non-zero values for the types corresponding to `unreliable`, `isReobservation` and `verificationState`?

It would help to make sure that these fields are actually getting tested, which is important because they're part of the pre-image which generates the signature for the Delegate Observations.

We've had issues in the past where these fields weren't properly serialized or tested so I want to make sure we're covered here. (e.g. https://github.com/wormhole-foundation/wormhole/pull/4428 for context)

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+func mustDecodeHex(t *testing.T, s string) []byte {
+	t.Helper()
+	b, err := hex.DecodeString(s)
+	require.NoError(t, err)
+	return b
+}
+
+// serializeVAABody reproduces vaa.VAA.serializeBody() without needing a full VAA struct.
+func serializeVAABody(ts time.Time, nonce uint32, chain vaa.ChainID, emitter vaa.Address, seq uint64, cl uint8, payload []byte) []byte {
```
johnsaigle: Similar to my other comment, it might be easier to use the VAA type and its methods instead of re-implementing.

johnsaigle (commented):
I think we should update the test coverage to make sure we're handling the non-VAA-hash-related fields correctly.

Otherwise, the rest of the comments are improvements but IMO they do not need to block the PR.

node/pkg/p2p/p2p.go
```diff
@@ -1551,6 +1665,108 @@ func processSignedDelegateObservation(d *gossipv1.SignedDelegateObservation, gs
...
+	var h gossipv1.DelegateSignaturesBroadcast
+	err = proto.Unmarshal(s.Broadcast, &h)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal delegate signatures broadcast: %w", err)
+	}
+
+	if time.Until(time.Unix(h.BroadcastTimestamp, 0)).Abs() > delegateSignaturesBroadcastMaxTimeDifference {
+		return nil, fmt.Errorf("delegate signatures broadcast is too old or too far into the future")
+	}
+
```
mdulin2, line 1766: Is it worth adding a length check on the number of signatures provided? If not, feel free to disregard.

node/pkg/p2p/p2p.go
```diff
@@ -1551,6 +1665,108 @@ func processSignedDelegateObservation(d *gossipv1.SignedDelegateObservation, gs
...
+		if gs != nil {
+			if _, ok := gs.KeyIndex(claimedAddr); !ok {
+				logger.Warn("broadcast delegate signature from address not in guardian set",
+					zap.String("addr", claimedAddr.Hex()))
+				continue
+			}
+		}
+
+		digest := signedDelegateObservationDigest(b)
+		pubKey, err := ethcrypto.Ecrecover(digest.Bytes(), sig.Signature)
```
mdulin2, line 1709: Mimics the single flow very well. I think this is pretty solid.

node/pkg/p2p/p2p.go
```diff
@@ -1175,6 +1233,62 @@ func Run(params *RunParams) func(ctx context.Context) error {
...
+								if logger.Level().Enabled(zapcore.DebugLevel) {
+									logger.Debug("dropping SignedDelegateSignaturesBroadcast - no guardian set",
+										zap.String("from", envelope.GetFrom().String()),
+									)
+								}
+								break
+							}
+
+							// SECURITY: Verify P2P peer matches known guardian peer(s) from heartbeats.
+							broadcastGuardianAddr := eth_common.BytesToAddress(s.GuardianAddr)
```
mdulin2, line 1250: I love this check! With this, I don't think that the replay attack is really an issue anymore, correct? It would be good to add this p2p matching check in more places, but that's a separate PR for sure.

mdulin2 (dismissed):
Looks pretty good to me! The checks on the single signature appear to all be done on the batch case as well.

node/pkg/p2p/p2p.go
evan-gray: that's my understanding, another peer could not replay a different guardian's message even within the 15 minute window. these must be signed 1) with a guardian evm key and 2) that guardian's attested (heartbeated) p2p key which was also signed by their evm key - this is only true because p2p verification is still enabled AFAIU

node/pkg/p2p/p2p.go
evan-gray: don't think we'd get this far in practice, but doesn't hurt

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,355 @@
...
+// TestBuildDelegateSignaturesBroadcasts_NonVAAFieldsAreSignaturePreimage verifies that
+// non-VAA fields (unreliable, isReobservation, verificationState) are part of the delegate
+// observation signature preimage. Changing any of these fields should cause signature
+// verification to fail, dropping all signatures.
+func TestBuildDelegateSignaturesBroadcasts_NonVAAFieldsAreSignaturePreimage(t *testing.T) {
+	vaaID := "50/00000000000000000000000062deeafee06c7442a21c93ededc79a0cb5791c83/1750"
+
+	tests := []struct {
+		name   string
+		mutate func(obs []wormholescanDelegateObservation)
```
johnsaigle, line 299: Could you add another case that mutates `TxID` as well? This is another value that is not part of the VAA's signing body but is part of the Preimage for the DGS signature pre-quorum checks.

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,355 @@
...
+	broadcasts, err := buildDelegateSignaturesBroadcasts(vaaID, observations)
+	require.NoError(t, err)
+	require.Len(t, broadcasts, 1)
+
+	// 5 of 6 should pass (the corrupted one is dropped).
+	assert.Equal(t, 5, len(broadcasts[0].Signatures))
+}
+
+// TestBuildDelegateSignaturesBroadcasts_NonVAAFieldsAreSignaturePreimage verifies that
+// non-VAA fields (unreliable, isReobservation, verificationState) are part of the delegate
```
johnsaigle:
```suggestion
// non-VAA fields (TxID, unreliable, isReobservation, verificationState) are part of the delegate
```

node/cmd/guardiand/adminclient_delegate_test.go
```diff
@@ -0,0 +1,355 @@
...
+// observation signature preimage. Changing any of these fields should cause signature
+// verification to fail, dropping all signatures.
+func TestBuildDelegateSignaturesBroadcasts_NonVAAFieldsAreSignaturePreimage(t *testing.T) {
```
johnsaigle, line 294: Could you add another test, or modify this one, to include observations that have mixed statuses for these values? 

The original problem I was worried about is that e.g. 2 messages in the same broadcast would have their non-VAA fields overwritten by the broadcast containing them. I believe that is no longer the case now that we have the message publication groups, but it would be good to prove that.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	conn, c, err := getAdminClient(ctx, *clientSocketPath)
+	if err != nil {
+		log.Fatalf("failed to get admin client: %v", err)
+	}
+	defer conn.Close()
+
+	for i, broadcast := range broadcasts {
```
johnsaigle, line 948: I'm wondering if we need to do a size check here when determining the number of batches to send, similar to `MaxObservationBatchSize` https://github.com/wormhole-foundation/wormhole/blob/937d18fcd9e3090cd6acaa79ff22f6a595615d0f/node/pkg/p2p/p2p.go#L52-L53

We have a 1 MB limit by default on the p2p network and the broadcast contains several Observations that include the payload. For broadcasts containing many signatures long payloads, maybe we could hit that limit? 

It might be good to log the size of the `broadcast` here so that if messages get dropped at the libp2p layer, we would be able to know and then debug or patch.

johnsaigle (commented):
A few small notes for test coverage and logging in case we hit size limits on the p2p layer.
Otherwise looks good and I'm happy to approve once these are resolved.

node/pkg/p2p/p2p.go
```diff
@@ -1551,6 +1665,112 @@ func processSignedDelegateObservation(d *gossipv1.SignedDelegateObservation, gs
...
+	err = proto.Unmarshal(s.Broadcast, &h)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal delegate signatures broadcast: %w", err)
+	}
+
+	if time.Until(time.Unix(h.BroadcastTimestamp, 0)).Abs() > delegateSignaturesBroadcastMaxTimeDifference {
+		return nil, fmt.Errorf("delegate signatures broadcast is too old or too far into the future")
+	}
+
+	if len(h.Signatures) == 0 {
```
johnsaigle, line 1767: We could reject here if `len(h.Signatues) > 19` as well, but not a big deal.

node/cmd/guardiand/adminclient.go
```diff
@@ -627,6 +641,325 @@ func runGetAndObserveMissingVAAs(cmd *cobra.Command, args []string) {
...
+	Sequence               uint64 `json:"sequence"`
+	EmitterChain           uint16 `json:"emitterChain"`
+	EmitterAddr            string `json:"emitterAddr"`
+	MessagePublicationHash string `json:"hash"`
+	TxHash                 string `json:"txHash"`
+	Payload                string `json:"payload"`
+	DelegatedGuardianAddr  string `json:"delegatedGuardianAddr"`
+	Signature              string `json:"signature"`
+	Nonce                  uint32 `json:"nonce"`
+	ConsistencyLevel       uint32 `json:"consistencyLevel"`
```
johnsaigle:
```suggestion
	ConsistencyLevel       uint8 `json:"consistencyLevel"`
```

node/pkg/p2p/p2p.go
evan-gray: I'm less sure about that. Is there a limit of 19 on the delegate set somewhere? Even so, we're not checking the uniqueness of the signatures' recovery, there's already the p2p message size limit. I just worry that check would be overkill and potentially bite some future case

johnsaigle: OK no problem

mdulin2 (approved):

johnsaigle (approved):

panoel (approved):

djb15 (approved):
Approving changes to `gossip.proto`

PR #4746, opened 2026-04-10
solana: reclaim rent for old accounts

kcsongor:
This PR extends the Wormhole program on Solana with new instructions to allow closing old accounts to reclaim their rent.
An old account is one that can safely be closed because it has already surved its purpose. This includes signature set accounts corresponding to VAAs that were submitted more than 30 days ago, the posted VAA data itself, and posted message accounts.
Since there is no on-chain guarantee that a posted message has actually been observed, when closing a posted message account, the handler re-emits an identical message via a space (and cost) efficient cpi event mechanism.

The reclaimed rent from these accounts goes into the fee collector account, which has been used to collect messaging fees.

### DIFF
--- a/solana/bridge/program/src/api/close_posted_message.rs
+++ b/solana/bridge/program/src/api/close_posted_message.rs
@@ -69,8 +69,8 @@ pub fn close_posted_message(
         return Err(InvalidProgramOwner.into());
     }
 
-    // 2-4. Parse and validate the message, extracting data for the event.
-    let message_data = {
+    // 2-4. Parse and validate the message, capturing full account data for the event.
+    let account_data = {
         let data = accs.message.data.borrow();
         if data.len() < 3 {
             return Err(InvalidAccountPrefix.into());
@@ -89,16 +89,16 @@ pub fn close_posted_message(
             return Err(MessageWithinRetentionWindow.into());
         }
 
-        message_data
+        data.to_vec()
         // data borrow dropped here
     };
 
-    // 5. Emit Anchor CPI event with the full message data.
+    // 5. Emit Anchor CPI event with the full account data (prefix + message).
     emit_message_event(
         ctx,
         &accs.event_authority,
         &accs.self_program,
-        &message_data,
+        &account_data,
     )?;
 
     // 6. Close the account: transfer lamports to fee_collector.
@@ -114,12 +114,12 @@ pub fn close_posted_message(
     Ok(())
 }
 
-/// Emit an Anchor-compliant CPI event containing the full MessageData.
+/// Emit an Anchor-compliant CPI event containing the full account data.
 fn emit_message_event(
     ctx: &ExecutionContext,
     event_authority: &Info,
     self_program: &Info,
-    message_data: &MessageData,
+    account_data: &[u8],
 ) -> Result<()> {
     // Verify self_program is actually our program.
     if self_program.key != ctx.program_id {
@@ -133,14 +133,11 @@ fn emit_message_event(
         return Err(InvalidEventAuthority.into());
     }
 
-    // Build instruction data: selector + discriminator + borsh(message_data).
-    let serialized = message_data
-        .try_to_vec()
-        .map_err(|_| SolitaireError::ProgramError(ProgramError::InvalidAccountData))?;
-    let mut ix_data = Vec::with_capacity(8 + 8 + serialized.len());
+    // Build instruction data: selector + discriminator + raw account data.
+    let mut ix_data = Vec::with_capacity(8 + 8 + account_data.len());
     ix_data.extend_from_slice(&EVENT_IX_TAG_LE);
     ix_data.extend_from_slice(&MESSAGE_ACCOUNT_CLOSED_DISCRIMINATOR);
-    ix_data.extend_from_slice(&serialized);
+    ix_data.extend_from_slice(account_data);
 
     let ix = Instruction {
         program_id: *ctx.program_id,
--- a/solana/bridge/program/src/api/close_posted_message.rs
+++ b/solana/bridge/program/src/api/close_posted_message.rs
@@ -27,8 +27,7 @@ use solitaire::{
     *,
 };
 
-/// 30 days in seconds.
-const RETENTION_PERIOD: i64 = 30 * 24 * 60 * 60;
+use super::RETENTION_PERIOD;
 
 /// Event discriminator: SHA256("event:MessageAccountClosed")[0..8].
 pub const MESSAGE_ACCOUNT_CLOSED_DISCRIMINATOR: [u8; 8] =
@@ -84,7 +83,7 @@ pub fn close_posted_message(
         let message_data = MessageData::try_from_slice(&data[3..])
             .map_err(|_| SolitaireError::ProgramError(ProgramError::InvalidAccountData))?;
 
-        // Check retention: submission_time must be at or before (now - 30 days).
+        // Check retention: submission_time must be at or before (now - RETENTION_PERIOD).
         if (message_data.submission_time as i64) > accs.clock.unix_timestamp - RETENTION_PERIOD {
             return Err(MessageWithinRetentionWindow.into());
         }
--- a/solana/bridge/program/src/api/close_posted_message.rs
+++ b/solana/bridge/program/src/api/close_posted_message.rs
@@ -1,0 +1,154 @@
+use crate::{
+    accounts::{
+        Bridge,
+        FeeCollector,
+    },
+    error::Error::{
+        InvalidAccountPrefix,
+        InvalidEventAuthority,
+        InvalidProgramOwner,
+        MessageWithinRetentionWindow,
+    },
+    MessageData,
+};
+use solana_program::{
+    instruction::{
+        AccountMeta,
+        Instruction,
+    },
+    program::invoke_signed,
+    program_error::ProgramError,
+    pubkey::Pubkey,
+    sysvar::clock::Clock,
+};
+use solitaire::{
+    EVENT_AUTHORITY_SEED,
+    EVENT_IX_TAG_LE,
+    *,
+};
+
+/// 30 days in seconds.
+const RETENTION_PERIOD: i64 = 30 * 24 * 60 * 60;
+
+/// Event discriminator: SHA256("event:MessageAccountClosed")[0..8].
+pub const MESSAGE_ACCOUNT_CLOSED_DISCRIMINATOR: [u8; 8] =
+    [0x9e, 0xf6, 0x1b, 0xc2, 0x24, 0x14, 0x28, 0xb9];
+
+#[derive(FromAccounts)]
+pub struct ClosePostedMessage<'b> {
+    /// Bridge config, used to update last_lamports.
+    pub bridge: Mut<Bridge<'b, { AccountState::Initialized }>>,
+
+    /// The posted message account to close. Must have "msg" or "msu" prefix.
+    /// Uses Info to avoid Persist serializing back into a zeroed account.
+    pub message: Mut<Info<'b>>,
+
+    /// Fee collector PDA to receive the reclaimed lamports.
+    pub fee_collector: Mut<FeeCollector<'b>>,
+
+    /// Clock for timestamp validation.
+    pub clock: Sysvar<'b, Clock>,
+
+    /// Event authority PDA for Anchor CPI event signing.
+    pub event_authority: Info<'b>,
+
+    /// This program (for self-CPI).
+    pub self_program: Info<'b>,
+}
+
+#[derive(Default, BorshSerialize, BorshDeserialize)]
+pub struct ClosePostedMessageData {}
+
+pub fn close_posted_message(
+    ctx: &ExecutionContext,
+    accs: &mut ClosePostedMessage,
+    _data: ClosePostedMessageData,
+) -> Result<()> {
+    // 1. Verify account is owned by this program.
+    if accs.message.owner != ctx.program_id {
+        return Err(InvalidProgramOwner.into());
+    }
+
+    // 2-4. Parse and validate the message, extracting data for the event.
+    let message_data = {
+        let data = accs.message.data.borrow();
+        if data.len() < 3 {
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let prefix = &data[0..3];
+        if prefix != b"msg" && prefix != b"msu" {
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let message_data = MessageData::try_from_slice(&data[3..])
+            .map_err(|_| SolitaireError::ProgramError(ProgramError::InvalidAccountData))?;
+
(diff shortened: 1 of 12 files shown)


johnsaigle:
@claude 

(I think this is broken unfortunately, but giving it a whirl)

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
```diff
@@ -0,0 +1,161 @@
...
+        accs.guardian_set
+            .verify_derivation(ctx.program_id, &gs_derivation)?;
+
+        // The guardian set must NOT be active (i.e. it must be expired).
+        if check_active(&accs.guardian_set, &accs.clock).is_ok() {
+            return Err(GuardianSetStillActive.into());
+        }
+    }
+
+    // 6. Close signature_set: transfer lamports to fee_collector.
```
johnsaigle: I think we should also assign ownership to the system program here (assuming Solitaire allows this)

e.g. example for closing a token account https://github.com/solana-program/token/blob/84769ca25bbf67aed7712dc3a3acac16e7415e81/program/src/processor.rs?ref=blog.asymmetric.re#L1035-L1039

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,153 @@
...
+    // 5. Emit Anchor CPI event with the full message data.
+    emit_message_event(
+        ctx,
+        &accs.event_authority,
+        &accs.self_program,
+        &message_data,
+    )?;
+
+    // 6. Close the account: transfer lamports to fee_collector.
+    let message_lamports = accs.message.lamports();
```
johnsaigle, line 123: Same comment, should we assign ownership to System Program when we cleanup here?

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    // 2-4. Parse and validate the message, extracting data for the event.
+    let message_data = {
+        let data = accs.message.data.borrow();
+        if data.len() < 3 {
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let prefix = &data[0..3];
+        if prefix != b"msg" && prefix != b"msu" {
```
johnsaigle, line 70: Do we want to cleanup `msu` accounts (unreliable message accounts)? I think this will delete the per-emitter accounts that shim integrators use.

I would need to double-check but I think these could be re-created if they're deleted, but it could be disruptive.

If we do want to cleanup whatever old `msu` accounts are lying around, maybe we only close ones that have a data length > 0? This would preserve the shim-related `msu` accounts which should always be empty.

kcsongor: `msu` accounts get transparently created by the instruction https://github.com/wormhole-foundation/wormhole/blob/07a2d0979dac6ce6f99ac691461a9ba2cb5c37e6/solana/bridge/program/src/api/post_message.rs#L160-L172
so closing them shouldn't cause any issues

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
kcsongor: done

solana/bridge/program/src/api/close_posted_message.rs
kcsongor: done

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // Compute event authority PDA and verify.
+    let (expected_authority, bump) =
+        Pubkey::find_program_address(&[EVENT_AUTHORITY_SEED], ctx.program_id);
+    if event_authority.key != &expected_authority {
+        return Err(InvalidEventAuthority.into());
+    }
+
+    // Build instruction data: selector + discriminator + borsh(message_data).
+    let serialized = message_data
```
evan-gray: this may be ill-advised. I recommend logging the whole account, otherwise the watcher is unable to determine the difference between a reliable and unreliable message. there are explicit checks in the guardian that prevent an `unreliable` message with a zero payload from being observed via the account which should be maintained as well as use in the processor that limit their ability to trigger re-observations. the latter is less important but the former would require that we essentially never allow _either_ type to be reobserved using this event if it has a zero length payload, and I'm not sure that's the intention of the change.

evan-gray (dismissed):

solana/bridge/program/src/api/close_posted_message.rs
kcsongor: addressed in the latest commit, please take a look!

aleksiko17:
I staked my W coins a few months ago on the official website [https://w.wormhole.com](https://t.co/dGxb5dq6yZ) in the Solana network. Everything was displaying correctly before, but now my staked coins are no longer visible in the dashboard. What could be the reason?

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,150 @@
...
+        &accs.event_authority,
+        &accs.self_program,
+        &account_data,
+    )?;
+
+    // 6. Close the account: transfer lamports to fee_collector.
+    let message_lamports = accs.message.lamports();
+    **accs.fee_collector.lamports.borrow_mut() += message_lamports;
+    **accs.message.lamports.borrow_mut() = 0;
+    accs.message.data.borrow_mut().fill(0);
```
johnsaigle: For all the cases where accounts are closed, are you able to use `resize()` or something similar to ensure that the length of the data buffer is also set to 0?

`fill()` should fill the allocated data with zeroes but I think the account will still have the same length as it had before this instruction.

Other parts of the core bridge (in this PR and in the currently deployed bridge) check whether an account is "initialized" by doing a length check on the buffer, so even if the buffer is zeroed-out, those checks would consider the account to be initialized or to "already exist" in some sense.

I don't think there's a direct interaction between those checks and these accounts, but IMO it would be better to be consistent with this so that the mental model of the bridge can be simpler. Note also that the token program uses `resize()` for account closing -- I am not sure whether that method is available in this version of Solana, but conceptually we should aim for the same action.

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,153 @@
...
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let message_data = MessageData::try_from_slice(&data[3..])
+            .map_err(|_| SolitaireError::ProgramError(ProgramError::InvalidAccountData))?;
+
+        // Check retention: submission_time must be at or before (now - 30 days).
+        if (message_data.submission_time as i64) > accs.clock.unix_timestamp - RETENTION_PERIOD {
+            return Err(MessageWithinRetentionWindow.into());
+        }
```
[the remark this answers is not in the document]
mdulin2: Non-issue. `post-message` always sets this to the proper time.

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
```diff
@@ -0,0 +1,164 @@
...
+    accs: &mut CloseSignatureSetAndPostedVAA,
+    _data: CloseSignatureSetAndPostedVAAData,
+) -> Result<()> {
+    // 1. Verify signature_set is owned by this program.
+    if accs.signature_set.owner != ctx.program_id {
+        return Err(InvalidProgramOwner.into());
+    }
+
+    // 2. Parse SignatureSetData.
+    let sig_data = {
```
mdulin2, line 87: The `try_from_slice()` function will not parse bytes into a data structure if there are too few or too many bytes. Because we lack discriminators on all but the VAA and post message accounts, this is really important for security. This length check, in combination with the `hash` check below, makes it impossible to spoof from what I can tell. The lack of discriminators is scary for arbitrary account closure attacks.

Could you add a test to confirm the length-check functionality? Having this would be helpful for future regression testing.

mdulin2: It might be good to document the constraints for the SignatureSet too. For instance: must parse into a sigantureset type, must have a valid expired guardian set index/have a hash associated with a valid VAA, and must be owned by the core bridge program. Given the weaving of security controls that make this secure, documentation is our friend.

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,150 @@
...
+    accs.message.assign(&solana_program::system_program::id());
+
+    // 7. Update bridge.last_lamports to prevent fee waiver.
+    accs.bridge.last_lamports = accs.fee_collector.lamports();
+
+    Ok(())
+}
+
+/// Emit an Anchor-compliant CPI event containing the full account data.
+fn emit_message_event(
```
mdulin2: Does it make sense to add this as a helper function to Solitaire instead of here as a standalone call? Seeing the `self_program` check, event authority derivation, and the event construction process, this seems like a good candidate to separate out.

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
```diff
@@ -0,0 +1,164 @@
...
+    let expected_vaa_seeds_ref: Vec<&[u8]> =
+        expected_vaa_seeds.iter().map(|s| s.as_slice()).collect();
+    let (expected_vaa_key, _) =
+        Pubkey::find_program_address(&expected_vaa_seeds_ref, ctx.program_id);
+    if accs.posted_vaa.key != &expected_vaa_key {
+        return Err(InvalidDerivedAccount.into());
+    }
+
+    // 4. Determine if PostedVAA is initialised
+    let vaa_initialized = accs.posted_vaa.data_len() > 0 && accs.posted_vaa.owner == ctx.program_id;
```
johnsaigle: Could you add a > 0 lamports check here as well? IMO it would be best to be consistent with our definition of an account being "initialized" or "closed" in the core bridge.

solana/bridge/program/tests/integration.rs
```diff
@@ -1578,3 +1599,608 @@ async fn upgrade_contract() {
...
+        &mut test_ctx.banks_client,
+        program,
+        &test_ctx.payer,
+        message_key.pubkey(),
+    )
+    .await
+    .unwrap();
+
+    // Verify account is closed.
+    let data = common::get_account_data_raw(&mut test_ctx.banks_client, message_key.pubkey()).await;
```
johnsaigle, line 1941: Could you add the same post-close checks in this test case as you have in the reliable case? e.g. that the balances have been updated etc.

solana/bridge/program/tests/integration.rs
```diff
@@ -1578,3 +1599,608 @@ async fn upgrade_contract() {
...
+        &test_ctx.payer,
+        signature_set,
+        posted_vaa_key,
+        0,
+    )
+    .await
+    .unwrap();
+
+    // Verify fee_collector received all lamports.
+    let fee_collector_after =
```
johnsaigle, line 2031: Can you add a check for the bridge's last lamports as well?

solana/bridge/program/src/api/close_posted_message.rs
```diff
@@ -0,0 +1,150 @@
...
+        if data.len() < 3 {
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let prefix = &data[0..3];
+        if prefix != b"msg" && prefix != b"msu" {
+            return Err(InvalidAccountPrefix.into());
+        }
+
+        let message_data = MessageData::try_from_slice(&data[3..])
```
mdulin2, line 74: Are there any other data validations that we can get to prevent type-cosplay issues? For instance, the emitter address is non-zero, vaa_time is zero, submission_time is non-zero, etc.

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fee_collector: Mut<FeeCollector<'b>>,
+
+    /// Clock for timestamp validation.
+    pub clock: Sysvar<'b, Clock>,
+}
+
+#[derive(Default, BorshSerialize, BorshDeserialize)]
+pub struct CloseSignatureSetAndPostedVAAData {}
+
+pub fn close_signature_set_and_posted_vaa(
```
mdulin2, line 76: Are there any other data validations we can implement to prevent type-cosplay issues for both accounts? I'm afraid of adding a new account in the future that can be closed through this function. The more checks we have, the less likely arbitrary-close bugs will happen.

solana/solitaire/program/src/macros.rs
```diff
@@ -24,9 +24,50 @@ macro_rules! trace_impl {
...
+#[macro_export]
+macro_rules! solitaire_event_guard {
+    (@event_cpi, $p:ident, $a:ident, $d:ident) => {
+        // Handle Anchor event CPI (self-invocation for event emission).
+        if $d.len() >= 8 && $d[..8] == $crate::EVENT_IX_TAG_LE {
+            if $a.is_empty() || !$a[0].is_signer {
+                return Err(SolitaireError::InvalidSigner(
+                    solana_program::pubkey::Pubkey::default(),
+                ));
+            }
```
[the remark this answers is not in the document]
kcsongor: done

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        if (submission_time as i64) > accs.clock.unix_timestamp - RETENTION_PERIOD {
+            return Err(MessageWithinRetentionWindow.into());
+        }
+
+        // Close PostedVAA: transfer lamports to fee_collector.
+        let vaa_lamports = accs.posted_vaa.lamports();
+        **accs.fee_collector.lamports.borrow_mut() += vaa_lamports;
+        **accs.posted_vaa.lamports.borrow_mut() = 0;
+        accs.posted_vaa.data.borrow_mut().fill(0);
```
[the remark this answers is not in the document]
kcsongor: I did this, although it's a bit of a silly check here

solana/bridge/program/src/api/close_posted_message.rs
kcsongor: good point, done

solana/bridge/program/src/api/close_posted_message.rs
kcsongor: moved out to a helper

solana/bridge/program/tests/integration.rs
kcsongor: done

solana/bridge/program/src/api/close_signature_set_and_posted_vaa.rs
kcsongor: I added a couple more checks

johnsaigle (approved):

mdulin2 (approved):

PR #4747, opened 2026-04-10
node: Support multiple transfers per tx in the Accountant audit flow

djb15:
The Accountant correctly tracks pending transfers by full Wormhole message ID, but its audit recovery path temporarily re-keys those transfers by `(emitterChain, txHash)` instead. That key is not unique when a single EVM transaction emits multiple covered token-bridge messages.

We make a change here to support the audit flow for multiple transfers in a single transaction.

### DIFF
--- a/node/pkg/accountant/audit.go
+++ b/node/pkg/accountant/audit.go
@@ -129,22 +129,33 @@ func (acct *Accountant) audit(ctx context.Context) error {
 // runAudit is the entry point for the audit of the pending transfer map. It creates a temporary map of all pending transfers and invokes the main audit function.
 func (acct *Accountant) runAudit() {
 	tmpMap := acct.createAuditMap(false)
-	acct.logger.Debug("in AuditPendingTransfers: starting base audit", zap.Int("numPending", len(tmpMap)))
+	acct.logger.Debug("in AuditPendingTransfers: starting base audit", zap.Int("numPending", auditMapSize(tmpMap)))
 	acct.performAudit(tmpMap, acct.wormchainConn, acct.contract)
 	acct.logger.Debug("in AuditPendingTransfers: finished base audit")
 
 	tmpMap = acct.createAuditMap(true)
-	acct.logger.Debug("in AuditPendingTransfers: starting ntt audit", zap.Int("numPending", len(tmpMap)))
+	acct.logger.Debug("in AuditPendingTransfers: starting ntt audit", zap.Int("numPending", auditMapSize(tmpMap)))
 	acct.performAudit(tmpMap, acct.nttWormchainConn, acct.nttContract)
 	acct.logger.Debug("in AuditPendingTransfers: finished ntt audit")
 }
 
+// auditMapSize returns the total number of pending entries across all keys in the audit map.
+func auditMapSize(tmpMap map[string][]*pendingEntry) int {
+	n := 0
+	for _, entries := range tmpMap {
+		n += len(entries)
+	}
+	return n
+}
+
 // createAuditMap creates a temporary map of all pending transfers. It grabs the pending transfer lock.
-func (acct *Accountant) createAuditMap(isNTT bool) map[string]*pendingEntry {
+// The map is keyed by chain-txHash (the audit key). Multiple transfers may share the same key if they
+// originate from the same transaction, so each key maps to a slice of pending entries.
+func (acct *Accountant) createAuditMap(isNTT bool) map[string][]*pendingEntry {
 	acct.pendingTransfersLock.Lock()
 	defer acct.pendingTransfersLock.Unlock()
 
-	tmpMap := make(map[string]*pendingEntry)
+	tmpMap := make(map[string][]*pendingEntry)
 	for _, pe := range acct.pendingTransfers {
 		if pe.isNTT == isNTT {
 			if pe.hasBeenPendingForTooLong() {
@@ -153,7 +164,7 @@ func (acct *Accountant) createAuditMap(isNTT bool) map[string]*pendingEntry {
 			}
 			key := pe.makeAuditKey()
 			acct.logger.Debug("will audit pending transfer", zap.String("msgId", pe.msgId), zap.String("moKey", key), zap.Bool("submitPending", pe.submitPending()), zap.Stringer("lastUpdateTime", pe.updTime()))
-			tmpMap[key] = pe
+			tmpMap[key] = append(tmpMap[key], pe)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


node/pkg/accountant/audit.go
```diff
@@ -197,18 +212,20 @@ func (acct *Accountant) performAudit(tmpMap map[string]*pendingEntry, wormchainC
 		}
 	}
 
-	if len(tmpMap) != 0 {
+	if auditMapSize(tmpMap) != 0 {
```
mdulin2: Do we actually need the real size to be calculated? My understanding is that this code should only run if there's a pending observation in the `tmpMap`. Couldn't you just use `len(tmpMap)` for this, for a binary, are there values or not? If there's a key in the map, we should iterate over it.

mdulin2 (dismissed):

node/pkg/accountant/audit.go
```diff
@@ -129,22 +129,33 @@ func (acct *Accountant) audit(ctx context.Context) error {
...
 
 	tmpMap = acct.createAuditMap(true)
-	acct.logger.Debug("in AuditPendingTransfers: starting ntt audit", zap.Int("numPending", len(tmpMap)))
+	acct.logger.Debug("in AuditPendingTransfers: starting ntt audit", zap.Int("numPending", auditMapSize(tmpMap)))
 	acct.performAudit(tmpMap, acct.nttWormchainConn, acct.nttContract)
 	acct.logger.Debug("in AuditPendingTransfers: finished ntt audit")
 }
 
+// auditMapSize returns the total number of pending entries across all keys in the audit map.
+func auditMapSize(tmpMap map[string][]*pendingEntry) int {
```
johnsaigle: Not sure about this implementation. It's not immediately clear to me what the size of a map should be given that the types here make it so that each key has 0 or more values.

It opens weird edge cases too (on the level of the type, not these particular callers). For example, you could have 5 keys each with a slice of 0 length, and this function would yield a size of `0`. On the other hand, it would be possible to add `100` instances of `nil` to a single key and get a length of `100` even if none of the entries are valid.

I get not wanting to scope-creep here, but minimally I'd suggested adding `nil` guards here and maybe changing the name to like "NumValuesInMap" or something, plus unit tests to avoid the degen cases I described.

Another thing to look into is https://pkg.go.dev/maps#Values in case that can address your problem.

node/pkg/accountant/audit.go
johnsaigle: Agreed, this is a semantic change from "does the map have any keys?" to "does the map have any values?". I think we want the former, so `len` is more correct given the current implementation.

node/pkg/accountant/audit.go
```diff
@@ -197,18 +212,20 @@ func (acct *Accountant) performAudit(tmpMap map[string]*pendingEntry, wormchainC
...
 	}
 
-	if len(tmpMap) != 0 {
+	if auditMapSize(tmpMap) != 0 {
 		var keys []TransferKey
 		var pendingTransfers []*pendingEntry
-		for _, pe := range tmpMap {
-			keys = append(keys, TransferKey{EmitterChain: uint16(pe.msg.EmitterChain), EmitterAddress: pe.msg.EmitterAddress, Sequence: pe.msg.Sequence})
-			pendingTransfers = append(pendingTransfers, pe)
+		for _, entries := range tmpMap {
```
johnsaigle, line 235: I think this section could be rewritten with https://pkg.go.dev/maps#Values

johnsaigle (changes requested):

node/pkg/accountant/audit.go
djb15: Fair, fixed in the latest commit

node/pkg/accountant/audit.go
djb15: Thanks for the suggestion, added `nil` guards and renamed for clarity

node/pkg/accountant/audit.go
djb15: Decided not to do this, I don't think it would massively change the readability

node/pkg/accountant/audit.go
```diff
@@ -169,26 +185,30 @@ func (pe *pendingEntry) hasBeenPendingForTooLong() bool {
...
-func (acct *Accountant) performAudit(tmpMap map[string]*pendingEntry, wormchainConn AccountantWormchainConn, contract string) {
+func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchainConn AccountantWormchainConn, contract string) {
 	acct.logger.Debug("entering performAudit", zap.String("contract", contract))
 	missingObservations, err := acct.queryMissingObservations(wormchainConn, contract)
 	if err != nil {
 		acct.logger.Error("unable to perform audit, failed to query missing observations", zap.Error(err))
-		for _, pe := range tmpMap {
-			acct.logger.Error("unsure of status of pending transfer due to query error", zap.String("msgId", pe.msgId))
+		for _, entries := range tmpMap {
+			for _, pe := range entries {
```
johnsaigle, line 198: There should be a nil check on `pe` so that the next line has no risk of nil dereference.

node/pkg/accountant/audit.go
```diff
@@ -129,22 +129,38 @@ func (acct *Accountant) audit(ctx context.Context) error {
...
 // createAuditMap creates a temporary map of all pending transfers. It grabs the pending transfer lock.
-func (acct *Accountant) createAuditMap(isNTT bool) map[string]*pendingEntry {
+// The map is keyed by chain-txHash (the audit key). Multiple transfers may share the same key if they
+// originate from the same transaction, so each key maps to a slice of pending entries.
+func (acct *Accountant) createAuditMap(isNTT bool) map[string][]*pendingEntry {
 	acct.pendingTransfersLock.Lock()
 	defer acct.pendingTransfersLock.Unlock()
 
-	tmpMap := make(map[string]*pendingEntry)
+	tmpMap := make(map[string][]*pendingEntry)
```
johnsaigle, line 163: Can you add a nil check on `pe` in this loop so that there's no risk nil dereference? I know this isn't code you wrote but I think it's a nice addition while we're here.

node/pkg/accountant/audit.go
```diff
@@ -200,15 +220,22 @@ func (acct *Accountant) performAudit(tmpMap map[string]*pendingEntry, wormchainC
...
+				keys = append(keys, TransferKey{EmitterChain: uint16(pe.msg.EmitterChain), EmitterAddress: pe.msg.EmitterAddress, Sequence: pe.msg.Sequence})
+				pendingTransfers = append(pendingTransfers, pe)
+			}
 		}
 
 		transferDetails, err := acct.queryBatchTransferStatus(keys, wormchainConn, contract)
 		if err != nil {
 			acct.logger.Error("unable to finish audit, failed to query for transfer statuses", zap.Error(err))
-			for _, pe := range tmpMap {
+			for _, pe := range pendingTransfers {
```
johnsaigle, line 250: Same here

node/pkg/accountant/audit.go
```diff
@@ -169,26 +185,30 @@ func (pe *pendingEntry) hasBeenPendingForTooLong() bool {
...
 		key := mo.makeAuditKey()
-		pe, exists := tmpMap[key]
+		entries, exists := tmpMap[key]
 		if exists {
-			if acct.submitObservation(pe) {
-				auditErrors.Inc()
-				acct.logger.Error("contract reported pending observation as missing, resubmitted it", zap.String("msgID", pe.msgId))
-			} else {
-				acct.logger.Info("contract reported pending observation as missing but it is queued up to be submitted, skipping it", zap.String("msgID", pe.msgId))
+			for _, pe := range entries {
```
johnsaigle, line 213: And here

johnsaigle (commented):
There are some places where it seems like `nil` is possible based on using a pointer type, so my suggestion is to handle these explicitly. I'm not sure if there's a better place to do this upstream or whether it's worth filtering out nil before we hit these processing steps.

johnsaigle (approved):

mdulin2 (approved):

PR #4717, opened 2026-03-12
node: Validate VAA key format in GetAndObserveMissingVAAs

djb15:
Resolves https://github.com/asymmetric-research/wormhole-llm-issues/issues/110.

PR #4752, opened 2026-04-15
sdk: Add VAA ID validation and use it everywhere

johnsaigle:
# Before (example)

```go
// Before: a maintainer had to manually parse the string, perform type
// conversions, decode hex, validate lengths, and build the canonical ID.
parts := strings.Split(msgID, "/")
if len(parts) != 3 {
	log.Fatalf("invalid message ID")
}
chain, err := strconv.ParseUint(parts[0], 10, 16)
if err != nil {
	log.Fatalf("invalid chain ID: %v", err)
}
addrBytes, err := hex.DecodeString(parts[1])
if err != nil {
	log.Fatalf("invalid emitter address: %v", err)
}
if len(addrBytes) != 32 {
	log.Fatalf("address must be 32 bytes")
}
var addr vaa.Address
copy(addr[:], addrBytes)
seq, err := strconv.ParseUint(parts[2], 10, 64)
if err != nil {
	log.Fatalf("invalid sequence: %v", err)
}
id := guardianDB.VAAID{
	EmitterChain:   vaa.ChainID(chain),
	EmitterAddress: addr,
	Sequence:       seq,
}
key := []byte(fmt.Sprintf("signed/%d/%s/%d", id.EmitterChain, id.EmitterAddress, id.Sequence))
```

# After (example)

```go
// After: all they need to do is parse once and use the typed ID everywhere.
id, err := vaa.VAAIDFromString(msgID)
if err != nil {
	log.Fatalf("invalid message ID %q: %v", msgID, err)
}
key := []byte("signed/" + id.String())
```

# Description

VAA and chain ID validation is replicated in a ton of places all over the Guardian code. These validations are so important for Wormhole's security that they are necessary to do across various trust boundaries.

This PR makes use of the VAA type within the SDK to do as much validation in a standardized way as possible using focused, well-tested implementations.

The PR fixes at least one existing bug in the Guardian: `GetAndObserveMissingVAAs` was incorrectly parsing the emitter value for VAAs, causing the DB to always miss stored VAAs. This occurred because the emitter address was converted into a byte slice according to its (ASCII) string representation and not its hex representation. (So e.g. the character `0` was represented as `0x30` rather than `0x00`)
It should also address problems like https://github.com/wormhole-foundation/wormhole/pull/4717

The goals of this refactoring is:
- Make it easier for maintainers to validate these crucial data types
- Hide implementation details for downstream code (not every maintainer needs to check for uint16 overflows)
- Allow code reviewers to trust that VAA and ChainID validation is done properly rather than the current status quo which is to cross-reference across e.g. all of the watchers for compliance
- Remove duplicated ad-hoc validation and formatting logic that has subtle differences between implementations
- Reduce the burden needed to test the ad-hoc validations
- Absolutely no existing behaviour should be modified. The "loose" versions of the SDK methods were used in order to preserve current behaviour rather than make validation more strict during refactoring

### DIFF
--- a/node/cmd/guardiand/adminclient.go
+++ b/node/cmd/guardiand/adminclient.go
@@ -811,15 +811,15 @@ func buildDelegateSignaturesBroadcasts(vaaID string, apiObservations []wormholes
 			d := &gossipv1.DelegateObservation{
 				Timestamp:         timestamp,
 				Nonce:             ref.Nonce,
-				EmitterChain:      uint32(chainID), // #nosec G115
+				EmitterChain:      uint32(chainID),
 				EmitterAddress:    emitterAddrBytes,
 				Sequence:          ref.Sequence,
-				ConsistencyLevel:  uint32(ref.ConsistencyLevel), // #nosec G115
+				ConsistencyLevel:  uint32(ref.ConsistencyLevel),
 				Payload:           payloadBytes,
 				TxHash:            txHashBytes,
 				Unreliable:        obs.Unreliable,
 				IsReobservation:   obs.IsReobservation,
-				VerificationState: uint32(obs.VerificationState), // #nosec G115
+				VerificationState: uint32(obs.VerificationState),
 				GuardianAddr:      guardianAddrBytes,
 				SentTimestamp:     sentTimestamp,
 			}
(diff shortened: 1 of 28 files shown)


node/cmd/guardiand/adminclient.go
```diff
@@ -829,7 +811,7 @@ func buildDelegateSignaturesBroadcasts(vaaID string, apiObservations []wormholes
 			d := &gossipv1.DelegateObservation{
 				Timestamp:         timestamp,
 				Nonce:             ref.Nonce,
-				EmitterChain:      uint32(chainIDParsed), // #nosec G115
+				EmitterChain:      uint32(chainID), // #nosec G115
```
djb15: Given you were able to remove the `#nosec G115` above, it's worth trying to remove here and below if you can

djb15 (commented):
Nice! Spent some time poking around, I think all the changes are solid

djb15 (dismissed):
Is the publicrpc unit test deliberate in this PR?

johnsaigle:
Yeah the refactoring added lines which caused a reduction in code coverage as a percentage of the file. I added some extra unit tests to make up for that. Incentives are working well 😄

PR #4762, opened 2026-04-21
node: April Governor limits update

djb15:


node/pkg/governor/mainnet_chains_test.go
```diff
@@ -73,6 +73,11 @@ func TestChainListBigTransfers(t *testing.T) {
 			continue
 		}
 
+		// If the chain has a daily limit of $10,000 or less we don't make the same contraints on big tx size
+		if e.DailyLimit <= 10_000 {
```
johnsaigle, line 77: Maybe we can change the assertion below to "assert less than or equal"?

This would allow us to set a big tx limit of `10_000_000` even if the daily limit was `1`

johnsaigle (approved):

pleasew8t (approved):

PR #4764, opened 2026-04-23
node(governor): Remove Linea

johnsaigle:
The WTT instance on Linea has never been cross registered to any of the other WTT instances, so no token transfers are possible


bemic (approved):

djb15 (changes requested):
I think we should instead set the limits to 0 rather than removing them altogether. At least then if it were registered in the future the WTT messages would be automatically governed, rather than skipping the Governor

djb15 (commented):
Alternatively I would remove the token bridge from the mainnet list in the SDK. But I don't think we should keep it there but not have Governor limits

johnsaigle:
> would be automatically governed

There are no tokens registered so this wouldn't work unless we also had either the manual or generated tokens lists populated with some Linea assets. When a PR would go through to register those tokens, it would fail due to this test: https://github.com/wormhole-foundation/wormhole/blob/main/node/pkg/governor/mainnet_tokens_test.go#L39. So if tokens get added in the future, we'd be forced to update the Governor's chains in the same PR.

I think it's probably fine to remove the Linea token bridge from the SDK though if that's what you want to do.

djb15 (approved):
Thanks for linking the test, happy that it'll cover any future registration changes that happen without us actively being in the loop

PR #4765, opened 2026-04-24
Dockerize go lint and tests

johnsaigle:
Wormhole is using go 1.24 as well as a slightly out of date version of
golangci-lint. This can make it annoying to run unit tests, code
coverage tools, or linting locally assuming that a dev wants to use more
recent toolchains on their system (e.g. by installing `go` or `golangci-lint`
via something like brew or apt).

This commit
- adds a new Dockerfile for running unit tests and test coverage tools
- modifies the local Dockerfile for linting to be up-to-date (its golangci-lint version was very out-of-date)
- adds cspell to that container
- adds `make` targets for each of these
- updates `goimports` to use the latest version that supports Go 1.24

All of this should make it easier for devs to run tests/linting locally
without needing to specify extra flags like `GOTOOLCHAIN=1.24.13`

It also adds a few more safe entries to `.dockerignore`, though this
should not affect the new Dockerfiles or the other ones in the repo,
given that they don't use `COPY .`

This commit also slightly bumps the sleep duration in the altpub tests
as it is more likely to run too slowly to complete properly within a
container. Ideally the test would be refactored to not use sleep, but
that's not in-scope.

---

See also #4701 as this PR is now failing with lints related to that version.


johnsaigle:
Note: this PR should ideally be kept in sync with the golangci-lint version here: #4701

Dockerfile.lint
```diff
@@ -0,0 +1,29 @@
...
+ENV GOPATH=/tmp/go
+ENV GOCACHE=/tmp/go-build
+ENV PATH=/tmp/go/bin:$PATH
+
+COPY --from=node-tools /usr/local/bin/node /usr/local/bin/node
+COPY --from=node-tools /usr/local/lib/node_modules /usr/local/lib/node_modules
+
+RUN ln -s /usr/local/lib/node_modules/cspell/bin.mjs /usr/local/bin/cspell
+
+RUN useradd -u 1000 -U -m -d /home/lint lint && mkdir -p /tmp/go /tmp/go-build && chmod -R 1777 /tmp/go /tmp/go-build
```
pleasew8t, line 22: nit: Mind just adding a one-liner comment describing the command. Could probably split it up into three commands.

Dockerfile.unit
```diff
@@ -0,0 +1,9 @@
+# syntax=docker.io/docker/dockerfile:1.3@sha256:42399d4635eddd7a9b8a24be879d2f9a930d0ed040a61324cfdf59ef1357b3b2
+FROM docker.io/golang:1.24.13-bookworm@sha256:1a6d4452c65dea36aac2e2d606b01b4a029ec90cc1ae53890540ce6173ea77ac
+
+ENV GOPATH=/tmp/go
+ENV GOCACHE=/tmp/go-build
+
+RUN mkdir -p /tmp/go /tmp/go-build && chmod -R 1777 /tmp/go /tmp/go-build
```
pleasew8t, line 12: nit: Add a comment

pleasew8t (dismissed):
Ran both the lint and unit test Makefile commands, and it works fine. It would be nice to have the lint command clean itself up (I see the unit test docker command uses `--rm`, can lint do this as well?).

.github/workflows/build.yml
```diff
@@ -303,7 +303,7 @@ jobs:
         with:
           go-version: "1.24.13"
       - name: Install formatter
-        run: go install golang.org/x/tools/cmd/goimports@v0.8.0
+        run: go install golang.org/x/tools/cmd/goimports@v0.42.0
```
djb15: Should probably target the commit hash instead whilst you're updating this

Dockerfile.lint
```diff
@@ -0,0 +1,34 @@
...
+# Direct Go caches to /tmp with world-writable perms so goimports/golangci-lint
+# can download modules at runtime regardless of user.
+RUN useradd -u 1000 -U -m -d /home/lint lint && mkdir -p /tmp/go /tmp/go-build && chmod -R 1777 /tmp/go /tmp/go-build
+
+# install goimports (pinned to match CI)
+RUN go install golang.org/x/tools/cmd/goimports@v0.42.0
+
+# install golangci-lint (pinned to match CI)
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+	sh -s -- -b $(go env GOPATH)/bin v2.8.0
```
djb15: I've always just used the tagged version when installing `golangci-lint` but wondering whether this also supports commit hashes?

djb15 (commented):
Minor suggestions, looks ok otherwise

Dockerfile.lint
johnsaigle: I think it does yeah. This install method is actually out of date and I had to fix it in #4701 anyway, so I'll update here too.

johnsaigle:
Converting to draft mode as this has potential sequencing conflicts with other open PRs. See https://github.com/wormhole-foundation/wormhole/pull/4701#issuecomment-4576517471 for more details. cc: @mdulin2

PR #4767, opened 2026-04-28
feat(pauser): design

evan-gray:
This PR is 1/6 implementing a Delegated Pauser functionality for Wormhole and WTT (the Token Bridge).

This includes the whitepaper (design) only.

- Design #4767
- Guardian #4769
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802
- EVM Pauser #4768
- SVM Pauser #4778

### DIFF
--- a/whitepapers/0018_pauser.md
+++ b/whitepapers/0018_pauser.md
@@ -1,0 +1,198 @@
+# WTT and NTT Pauser
+
+## Objective
+
+Develop a mechanism for a set of keys designated by the guardians that can be granted the authority to pause and unpause Wormhole protocols, such as WTT (Token Bridge) and NTT.
+
+## Background
+
+Wormhole protocols operate with a number of defense-in-depth measures, but during an active exploit each protection or action has a different scope and time to implement, including:
+
+- Disconnecting RPCs - a super-minority of guardians can disconnect the chain from the guardian node to stop new messages from generating VAAs.
+- [Governor](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0007_governor.md) - WTT transfers may be made subject to a delay. This limit can be lowered further in response to an exploit but requires a super-minority of guardians to take effect.
+- [Global Accountant](https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0011_accountant.md) - WTT transfers and some NTT deployments’ transfers are first checked against the Global Accountant which ensures that the number of wrapped tokens capable of being moved out of a chain never exceeds the number sent to that chain.
+- Contract Upgrade - WTT contracts and some NTT deployments are administered by guardian governance which can perform upgrades to mitigate attacks or disable functionality. This requires code changes, review, and governance by a super-majority of guardians.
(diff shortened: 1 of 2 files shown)


whitepapers/0018_pauser.md
```diff
@@ -0,0 +1,246 @@
+# WTT and NTT Pauser
+
+## Objective
+
+Develop a mechanism for a set of keys designated by the guardians can be granted the authority to pause and unpause Wormhole protocols, such as WTT (Token Bridge) and NTT.
```
johnsaigle:
```suggestion
Develop a mechanism for a set of keys designated by the guardians that can be granted the authority to pause and unpause Wormhole protocols, such as WTT (Token Bridge) and NTT.
```

whitepapers/0018_pauser.md
```diff
@@ -0,0 +1,246 @@
...
+
+This change may increase cost on some platforms, such as EVM, by introducing a new state read and check on every call.
+
+### Backwards Compatibility
+
+This change should be fully backwards compatible on EVM and SVM for both WTT and NTT. For WTT, Solana already passes the `config` on all relevant calls and NTT already incorporates a `pauser` role.
+
+### Unpauser Contract
+
+This design does not introduce an Unpauser delegation at this time, as the Wormhole Governance can serve that higher threshold.
```
johnsaigle, line 151: So long as Wormhole Governance is the unpauser, I think that avoiding the 'temporary pause' makes sense. However, if the unpauser role changes to another contract which might have a bug or else a different set of actors, it might be good to revisit the idea of a temporary pause. Essentially, there's a small but IMO meaningful risk that a bad unpause implementation could brick the paused contract.

I'm not sure how to capture this in the whitepaper exactly, but I think that the idea of the role of the Unpauser is inherently linked to the decision about whether the pause is temporary.

evan-gray: but isn't that mitigated by being able to rotate the pauser / unpauser via governance anyway?

johnsaigle: Yeah fair

whitepapers/0018_pauser.md
```diff
@@ -0,0 +1,246 @@
...
+
+#### WTT Governance
+
+```solidity
+// Wire header:
+//   module:   "TokenBridge" left-padded to 32 bytes
+//   action:   uint8 - 4 = SetPauserAddressesEvm, 5 = SetPauserAddressesSolana
+//   chainId:  uint16 target chain (the chain whose token bridge is being configured)
+
+// Action 4: EVM token bridge pauser config (20-byte addresses)
```
johnsaigle: Action 3 is missing?

evan-gray: no, this is unfortunate semantics but action 1 and 2 here belong to `DelegatedPauser` (which is why they're in the `Delegated Pauser Governance` section), while 4 and 5 belong to `TokenBridge`, which already has 1, 2, and 3 (and why it's under `WTT Governance` here).

johnsaigle: Ah I see, thanks for explaining.

johnsaigle (dismissed):

whitepapers/0018_pauser.md
```diff
@@ -0,0 +1,246 @@
...
+4. Cancel an approval
+   1. MUST verify that the caller is part of the signing set
+   2. MUST require that the proposal is active
+   3. MUST verify that the caller has already approved
+   4. MUST unmark the caller as having approved and decrement the approval count
+   5. MUST emit an event indicating that the caller has cancelled their approval of this proposal
+
+#### WTT (Token Bridge)
+
+1. Add a `notPaused` check to every entry point that is not governance or `unpause`
```
johnsaigle: Just for clarity, this means that actions like the following are allowed while paused:

* registerChain
* upgrade
* submitRecoverChainId

I think this is probably OK since they are gated by Guardian Governance, similar to pausing and unpausing.

Maybe worth adding under the "Alternatives Considered" section.

johnsaigle: The implementations also omit a `notPaused` check for `pause()` because it is idempotent.

PR #4769, opened 2026-04-28
feat(node): pauser governance

evan-gray:
This PR is 2/6 implementing a Delegated Pauser functionality for Wormhole and WTT (the Token Bridge).

This includes the guardian governance message implementation only.

- Design #4767
- Guardian #4769
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802
- EVM Pauser #4768
- SVM Pauser #4778


sdk/vaa/payloads.go
```diff
@@ -683,6 +736,50 @@ func (r *BodyManagerSetUpdate) Deserialize(bz []byte) error {
 	return nil
 }
 
+func (r BodyDelegatedPauserSetConfigEvm) Serialize() ([]byte, error) {
+	if len(r.Signers) > math.MaxUint8 {
```
johnsaigle, line 789: We could add a 'known ChainID' check here, as well as check for duplicated signers.

On the other hand, if the idea is that this is only a low-level serialization function, it could be documented that these sorts of checks are explicitly out-of-scope.

node/pkg/adminrpc/adminserver.go
```diff
@@ -872,6 +872,193 @@ func suiCallToVaa(suiCall *nodev1.SuiCall, timestamp time.Time, guardianSetIndex
 	return v, nil
 }
 
+// delegatedPauserSetConfigEvmToVaa builds the SetConfigEvm governance VAA for the WormholePauser contract.
+// See whitepapers/0018_pauser.md.
+func delegatedPauserSetConfigEvmToVaa(req *nodev1.DelegatedPauserSetConfigEvm, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+	if req.ChainId == 0 || req.ChainId > math.MaxUint16 {
```
johnsaigle: Could you refactor these checks to the SDK's known chain ID function?

Same for the other instances.

node/pkg/adminrpc/adminserver.go
```diff
@@ -872,6 +872,193 @@ func suiCallToVaa(suiCall *nodev1.SuiCall, timestamp time.Time, guardianSetIndex
...
+	}
+	index := uint16(req.Index)
+
+	if req.Threshold == 0 || req.Threshold > math.MaxUint8 {
+		return nil, errors.New("invalid threshold (must be > 0 and <= uint8 max)")
+	}
+	threshold := uint8(req.Threshold)
+
+	if req.ExpiryDuration == 0 {
+		return nil, errors.New("invalid expiry_duration (must be > 0)")
```
johnsaigle, line 937: Maybe it should be at least an hour? Same for the below instances.

node/pkg/adminrpc/adminserver.go
```diff
@@ -872,6 +872,193 @@ func suiCallToVaa(suiCall *nodev1.SuiCall, timestamp time.Time, guardianSetIndex
...
```
johnsaigle, line 996: Same comment here on duration -- maybe an hour makes sense as a minimum?

johnsaigle:
@claude

node/cmd/guardiand/admintemplate.go
```diff
@@ -1620,3 +1678,237 @@ func randSeqNonce() (uint64, uint32) {
...
+	}
+	if len(out) == 0 {
+		log.Fatal("at least one signer is required")
+	}
+	return out
+}
+
+// parseEvmHexAddress decodes a 20-byte EVM address from `raw`, accepting an optional 0x prefix.
+// The returned string is the canonical 40-char lowercase hex with no prefix.
+func parseEvmHexAddress(raw string) (string, error) {
```
johnsaigle, line 1743: This should be replaced with something like `IsHexAddress()` from go-ethereum.

node/cmd/guardiand/admintemplate.go
```diff
@@ -1620,3 +1678,237 @@ func randSeqNonce() (uint64, uint32) {
...
+		out = append(out, s)
+	}
+	if len(out) == 0 {
+		log.Fatal("at least one signer is required")
+	}
+	return out
+}
+
+// parseEvmHexAddress decodes a 20-byte EVM address from `raw`, accepting an optional 0x prefix.
+// The returned string is the canonical 40-char lowercase hex with no prefix.
```
johnsaigle, line 1742: Using the checksummed address is probably better. I wouldn't consider all-lowercase addresses to be canonical

node/cmd/guardiand/admintemplate.go
```diff
@@ -94,6 +94,17 @@ var delegatedManagerThreshold *string
 var delegatedManagerNumKeys *string
 var delegatedManagerPublicKeys *string
 
+var delegatedPauserChainId *string
+var delegatedPauserIndex *string
+var delegatedPauserThreshold *string
+var delegatedPauserExpiryDuration *string
+var delegatedPauserSigners *string
```
johnsaigle, line 101: I think it would be better to use `*[]string` here and cobra.Command's built-in parsing for arrays on CLI rather than implement custom parsing logic and unit tests. Otherwise this file is going to keep ballooning

node/cmd/guardiand/admintemplate.go
```diff
@@ -94,6 +94,17 @@ var delegatedManagerThreshold *string
 var delegatedManagerNumKeys *string
 var delegatedManagerPublicKeys *string
 
+var delegatedPauserChainId *string
```
johnsaigle, line 97: Probably better to use the smallest-possible int types for these variables rather than strings so you can skip validation later.

node/cmd/guardiand/admintemplate_test.go
```diff
@@ -0,0 +1,173 @@
...
+	})
+
+	t.Run("hex too long", func(t *testing.T) {
+		_, err := parseAddress("0x" + strings.Repeat("ab", 33)) // 33 bytes > 32
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "longer than 32 bytes")
+	})
+}
+
+func TestParseChainID(t *testing.T) {
```
johnsaigle, line 50: You can use the SDK functions instead in the template and delete this test. `StringToKnownChainID`

node/pkg/adminrpc/adminserver.go
```diff
@@ -915,6 +915,206 @@ func suiCallToVaa(suiCall *nodev1.SuiCall, timestamp time.Time, guardianSetIndex
...
+func bridgeSetPauserAddressesEvmToVaa(req *nodev1.BridgeSetPauserAddressesEvm, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+	if req.Module == "" {
+		return nil, errors.New("module is required")
+	}
+	targetChainID, err := vaa.KnownChainIDFromNumber[uint32](req.TargetChainId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert chain id: %w", err)
+	}
+
+	parseEvm := func(name, raw string) (ethcommon.Address, error) {
```
johnsaigle, line 1048: Should use go-ethereum's validation function(s)

node/pkg/adminrpc/adminserver.go
```diff
@@ -915,6 +915,206 @@ func suiCallToVaa(suiCall *nodev1.SuiCall, timestamp time.Time, guardianSetIndex
...
+func bridgeSetPauserAddressesSolanaToVaa(req *nodev1.BridgeSetPauserAddressesSolana, timestamp time.Time, guardianSetIndex uint32, nonce uint32, sequence uint64) (*vaa.VAA, error) {
+	if req.Module == "" {
+		return nil, errors.New("module is required")
+	}
+	targetChainID, err := vaa.KnownChainIDFromNumber[uint32](req.TargetChainId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert chain id: %w", err)
+	}
+
+	parse := func(name, raw string) (vaa.Address, error) {
```
johnsaigle, line 1088: It'd be better to turn this into an SDK function rather than in-lining here: we use this type of validation elsewhere and it would allow for a more focused unit tests.

 vaa.StringToAddress is close but it would allow for inputs that are <= 32, not strictly == 32.

johnsaigle (changes requested):

node/cmd/guardiand/admintemplate.go
```diff
@@ -1620,3 +1678,237 @@ func randSeqNonce() (uint64, uint32) {
...
+		}
+		return out
+	}
+	pauser = parse("pauser", *bridgeSetPauserAddressesPauser)
+	unpauser = parse("unpauser", *bridgeSetPauserAddressesUnpauser)
+	return
+}
+
+func runBridgeSetPauserAddressesEvmTemplate(cmd *cobra.Command, args []string) {
+	module, chainId, pauser, unpauser := parseBridgeSetPauserCommonFlags(parseEvmHexAddress)
```
johnsaigle, line 1863: For both this function and the Solana one, we should log a warning if `pauser == unpauser`. 

While this is allowed in the design spec for flexibility, IMO it should be discouraged given that unpausing is a more highly-privileged action than pausing (which is also in the spec).

e.g. `"WARNING: pauser and unpauser are the same address. This removes the intended higher-trust unpause path. Use only if symmetric pause/unpause authority is intentional."`

johnsaigle (changes requested):

johnsaigle:
@claude triggering a second run as an experiment to see if it has any new suggestions based on the comments I made above. It might be able to help give more precise instructions for avoiding duplicated validation code or adding more sanity checks

PR #4772, opened 2026-04-28
ci: Relax delta for unit test coverage

johnsaigle:
The current code coverage skew is very tight, causing CI to fail even for small diffs that don't meaningfully reduce test coverage.

This bumps the skew up a bit so that there's less noise in the PR process.

e.g. https://github.com/wormhole-foundation/wormhole/actions/runs/25067571362/job/73439828216?pr=4711

### DIFF
--- a/scripts/coverage-check/README.md
+++ b/scripts/coverage-check/README.md
@@ -6,7 +6,7 @@ A minimal Go tool to enforce incremental test coverage improvements without bloc
 
 This tool enforces two simple rules:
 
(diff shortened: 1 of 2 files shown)


djb15 (approved):

djb15:
To be fair, that other PR's coverage drops 1.1% so I probably just have to write some tests 😅 but I still think a 1% tolerance is fair

bemic (approved):

PR #3253, opened 2023-08-02
Solana: add legacy directory for old programs and fix CI

barnjamin:

PR #4775, opened 2026-04-28
node(svm-watcher): Remove unused code referring to Anchor version of core bridge

johnsaigle:
## Summary
- Remove the Solana watcher finalization check for the abandoned 2023 core bridge rewrite layout.
- Restore the parsed message account field names to the current legacy layout (`VaaTime`, `VaaSignatureAccount`).
- Remove tests that only covered the unused rewrite-specific `EmitterAuthority` / `MessageStatus` / `Gap` behavior.
## Context
The watcher check was introduced directly on `main` in 85e785a13 (`node/watchers/solana: additional safety check`) and later given a Pythnet carve-out in 4bfcb89d4 (`node/watchers/solana: pythnet safety carve out`). GitHub does not report an associated PR for either commit, and 4bfcb89d4 is the v2.23.16 tag, which is why this shows up as release history rather than a PR merge.
The check appears to have been intended to support the 2023 Anchor-based Solana core bridge rewrite message account layout. That rewrite work was developed under the `wen/solana-rewrite` branch family. The original scaffolding PR was #3253 (`Solana: add legacy directory for old programs and fix CI`), which merged into `wen/solana-rewrite`, not `main`.
The rewrite never landed on `main`; the current Solana core bridge still uses the legacy message account layout.

## Layout Difference
Current `main` core bridge message layout is:
https://github.com/wormhole-foundation/wormhole/blob/main/solana/bridge/program/src/accounts/posted_message.rs#L45-L76
```rust
pub struct MessageData {
    pub vaa_version: u8,
    pub consistency_level: u8,
    pub vaa_time: u32,
    pub vaa_signature_account: Pubkey,
    pub submission_time: u32,
    pub nonce: u32,
    pub sequence: u64,
    pub emitter_chain: u16,
    pub emitter_address: [u8; 32],
    pub payload: Vec<u8>,
}
```

The abandoned wen/solana-rewrite layout instead used:
https://github.com/wormhole-foundation/wormhole/blob/wen/solana-rewrite/solana/programs/core-bridge/src/legacy/state/posted_message_v1/mod.rs#L36-L69

```rust
pub struct PostedMessageV1Info {
    pub consistency_level: u8,
    pub emitter_authority: Pubkey,
    pub status: MessageStatus,
    pub _gap_0: [u8; 3],
    pub posted_timestamp: Timestamp,
    pub nonce: u32,
    pub sequence: u64,
    pub solana_chain_id: ChainIdSolanaOnly,
    pub emitter: Pubkey,
}
```

The removed watcher check interpreted the current legacy vaa_time / vaa_signature_account byte range as rewrite-only emitter_authority / status / gap state. That state machine does not exist in the current core bridge code on main.

### DIFF
--- a/node/pkg/watchers/solana/client_test.go
+++ b/node/pkg/watchers/solana/client_test.go
@@ -196,7 +196,9 @@ func Test_validateTransactionMeta(t *testing.T) {
 
 func TestParseMessagePublicationAccount(t *testing.T) {
 
-	// Define well-formed Solana mainnet PostedMessage account data for reliable and unreliable messages
+	// Define well-formed Solana mainnet PostedMessage account data for reliable and unreliable messages.
+	// The SVM core bridge leaves vaa_time at zero for msg/msu accounts; post_vaa writes
+	// non-zero vaa_time to vaa-prefixed accounts instead.
 	var (
 		// Solana mainnet PostedMessage account: `GU76rcJ4rgw5sZQ2efVRC4yUZyhwrVM6pTGs2kFhckYy`
 		validMessageAccountDataReliable, _ = hex.DecodeString("6d73670020000000000000000000000000000000000000000000000000000000000000000000000000404d836900000000f5de1400000000000100ec7372995d5cc8732397fb0ad35c0121e0eaa90d26f828a534cab54391b3a4f5850000000100000000000000000000000000000000000000000000000000009ed268a30dd700000000000000000000000089f4e8011c35831130c4c3ab95e53de9411d2fcc00040000000000000000000000006722b2c28d7d299b56a5febedbbe865a84ee0d7d00040000000000000000000000000000000000000000000000000000000000000000")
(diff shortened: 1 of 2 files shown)


johnsaigle:
The node coverage check failure is very minor and should be resolved after #4772 is merged.

djb15 (dismissed):
Checked the deserialisation will still yield the same layout for the other fields and that none of the removed or added fields have any references elsewhere in the codebase

djb15 (approved):

node/pkg/watchers/solana/client_test.go
```diff
@@ -237,17 +237,16 @@ func TestParseMessagePublicationAccount(t *testing.T) {
...
-				Gap:              [3]byte{0, 0, 0},
-				SubmissionTime:   1770212672,
-				Nonce:            0,
-				Sequence:         1367797,
-				EmitterChain:     1,
-				EmitterAddress:   emitterAddrReliable,
-				Payload:          payload,
+				VaaVersion:          0,
+				ConsistencyLevel:    32,
+				VaaTime:             0,
```
mdulin2, line 244: Having this be non-zero for testing seems wise. It would prevent default zero cases that we hadn't noticed before.

mdulin2 (changes requested):
The format changes seem fine. Otherwise, the `TestParseMessagePublicationAccount` would have failed on deserialization.

mdulin2:
Solana watcher integration tests work fine too. That's a good sign of this being valid!

node/pkg/watchers/solana/client_test.go
johnsaigle: My understanding is that message accounts will always have this value set to 0. This gets enforced in the latest, post-rent-reclaim update in the core bridge: https://github.com/wormhole-foundation/wormhole/blob/6cc8db2013e72fcb2b4e4c34e7a5933b535d56d7/solana/bridge/program/src/api/close_posted_message.rs#L89

I added some test cases around this anyway since it shouldn't matter for the changes made in this PR.

I also added comments on the field itself.

djb15 (approved):

johnsaigle:
@mdulin2 can you re-review?

mdulin2 (approved):

PR #4776, opened 2026-04-28
Evm watcher unit tests

mdulin2:
This is a small rewrite that moves large amounts of code from goroutines into a separate function to make it more testable. 

The main reason for this is to make the code more testable. This includes a large set of unit tests for the EVM watcher paths from this. 

There are two main non-test changes: 
* Error handing on tx == nil receipt used to be deleted from the pending list. However, transient errors also return a nil receipt. So, I changed the logic to actually have transient errors stay in the watcher. 
* Added a receipt.txhash == pending.txhash check. This should probably never happen but with the amount of chains we support, we need to be sure.


johnsaigle:
@mdulin2 This should be good to rebase and continue working on now that your other PR was merged.

node/pkg/watchers/evm/watcher.go
```diff
@@ -1042,6 +867,217 @@ func (w *Watcher) postMessage(
...
+
+		// If the node returns an error after waiting expectedConfirmation blocks,
+		// it means the chain reorged and the transaction was orphaned. The
+		// TransactionReceipt call is using the same websocket connection than the
+		// head notifications, so it's guaranteed to be atomic.
+		//
+		// Check multiple possible error cases - the node seems to return a
+		// "not found" error most of the time, but it could conceivably also
+		// return a rpc.ErrNoResult.
+		if errors.Is(err, rpc.ErrNoResult) || errors.Is(err, ethereum.NotFound) || (err != nil && err.Error() == "not found") {
```
mdulin2, line 940: Moved the tx == nil check to be explicit below this for better handling of transient failures.

node/pkg/watchers/evm/watcher.go
```diff
@@ -1042,6 +867,217 @@ func (w *Watcher) postMessage(
...
+			)
+			delete(w.pending, key)
+			ethMessagesOrphaned.WithLabelValues(w.networkName, "nil_receipt").Inc()
+			continue
+		}
+
+		// SECURITY: Defense in depth against a buggy or malicious connector returning
+		// a receipt for a different transaction than the one we queried.
+		expectedTxHash := eth_common.BytesToHash(pLock.message.TxID)
+		if tx.TxHash != expectedTxHash {
```
mdulin2: New check as well.

node/pkg/watchers/evm/watcher.go
```diff
@@ -533,185 +533,11 @@ func (w *Watcher) Run(parentCtx context.Context) error {
...
-				case connectors.Finalized:
-					thisConsistencyLevel = vaa.ConsistencyLevelFinalized
-					atomic.StoreUint64(&w.latestFinalizedBlockNumber, blockNumberU)
-					currentEthFinalizedHeight.WithLabelValues(w.networkName).Set(float64(blockNumberU))
-					stats.FinalizedHeight = int64(blockNumberU) // #nosec G115 -- This conversion is safe indefinitely
-				default:
-					logger.Error("unexpected finality in block", zap.Stringer("finality", ev.Finality), zap.Any("event", ev))
-					errC <- fmt.Errorf("unexpected finality in block: %v", ev.Finality)
+				if err := w.processNewBlock(ctx, ev, &stats); err != nil {
+					errC <- err //nolint:channelcheck // The watcher will exit anyway
```
johnsaigle:
```suggestion
					errC <- err
```

Since channelcheck doesn't exist, it causes problems in future versions of golangci-lint #4701

node/pkg/watchers/evm/watcher.go
```diff
@@ -1042,6 +868,215 @@ func (w *Watcher) postMessage(
...
+			zap.Stringer("blockHash", key.BlockHash),
+			zap.Uint64("observedHeight", pLock.height),
+			zap.Uint64("additionalBlocks", pLock.additionalBlocks),
+			zap.Stringer("current_blockNum", ev.Number),
+			zap.Stringer("finality", ev.Finality),
+			zap.Stringer("current_blockHash", currentHash),
+		)
+		delete(w.pending, key)
+
+		// Note that `tx` here is actually a receipt
```
johnsaigle: While we're here, I wonder if we just rename this variable to like `txReceipt` or just `receipt`? I think it could make the code easier to understand.

johnsaigle (commented):
I think the refactor looks good, just left a couple of suggestions.

johnsaigle:
Please also run `make update-coverage-baseline` and commit the results so that we can lock-in the improved coverage 🔒 ✅

pleasew8t (dismissed):

node/pkg/watchers/evm/watcher.go
mdulin2: Yeah, seems like an easy change to me. I'll do it.

johnsaigle (dismissed):
My previous comments are addressed.

djb15 (commented):
The refactor looks safe and happy to see more tests! Before I approve, what's the reasoning for removing some of the prior `SECURITY` comments? I like the explicitness and in prior conversations we agreed to flag critical checks in this way so they're less likely to be removed in future changes

mdulin2:
I'll add in the `SECURITY` comment back in. No problem.

mdulin2:
I messed up the history of the PR with something else I'm working on. I'll fix this in a little bit.

mdulin2:
Rebase shouldn't include the Aptos PR changes anymore. This should be rebased on main.

djb15 (dismissed):

node/pkg/watchers/evm/watcher.go
```diff
@@ -1043,6 +869,220 @@ func (w *Watcher) postMessage(
...
+			logger.Error("transaction receipt with non-success status",
+				zap.String("msgId", pLock.message.MessageIDString()),
+				zap.String("txHash", pLock.message.TxIDString()),
+				zap.Stringer("blockHash", key.BlockHash),
+				zap.Uint64("observedHeight", pLock.height),
+				zap.Uint64("additionalBlocks", pLock.additionalBlocks),
+				zap.Stringer("current_blockNum", ev.Number),
+				zap.Stringer("finality", ev.Finality),
+				zap.Stringer("current_blockHash", currentHash),
+				zap.Error(err))
```
johnsaigle: This `err` comes from an earlier check so I think it's guaranteed to be `nil` here. It would make more sense to log the `txReceipt.Status` so we can see what we got instead of ReceiptStatusSuccessful

node/pkg/watchers/evm/watcher.go
```diff
@@ -1043,6 +869,220 @@ func (w *Watcher) postMessage(
...
+					zap.Stringer("current_blockNum", ev.Number),
+					zap.Stringer("finality", ev.Finality),
+					zap.Stringer("current_blockHash", currentHash),
+					zap.Error(err))
+			}
+			continue
+		}
+
+		// txreceipt is nil and err is nil. A check to prevent panics later
+		if txreceipt == nil {
```
johnsaigle, line 988: Can you add this check after the other calls to `ethConn.TransactionReceipt`?

I see 3 total:

```
node/pkg/watchers/evm/by_transaction.go:56:18: receipt, err := ethConn.TransactionReceipt(ctx, tx)                                                                                                                
node/pkg/watchers/evm/watcher.go:919:23:  txreceipt, err := w.ethConn.TransactionReceipt(timeout, eth_common.BytesToHash(pLock.message.TxID))                                                                     
node/pkg/watchers/evm/watcher.go:799:21:  receipt, err := w.ethConn.TransactionReceipt(receiptCtx, ev.Raw.TxHash)                                                                                                 
```

node/pkg/watchers/evm/watcher.go
mdulin2: Good call out. Happy to do.

node/pkg/watchers/evm/watcher.go
mdulin2: The 919 is the one above. So, it should just be the first and third one.

mdulin2: Should be resolved.

node/pkg/watchers/evm/watcher.go
mdulin2: Should be resolved.

johnsaigle (dismissed):

djb15 (dismissed):

panoel (dismissed):

panoel (approved):

johnsaigle (approved):
Re-approving. Minor changes since my last approval.

PR #4779, opened 2026-04-29
feat(node): Tron support

evan-gray:
Tron support required a new poller implementation. This implementation has been tested against Nile Testnet.

### Finalized

https://nile.tronscan.org/#/transaction/3db6c638da4cc83b0cf7f58c77cd3ac0e7d67e19005a841f07ae7f75f2ccbbc4

https://api.testnet.wormholescan.io/api/v1/observations/70/0000000000000000000000008f26a0025dccc6cfc07a7d38756280a10e295ad7/1

https://api.testnet.wormholescan.io/v1/signed_vaa/70/0000000000000000000000008f26a0025dccc6cfc07a7d38756280a10e295ad7/1

### Latest (Instant)

https://nile.tronscan.org/#/transaction/202f337036f1f245151818a1db1b05c6d899c54be51b0af50bbb680842a68759

https://api.testnet.wormholescan.io/api/v1/observations/70/0000000000000000000000008f26a0025dccc6cfc07a7d38756280a10e295ad7/2

https://api.testnet.wormholescan.io/v1/signed_vaa/70/0000000000000000000000008f26a0025dccc6cfc07a7d38756280a10e295ad7/2


node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,422 @@
...
+			L1BlockNumber: l1bn,
+			Finality:      finality,
+		}
+	}
+
+	return ret, nil
+}
+
+// pollBlocks polls for the latest set of blocks, compares to previous, and publishes new ones with gap filling.
+func (p *PollConnector) pollBlocks(ctx context.Context, sink chan<- *NewBlock, prevBlocks Blocks) (Blocks, error) {
```
mdulin2: This code is line-by-line the same as the `batch_poller.go` version. Consider consolidating the logic/code for this and other locations where the code was copied.

node/pkg/watchers/evm/tron_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+	assert.Equal(t, uint64(0), msg.Sequence, "sequence")
+	assert.False(t, msg.IsReobservation)
+
+	// Verify the emitter address is the 32-byte left-padded sender.
+	sender := ethCommon.HexToAddress("0x8F26A0025dcCc6Cfc07A7d38756280a10E295ad7")
+	expectedEmitter := PadAddress(sender)
+	assert.Equal(t, expectedEmitter, msg.EmitterAddress, "emitter address")
+
+	t.Logf("block=%d emitter=%x nonce=%d seq=%d consistency=%d payload=%q",
+		blockNum, msg.EmitterAddress, msg.Nonce, msg.Sequence, msg.ConsistencyLevel, string(msg.Payload))
```
mdulin2, line 75: Emitter address is showing: 30303030303030303030303030303030303030303030303038663236613030323564636363366366633037613764333837353632383061313065323935616437

Maybe it's worth using a %s or something similar instead.

node/pkg/watchers/evm/tron_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	logger, _ := zap.NewDevelopment()
+	coreAddr := ethCommon.HexToAddress(coreAddrHex)
+
+	base, err := connectors.NewEthereumBaseConnector(ctx, "tron-nile-test", rpcURL, coreAddr, nil, logger)
+	require.NoError(t, err)
+
+	// Build a PollConnector but we won't use SubscribeForBlocks — we'll
+	// directly test that FilterLogs over HTTP works for the known block range.
+	poll := connectors.NewPollConnector(ctx, logger, base, false, time.Second)
```
mdulin2: I feel like this test code doesn't use `WatchLogMessagePublished()`. It copies the code from it and uses. From manual review, this seems fine but this wouldn't work for regression-based issues in the future. Does it make sense to rework `WatchLogMessagePublished` so that a specific block can be injected for testing purposes?

mdulin2: Thank you for writing tests 🙏

mdulin2 (changes requested):

mdulin2 (dismissed):

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+	return sub, nil
+}
+
+// TimeOfBlockByHash overrides the base connector's implementation because some
+// HTTP-only chains (e.g. Tron) return non-standard header fields (such as an
+// empty stateRoot "0x") that cause go-ethereum's HeaderByHash to fail. This
+// version uses a raw RPC call and only extracts the timestamp.
+func (p *PollConnector) TimeOfBlockByHash(ctx context.Context, hash ethCommon.Hash) (uint64, error) {
+	var m BlockMarshaller
+	err := p.RawCallContext(ctx, &m, "eth_getBlockByHash", hash, false)
```
johnsaigle, line 258: It's not clear to be that this will return `err` when a block isn't found. Do you know if that's the case?

If it doesn't, I think this implementation will marshal an `m` with zero-values, so this function will return `0, nil` for missing blocks.

It looks like other caller rely on the error value returned here, and so they might end up building a message publication with a timestamp of 0 if that happens.

It could be worth adding a check for a 0 value block timestamp in `getBlockTime` while we're at it (unless you know of a case where we do want to support this).

```go
// getBlockTime reads the time of a block.
func (w *Watcher) getBlockTime(ctx context.Context, blockHash eth_common.Hash) (uint64, error) {
	msm := time.Now()
	timeout, cancel := context.WithTimeout(ctx, 15*time.Second)
	blockTime, err := w.ethConn.TimeOfBlockByHash(timeout, blockHash)
	cancel()
	queryLatency.WithLabelValues(w.networkName, "block_by_number").Observe(time.Since(msm).Seconds())
	return blockTime, err
}
```

https://developers.tron.network/reference/eth_getblockbyhash

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+// It also replaces WatchLogMessagePublished (which calls eth_subscribe "logs")
+// with a polling loop that uses eth_getLogs.
+type PollConnector struct {
+	*BatchPollConnector
+
+	// MaxLogScanBlocks caps the per-iteration eth_getLogs range in
+	// WatchLogMessagePublished. Zero means no cap. Useful as a safety net
+	// against pathological requests and to constrain tests scanning historic
+	// blocks.
+	MaxLogScanBlocks uint64
```
johnsaigle, line 51: Right now this is only used in the tests and it isn't configurable in production because it's unset by `NewPollConnector`. I think it makes sense to add this as an argument in `NewPollConnector` or else make it constant so we understand that there's no cap in production.

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	})
+
+	return sub, nil
+}
+
+// WatchLogMessagePublished polls for LogMessagePublished events via eth_getLogs,
+// replacing the WebSocket-based subscription on the base connector. It begins
+// at the current finalized block; callers that need a different starting block
+// (e.g. tests) should use WatchLogMessagePublishedFrom directly.
```
johnsaigle: Do you expect a caller to need `WatchLogMessagePublishedFrom`? If it's only here for tests, we could consider making it private instead of exporting it.

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return NewPollSubscription(), fmt.Errorf("failed to get initial block for log polling: %w", err)
+	}
+	return p.WatchLogMessagePublishedFrom(ctx, errC, sink, block.Number.Uint64())
+}
+
+// WatchLogMessagePublishedFrom is the testable variant of
+// WatchLogMessagePublished: the caller specifies the starting block instead of
+// using the current finalized block. The scan respects MaxLogScanBlocks.
+func (p *PollConnector) WatchLogMessagePublishedFrom(ctx context.Context, errC chan error, sink chan<- *ethAbi.AbiLogMessagePublished, fromBlock uint64) (ethEvent.Subscription, error) {
```
johnsaigle: `errC` is unused here so I don't think that persistent errors will cause a watcher restart and instead I suppose it'll just keep logging errors and fail to process messages. Is there a reason to not use `errC` here?

node/pkg/watchers/evm/connectors/poller.go
evan-gray: thanks for the catch, I've updated it to handle the null case, could you confirm this seems like it should work? I'd prefer to avoid additional checks that aren't at parity with the other connectors / go-eth library

node/pkg/watchers/evm/connectors/poller.go
evan-gray: updated to match batch_poller

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,250 @@
...
+				// idle poll and falls through to reset errCount below.
+				toBlock := latest.Number.Uint64()
+				if toBlock >= fromBlock {
+					if p.MaxLogScanBlocks > 0 && toBlock-fromBlock+1 > p.MaxLogScanBlocks {
+						toBlock = fromBlock + p.MaxLogScanBlocks - 1
+					}
+
+					from := new(big.Int).SetUint64(fromBlock)
+					to := new(big.Int).SetUint64(toBlock)
+					logs, err := p.Client().FilterLogs(ctx, ethereum.FilterQuery{
```
johnsaigle: I think this is a blocking call so you might want to do a context cancel timeout construction here in case this stalls. Compare with the body of the `GetBlock` call below which handles this case.

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,250 @@
...
+				sub.unsubDone <- struct{}{}
+				return nil
+			case <-timer.C:
+				lastBlocks, err = p.pollBlocks(ctx, sink, lastBlocks)
+				if err != nil {
+					errCount++
+					p.logger.Error("poll connector encountered an error", zap.Int("errCount", errCount), zap.Error(err))
+					if errCount > pollMaxErrors {
+						errC <- fmt.Errorf("polling encountered too many errors: %w", err)
+						return nil
```
johnsaigle: I think this needs to write to `sub.unsubDone` or else the watcher may not be able to restart properly even if we signal to it via `errC` due to a deadlock.

Otherwise when we return, the `defer`s start, and I think the [`defer` of `Unsubscribe()`](https://github.com/wormhole-foundation/wormhole/blob/aa2aac0100c0a3f28713cea0ae3482ea04f1a28e/node/pkg/watchers/evm/watcher.go#L502) will block the `defer` of the watcher context cancel function (defers are LIFO ordering). So if we signal error here without resolving `Unsubscribe()`, the supervisor won't actually restart the watcher and it'll just run without processing anything.

If the above is correct, the other lines checking `pollMaxErrors` also need to be fixed.

johnsaigle (commented):
Minor follow up comments around some of the async operations

node/pkg/watchers/evm/connectors/poller.go
johnsaigle: Fixed in latest commit

johnsaigle (dismissed):
Code looks good now, barring one concurrency edge-case which can be addressed in a follow-up PR since the pattern it identifies is also present in code that is not directly related to adding Tron support.

node/pkg/watchers/evm/connectors/poller.go
```diff
@@ -0,0 +1,266 @@
...
+				return nil
+			case <-sub.quit:
+				return nil
+			case <-timer.C:
+				latest, err := GetBlockByFinality(ctx, p.Connector, Latest)
+				if err != nil {
+					errCount++
+					p.logger.Error("log poller failed to get latest block", zap.Int("errCount", errCount), zap.Error(err))
+					if errCount > pollMaxErrors {
+						errC <- fmt.Errorf("log polling encountered too many errors: %w", err)
```
johnsaigle: (edge-case: We can track this separately in a GitHub issue if you want -- I don't want to block this PR)

After looking at the unsub behaviour during the review yesterday, I think there's one other edge-case here. If the watcher is _already_ exiting and its `errC` reader is closed, then writing to `errC` can block here and we will get the same deadlock behaviour where `Unsubscribe()` does not execute successfully.

That might happen if some other goroutine within the watcher has already started shutdown operations, while the Poller here also hits pollMaxErrors before the shutdown has completed. In that case, there is no `errC` consumer and this locks. 
(For reference, see also [this code block](https://github.com/wormhole-foundation/wormhole/blob/194fe7ba92e706523d67c94d7e8dedd10974aa74/node/pkg/common/scissors.go#L68-L72) from `RunWithScissors`. We're running into the same case here where the listener is gone)

The same code is in the batch poller but the consequences are different. In this case, the poller goroutine and watcher are coupled through Wormhole’s `PollSubscription`, whereas geth’s subscription is not coupled to the Wormhole batch poller goroutine, and does not write to errC during shutdown.

Sorry about the churn here -- I know in a separate review I asked about writing to `errC` because the code at that time was only logging errors and not signalling them to the watcher. I still think we ought to signal errors but we want to do so in a non-blocking way. Probably the cleanest way to do this is to instead return error here instead of nil and instead of writing directly in a blocking way to `errC`.

evan-gray: I think [we recently identified a memory leak](https://github.com/wormholelabs-xyz/wormhole/pull/20/changes/5a45e68115c2af2e910c71c4f7064db0c106cc75) related to that, so perhaps we can fix them together.

I do wonder if we should fix it now though, will investigate

evan-gray: Fixed here, I think?

johnsaigle (approved):
Edge-case deadlock should be resolved via non-blocking error sends

mdulin2 (approved):

panoel (approved):

PR #4780, opened 2026-04-29
node: Add watcher interface and implement in Aptos

johnsaigle:
#4761


node/pkg/watchers/observation.go
```diff
@@ -0,0 +1,110 @@
...
+// publish as a reobservation for the validated watcher chain.
+func ValidateReobservedMessage(observation ValidObservation, msg *common.MessagePublication) error {
+	if msg == nil {
+		return errors.New("message publication is nil")
+	}
+
+	if msg.EmitterChain != observation.chainID {
+		return fmt.Errorf("message publication emitter chain %v does not match validated observation chain %v", msg.EmitterChain, observation.chainID)
+	}
+
```
bemic, line 95: Do we want to add check like this `bytes.Equal(msg.TxID, observation.txHash)` ??

node/pkg/watchers/aptos/watcher.go
```diff
@@ -76,6 +78,54 @@ func NewWatcher(
...
+
+func (e *Watcher) ChainID() vaa.ChainID {
+	return e.chainID
+}
+
+func (e *Watcher) PublishMessage(msg *common.MessagePublication) error {
+	if msg == nil {
+		return fmt.Errorf("message publication is nil")
+	}
+
```
bemic, line 123: One more extra defensive idea. 
```
if msg.EmitterChain != e.chainID {
		return fmt.Errorf(
			"message publication emitter chain %v does not match watcher chain %v",
			msg.EmitterChain,
			e.chainID,
		)
}
```

bemic (dismissed):
I left two defense-check ideas to consider, otherwise I think its a good change

johnsaigle:
Converting to draft mode:
- #4838 is in late stages of review and modifies some of the same code. It can be merged first
- lint errors
- code coverage conflict.

PR #4783, opened 2026-04-30
Adding more Solana tests + bug fix + hardening

mdulin2:
Added several new tests, building off of Dirk's PR from a few weeks back.

Two other small changes were made to the watcher:
* Add ALT address ownership check. If we add another Solana blockchain that uses a different address, we will need to update this code.
* Fixed the 0-index bug in the Solana watcher. This is impossible to hit because the first account must be a signer from whom the signature comes.


johnsaigle (approved):
Nice - wanna try updating the code coverage too?

You should be able to run `make update-coverage-baseline` from the repo root and committing the changed file

pleasew8t (approved):

PR #4790, opened 2026-05-04
feat(node): solana core close event handling

evan-gray:
This PR adds support for re-observing a Solana message based on message closure CPI event data. It follows a similar pattern as the shim and reuses the same account parsing / publication logic.

This is the guardian portion of https://github.com/wormhole-foundation/wormhole/pull/4746.

I ran the additional Solana test with the following command in order to get the instruction data.
```bash
cd solana
docker run --rm -it -v "$(pwd):/usr/src/bridge" -w /usr/src/bridge "ghcr.io/wormhole-foundation/solana:1.10.31@sha256:d31e8db926a1d3fbaa9d9211d9979023692614b7b64912651aba0383e8c01bad" bash -c 'CHAIN_ID=1 EMITTER_ADDRESS="11111111111111111111111111111115" cargo test-bpf --manifest-path bridge/program/Cargo.toml --features instructions -- test_close_posted_message_capture_fixture --nocapture'
```

The integration test for the close event message generation can be run with 
```bash
cd node
go test -tags=integration -run TestReobservationDevnet -v ./pkg/watchers/solana/
```

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -22,7 +22,6 @@ import (
 	"github.com/certusone/wormhole/node/pkg/readiness"
 	"github.com/certusone/wormhole/node/pkg/supervisor"
 	"github.com/certusone/wormhole/node/pkg/watchers"
-	eth_common "github.com/ethereum/go-ethereum/common"
 	"github.com/gagliardetto/solana-go"
 	lookup "github.com/gagliardetto/solana-go/programs/address-lookup-table"
 	"github.com/gagliardetto/solana-go/rpc"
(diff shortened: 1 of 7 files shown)


kcsongor (approved):
looked at the rust!

johnsaigle (approved):
Approving based on an earlier review. My suggestions have been applied.

One question -- with this new mechanism for SVM messages, are all of the Message Publications coming from this watcher "Reliable" now? Or are they all unreliable? I think that this distinction was easier to reason about with the older behaviour where Message Accounts stuck around forever. Now they'll presumably always be closed, but we should have the shim event data for a while.

evan-gray:
> Approving based on an earlier review. My suggestions have been applied.
> 
> One question -- with this new mechanism for SVM messages, are all of the Message Publications coming from this watcher "Reliable" now? Or are they all unreliable? I think that this distinction was easier to reason about with the older behaviour where Message Accounts stuck around forever. Now they'll presumably always be closed, but we should have the shim event data for a while.

this doesn't change the reliable/unreliable marking or logic - I think it can be thought of as keeping the reliable messages theoretically re-observable - i.e. keeping reliable messages reliable, at least if an archive node exists.

PR #4797, opened 2026-05-07
node: Remove minimum TxID length

johnsaigle:
https://github.com/wormhole-foundation/wormhole/pull/4428 added validation for Message Publications. This included a minimum Tx ID length check. However, this should not have been added as the Tx ID should be controlled by each watcher rather than enforced in the Message Publication itself.

In theory there is a current minimum length for TxIDs across all of the implementations but we should keep this field flexible.

Right now, Aptos is likely not marshalling correctly because its TxID is based on a uint64 value. In some cases it may be 'normalized' i.e. padded to 32 bytes to conform to other chains' IDs, but this shouldn't be necessary and enforced from a distance.

### DIFF
--- a/node/pkg/common/chainlock.go
+++ b/node/pkg/common/chainlock.go
@@ -17,9 +17,6 @@ import (
 )
 
 const (
-	// AddressLength is the length of a normalized Wormhole address in bytes.
-	AddressLength = 32
-
 	// Wormhole supports arbitrary payloads due to the variance in transaction and block sizes between chains.
 	// However, during serialization, payload lengths are limited by Go slice length constraints and violations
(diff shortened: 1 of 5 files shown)


mdulin2 (dismissed):

node/pkg/common/chainlock.go
mdulin2: Is this missing `verificationState` in the length check?

johnsaigle: Yeah it looks like it. I'll add it to this PR, thanks for raising

johnsaigle:
@mdulin2 nice spot, I fixed the bug in the recent commits and also added a unit test for that particular case.

PR #4801, opened 2026-05-07
feat(evm): wtt pauser role

evan-gray:
This PR implements pauser functionality for the EVM Token Bridge.

- Design #4809
- Guardian #4810
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802

Unfortunately, the pauser change alone put the contract over the 24,576-byte EIP-170 limit. e.g.

```
Error: `BridgeImplementation` is above the contract size limit (25714 > 24576).
```

So, this proposal also modernizes the contract by switching all of the reverts to custom errors leaving over 1,000 bytes of headroom.

### DIFF
--- a/ethereum/contracts/bridge/Bridge.sol
+++ b/ethereum/contracts/bridge/Bridge.sol
@@ -52,7 +52,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     ///         so that an all-zero `unpauser` is never treated as an authorized caller.
     error NotUnpauser();
     /// @notice Reverts when `msg.value` does not cover the wormhole message fee.
-    error WormholeFeeTooLarge();
+    error InsufficientFee();
     /// @notice Reverts when an arbiter / relayer fee exceeds the transfer amount.
     error FeeExceedsAmount();
     /// @notice Reverts when a transfer VAA is not from a registered token-bridge emitter.
@@ -99,6 +99,10 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     ///         via the `SetPauserAddresses` (action 4) governance VAA and may be left unassigned;
     ///         when unassigned this entry point reverts before comparing `msg.sender`, so an
     ///         all-zero `pauser` is never authorized.
+    /// @dev Intentionally does not check `isFork()`. On a forked chain the pre-fork pauser can
+    ///      still pause the bridge without first waiting for a `submitRecoverChainId` governance
+    ///      VAA — letting whoever holds the pauser key shut the bridge down on the fork
+    ///      immediately, before chain-id recovery completes.
     function pause() external {
         address p = pauser();
         if (p == address(0)) revert NotPauser();
@@ -112,6 +116,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     ///         unassigned; when unassigned this entry point reverts before comparing `msg.sender`,
     ///         so an all-zero `unpauser` is never authorized. Note that `unpause` is intentionally
     ///         exempt from the `notPaused` modifier so it remains callable while paused.
+    /// @dev Intentionally does not check `isFork()`; see the matching note on `pause()`.
     function unpause() external {
         address u = unpauser();
         if (u == address(0)) revert NotUnpauser();
@@ -216,7 +221,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     function _wrapAndTransferETH(uint256 arbiterFee) internal returns (BridgeStructs.TransferResult memory transferResult) {
         uint wormholeFee = wormhole().messageFee();
 
-        if (wormholeFee >= msg.value) revert WormholeFeeTooLarge();
+        if (wormholeFee >= msg.value) revert InsufficientFee();
 
         uint amount = msg.value - wormholeFee;
 
--- a/ethereum/contracts/bridge/Bridge.sol
+++ b/ethereum/contracts/bridge/Bridge.sol
@@ -38,16 +38,63 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     /// @notice Emitted when the bridge is unpaused.
     event Unpaused(address indexed by);
 
+    /// @dev Custom errors are used in place of revert strings to keep `BridgeImplementation` under
+    ///      the 24,576-byte EIP-170 limit.
+
+    /// @notice Reverts when a `notPaused`-guarded entry point is called while the bridge is paused.
+    error BridgePaused();
+    /// @notice Reverts when `pause()` is called by anyone other than the configured pauser.
+    error NotPauser();
+    /// @notice Reverts when `unpause()` is called by anyone other than the configured unpauser.
+    error NotUnpauser();
+    /// @notice Reverts when `msg.value` does not cover the wormhole message fee.
+    error WormholeFeeTooLarge();
+    /// @notice Reverts when an arbiter / relayer fee exceeds the transfer amount.
+    error FeeExceedsAmount();
+    /// @notice Reverts when a transfer VAA is not from a registered token-bridge emitter.
+    error InvalidEmitter();
+    /// @notice Reverts when no wrapped asset is registered for a given (chain, token) pair.
+    error WrappedAssetNotFound();
+    /// @notice Reverts when `createWrapped` is called for a token native to this chain.
+    error OnlyForeignTokens();
+    /// @notice Reverts when `createWrapped` is called for a (chain, token) that already has a wrapper.
+    error WrappedAssetAlreadyExists();
+    /// @notice Reverts when truncating a 32-byte value to an EVM address loses non-zero high bytes.
+    error InvalidEVMAddress();
+    /// @notice Reverts when a payload-3 transfer is redeemed by anyone other than the recipient.
+    error InvalidSender();
+    /// @notice Reverts when a transfer VAA has already been redeemed.
+    error TransferAlreadyCompleted();
+    /// @notice Reverts when redeeming a transfer whose target chain is not this chain.
+    error InvalidTargetChain();
+    /// @notice Reverts when an unwrap-ETH path is taken with a token other than WETH.
+    error OnlyWETH();
+    /// @notice Reverts when an outbound transfer would exceed the per-token outstanding bridged cap.
+    error OutstandingExceedsMax();
+    /// @notice Reverts when an AssetMeta payload is malformed.
+    error InvalidAssetMeta();
+    /// @notice Reverts when a Transfer payload is malformed.
+    error InvalidTransferPayload();
+    /// @notice Reverts when an unknown payload id is encountered.
+    error InvalidPayloadId();
+
     /// @dev Reverts if the bridge is paused. See whitepapers/0018_pauser.md.
+    /// @dev Implemented as a shared internal function (rather than inlining the check in the
+    ///      modifier body) so the bytecode is emitted once and JUMPed to from every callsite,
+    ///      keeping `BridgeImplementation` under the 24,576-byte EIP-170 limit.
+    function _requireNotPaused() internal view {
+        if (paused()) revert BridgePaused();
+    }
+
     modifier notPaused() {
-        require(!paused(), "paused");
+        _requireNotPaused();
         _;
     }
 
     /// @notice Pause the bridge. Only callable by the configured pauser. Configured via the
     ///         SetPauserAddressesEvm (action 4) governance VAA.
     function pause() external {
-        require(msg.sender == pauser(), "not pauser");
+        if (msg.sender != pauser()) revert NotPauser();
         setPaused(true);
         emit Paused(msg.sender);
     }
@@ -56,7 +103,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     ///         SetPauserAddressesEvm (action 4) governance VAA. Note that `unpause` is intentionally
     ///         exempt from the `notPaused` modifier.
     function unpause() external {
-        require(msg.sender == unpauser(), "not unpauser");
+        if (msg.sender != unpauser()) revert NotUnpauser();
         setPaused(false);
         emit Unpaused(msg.sender);
     }
@@ -157,11 +204,11 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
     function _wrapAndTransferETH(uint256 arbiterFee) internal returns (BridgeStructs.TransferResult memory transferResult) {
         uint wormholeFee = wormhole().messageFee();
 
-        require(wormholeFee < msg.value, "value is smaller than wormhole fee");
+        if (wormholeFee >= msg.value) revert WormholeFeeTooLarge();
 
         uint amount = msg.value - wormholeFee;
 
-        require(arbiterFee <= amount, "fee is bigger than amount minus wormhole fee");
+        if (arbiterFee > amount) revert FeeExceedsAmount();
 
         uint normalizedAmount = normalizeAmount(amount, 18);
         uint normalizedArbiterFee = normalizeAmount(arbiterFee, 18);
@@ -338,7 +385,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
         uint256 callValue,
         uint32 nonce
     ) internal returns (uint64 sequence) {
-        require(fee <= amount, "fee exceeds amount");
+        if (fee > amount) revert FeeExceedsAmount();
 
         BridgeStructs.Transfer memory transfer = BridgeStructs.Transfer({
             payloadID: 1,
@@ -395,7 +442,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
         (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(encodedVm);
 
         require(valid, reason);
-        require(verifyBridgeVM(vm), "invalid emitter");
+        if (!verifyBridgeVM(vm)) revert InvalidEmitter();
 
         BridgeStructs.AssetMeta memory meta = parseAssetMeta(vm.payload);
         return _updateWrapped(meta, vm.sequence);
@@ -403,7 +450,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
 
     function _updateWrapped(BridgeStructs.AssetMeta memory meta, uint64 sequence) internal returns (address token) {
         address wrapped = wrappedAsset(meta.tokenChain, meta.tokenAddress);
-        require(wrapped != address(0), "wrapped asset does not exists");
+        if (wrapped == address(0)) revert WrappedAssetNotFound();
 
         // Update metadata
         TokenImplementation(wrapped).updateDetails(bytes32ToString(meta.name), bytes32ToString(meta.symbol), sequence);
@@ -415,7 +462,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
         (IWormhole.VM memory vm, bool valid, string memory reason) = wormhole().parseAndVerifyVM(encodedVm);
 
         require(valid, reason);
-        require(verifyBridgeVM(vm), "invalid emitter");
+        if (!verifyBridgeVM(vm)) revert InvalidEmitter();
 
         BridgeStructs.AssetMeta memory meta = parseAssetMeta(vm.payload);
         return _createWrapped(meta, vm.sequence);
@@ -423,8 +470,8 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
 
     // Creates a wrapped asset using AssetMeta
     function _createWrapped(BridgeStructs.AssetMeta memory meta, uint64 sequence) internal returns (address token) {
-        require(meta.tokenChain != chainId(), "can only wrap tokens from foreign chains");
-        require(wrappedAsset(meta.tokenChain, meta.tokenAddress) == address(0), "wrapped asset already exists");
+        if (meta.tokenChain == chainId()) revert OnlyForeignTokens();
+        if (wrappedAsset(meta.tokenChain, meta.tokenAddress) != address(0)) revert WrappedAssetAlreadyExists();
 
         // initialize the TokenImplementation
         bytes memory initialisationArgs = abi.encodeWithSelector(
@@ -516,7 +563,7 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
      * @param bytes32 bytes The 32 byte array to be converted.
      */
     function _truncateAddress(bytes32 b) internal pure returns (address) {
-        require(bytes12(b) == 0, "invalid EVM address");
+        if (bytes12(b) != 0) revert InvalidEVMAddress();
         return address(uint160(uint256(b)));
     }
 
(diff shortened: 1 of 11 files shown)


ethereum/contracts/bridge/BridgeState.sol
```diff
@@ -49,6 +49,19 @@ contract BridgeStorage {
 
         // EIP-155 Chain ID
         uint256 evmChainId;
+
+        // Address authorized to call pause(). Configured via the SetPauserAddresses governance action.
+        // Typically the WormholePauser contract.
+        // See whitepapers/0018_pauser.md.
+        address pauser;
```
johnsaigle: These changes will modify the storage layout for the contract.

I used `forge inspect contracts/bridge/BridgeImplementation.sol:BridgeImplementation storageLayout
` on both this PR and `main` that shows that the slot for `_state` includes 0-14 in this commit but only 0-12 on main.

```
forge inspect contracts/bridge/BridgeImplementation.sol:BridgeImplementation storageLayout


╭---------+----------------------------+------+--------+-------+----------------------------------------------------------------╮
| Name    | Type                       | Slot | Offset | Bytes | Contract                                                       |
+===============================================================================================================================+
| _state  | struct BridgeStorage.State | 0    | 0      | 480   | contracts/bridge/BridgeImplementation.sol:BridgeImplementation |
|---------+----------------------------+------+--------+-------+----------------------------------------------------------------|
| _status | uint256                    | 15   | 0      | 32    | contracts/bridge/BridgeImplementation.sol:BridgeImplementation |
╰---------+----------------------------+------+--------+-------+----------------------------------------------------------------╯

```

(The `_status` variable comes from reentrancy guard  which is set to `0x0..1` on mainnet.) 

The change to the storage slots will mean that the `pauser` slot will initially be set to the old slot used by status, making the pauser start as `0x0..1`. Also, `_status` will be set to `0x0..0`, though it gets reinitialized as `0x0..1` from my understanding.

Anyway, I don't think there is any direct/immediate impact as a result of changing the storage layout here -- however, the whitepaper states that explicit backward compatibility is the goal, and I believe this is a breaking change (especially if we want to revert the change).

I'm not an expert on upgrade semantics so I'd appreciate your thoughts on this vs. maybe using an EIP-1967 storage slot for the new variables.

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,112 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+    /// @notice Emitted when the pauser/unpauser addresses are updated via governance.
+    /// @param pauser The address authorized to call pause().
+    /// @param unpauser The address authorized to call unpause().
+    event PauserAddressesSet(address indexed pauser, address indexed unpauser);
+
+    /// @notice Set the pauser and unpauser addresses via a SetPauserAddressesEvm (action 4) governance VAA.
+    /// @dev Payload layout: module(32) | action(1)=4 | chainId(2) | pauser(20) | unpauser(20). Parsed
+    ///      inline (no separate `parseSetPauserAddresses` / struct) to keep `BridgeImplementation` under
+    ///      the 24,576-byte EIP-170 limit. See whitepapers/0018_pauser.md.
+    function submitSetPauserAddresses(bytes memory encodedVM) public {
```
johnsaigle, line 91: I think this function should revert if either unpauser or pauser is equal to the zero address. (I left similar comments on the Guardian/Governance PR).

A more nuanced check might be that if the Pauser is non-zero, Unpauser must also be non-zero. I don't think there's a valid case where we would want the ability to Pause without being able to Unpause (unless another Governance VAA is issued).

ethereum/contracts/bridge/BridgeState.sol
```diff
@@ -49,6 +49,19 @@ contract BridgeStorage {
 
         // EIP-155 Chain ID
         uint256 evmChainId;
+
```
johnsaigle: We could consider explicitly setting these values to be zero/false when initializing or upgrading the contract. This could prevent a weird base where dirty slots cause the contract to become paused by mistake after an upgrade. 

On the other hand, if "upgrade while paused" is a pathway we want to leave open: a different way to approach this would be to find a way to make sure that if we upgrade while paused the contract doesn't become unpaused, or vice-versa.

I haven't reviewed the Solana implementation yet, but I noticed that it does zero-out the new memory representing the paused state. It would be good to be consistent across the implementation with respect to potentially non-zeroed out memory.

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -32,10 +32,86 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+    ///      the 24,576-byte EIP-170 limit.
+
+    /// @notice Reverts when a `notPaused`-guarded entry point is called while the bridge is paused.
+    error BridgePaused();
+    /// @notice Reverts when `pause()` is called by anyone other than the configured pauser.
+    error NotPauser();
+    /// @notice Reverts when `unpause()` is called by anyone other than the configured unpauser.
+    error NotUnpauser();
+    /// @notice Reverts when `msg.value` does not cover the wormhole message fee.
+    error WormholeFeeTooLarge();
```
johnsaigle: This name is a bit confusing. The error is that the caller's fee was insufficient, not that the fee is set to a value that's too high, right? This could be something like `InsufficientFee()`

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -32,10 +32,86 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+    }
+
+    modifier notPaused() {
+        _requireNotPaused();
+        _;
+    }
+
+    /// @notice Pause the bridge. Only callable by the configured pauser. Configured via the
+    ///         SetPauserAddressesEvm (action 4) governance VAA.
+    function pause() external {
```
johnsaigle, line 106: Should we have isFork checks on the pause/unpause? It might be good to document the intention one way or the other.

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -32,10 +32,86 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
         uint64 indexed sequence
     );
 
+    /// @notice Emitted when the bridge is paused.
+    event Paused(address indexed by);
+
+    /// @notice Emitted when the bridge is unpaused.
+    event Unpaused(address indexed by);
+
+    /// @dev Custom errors are used in place of revert strings to keep `BridgeImplementation` under
```
johnsaigle, line 41: The whitepaper says that the changes should be backward compatible. I think that changing the error strings can affect other contracts' error handling as well as off-chain tooling.

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,112 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
-        require(rci.evmChainId == block.chainid, "invalid EVM Chain");
+        if (rci.evmChainId != block.chainid) revert InvalidEVMChain();
 
         // Update the chainIds
         setEvmChainId(rci.evmChainId);
         setChainId(rci.newChainId);
     }
 
-    function verifyGovernanceVM(bytes memory encodedVM) internal view returns (IWormhole.VM memory parsedVM, bool isValid, string memory invalidReason){
+    /// @dev Reverts directly on any failure path. Returns the parsed VM on success. Callers no longer
```
johnsaigle: The "Callers no longer..." comment should be removed or reworded so it reads as actual behaviour, not a changelog. It could maybe added to the commit history if you think it's important.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    address constant PAUSER = address(0xCAFE);
+    address constant UNPAUSER = address(0xBEEF);
+
+    // Re-declared so vm.expectEmit can match (Solidity 0.8.4 doesn't allow IName.Event references).
+    event PauserAddressesSet(address indexed pauser, address indexed unpauser);
+    event Paused(address indexed by);
+    event Unpaused(address indexed by);
+
+    uint256 constant testGuardian =
```
johnsaigle, line 44: It would be cool to generate the actual bytes from #4769 and parse them as a unit test here.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+        vm.expectEmit(true, true, true, true);
+        emit PauserAddressesSet(PAUSER, UNPAUSER);
+        bridge.submitSetPauserAddresses(vaa);
+
+        assertEq(bridge.pauser(), PAUSER);
+        assertEq(bridge.unpauser(), UNPAUSER);
+        assertFalse(bridge.paused());
+    }
+
+    function testSubmitSetPauserAddresses_Revert_WrongChain() public {
```
johnsaigle, line 159: Depending on which way we go with the zero-address checks, it would be good to confirm that behaviour in a test like this one.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+            ACTION_SET_PAUSER_ADDRESSES_SOLANA,
+            testChainId,
+            PAUSER,
+            UNPAUSER
+        );
+        vm.expectRevert(ITokenBridge.WrongAction.selector);
+        bridge.submitSetPauserAddresses(_signAndEncodeVM(payload, 0));
+    }
+
+    function testSubmitSetPauserAddresses_Revert_WrongModule() public {
```
johnsaigle, line 181: We can add tests also for Wrong Governance Emitter, Wrong Governance Contract.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    function testNotPaused_GovernanceStillWorksWhenPaused() public {
+        _configurePauser();
+        vm.prank(PAUSER);
+        bridge.pause();
+
+        // submitSetPauserAddresses is governance and must remain callable when paused.
+        bytes memory payload = _setPauserAddressesPayload(testChainId, address(0xFEED), UNPAUSER);
+        bridge.submitSetPauserAddresses(_signAndEncodeVM(payload, 7));
+        assertEq(bridge.pauser(), address(0xFEED));
```
johnsaigle: This could be expanded to also update and assert the unpauser gets modified too.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    function testUnpause_Revert_NotUnpauser() public {
+        _configurePauser();
+        vm.prank(PAUSER);
+        bridge.pause();
+        vm.expectRevert(ITokenBridge.NotUnpauser.selector);
+        bridge.unpause();
+    }
+
+    function testUnpause_NotGuardedByNotPaused() public {
```
johnsaigle, line 355: We should add a test one way or the other for:
* registerChain
* upgrade
* submitRecoverChainId

I assume we want to be able to upgrade while paused, but registration and recovery should not work when paused.

UPDATE: I went back to the whitepaper which specifies all governance messages should be allowed while paused. Cross-referencing this comment with another one that I left on the whitepaper: https://github.com/wormhole-foundation/wormhole/pull/4767/changes#r3235302871

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+            1
+        );
+        bridge.submitSetPauserAddresses(v2);
+        assertEq(bridge.pauser(), newPauser);
+        assertEq(bridge.unpauser(), newUnpauser);
+    }
+
+    // ============================ pause / unpause ============================
+
+    function testPause_Success() public {
```
johnsaigle, line 319: We could add tests here for idempotency of pause and unpause. Basically just to lock in that this works like a no-op rather than a toggle if two pauses or unpauses are submitted in a row.

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    function testSubmitSetPauserAddresses_Revert_AlreadyConsumed() public {
+        bytes memory payload = _setPauserAddressesPayload(testChainId, PAUSER, UNPAUSER);
+        bytes memory vaa = _signAndEncodeVM(payload, 0);
+        bridge.submitSetPauserAddresses(vaa);
+        vm.expectRevert(ITokenBridge.GovernanceActionConsumed.selector);
+        bridge.submitSetPauserAddresses(vaa);
+    }
+
+    function testSubmitSetPauserAddresses_CanRotate() public {
```
johnsaigle, line 302: it would be good to extend this with a case where the bridge is paused.

johnsaigle (commented):
Mostly minor comments regarding backward-compatibility, upgrade concerns, and increasing the test coverage.

johnsaigle:
@claude

ethereum/forge-test/BridgePauser.t.sol
```diff
@@ -0,0 +1,333 @@
...
+    function testNotPaused_GovernanceStillWorksWhenPaused() public {
+        _configurePauser();
+        vm.prank(PAUSER);
+        bridge.pause();
+
+        // submitSetPauserAddresses is governance and must remain callable when paused.
+        bytes memory payload = _setPauserAddressesPayload(testChainId, address(0xFEED), UNPAUSER);
+        bridge.submitSetPauserAddresses(_signAndEncodeVM(payload, 7));
+        assertEq(bridge.pauser(), address(0xFEED));
+    }
```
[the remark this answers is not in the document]
johnsaigle: Damn I didn't know that this plugin read my reviews.

ethereum/contracts/bridge/BridgeState.sol
bemic: I agree we should not append to a struct if later storage variables exist in the linearized contract layout. If we clear reentrancy guard there might be a tiny vulnerable window allowing reentrancy during the upgrade. EIP-1967 or ERC-7201 is not a bad idea

evan-gray: thanks for the comments, I intended to pack in paused and move the addresses to ERC-7201. updating.

ethereum/contracts/bridge/BridgeGovernance.sol
evan-gray: the design was updated and this was updated to match

ethereum/contracts/bridge/BridgeState.sol
evan-gray: on the initial upgrade this would be unpaused, though I don't think we should be prescriptive about future upgrades. leaving as is (though the storage slot moved since this review)

ethereum/contracts/bridge/Bridge.sol
evan-gray: concur, updating

ethereum/contracts/bridge/Bridge.sol
evan-gray: I don't think so, might be good to be able to pause the fork without needing full governance. will add a comment

ethereum/contracts/bridge/Bridge.sol
evan-gray: unfortunately there's not a clean way around changing at least some of these, and I'd like to stick to doing them all. are you aware of any integrations that check the individual revert strings on-chain? that seems like it wouldn't be "compatible" anyway since a new revert is being added here, but also seems like a poor integration approach for an upgradeable contract

ethereum/contracts/bridge/BridgeState.sol
```diff
@@ -49,6 +49,19 @@ contract BridgeStorage {
...
+        // See whitepapers/0018_pauser.md.
+        address pauser;
+
+        // Address authorized to call unpause(). Configured via the SetPauserAddresses governance action.
+        // Typically the guardian governance contract for a higher-trust unpause path.
+        address unpauser;
+
+        // Whether the Token Bridge is currently paused. When true, all entry points except governance
+        // and unpause revert.
+        bool paused;
```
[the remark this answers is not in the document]
evan-gray: duplicate, moving `pause` to the Provider storage so it doesn't add an additional SLOAD and the pauser and unpauser addresses to deterministic slots

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -32,10 +32,86 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+    }
+
+    /// @notice Unpause the bridge. Only callable by the configured unpauser. Configured via the
+    ///         SetPauserAddressesEvm (action 4) governance VAA. Note that `unpause` is intentionally
+    ///         exempt from the `notPaused` modifier.
+    function unpause() external {
+        if (msg.sender != unpauser()) revert NotUnpauser();
+        setPaused(false);
+        emit Unpaused(msg.sender);
+    }
```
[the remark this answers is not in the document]
evan-gray: duplicate

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,112 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+    ///      need to unpack a `(vm, valid, reason)` tuple — saves bytecode at every call site.
```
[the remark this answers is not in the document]
evan-gray: why is this bot just regurgitating @johnsaigle 's comments? 😕

ethereum/contracts/bridge/BridgeImplementation.sol
```diff
@@ -10,6 +10,8 @@ import "./Bridge.sol";
 
 
 contract BridgeImplementation is Bridge {
+    error AlreadyInitialized();
```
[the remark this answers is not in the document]
evan-gray: won't do this, fixing the storage instead

ethereum/forge-test/BridgePauser.t.sol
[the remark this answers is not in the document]
evan-gray: duplicate 😕

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,112 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+        if (payload.length != 75) revert WrongLength();
+        if (payload.toBytes32(0) != module) revert WrongModule();
+        if (payload.toUint8(32) != 4) revert WrongAction();
+        if (payload.toUint16(33) != chainId()) revert WrongChainId();
+
+        address newPauser = payload.toAddress(35);
+        address newUnpauser = payload.toAddress(55);
+
+        setPauser(newPauser);
+        setUnpauser(newUnpauser);
```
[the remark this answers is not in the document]
evan-gray: no.

ethereum/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
evan-gray: dup

ethereum/contracts/bridge/Bridge.sol
johnsaigle: I'm not aware of anything specifically. Maybe this could upgrade can be communicated somewhere as a 'breaking change' at least when it comes to error handling. I suppose a GitHub Issue works unless you're aware of a better medium?

ethereum/contracts/bridge/BridgePauserStorage.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: Apache 2
+pragma solidity ^0.8.0;
+
+/// @notice ERC-7201 namespaced storage for the token bridge `pauser` / `unpauser` roles.
+/// @dev Decoupling these roles from `BridgeStorage.State` means adding pauser support does not
+///      shift any pre-existing storage slot (notably `_status` from `ReentrancyGuard` at slot 13)
+///      on an in-place proxy upgrade. The `paused` flag itself lives in `BridgeStorage.Provider`
```
johnsaigle, line 7: I imagine this value will be 0 for all mainnet deployments, so WTTs should continue to be unpaused after an upgrade. Even if the slot was dirty somehow, and the bridge became paused after an upgrade, it could be fixed by governance. 

So no problem here, just wanted to leave a note for completeness.

johnsaigle (approved):
My earlier notes have all been addressed and i don't see any problems with the latest implementations.

bemic (approved):
New storage implementation ERC-7201 looks good and is covered by tests. The rest of the code looks fine as well.

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,141 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
-        require(valid, reason);
+        IWormhole.VM memory vm = verifyGovernanceVM(encodedVM);
 
         setGovernanceActionConsumed(vm.hash);
 
         BridgeStructs.RegisterChain memory chain = parseRegisterChain(vm.payload);
 
-        require((chain.chainId == chainId() && !isFork()) || chain.chainId == 0, "invalid chain id");
-        require(bridgeContracts(chain.emitterChainID) == bytes32(0), "chain already registered");
+        if (!((chain.chainId == chainId() && !isFork()) || chain.chainId == 0)) revert InvalidChainId();
```
kcsongor, line 54: I appreciate not distributing this negation over the original formula 😄 🙏

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -24,77 +24,141 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+        if (rci.evmChainId != block.chainid) revert InvalidEVMChain();
 
         // Update the chainIds
         setEvmChainId(rci.evmChainId);
         setChainId(rci.newChainId);
     }
 
-    function verifyGovernanceVM(bytes memory encodedVM) internal view returns (IWormhole.VM memory parsedVM, bool isValid, string memory invalidReason){
+    /// @dev Verify a governance VM. Reverts on any failure path; returns the parsed VM on success.
+    function verifyGovernanceVM(bytes memory encodedVM) internal view returns (IWormhole.VM memory) {
```
kcsongor, line 152: much better this way 🙏

ethereum/contracts/bridge/BridgePauserStorage.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: Apache 2
+pragma solidity ^0.8.0;
+
+/// @notice ERC-7201 namespaced storage for the token bridge `pauser` / `unpauser` roles.
+/// @dev Decoupling these roles from `BridgeStorage.State` means adding pauser support does not
+///      shift any pre-existing storage slot (notably `_status` from `ReentrancyGuard` at slot 13)
```
kcsongor, line 6: as long as the slots are only appended, shifting the ReentrancyGuard should not be a problem, but agree with using a const slot here

kcsongor (approved):

ethereum/contracts/bridge/BridgePauserStorage.sol
evan-gray: I agree, but the reviewers above were nervous and this didn't seem to hurt.

ethereum/contracts/bridge/Bridge.sol
evan-gray: #4840

Issue #4065, opened 2024-08-07
node: Solana watcher filters transactions based on logs which may be truncated, leading to missed messages


evan-gray:
#### Wormhole

The guardian is responsible for witnessing all core contract message emissions on all chains. On Solana, the message data is stored in a Solana account via the `postMessage` or `postMessageUnreliable` instruction. 

https://github.com/wormhole-foundation/wormhole/blob/46bcc70e9563121eb30b09797ffe5582474fe8ab/node/pkg/watchers/solana/client.go#L192-L193

It must also be able to detect and process all Wormhole messages within a block in less time than it takes the following block to be produced, so as not to fall behind, in addition to all of the other guardian node tasks. As of this writing, according to the [Solana Explorer](https://explorer.solana.com/), the average block time is ~422ms with about 3.7k transactions per second.

Notably, the Solana contract includes only one log (`msg!`) which logs the sequence number. 

https://github.com/wormhole-foundation/wormhole/blob/46bcc70e9563121eb30b09797ffe5582474fe8ab/solana/bridge/program/src/api/post_message.rs#L224-L225

#### Solana

With the advent of [Versioned Transactions](https://solana.com/docs/advanced/versions), `v0` transactions added support for [Address Lookup Tables](https://solana.com/docs/advanced/lookup-tables). This means that, for `v0` transactions, an instruction's program index may be located in the lookup table, and populating that table requires an additional RPC call for the given account. A quick check of a [recent block](https://explorer.solana.com/block/282157964) shows a mix of 172 `v0` transactions, 79 of those with address table lookups, to 3074 `legacy` transactions.

Additionally, there is a long standing Solana node DoS prevention around log messages, [truncating the message log](https://github.com/solana-labs/solana/blob/e0203f22dc83cb792fa97f91dbe6e924cbd08af1/program-runtime/src/log_collector.rs#L31-L40) to a [default of 10k](https://github.com/solana-labs/solana/blob/e0203f22dc83cb792fa97f91dbe6e924cbd08af1/program-runtime/src/log_collector.rs#L4) log lines.

### Current Watcher Implementation

The current guardian watcher (note: there are two, one for `confirmed` and one for `finalized`) performs the following steps:

[...]

PR #4802, opened 2026-05-07
feat(solana): wtt pauser role

evan-gray:
This PR implements pauser functionality for the SVM Token Bridge.

- Design #4809
- Guardian #4810
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802

Note, this change is intended to maintain full backwards compatibility for the program instructions, however it does so by adding the new fields to the existing Config which may break any consumer's borsh deserialization. In order to support this extension within the program, it adds custom parsing to not error when the length is not yet migrated length. The account is extended to the new length on the first set pauser governance. Until the first pauser is set, it is considered unpaused with no pauser/unpauser.

### DIFF
--- a/solana/modules/token_bridge/program/src/api/governance.rs
+++ b/solana/modules/token_bridge/program/src/api/governance.rs
@@ -15,7 +15,7 @@ use crate::{
     },
     TokenBridgeError::{
         InvalidGovernanceKey,
-        InvalidProgramOwner,
+        InvalidSelfProgram,
         InvalidVAA,
     },
     INVALID_VAAS,
@@ -254,7 +254,7 @@ pub fn set_pauser_addresses(
     }
 
     if accs.self_program.key != ctx.program_id {
-        return Err(InvalidProgramOwner.into());
+        return Err(InvalidSelfProgram.into());
     }
     let mut payload = [0u8; 64];
     payload[..32].copy_from_slice(&accs.vaa.pauser.to_bytes());
--- a/solana/modules/token_bridge/program/src/api/governance.rs
+++ b/solana/modules/token_bridge/program/src/api/governance.rs
@@ -11,7 +11,7 @@ use crate::{
     },
     types::{
         write_pauser_addresses,
-        CONFIG_FULL_LEN,
+        CONFIG_WITH_PAUSER_LEN,
     },
     TokenBridgeError::{
         InvalidGovernanceKey,
@@ -182,7 +182,7 @@ pub const PAUSER_ADDRESSES_SET_EVENT_DISCRIMINATOR: [u8; 8] =
 pub struct SetPauserAddresses<'b> {
     pub payer: Mut<Signer<AccountInfo<'b>>>,
 
-    /// Existing token bridge `Config` PDA. Realloc'd from 32 → `CONFIG_FULL_LEN` bytes the first
+    /// Existing token bridge `Config` PDA. Realloc'd from 32 → `CONFIG_WITH_PAUSER_LEN` bytes the first
     /// time this governance VAA is processed; subsequent VAAs only update the tail.
     pub config: Mut<ConfigAccount<'b, { AccountState::Initialized }>>,
 
@@ -227,13 +227,13 @@ pub fn set_pauser_addresses(
     // pauser state must use the `paused()` / `pauser()` / `unpauser()` helpers in `types.rs`,
     // or similar, which check the account length and treat un-migrated accounts as unassigned.
     let config_info = accs.config.info();
-    if config_info.data_len() < CONFIG_FULL_LEN {
+    if config_info.data_len() < CONFIG_WITH_PAUSER_LEN {
         // Top up the lamport balance so the account remains rent-exempt at the new size, then
         // realloc. `zero_init = true` zeroes the new tail bytes — that's the desired
         // initial state for `paused = false`.
         use solana_program::sysvar::Sysvar;
         let rent = solana_program::sysvar::rent::Rent::get()?;
-        let required = rent.minimum_balance(CONFIG_FULL_LEN);
+        let required = rent.minimum_balance(CONFIG_WITH_PAUSER_LEN);
         let current = config_info.lamports();
         if current < required {
             let topup = required - current;
@@ -244,7 +244,7 @@ pub fn set_pauser_addresses(
             );
             invoke_signed(&topup_ix, ctx.accounts, &[])?;
         }
-        config_info.realloc(CONFIG_FULL_LEN, true)?;
+        config_info.realloc(CONFIG_WITH_PAUSER_LEN, true)?;
     }
 
     {
--- a/solana/modules/token_bridge/program/src/api/governance.rs
+++ b/solana/modules/token_bridge/program/src/api/governance.rs
@@ -7,6 +7,11 @@ use crate::{
     messages::{
         GovernancePayloadUpgrade,
         PayloadGovernanceRegisterChain,
+        PayloadSetPauserAddresses,
+    },
+    types::{
+        write_pauser_addresses,
+        CONFIG_FULL_LEN,
     },
     TokenBridgeError::{
         InvalidGovernanceKey,
@@ -162,6 +167,67 @@ pub fn register_chain(
 
     accs.endpoint.chain = accs.vaa.chain;
     accs.endpoint.contract = accs.vaa.endpoint_address;
+
+    Ok(())
+}
+
+#[derive(FromAccounts)]
+pub struct SetPauserAddresses<'b> {
+    pub payer: Mut<Signer<AccountInfo<'b>>>,
+
+    /// Existing token bridge `Config` PDA. Realloc'd from 32 → `CONFIG_FULL_LEN` bytes the first
+    /// time this governance VAA is processed; subsequent VAAs only update the tail.
+    pub config: Mut<ConfigAccount<'b, { AccountState::Initialized }>>,
+
+    /// Governance VAA carrying a `PayloadSetPauserAddresses` (action 5).
+    pub vaa: PayloadMessage<'b, PayloadSetPauserAddresses>,
+
+    /// VAA replay-protection claim, consistent with the rest of the token bridge governance.
+    pub claim: Mut<Claim<'b>>,
+}
+
+#[derive(BorshDeserialize, BorshSerialize, Default)]
+pub struct SetPauserAddressesData {}
+
+pub fn set_pauser_addresses(
+    ctx: &ExecutionContext,
+    accs: &mut SetPauserAddresses,
+    _data: SetPauserAddressesData,
+) -> Result<()> {
+    if INVALID_VAAS.contains(&&*accs.vaa.info().key.to_string()) {
+        return Err(InvalidVAA.into());
+    }
+
+    verify_governance(&accs.vaa)?;
+    claim::consume(ctx, accs.payer.key, &mut accs.claim, &accs.vaa)?;
+
+    // One-time migration: legacy 32-byte Config accounts grow to 97 bytes the first time this
+    // VAA is processed. Subsequent VAAs reuse the already-extended layout. The `paused` flag
+    // is preserved across updates — rotating the pauser/unpauser keys does not implicitly
+    // unpause the bridge.
+    let config_info = accs.config.info();
+    if config_info.data_len() < CONFIG_FULL_LEN {
+        // Top up the lamport balance so the account remains rent-exempt at the new size, then
+        // realloc. `zero_init = true` zeroes the new tail bytes — that's the desired
+        // initial state for `paused = false`.
+        use solana_program::sysvar::Sysvar;
+        let rent = solana_program::sysvar::rent::Rent::get()?;
+        let required = rent.minimum_balance(CONFIG_FULL_LEN);
+        let current = config_info.lamports();
+        if current < required {
+            let topup = required - current;
+            let topup_ix = solana_program::system_instruction::transfer(
+                accs.payer.key,
+                config_info.key,
+                topup,
+            );
+            invoke_signed(&topup_ix, ctx.accounts, &[])?;
+        }
+        config_info.realloc(CONFIG_FULL_LEN, true)?;
+    }
+
+    let mut data = config_info.data.borrow_mut();
+    write_pauser_addresses(&mut data, &accs.vaa.pauser, &accs.vaa.unpauser);
 
     Ok(())
 }
--- a/solana/modules/token_bridge/program/src/api/governance.rs
+++ b/solana/modules/token_bridge/program/src/api/governance.rs
@@ -15,6 +15,7 @@ use crate::{
     },
     TokenBridgeError::{
         InvalidGovernanceKey,
+        InvalidProgramOwner,
         InvalidVAA,
     },
     INVALID_VAAS,
@@ -171,6 +172,12 @@ pub fn register_chain(
     Ok(())
 }
 
+/// Event discriminator: `SHA256("event:PauserAddressesSet")[0..8]`. Emitted via Anchor-style
+/// self-CPI when `set_pauser_addresses` succeeds. Payload: 32-byte pauser pubkey followed by the
+/// 32-byte unpauser pubkey, in that order.
+pub const PAUSER_ADDRESSES_SET_EVENT_DISCRIMINATOR: [u8; 8] =
+    [0xb9, 0xcf, 0x4f, 0x8f, 0x6c, 0x76, 0xdf, 0x6d];
+
 #[derive(FromAccounts)]
 pub struct SetPauserAddresses<'b> {
     pub payer: Mut<Signer<AccountInfo<'b>>>,
@@ -179,11 +186,17 @@ pub struct SetPauserAddresses<'b> {
     /// time this governance VAA is processed; subsequent VAAs only update the tail.
     pub config: Mut<ConfigAccount<'b, { AccountState::Initialized }>>,
 
-    /// Governance VAA carrying a `PayloadSetPauserAddresses` (action 5).
+    /// Governance VAA carrying a `PayloadSetPauserAddresses` (action 4 per whitepaper 0003).
     pub vaa: PayloadMessage<'b, PayloadSetPauserAddresses>,
 
     /// VAA replay-protection claim, consistent with the rest of the token bridge governance.
     pub claim: Mut<Claim<'b>>,
+
+    /// Event authority PDA for Anchor CPI event signing.
+    pub event_authority: Info<'b>,
+
+    /// This program (for self-CPI).
+    pub self_program: Info<'b>,
 }
 
 #[derive(BorshDeserialize, BorshSerialize, Default)]
@@ -205,6 +218,14 @@ pub fn set_pauser_addresses(
     // VAA is processed. Subsequent VAAs reuse the already-extended layout. The `paused` flag
     // is preserved across updates — rotating the pauser/unpauser keys does not implicitly
     // unpause the bridge.
+    //
+    // Borsh back-compat: the first 32 bytes of the account (the original `Config` struct) are
+    // never moved or rewritten, and the `Config::BorshDeserialize` impl in `types.rs` is
+    // hand-rolled to tolerate the trailing tail (instead of the derived impl that would reject
+    // any unread bytes). Off-chain clients that deserialize Config via Borsh against the raw
+    // account data continue to round-trip both before and after this migration. SDKs that read
+    // pauser state must use the `paused()` / `pauser()` / `unpauser()` helpers in `types.rs`,
+    // or similar, which check the account length and treat un-migrated accounts as unassigned.
     let config_info = accs.config.info();
     if config_info.data_len() < CONFIG_FULL_LEN {
         // Top up the lamport balance so the account remains rent-exempt at the new size, then
@@ -226,8 +247,22 @@ pub fn set_pauser_addresses(
         config_info.realloc(CONFIG_FULL_LEN, true)?;
     }
 
-    let mut data = config_info.data.borrow_mut();
-    write_pauser_addresses(&mut data, &accs.vaa.pauser, &accs.vaa.unpauser);
+    {
+        let mut data = config_info.data.borrow_mut();
+        write_pauser_addresses(&mut data, &accs.vaa.pauser, &accs.vaa.unpauser);
+        // `data` borrow dropped here so the self-CPI below can re-borrow the account.
+    }
 
-    Ok(())
+    if accs.self_program.key != ctx.program_id {
+        return Err(InvalidProgramOwner.into());
+    }
+    let mut payload = [0u8; 64];
+    payload[..32].copy_from_slice(&accs.vaa.pauser.to_bytes());
+    payload[32..].copy_from_slice(&accs.vaa.unpauser.to_bytes());
+    emit_event_cpi(
+        ctx,
+        &accs.event_authority,
+        &PAUSER_ADDRESSES_SET_EVENT_DISCRIMINATOR,
+        &payload,
+    )
 }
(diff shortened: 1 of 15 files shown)


johnsaigle:
@claude (experiment) 
Analyze the PR with respect to compliance with #4767. Cross reference this Solana implementation with the EVM implementation in #4801 and identify any important divergences between the two implementations.

solana/modules/token_bridge/program/src/api/governance.rs
```diff
@@ -165,3 +170,64 @@ pub fn register_chain(
...
+    _data: SetPauserAddressesData,
+) -> Result<()> {
+    if INVALID_VAAS.contains(&&*accs.vaa.info().key.to_string()) {
+        return Err(InvalidVAA.into());
+    }
+
+    verify_governance(&accs.vaa)?;
+    claim::consume(ctx, accs.payer.key, &mut accs.claim, &accs.vaa)?;
+
+    // One-time migration: legacy 32-byte Config accounts grow to 97 bytes the first time this
```
johnsaigle, line 217: I suppose this might break off-chain code that is tries to deserialize this with Borsh. Probably worth noting as a code comment or in the PR description in case we need to trace this back later.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    {
+        let data = config_info.data.borrow();
+        let configured = if is_pauser_path {
+            read_pauser(&data)
+        } else {
+            read_unpauser(&data)
+        };
+        if configured == Pubkey::default() {
```
johnsaigle, line 120: I think this is redundant given that the `Pubkey::default()` can't be a Signer anyway, so the check on the next line will fail.

solana/modules/token_bridge/program/src/api/governance.rs
```diff
@@ -165,3 +170,64 @@ pub fn register_chain(
...
+    pub vaa: PayloadMessage<'b, PayloadSetPauserAddresses>,
+
+    /// VAA replay-protection claim, consistent with the rest of the token bridge governance.
+    pub claim: Mut<Claim<'b>>,
+}
+
+#[derive(BorshDeserialize, BorshSerialize, Default)]
+pub struct SetPauserAddressesData {}
+
+pub fn set_pauser_addresses(
```
johnsaigle, line 205: This entrypoint allows for Pubkey::default to be the pauser or unpauser. Later during the pause/unpause operation, a zero-value Pubkey gives the "PauserNotConfigured" error. That error elsewhere means "we haven't upgraded the smart contract yet".

The EVM and Guardian implementations do not (yet) forbid zero-addresses, though I left reviews in those PRs suggesting we could add those checks.

If the zero address is _not_ allowed, this function should enforce that.
If the zero address _is_ allowed, then the pause/unpause error message should be changed from `PauserNotConfigured`.

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -92,6 +101,12 @@ pub enum TokenBridgeError {
     NonexistentTokenMetadataAccount,
     NotMetadataV1Account,
     WrongMetadataAccount,
+    /// `paused == true`. Lifted only by a successful `unpause` call from the configured unpauser.
+    Paused,
+    /// Caller of `pause` / `unpause` is not the configured pauser / unpauser.
+    InvalidPauser,
+    /// Configured pauser / unpauser is the zero pubkey, so that side of the pair is disabled.
```
johnsaigle, line 108: Do we want situations where we have pausing without unpausing? Both options seem bad. We might want to re-evaluate here so that if one of the 'sides' is configured, both are.

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -92,6 +101,12 @@ pub enum TokenBridgeError {
...
+    PauserNotConfigured,
```
johnsaigle, line 109: As noted elsewhere, this also means "contract has not been updated yet".
This should also be brought into alignment with the EVM side which currently allows for the zero address to be configured.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
     }
 }
 
+/// Pauser tail layout written at offset `CONFIG_BORSH_LEN` of the existing `Config` PDA.
+///
+/// Backwards-compatible extension: the `Config` Borsh struct itself stays 32 bytes, so existing
+/// SDK transactions that pass the Config as writable continue to round-trip — solitaire's
+/// `Persist::persist` calls Borsh `serialize` which writes only the first 32 bytes and leaves
+/// the tail untouched.
```
johnsaigle, line 79: I'm not sure if there are cases where strict length-based validation is done. In that case, the presence of a tail might cause problems. It's worth reviewing https://github.com/wormhole-foundation/wormhole-scaffolding at least to see if there's any potential issues there.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+///
+/// Lazy migration: legacy 32-byte accounts are treated as "unpaused, no pauser configured".
+/// The first `SetPauserAddresses` governance VAA realloc's the account to
+/// `CONFIG_BORSH_LEN + PAUSER_TAIL_LEN` bytes and writes the tail.
+///
+/// Tail layout (97-byte total Config account):
+///   bytes 32 .. 33 : `paused` (u8, 0 = false, anything else = true)
```
johnsaigle: This is a bit inconsistent with the writer below. If paused, it writes exactly `1`, otherwise `0`. Then read paused checks for either 0 or _any_ non-zero value.

If a value ends up as neither `0` nor `1` then we probably have corrupted state (somehow). IIRC Rust conventions are that bools are exactly 0x00 or 0x01 and anything else is considered undefined behaviour (in contrast to e.g. C)

In practice, I do think it's probably better to pause the contract if we get an unusual `bool` value, but it would also be valid to panic in this case IMO or do some other error-handling.

Up to you on how you want to handle it, but let's update the code comments here to be explicit that we're doing something a little weird with bools here relative to Rust conventions.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+pub const UNPAUSER_OFFSET: usize = PAUSER_OFFSET + 32;
+pub const PAUSER_TAIL_LEN: usize = 1 + 32 + 32;
+pub const CONFIG_FULL_LEN: usize = CONFIG_BORSH_LEN + PAUSER_TAIL_LEN;
+
+/// Read the `paused` flag, returning `false` for legacy (un-extended) accounts.
+pub fn read_paused(config_data: &[u8]) -> bool {
+    config_data.len() > PAUSED_OFFSET && config_data[PAUSED_OFFSET] != 0
+}
+
+/// Read the configured pauser. Returns the all-zero pubkey for legacy accounts (which pause.rs
```
johnsaigle: As mentioned elsewhere it's also possible at the moment to configure the Pauser to Pubkey::default() even after the Config is extended (and is no longer "legacy"). So this comment could be clarified after we are set on that design decision.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+/// rejects via `PauserNotConfigured`), so callers don't need to special-case un-extended state.
+pub fn read_pauser(config_data: &[u8]) -> Pubkey {
+    if config_data.len() < UNPAUSER_OFFSET {
```
johnsaigle:
```suggestion
    if config_data.len() < CONFIG_FULL_LEN {
```

I don't think it's valid for the Config accounts to be >= UNPAUSER_OFFSET but < CONFIG_FULL_LEN.

This should be safe because CONFIG_FULL_LEN is always strictly greater than UNPAUSER_OFFSET

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+/// The first `SetPauserAddresses` governance VAA realloc's the account to
+/// `CONFIG_BORSH_LEN + PAUSER_TAIL_LEN` bytes and writes the tail.
+///
+/// Tail layout (97-byte total Config account):
+///   bytes 32 .. 33 : `paused` (u8, 0 = false, anything else = true)
+///   bytes 33 .. 65 : `pauser`   (Pubkey)
+///   bytes 65 .. 97 : `unpauser` (Pubkey)
+///
+/// See whitepapers/0018_pauser.md.
+pub const CONFIG_BORSH_LEN: usize = 32;
```
johnsaigle, line 98: This is a nit but it would be nicer to group the LEN variables separately from the OFFSET variables

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+    }
+    Pubkey::new(&config_data[UNPAUSER_OFFSET..UNPAUSER_OFFSET + 32])
+}
+
+pub fn write_paused(config_data: &mut [u8], paused: bool) {
+    config_data[PAUSED_OFFSET] = if paused { 1 } else { 0 };
+}
+
+pub fn write_pauser_addresses(config_data: &mut [u8], pauser: &Pubkey, unpauser: &Pubkey) {
+    config_data[PAUSER_OFFSET..PAUSER_OFFSET + 32].copy_from_slice(&pauser.to_bytes());
```
johnsaigle: Nit: I got confused by this at first because of how the line splits (on GitHub anyway) 

```suggestion
pub fn write_pauser_addresses(config_data: &mut [u8], pauser: &Pubkey, unpauser: &Pubkey) {
    config_data[PAUSER_OFFSET..(PAUSER_OFFSET + 32)].copy_from_slice(&pauser.to_bytes());
```

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+    config_data[UNPAUSER_OFFSET..UNPAUSER_OFFSET + 32].copy_from_slice(&unpauser.to_bytes());
```
johnsaigle:
```suggestion
    config_data[UNPAUSER_OFFSET..(UNPAUSER_OFFSET + 32)].copy_from_slice(&unpauser.to_bytes());
```

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+pub const CONFIG_BORSH_LEN: usize = 32;
+pub const PAUSED_OFFSET: usize = CONFIG_BORSH_LEN;
+pub const PAUSER_OFFSET: usize = PAUSED_OFFSET + 1;
+pub const UNPAUSER_OFFSET: usize = PAUSER_OFFSET + 32;
+pub const PAUSER_TAIL_LEN: usize = 1 + 32 + 32;
+pub const CONFIG_FULL_LEN: usize = CONFIG_BORSH_LEN + PAUSER_TAIL_LEN;
+
+/// Read the `paused` flag, returning `false` for legacy (un-extended) accounts.
+pub fn read_paused(config_data: &[u8]) -> bool {
+    config_data.len() > PAUSED_OFFSET && config_data[PAUSED_OFFSET] != 0
```
johnsaigle:
```suggestion
    config_data.len() > CONFIG_FULL_LEN && config_data[PAUSED_OFFSET] != 0
```

Similar to my other comment. It's not valid for the Config data to be anything other than 32 or CONFIG_FULL_LEN, and CONFIG_FULL_LEN is greater than PAUSED_OFFSET.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -618,3 +626,328 @@ async fn transfer_native_with_payload_in() {
     .await
     .unwrap();
 }
+
+// ============================ Pauser tests (whitepapers/0018_pauser.md) ============================
+//
+// These tests exercise the lazy migration of the `Config` PDA from the legacy 32-byte layout to
+// the 97-byte layout, the configured pauser/unpauser flow, and the `notPaused` gate on a
+// representative non-governance entry point. The legacy compatibility test then confirms an
```
johnsaigle, line 640: It would be good to get an LLM to do each entrypoint. Should be quick to get an exhaustive error message for each entrypoint, even if it makes the integration test a little long.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -618,3 +626,328 @@ async fn transfer_native_with_payload_in() {
...
+// un-migrated bridge still behaves identically to pre-upgrade.
+
+const SET_PAUSER_ADDRESSES_ACTION: u8 = 5;
+
+/// Build a `SetPauserAddresses` governance VAA, post it through the core bridge, and submit it
+/// to the token bridge. Returns the serialized payload so callers can assert what got applied.
+async fn submit_set_pauser_addresses(context: &mut Context, pauser: Pubkey, unpauser: Pubkey) {
```
johnsaigle, line 647: We should add some failure vases for this, similar to what's done on the EVM side: rejecting the wrong addresses, (maybe) rejecting zero addresses.

We should also add the test to make sure it's possible to rotate the keys while paused.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+/// See whitepapers/0018_pauser.md.
+pub const CONFIG_BORSH_LEN: usize = 32;
+pub const PAUSED_OFFSET: usize = CONFIG_BORSH_LEN;
+pub const PAUSER_OFFSET: usize = PAUSED_OFFSET + 1;
+pub const UNPAUSER_OFFSET: usize = PAUSER_OFFSET + 32;
+pub const PAUSER_TAIL_LEN: usize = 1 + 32 + 32;
+pub const CONFIG_FULL_LEN: usize = CONFIG_BORSH_LEN + PAUSER_TAIL_LEN;
+
+/// Read the `paused` flag, returning `false` for legacy (un-extended) accounts.
+pub fn read_paused(config_data: &[u8]) -> bool {
```
johnsaigle: Nit, feel free to ignore:

Because `types` is a public module, these pub functions become part of the Rust crate API for downstream crates. Consider making write/offset helpers `pub(crate)` and only exposing semantic read helpers if downstream crates are expected to inspect the pauser tail. (I'm not sure it makes sense to expose writer methods that refer to on-chain state.)

In terms of naming the public API, I'd prefer names like `paused` or `pauser` to `read_paused` or `read_pauser`. Callers mostly care about the semantic state, not that the implementation reads specific offsets. Code like `assert!(!paused(&data))` is more natural than `assert!(!read_paused(&data))`.

The low-level offset operations can stay private/internal, but the exposed helpers for other crates could be:

```rust
pub fn paused(config_data: &[u8]) -> bool
pub fn pauser(config_data: &[u8]) -> Pubkey
pub fn unpauser(config_data: &[u8]) -> Pubkey
```

Otherwise, these functions could be limited to `pub(crate)` since they're only available as internal helper functions for modifying the on-chain state.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+///   bytes 33 .. 65 : `pauser`   (Pubkey)
+///   bytes 65 .. 97 : `unpauser` (Pubkey)
+///
+/// See whitepapers/0018_pauser.md.
+pub const CONFIG_BORSH_LEN: usize = 32;
+pub const PAUSED_OFFSET: usize = CONFIG_BORSH_LEN;
+pub const PAUSER_OFFSET: usize = PAUSED_OFFSET + 1;
+pub const UNPAUSER_OFFSET: usize = PAUSER_OFFSET + 32;
+pub const PAUSER_TAIL_LEN: usize = 1 + 32 + 32;
+pub const CONFIG_FULL_LEN: usize = CONFIG_BORSH_LEN + PAUSER_TAIL_LEN;
```
johnsaigle: Maybe overkill but you could consider a `const` assertion here in case these values ever change. This makes sure we don't have offset issues with CONFIG_FULL_LEN relative to the offsets.

e.g.

```rust
const _: () = {
    assert!(PAUSED_OFFSET < CONFIG_FULL_LEN);
    assert!(PAUSER_OFFSET + PUBKEY_LEN <= CONFIG_FULL_LEN);
    assert!(UNPAUSER_OFFSET + PUBKEY_LEN <= CONFIG_FULL_LEN);
};
```

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +68,73 @@ impl Owned for Config {
...
+
+/// Read the `paused` flag, returning `false` for legacy (un-extended) accounts.
+pub fn read_paused(config_data: &[u8]) -> bool {
+    config_data.len() > PAUSED_OFFSET && config_data[PAUSED_OFFSET] != 0
+}
+
+/// Read the configured pauser. Returns the all-zero pubkey for legacy accounts (which pause.rs
+/// rejects via `PauserNotConfigured`), so callers don't need to special-case un-extended state.
+pub fn read_pauser(config_data: &[u8]) -> Pubkey {
```
johnsaigle: Each of the `read_*` functions could have `#[must_use]` so that callers are warned if they don't actually read the return value.

https://std-dev-guide.rust-lang.org/policy/must-use.html

johnsaigle (commented):
Minor comments around edge-cases, design intent, and some clarity around comments and the public APIs created in the diff.

From a security perspective things largely look good and conform to the whitepaper and correlate with the EVM implementation.

solana/modules/token_bridge/program/src/api/governance.rs
```diff
@@ -165,3 +170,64 @@ pub fn register_chain(
...
+                config_info.key,
+                topup,
+            );
+            invoke_signed(&topup_ix, ctx.accounts, &[])?;
+        }
+        config_info.realloc(CONFIG_FULL_LEN, true)?;
+    }
+
+    let mut data = config_info.data.borrow_mut();
+    write_pauser_addresses(&mut data, &accs.vaa.pauser, &accs.vaa.unpauser);
```
johnsaigle: We should emit an event here. We could _maybe_ use the Anchor-style self-CPI machinery introduced in #4746 to get a more reliable event that's not subject to log truncation (https://github.com/wormhole-foundation/wormhole/issues/4065).

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -0,0 +1,89 @@
...
+            return Err(PauserNotConfigured.into());
+        }
+        if &configured != signer {
+            return Err(InvalidPauser.into());
+        }
+    }
+
+    let mut data = config_info.data.borrow_mut();
+    write_paused(&mut data, paused);
```
johnsaigle, line 129: We should emit an event here.

johnsaigle (commented):
Notes on event emission.

johnsaigle:
We could consider bumping the Cargo.toml version of the token bridge but maybe that's a lost cause at this point 😅

solana/modules/token_bridge/program/src/api/governance.rs
evan-gray: spec was updated

solana/modules/token_bridge/program/src/lib.rs
evan-gray: yes, spec was updated

solana/modules/token_bridge/program/src/api/pause.rs
evan-gray: spec calls out that this should explicitly be done

solana/modules/token_bridge/program/src/lib.rs
evan-gray: effectively the same as it may be unset.

solana/modules/token_bridge/program/src/types.rs
evan-gray: updated to match rust convention

solana/modules/token_bridge/program/src/types.rs
evan-gray: the design was updated to explicitly allow this.

solana/modules/token_bridge/program/src/api/governance.rs
evan-gray: updated the PR description to make this more clear

solana/modules/token_bridge/program/src/types.rs
evan-gray: hmm, searched [here](https://github.com/search?q=repo%3Awormhole-foundation%2Fwormhole-scaffolding%20token_bridge_config&type=code) which seems [may parse](https://github.com/wormhole-foundation/wormhole-scaffolding/blob/1e5f6d6db2e1d77cca291cd77b5b2e05e97b905e/solana/modules/wormhole-anchor-sdk/src/token_bridge/accounts.rs#L11) the config, but my understanding was that Anchor deserialize never checks if the cursor is exhausted.

Would be good to confirm though 😬

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -50,6 +71,112 @@ impl Owned for Config {
...
+///   bytes 65 .. 97 : `unpauser` (Pubkey)
+///
+/// See the "Pausing" section of whitepapers/0003_token_bridge.md.
+
+// Account-size sentinels: a Config account is either CONFIG_BORSH_LEN (legacy / un-migrated)
+// or CONFIG_FULL_LEN (post-migration). Any other length on a successfully-deserialized account
+// would indicate corruption.
+pub const CONFIG_BORSH_LEN: usize = 32;
+pub const PAUSER_TAIL_LEN: usize = 1 + PUBKEY_BYTES + PUBKEY_BYTES;
+pub const CONFIG_FULL_LEN: usize = CONFIG_BORSH_LEN + PAUSER_TAIL_LEN;
```
kcsongor: small nit: we make this more descriptive, e.g. `CONFIG_WITH_PAUSER_LEN`. The rationale is if we need to do another extension, then we would likely rename this anyway, and it's easier to just do it now

evan-gray: updated, thanks

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -92,6 +101,15 @@ pub enum TokenBridgeError {
...
+    /// `self_program` account passed by the caller does not match this program's `program_id`.
+    /// Only used to defensively guard the self-CPI that emits Anchor-style events.
+    InvalidProgramOwner,
```
johnsaigle:
```suggestion
    InvalidSelfProgram
```

This conflicts a bit with the same error used in the core bridge. In that case it really is doing an ownership check on an account. In this PR, this means that the "self program" value used in the anchor-style event CPI does not match the running program. The use of the word "Owner" is misleading in that case.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -618,3 +632,980 @@ async fn transfer_native_with_payload_in() {
     .await
     .unwrap();
 }
+
+// ============== Pauser tests (whitepapers/0003_token_bridge.md Pausing) ==============
```
johnsaigle, line 636: You could consider adding idempotency tests here similar to the ones in EVM

johnsaigle (commented):
Two small comments, otherwise I'm ready to approve.

solana/modules/token_bridge/program/src/types.rs
johnsaigle: Should we track this in an issue somewhere just to be safe? Either in this repo or on the scaffolding one?

evan-gray: opened one here: https://github.com/wormhole-foundation/wormhole/issues/4837

johnsaigle (approved):

kcsongor (approved):

PR #4804, opened 2026-05-08
node: Add Sui gRPC package

pleasew8t:
This pull request adds a new package called `suiclient`. This is meant to support the migration from Sui JSON RPC to gRPC. It adds a `SuiClient` interface, as well as an implementation supporting the Sui gRPC protocol. After this PR is merged, the next step would be to first update the Sui watcher, and then the Sui transfer verifier - these components might require significant refactoring.

- `suiclient.go` - defines the new `SuiClient` interface, and standard objects that the implementations should return from their underlying queries.
- `suigrpc.go` - defines the (main) Sui gPRC implementation.
- `suigrpc_services.go` - defines service interfaces used by the gRPC implementation. While not strictly necessary, wrapping these services in interfaces allows properly mocking the upstream gRPC service in tests (see `suigrpc_test.go`).
- `suigrpc_test.go` - implements a number of fuzz tests for the gRPC implementation, with a number of seeds to ensure testing is done when only running `go test`.
- `suigrpc_live_test.go` - implements tests that interact with a live Sui network, if the `SUI_GRPC_TEST_LIVE` environment variable is set.
- `utils.go` - provides a helper function for unmarshalling Bcs bytes.

# Testing

Testing for the client is done using fuzz tests, but each with a number of seeds to ensure that `go test` will test for cases where the server might respond with nil.

Note that `suigrpc_live_test.go` includes tests that can be used for interacting with a live Sui network. It showcases how to use the new client, as well as basic unmarshalling of Bcs bytes into instances of Go structs.

# New Dependencies

The pull request adds the `github.com/block-vision/sui-go-sdk` package. This package is maintained by the same folks as Suivision, as far as I can tell. The package adds the following:

- `mystenbcs`: Allows deserializing Bcs bytes into Go structs. Implementation-wise, this is much cleaner than parsing the response JSON, which typically involves hard-to-follow nested JSON objects.
- `pb/sui/rpc/v2` : Pre-built protobuf files, which we don't need to maintain or build ourselves.

```
replace google.golang.org/grpc => google.golang.org/grpc v1.57.1

go get github.com/block-vision/sui-go-sdk
go mod tidy
```

The package is as optional as how badly we want to maintain this ourselves. If we want to avoid the `sui-go-sdk` package as a dependency, we would need to:

- Add the Sui gRPC protobuf files under `/proto`, and add the `option go_package = "github.com/certusone/wormhole/node/pkg/proto/sui/v2;suigrpc"` directive.
- Copy over the deserialization logic from `mystenbcs`. 

# Broken Dependencies

With the new package, `go mod tidy` bumped the versions of a few packages, and then added/removed some. The google grpc package (`google.golang.org/grpc`) got bumped as well, which breaks `kuberesolver/v4` - it relies on an older version of grpc that is not compatible with new versions. To remedy this, the PR adds a `replace` directive, to point `kubersolver` to a newer version.

```
replace github.com/sercand/kuberesolver/v4 => github.com/sercand/kuberesolver/v6 v6.0.1
```

The root cause, however, is old Loki packages that need to be updated. This can be updated in a follow-up PR.

```
$ go mod why github.com/sercand/kuberesolver/v4
# github.com/sercand/kuberesolver/v4
github.com/certusone/wormhole/node/pkg/telemetry
github.com/grafana/loki/clients/pkg/promtail/client
github.com/grafana/loki/pkg/util
github.com/grafana/loki/pkg/util/log
github.com/weaveworks/common/server
github.com/weaveworks/common/httpgrpc/server
github.com/sercand/kuberesolver/v4
```

# Sidequest

Additional files were modified during this pull request. The PR includes changing the use of  `grpc.DialContext()` (deprecated) to `grpc.NewClient()`. Note that for the `publicWebRunnable()` this includes a hand-rolled version of the blocking dial, which isn't officially supported anymore.


johnsaigle:
@claude

node/pkg/node/publicwebRunnable.go
```diff
@@ -57,15 +60,32 @@ func publicwebServiceRunnable(
...
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+		)
 		if dialErr != nil {
 			return fmt.Errorf("failed to dial upstream: %s", dialErr)
 		}
 
+		// grpc.NewClient does not support blocking dials. The following code attempts to preserve the old behaviour.
+		conn.Connect()
+		ctx, cancel := context.WithTimeout(ctx, DialConnectTimeout)
+		defer cancel()
```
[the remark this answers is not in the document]
johnsaigle: We have linting for shadowed variables. Unfortunate that it wasn't caught there.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+	pb "github.com/block-vision/sui-go-sdk/pb/sui/rpc/v2"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+const (
+	SuiGrpcTimeout           = 10 * time.Second
+	SuiGrpcSteamNilThreshold = 100
```
bengtlofgren: Steam or stream?

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+				}
+			}
+		}
+
+	}()
+
+	return subscription, nil
+}
+
+// Create a new SuiClient, with the gRPC service as iplementation.
```
bengtlofgren:
```suggestion
// Create a new SuiClient, with the gRPC service as implementation.
```

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+func NewSuiGrpcClient(rpcURL string, logger *zap.Logger) (SuiClient, error) {
+
+	// Setting the minimum TLS version is a linting requirement, but should ideally be adheared to in production.
```
bengtlofgren:
```suggestion
	// Setting the minimum TLS version is a linting requirement, but should ideally be adhered to in production.
```

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+			}
+		}
+
+	}()
+
+	return subscription, nil
+}
+
+// Create a new SuiClient, with the gRPC service as iplementation.
+func NewSuiGrpcClient(rpcURL string, logger *zap.Logger) (SuiClient, error) {
```
bengtlofgren: It would be great if we could pass custom metadata and dial options to this function (or alternatively creating a `_WithMetadata` fn) so that a user can pass headers

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+		requestedVersion = nil
+	}
+
+	fields := []string{
+		"object_id",
+		// NOTE: "contents" exists only for Move structs. If other types of
+		// objects are desired, "bcs" needs to be specified
+		"contents",
+		"object_type",
+	}
```
bengtlofgren: Instead of hardcoding this `ReadMask` - I would suggest allowing for a custom one passed in. We realised that for example on the `ExecutedTransaction` proto, we wanted `checkpoint` and `timestamp` exposed.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+
+// Replaces `sui_getTransactionBlock`.
+// Docs: https://www.quicknode.com/docs/sui/sui-grpc/ledger/get-transaction
+func (s *SuiGrpcClient) GetTransaction(ctx context.Context, digest string) (SuiTransaction, error) {
+
+	fields := []string{
+		"digest",
+		"events",
+		"effects",
+	}
```
bengtlofgren: Same here about `ReadMask`

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+		BcsBytes:   resp.Object.Contents.Value,
+	}, nil
+}
+
+func (s *SuiGrpcClient) GetLatestCheckpointSN(ctx context.Context) (uint64, error) {
+
+	// NOTE: "digest" can be included here, if there is any need in the future.
+	fields := []string{
+		"sequence_number",
+	}
```
bengtlofgren: Ditto `ReadMask`

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,409 @@
...
+
+	return s.SubscribeToEvents(ctx, eventTypes, eventWriteChannel)
+}
+
+func (s *SuiGrpcClient) SubscribeToEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
+
+	// This stream is only concerned with transaction events
+	fields := []string{
+		"transactions.events",
+	}
```
bengtlofgren, line 190: Same here with readmask, although I see now there is some code in place for parameterising it in e.g `createCheckpointStream` - maybe im missing the big picture

node/pkg/suiclient/suigrpc_live_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Definition converted from the Sui module's definition.
+type WormholeMessage struct {
+	Sender           [32]byte
+	Sequence         uint64
+	Nonce            uint32
+	Payload          []byte
+	ConsistencyLevel uint8
+	Timestamp        uint64
+}
```
bengtlofgren, line 96: It could be worth documenting how to get these bcs structs - like `sui move summary --bytecode` for verification at least if you do it by hand?

bengtlofgren (changes requested):

bengtlofgren (dismissed):
It looks like it's functional but can use some improvements

bengtlofgren (commented):
[deleted]

node/pkg/suiclient/suigrpc.go
djb15: Added the ability to pass in additional options

node/pkg/suiclient/suigrpc.go
djb15: Re all the `ReadMask` suggestions, the returned structs from each of these methods only expose the fields requested by the hard coded masks. So being able to request additional fields wouldn't actually return any additional data to the caller unless we also changed the return structs.

I would suggest we leave the methods as they are, but I can create `_Raw` variants to pass custom masks if there's a use case you can see? Or instead we just add new methods with the required `ReadMask`s as/when required which would be my preferred path forward

djb15:
@claude

node/pkg/suiclient/suigrpc_live_test.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+		case subErr := <-subscription.Err():
+			{
+				fmt.Printf("subscription reported an error: %v\n", subErr)
+			}
+		case <-ticker.C:
+			{
+				fmt.Println("calling unsubscribe")
+				subscription.Unsubscribe()
+				time.Sleep(2 * time.Second) //nolint:forbidigo // Using time.Sleep here as an arbitrary delay to ensure the subscription closes internally
```
johnsaigle: Can this be reworked? I think this is exactly why we added the foribidigo rule.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
+package suiclient
+
+// TODO: better error handling
```
johnsaigle: Can these be resolved? Specifically the error handling and safe channels

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+const (
+	SuiGrpcTimeout            = 10 * time.Second // TODO: To be used from calling sites
+	SuiGrpcStreamNilThreshold = 100
+	SuiGrpcInvalidVersion     = math.MaxUint64
```
johnsaigle: It's worth a comment with a hyperlink so we know where this value comes from

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
...
+				return
+			}
+
+			// Check that the response and Checkpoint are non-nil before further processing.
+			if resp == nil || resp.Checkpoint == nil {
+
+				// Whenever the stream produces nil, the nil responses counter is incremented. When the counter
+				// reaches a certain threshold, a debug log is produced.
+				streamNilRespCounter = streamNilRespCounter + 1
+				if streamNilRespCounter%SuiGrpcStreamNilThreshold == 0 {
```
johnsaigle:
```suggestion
				if streamNilRespCounter > SuiGrpcStreamNilThreshold {
```

Checking that a count is greater than the threshold is more natural IMO than checking if it's an even division.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
...
+					s.logger.Debug("Sui gRPC nil response update", zap.Uint64("streamNilRespCounter", uint64(streamNilRespCounter)))
```
johnsaigle: It it worth refactoring this into a Prometheus metric? We won't see this value in production.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
...
+
+	subscription := SuiSubscription{
+		err:       errorChannel,
+		ctxCancel: cancel,
+	}
+
+	go func() {
+		defer cancel()
+
+		streamNilRespCounter := uint64(0)
```
johnsaigle: I wonder if this should be refactored into a nils-per-hour or something like that. Otherwise this value appears to increase forever and only reset after a Guardian restart.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,419 @@
...
+					// If the event type does not match the event types passed to the function, ignore it.
+					if !slices.Contains(eventTypes, *grpcEvent.EventType) {
+						continue
+					}
+
+					suiEvent := grpcEventToSuiEvent(grpcEvent)
+
+					// The grpcEvent was malformed, so ignore it.
+					if suiEvent == nil {
+						s.logger.Debug("Sui gRPC event was malformed")
```
johnsaigle: Is this worth turning into a Warning? How often is this expected to happen?

node/pkg/suiclient/suigrpc_services.go
```diff
@@ -0,0 +1,31 @@
+package suiclient
+
+import (
+	"context"
+
+	pb "github.com/block-vision/sui-go-sdk/pb/sui/rpc/v2"
+)
+
```
johnsaigle, line 8: Could you explain why these types exist? Functionally they seem identical to the wrapped types from `pb`. 

Is the idea here to separate the Grpc implementation from that type? Or to otherwise help with testing? It would be good to add some comments about the motivation here because the code as written seems redundant otherwise.

node/pkg/suiclient/suigrpc.go
djb15: I actually think these are stale TODOs, I don't think the error handling is that poor. Is there anything specific you had in mind?

I made a couple of small changes to the channel handling but it also look generally ok

node/pkg/suiclient/suigrpc.go
djb15: It doesn't actually come from anywhere, it's just an infeasible version number that's used to signal intent to query for the latest version. I will add a comment though

node/pkg/suiclient/suigrpc_services.go
djb15: https://github.com/pleasew8t/wormhole/blob/2552220bf5ddccc8028e3f00bac7849ed85cbd8d/node/pkg/suiclient/suigrpc.go#L44-L48

node/pkg/suiclient/suigrpc.go
djb15: I think `Warning` is fair here, updated

node/pkg/suiclient/suigrpc.go
djb15: Replaced with Prometheus metric

node/pkg/suiclient/suigrpc.go
djb15: Prometheus metric will do the bucketing for us

djb15:
@claude

node/pkg/suiclient/suigrpc.go
johnsaigle: If they're stale then I think we're good to resolve this comment and delete those lines.

mdulin2:
I did a quick sanity check on the flow speed for retrieving all events within a Checkpoint. This actually works pretty well, even over a public RPC. The `Recv()` doesn't seem to lag behind the active Sui checkpoints, which can contain something like 10K txs.

djb15:
Fwiw we now have a combined 10B+ executions across the fuzz harnesses so far with 90% coverage and no bugs surfaced yet

node/pkg/suiclient/suigrpc.go
mdulin2: I agree with the custom value take. The tighter we can make the mask, the more efficient the watcher and other usages of it can be. Hardcoding this doesn't seem like a scalable option imo.

djb15: @mdulin2 I guess my point was that the hardcoded mask is already efficient in the sense it's querying the minimum required fields to perform the intended function. And we'd have to change the return types of all the functions to return fields the caller may or may not want.

But I'm happy to make the changes if you both disagree with my justification

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+// docs: https://www.quicknode.com/docs/sui/sui-grpc/ledger/get-object
+func (s *SuiGrpcClient) GetObject(ctx context.Context, objectID string) (SuiObject, error) {
+	return s.GetObjectAtVersion(ctx, objectID, SuiGrpcInvalidVersion)
+}
+
+// To replace the old `suix_tryMultiGetPastObjects` behaviour, this method needs to be called
+// for each version of interest.
+//
+// docs: https://www.quicknode.com/docs/sui/sui-grpc/ledger/get-object
+func (s *SuiGrpcClient) GetObjectAtVersion(ctx context.Context, objectID string, version uint64) (SuiObject, error) {
```
johnsaigle: I think `SuiGrpcInvalidVersion` is not a good way to signal the intent of getting the latest object.

My suggestion would be to change version to `*uint64` and pass it through as a pointer, which should preserve the manual `nil` behaviour below. `GetObject` can be changed to use `nil` as its 3rd paramter, and `SuiGrpcInvalidVersion` can be deleted.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+// of mocks that return different data. Requiring the implementations of the interfaces to do additional parsing
+// is unnecessary.
+type SuiGrpcClient struct {
+	conn                        *grpc.ClientConn
+	logger                      *zap.Logger
+	pbLedgerServiceClient       GrpcLedgerServiceClientInterface       //pb.LedgerServiceClient
+	pbSubscriptionServiceClient GrpcSubscriptionServiceClientInterface //pb.SubscriptionServiceClient
+}
+
+// Calls `GetObjectAtVersion` with an invalid version, used to signal retrieving the
```
johnsaigle: `// GetObject retrieves the most recent version of the object.`

I think the 'invalid version' is an implementation detail and this comment can instead just describe what the function does.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+					// If the event type does not match the event types passed to the function, ignore it.
+					if !slices.Contains(eventTypes, *grpcEvent.EventType) {
+						continue
+					}
+
+					suiEvent := grpcEventToSuiEvent(grpcEvent)
+
+					// The grpcEvent was malformed, so ignore it.
+					if suiEvent == nil {
+						s.logger.Warn("Sui gRPC event was malformed")
```
johnsaigle: This log probably needs some details in order to be actionable

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+
+func (s *SuiGrpcClient) SubscribeToEvent(ctx context.Context, event string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
+	eventTypes := []string{
+		event,
+	}
+
+	return s.SubscribeToEvents(ctx, eventTypes, eventWriteChannel)
+}
+
+func (s *SuiGrpcClient) SubscribeToEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
```
johnsaigle:
```suggestion
func (s *SuiGrpcClient) SubscribeToTransactionsEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
```

Probably better to expose that this method has a hard-coded filter.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// Replaces `sui_getTransactionBlock`.
+// Docs: https://www.quicknode.com/docs/sui/sui-grpc/ledger/get-transaction
+func (s *SuiGrpcClient) GetTransaction(ctx context.Context, digest string) (SuiTransaction, error) {
+
+	fields := []string{
+		"digest",
+		"events",
+		"effects",
```
johnsaigle: `"effects"` doesn't get used by `grpcExecutedTransactionToSuiTransaction` and `SuiTransaction` does not have a field that maps to `effects.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+	getTransactionRequest := pb.GetTransactionRequest{
+		Digest:   &digest,
+		ReadMask: fieldMask(fields),
+	}
+
+	resp, err := s.pbLedgerServiceClient.GetTransaction(ctx, &getTransactionRequest)
+
+	// gRPC call error check
+	if err != nil {
+		return SuiTransaction{}, fmt.Errorf("sui gRPC GetTransaction failed: %w", err)
```
johnsaigle: It would be nice to include the digest in each of these error messages.

node/pkg/suiclient/suigrpc.go
johnsaigle: We should return an error here if `eventTypes` is empty, as well as if eventWriteChannel is nil

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+	if s.conn != nil {
+		return s.conn.Close()
+	}
+	return nil
+}
+
+// Create a new SuiClient, with the gRPC service as implementation. Additional gRPC dial options
+// (e.g. for custom transport credentials, interceptors, or per-RPC metadata via interceptors) may
+// be supplied; they are appended after the defaults so callers can override them.
+func NewSuiGrpcClient(rpcURL string, logger *zap.Logger, extraOpts ...grpc.DialOption) (SuiClient, error) {
```
johnsaigle, line 304: `logger` should be set to zap.NewNop() if nil, otherwise this will panic if a nil logger is specified

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,430 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+const (
+	suiGrpcTimeout        = 10 * time.Second // TODO: Export in future to be used from calling sites
```
johnsaigle: This can be deleted, it seems unused.

node/pkg/suiclient/suigrpc_test.go
```diff
@@ -0,0 +1,514 @@
...
+				}
+
+			}
+		}
+
+		ledgerService.SetNextGetObjectResponse(resp)
+
+		// Request GetObject if version is even
+		if input_version%2 == 0 {
+			grpcClient.GetObject(context.Background(), input_objectId) //nolint:errcheck // The function returning an error is expected
```
johnsaigle:
```suggestion
			_, _ := grpcClient.GetObject(context.Background(), input_objectId)
```

I think it's better to suppress these using the Go syntax rather than fight the linter. We can add a comment at the top of this block to note that we're checking for panics and dropping errors.

node/pkg/suiclient/suigrpc.go
djb15: Nice suggestion, agree that seems like an improvement

node/pkg/suiclient/suigrpc.go
mdulin2: Minimum fields required for what? I think it would be better to have a generic function, where you set the mask, to use across the Sui package. This will need to be used within the watcher and the transfer verifier. So, I think this small rework is worth the effort.

djb15:
@mdulin2 @bengtlofgren I've added generic ReadMask support in https://github.com/wormhole-foundation/wormhole/pull/4804/commits/ee02ba7c06246c2734a804c706b37a235970a88a. Please let me know if this satisfies what you're looking for?

bengtlofgren (dismissed):
Couldn't have done it better myself

mdulin2 (dismissed):

node/pkg/node/publicwebRunnable.go
```diff
@@ -57,18 +60,37 @@ func publicwebServiceRunnable(
...
+			state := conn.GetState()
+
+			// If the connection is ready, just break from the loop.
+			if state == connectivity.Ready {
+				break
+			}
+
+			// Wait until the state changes.
+			if !conn.WaitForStateChange(dialCtx, state) {
+				conn.Close()
```
evan-gray: there seem to be other cases below where an error is returned but `conn.Close()` is not called. 

  - return fmt.Errorf("failed to get systemd listeners: %w", sdErr) (line ~142)
  - return fmt.Errorf("no valid systemd listeners ...") (line ~158)
  - return fmt.Errorf("failed to listen: %v", err) (line ~163)

should those have a `defer func` to close the connection or is it not an issue after `RegisterPublicRPCServiceHandler`?

it was like that before, so perhaps it could stay that way for the time being

node/pkg/node/publicwebRunnable.go
```diff
@@ -16,9 +16,12 @@ import (
 	"golang.org/x/crypto/acme"
 	"golang.org/x/crypto/acme/autocert"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const DialConnectTimeout = 10 * time.Second
```
evan-gray: is this used anywhere else? if not, lowercase it

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,460 @@
...
+// grpcObjectToSuiObject maps a gRPC object into the wrapper struct. Fields are
+// populated only when present in the proto; absent fields are left as their
+// zero value (nil/empty). The Bcs and Contents sub-messages are flattened into
+// the corresponding {Bcs,Contents}{Type,Bytes} pairs. The caller is
+// responsible for nil-checking the fields they read.
+func grpcObjectToSuiObject(grpcObject *pb.Object) SuiObject {
+	out := SuiObject{}
+
+	if grpcObject == nil {
+		return out
```
evan-gray, line 421: I find this design pattern a little awkward in that it coalesces a nil to an empty object, but I understand the desire of not returning a pointer. that said, it does mean the caller is responsible both for nil checking the fields they need and probably nil checking their object first.

node/pkg/suiclient/suigrpc.go
```diff
@@ -0,0 +1,460 @@
...
+	return out
+}
+
+// grpcObjectToSuiObject maps a gRPC object into the wrapper struct. Fields are
+// populated only when present in the proto; absent fields are left as their
+// zero value (nil/empty). The Bcs and Contents sub-messages are flattened into
+// the corresponding {Bcs,Contents}{Type,Bytes} pairs. The caller is
+// responsible for nil-checking the fields they read.
+func grpcObjectToSuiObject(grpcObject *pb.Object) SuiObject {
+	out := SuiObject{}
```
evan-gray, line 418: I thought @johnsaigle had a recent linter change that required these to always have all the fields defined, but maybe that doesn't apply here for some reason

evan-gray (dismissed):
the suiclient package is unused so far, so the ~devil~ bugs may be in the details, but this seams feasible.

node/pkg/suiclient/suigrpc.go
johnsaigle: It's allow-listed to only certain files/structs otherwise there was no way to enable it since we'd have 100s or 1000s of violations. 

I agree though that for new packages we should be using it, and it can be added here: https://github.com/wormhole-foundation/wormhole/blob/aa2aac0100c0a3f28713cea0ae3482ea04f1a28e/.golangci.yml#L119

node/pkg/suiclient/suigrpc.go
djb15: Yeah I agree, that was one of the downsides of the latest change to accept an arbitrary read mask that the other reviewers requested. Unfortunately we can't easily nil check the fields here since we don't know what fields the user will be requesting; we could theoretically nil pass in the requested fields and use that to choose what nil checks to make, but that felt even more clunky to me

node/pkg/suiclient/suigrpc.go
djb15: I attempted to include `suiclient` but it broke a bunch of other lints in other completely unrelated packages. And when I tried to remove the directives that is suggested it errored for the same reason the directive was added in the first place 🤔 So I gave up but will try again in the follow on PR

node/pkg/node/publicwebRunnable.go
djb15: Good catch, I actually think we need to defer the close across the whole function, but as suggested I've just taken out the `Close` calls to avoid fixing what ain't broke

evan-gray: perhaps that `.Close()` was needed as part of reproducing the block? but idk how this code is realistically hit

evan-gray (approved):

node/pkg/node/publicwebRunnable.go
djb15: Nothing special about the blocking dial as far as I could tell, no different to leaking `conn` in the other error paths

mdulin2 (approved):

PR #4805, opened 2026-05-08
fix(node): delegate quorum should use message publication digest

evan-gray:
On a chain with delegated guardians, when some delegates sign a message as an original observation (IsReobservation=false) and others as a re-observation (IsReobservation=true), the canonical guardians never reach delegate quorum even when a quorum of delegates have signed.

This is due to the canonical's per-message delegateState bucket being [keyed on](https://github.com/wormhole-foundation/wormhole/blob/f1da2892a723aa49c7eab2728415bda078356678/node/pkg/processor/observation.go#L641-L648) Keccak256(MarshalBinary(mp)). MarshalBinary [writes](https://github.com/wormhole-foundation/wormhole/blob/f1da2892a723aa49c7eab2728415bda078356678/node/pkg/common/chainlock.go#L332-L337) IsReobservation as a byte, so two otherwise-identical MessagePublications produce different bucket keys. Signatures get distributed across two buckets and neither reaches quorum. 

Here is an example of the mismatched case resulting in a loss of quorum: https://api.wormholescan.io/api/v1/observations/delegate/16/000000000000000000000000b1731c586ca89a23809861c6103f0b96b3f57d92/95838

This PR
1. Fixes the bug by using `CreateDigest` as the digest.
2. Skips the Notary on the delegate-consensus path. It cannot be guaranteed that there is a consensus of `verificationState`s across different guardian configurations and versions. The Notary is also run by the delegates before broadcasting their delegate observations. The Notary functionality can be amended in the future to handle canonical-only specific checks if desired.
3. Adds `MessagePublication.NormalizeForDelegateConsensus()` which resets per-guardian metadata (IsReobservation=true, Unreliable=false, verificationState=NotApplicable) on the canonical's post-quorum MP so downstream behavior doesn't depend on whichever delegate's observation arrived last. 
4. Resolves the TxID via `delegateState.consensusTxID()` which picks the bucket-majority TxID with a lex-smallest tie-break.
5. Adds a comment calling out that the checks enforced in `delegateObservationToMessagePublication` are **not** forward-compatible. I'm not comfortable with the forward incompatible nature of these checks, but this PR is about fixing a more obvious consensus failure whereas these are debatably desirable defensive checks.

A note on the accountant impact:
> While TxID isn't part of the VAA signing digest, it is part of the global-accountant contract's per-entry match: PENDING_TRANSFERS is keyed by (chain, emitter, sequence) but stores a Vec<Data> whose entries are uniquely identified by (guardian_set_index, digest, tx_hash), and signatures only merge across observations with matching tx_hash (cosmwasm/contracts/global-accountant/src/contract.rs:184-205). If canonicals submit divergent TxIDs for the same VAA, their signatures land in separate Data entries and quorum is counted per entry — so the transfer can stall at the accountant even with enough total signatures for the same digest. consensusTxID makes every canonical deterministically pick the same majority TxID from the delegate bucket, so all canonicals submit matching tx_hash and signatures merge into one entry. Recovery via the audit's missing_observations query exists if a split does happen (each guardian re-observation-requests the other side's TxID), but it's noisy and depends on the audit retry loop rather than the delegate-quorum path.

### DIFF
--- a/node/pkg/common/chainlock.go
+++ b/node/pkg/common/chainlock.go
@@ -213,6 +213,37 @@ func (msg *MessagePublication) VerificationState() VerificationState {
 	return msg.verificationState
 }
 
+// NormalizeForDelegateConsensus resets per-guardian metadata to safe defaults
+// for an MP that is the product of delegate-guardian quorum on a canonical
+// guardian.
+//
+// The fields hashed into the VAA signing digest (Timestamp, Nonce, Sequence,
+// ConsistencyLevel, EmitterChain, EmitterAddress, Payload) are the consensus
+// inputs and are preserved. TxID is preserved as-is here; it is NOT part of
+// the VAA signing digest and is NOT part of accountant's aggregation key, so
+// its value here is informational only (logs, persistence, the bytes in the
+// canonical's accountant submission payload). Callers may set it to a
+// deterministic value before invoking this method for tidiness, but no
+// downstream consensus depends on it.
+//
+// The fields reset here are also per-guardian metadata, but unlike TxID they
+// have safe canonical defaults that don't depend on any single delegate:
+//
+//   - IsReobservation: forced true. Canonicals can't trigger a reobservation
+//     for a delegated chain (no watcher is running for it), so treat the
+//     consensus result as already a reobservation and avoid further cycles.
+//   - Unreliable: forced false. Solana-only shortcut for skipping cleanup;
+//     never appropriate for a delegate-derived MP.
+//   - verificationState: forced NotApplicable. The canonical did not verify
+//     the transaction itself, and per-guardian verification results are not
+//     part of consensus. NotApplicable matches the documented contract for
+//     "no verification was performed on this guardian."
+func (msg *MessagePublication) NormalizeForDelegateConsensus() {
+	msg.IsReobservation = true
+	msg.Unreliable = false
+	msg.verificationState = NotApplicable
+}
+
 // SetVerificationState is the setter for verificationState. Returns an error if called in a way that likely indicates a programming mistake.
 // This includes cases where:
 // - an existing state would be overwritten by the NotVerified state
--- a/node/pkg/common/chainlock.go
+++ b/node/pkg/common/chainlock.go
@@ -163,6 +163,34 @@ func (v VerificationState) String() string {
 }
 
 type MessagePublication struct {
+	// TxID is the chain-native transaction identifier where the wormhole
+	// message was emitted. TxID is per-guardian metadata: it is NOT hashed
+	// into the VAA signing digest (see CreateVAA / CreateDigest), so guardians
+	// can disagree on TxID without producing different VAAs.
+	//
+	// TxID IS, however, part of the global-accountant contract's matching key
+	// within a pending transfer. PENDING_TRANSFERS is keyed by
+	// (emitter_chain, emitter_address, sequence), but within that key the
+	// contract stores a Vec<Data> whose entries are uniquely identified by
+	// (guardian_set_index, digest, tx_hash) — see
+	// cosmwasm/contracts/global-accountant/src/contract.rs. Signatures only
+	// merge across observations with matching tx_hash, so if guardians submit
+	// observations for the same VAA but with different TxIDs, their signatures
+	// land in separate Data entries and quorum is counted independently per
+	// entry. If no single entry reaches quorum on its own, the transfer
+	// stalls at the accountant — even when there are enough total signatures
+	// for the same VAA digest.
+	//
+	// Code paths that aggregate observations across guardians (e.g. delegate
+	// quorum on a canonical) should converge on a single TxID before
+	// submitting to the accountant; see the consensusTxID helper and
+	// NormalizeForDelegateConsensus in pkg/processor. That helper picks the
+	// bucket-majority TxID deterministically, which is sufficient for
+	// convergence when one TxID has a clear majority that's stable under
+	// observation arrival order — but with a near-even split, canonicals
+	// whose buckets reach quorum at different observation subsets can still
+	// pick different majorities; recovery for that case is the audit /
+	// missing_observations reobs flow rather than the delegate-quorum path.
 	TxID      []byte
 	Timestamp time.Time
 
@@ -213,18 +241,27 @@ func (msg *MessagePublication) VerificationState() VerificationState {
 	return msg.verificationState
 }
 
-// NormalizeForDelegateConsensus resets per-guardian metadata to safe defaults
-// for an MP that is the product of delegate-guardian quorum on a canonical
-// guardian.
+// NormalizeForDelegateConsensus resets per-guardian metadata on the consensus
+// MP that a canonical guardian forwards downstream after delegate quorum.
+//
+// The purpose is to keep the downstream pipeline (and any state the canonical
+// persists about the message) independent of *which* delegate's observation
+// happened to push the bucket over quorum. Without this, every per-guardian
+// field on the MP would last-writer-wins — different canonicals would
+// disagree on IsReobservation / Unreliable / verificationState based purely
+// on observation arrival order, and the same canonical's logs/database row
+// would depend on a delegate-set race. Resetting these fields to canonical-
+// safe defaults gives a single, reproducible consensus MP across canonicals.
 //
 // The fields hashed into the VAA signing digest (Timestamp, Nonce, Sequence,
 // ConsistencyLevel, EmitterChain, EmitterAddress, Payload) are the consensus
-// inputs and are preserved. TxID is preserved as-is here; it is NOT part of
-// the VAA signing digest and is NOT part of accountant's aggregation key, so
-// its value here is informational only (logs, persistence, the bytes in the
-// canonical's accountant submission payload). Callers may set it to a
-// deterministic value before invoking this method for tidiness, but no
-// downstream consensus depends on it.
+// inputs and are preserved. TxID is preserved as-is here; it is the caller's
+// responsibility to pick a deterministic TxID before invoking this method,
+// because while TxID is not in the VAA signing digest, it IS part of the
+// global-accountant contract's per-entry match — different TxIDs across
+// canonicals split signatures at the accountant and can prevent quorum
+// there. See the TxID field godoc and the consensusTxID helper in
+// pkg/processor.
 //
 // The fields reset here are also per-guardian metadata, but unlike TxID they
 // have safe canonical defaults that don't depend on any single delegate:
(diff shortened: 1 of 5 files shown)


evan-gray:
proven failure: https://github.com/wormhole-foundation/wormhole/actions/runs/25577643645/job/75088635252?pr=4805

mdulin2 (dismissed):

node/pkg/common/chainlock.go
```diff
@@ -213,6 +213,37 @@ func (msg *MessagePublication) VerificationState() VerificationState {
...
+//   - Unreliable: forced false. Solana-only shortcut for skipping cleanup;
+//     never appropriate for a delegate-derived MP.
+//   - verificationState: forced NotApplicable. The canonical did not verify
+//     the transaction itself, and per-guardian verification results are not
+//     part of consensus. NotApplicable matches the documented contract for
+//     "no verification was performed on this guardian."
+func (msg *MessagePublication) NormalizeForDelegateConsensus() {
+	msg.IsReobservation = true
+	msg.Unreliable = false
+	msg.verificationState = NotApplicable
```
johnsaigle: I think `NotVerified` is more appropriate here. The idea with `NotApplicable` is more like "I determined this message does not need to be verified" rather than "I'm not going to try to verify this."

node/pkg/common/chainlock.go
```diff
@@ -213,6 +213,37 @@ func (msg *MessagePublication) VerificationState() VerificationState {
...
+// guardian.
+//
+// The fields hashed into the VAA signing digest (Timestamp, Nonce, Sequence,
+// ConsistencyLevel, EmitterChain, EmitterAddress, Payload) are the consensus
+// inputs and are preserved. TxID is preserved as-is here; it is NOT part of
+// the VAA signing digest and is NOT part of accountant's aggregation key, so
+// its value here is informational only (logs, persistence, the bytes in the
+// canonical's accountant submission payload). Callers may set it to a
+// deterministic value before invoking this method for tidiness, but no
+// downstream consensus depends on it.
```
johnsaigle: Probably worth documenting this info directly on TxID so it's clearer for future changes.

node/pkg/processor/delegate_obs.go
```diff
@@ -30,7 +30,30 @@ func (p *Processor) publishDelegateObservation(d *gossipv1.DelegateObservation)
...
 // delegateObservationToMessagePublication converts a DelegateObservation into a MessagePublication that can be passed through the normal processor pipeline.
-// Returns error on invalid delegate observation input
+// Returns error on invalid delegate observation input.
+//
+// VERSION SKEW: several of the checks below are forward-incompatible — a
+// delegate guardian on a newer build can produce values that an older
+// canonical rejects here, causing the observation to be dropped before it
+// reaches the delegate-quorum bucket. Two cases are particularly relevant:
+//
+//   - KnownChainIDFromNumber: a new chain ID added on delegates ahead of
```
johnsaigle: Note that this function is also used in other places, e.g. `handleSignedDelegateObservation`.

If more flexibility is better, `ChainIDFromNumber` can be used instead to skip the check regarding whether the ChainID is registered. That probably makes sense for this method specifically since it's converting one type to another and doesn't necessarily need to know about the configured state of the network.

On the larger point about canonicals needing to be updated before delegates, I'm not sure how we get around that scenario when we're checking an emitter chain. It seems like that would cause problems during downstream processing anyway, e.g. the Canonical Guardians' governors won't be able to catch transfers from an emitting chain that they don't know about.

node/pkg/processor/observation.go
```diff
@@ -121,6 +122,29 @@ func (p *Processor) handleMessagePublication(ctx context.Context, k *node_common
 	return p.processWithAccountant(ctx, k)
 }
 
+// handleDelegateConsensusMessagePublication runs an MP that has reached
+// delegate-guardian quorum on a canonical guardian through governor and
+// accountant only. The notary is intentionally skipped: each delegate already
+// gated its own broadcast on its local notary (processor.go), and per-guardian
+// notary-managed fields (e.g. verificationState) are not part of consensus, so
+// running the canonical's notary here would couple signing to delegate build
+// lockstep and re-introduce the silent-stall failure mode.
```
johnsaigle: The comments aren't very clear on what the silent-stall failure mode is. Ideally we can come back to these comments are capture them somewhere else where a future maintainer has complete context on the bug without needing to come back to this PR.

node/pkg/processor/observation.go
```diff
@@ -638,14 +662,11 @@ func (p *Processor) handleSignedDelegateObservation(ctx context.Context, m *goss
...
-	buf, err := mp.MarshalBinary()
-	if err != nil {
-		p.logger.Warn("failed to marshal message publication", mp.ZapFields(zap.Error(err))...)
-		delegateObservationsFailedTotal.WithLabelValues("invalid_message_publication").Inc()
-		return nil
-	}
-
-	hash := crypto.Keccak256Hash(buf).Hex()
+	// Key the delegate-state bucket by the VAA signing digest, matching the
+	// canonical observation path (pkg/processor/message.go:70). This excludes
```
johnsaigle:
```suggestion
	// canonical observation path (pkg/processor/message.go). This excludes
```

Just to avoid a line number that is likely to become stale before too long.

node/pkg/processor/observation.go
```diff
@@ -757,12 +778,11 @@ func (p *Processor) handleCanonicalDelegateObservation(ctx context.Context, cfg
...
 	delegateObservationsReceivedByGuardianAddressTotal.WithLabelValues(addr.Hex()).Inc()
 
-	buf, err := mp.MarshalBinary()
-	if err != nil {
-		p.logger.Warn("failed to marshal message publication", mp.ZapFields(zap.Error(err))...)
-		return nil
-	}
-	hash := crypto.Keccak256Hash(buf).Hex()
+	// Key the delegate-state bucket by the VAA signing digest, matching the
+	// canonical observation path (pkg/processor/message.go:70). This excludes
```
johnsaigle:
```suggestion
	// canonical observation path (pkg/processor/message.go). This excludes
```

node/pkg/common/chainlock_test.go
```diff
@@ -965,3 +965,140 @@ func TestMessagePublication_IsWTT(t *testing.T) {
...
+	// Fields hashed into the VAA digest are preserved — verified by digest stability.
+	require.Equal(t, digestBefore, mp.CreateDigest(),
+		"CreateDigest must be unchanged because the normalized fields are not part of the VAA digest")
+}
+
+// TestIsReobservationDoesNotSplitDelegateQuorumBucket asserts the protocol invariant
+// that two otherwise-identical MessagePublications differing only in IsReobservation
+// produce the SAME bucket key for the canonical guardian's delegate-quorum check
+// (pkg/processor/observation.go). The bucket key must match the canonical
+// observation path's key (pkg/processor/message.go:70 — the VAA SigningDigest,
```
johnsaigle:
```suggestion
// observation path's key (pkg/processor/message.go — the VAA SigningDigest,
```

node/pkg/common/chainlock.go
```diff
@@ -213,6 +213,37 @@ func (msg *MessagePublication) VerificationState() VerificationState {
...
+//
+// The fields reset here are also per-guardian metadata, but unlike TxID they
+// have safe canonical defaults that don't depend on any single delegate:
+//
+//   - IsReobservation: forced true. Canonicals can't trigger a reobservation
+//     for a delegated chain (no watcher is running for it), so treat the
+//     consensus result as already a reobservation and avoid further cycles.
+//   - Unreliable: forced false. Solana-only shortcut for skipping cleanup;
```
johnsaigle: This affects Fogo too and so I think this opens up a memory leak if messages aren't cleaned. Given that there isn't a lot of volume on that chain and that not many messages will be Unreliable, this probably won't be a problem in practice.

node/pkg/common/chainlock.go
```diff
@@ -213,6 +213,37 @@ func (msg *MessagePublication) VerificationState() VerificationState {
 	return msg.verificationState
 }
 
+// NormalizeForDelegateConsensus resets per-guardian metadata to safe defaults
```
johnsaigle: It'd be nice to add a comment here explaining that the purpose of the method is to prevent the specific last-writer-win case when aggregating the signatures. Otherwise IMO there's some context missing about why these values are safe and the other ones aren't.

node/pkg/processor/delegate_obs.go
```diff
@@ -30,7 +30,30 @@ func (p *Processor) publishDelegateObservation(d *gossipv1.DelegateObservation)
...
+//     to be updated first.
+//
+//   - NumVariantsVerificationState: a new VerificationState variant added on
+//     delegates is rejected even though the canonical's consensus path
+//     normalizes this field away (NormalizeForDelegateConsensus → NotApplicable).
+//     If enough delegates upgrade ahead of canonicals, the silently-dropped
+//     observations reduce the effective quorum count and re-introduce the
+//     stall pattern — operators only see the per-observation
+//     "failed to convert delegate observation to message publication" warns,
+//     not a per-message symptom.
```
johnsaigle: Not sure I understand the 'per-message symptom' part. These warnings also log the error returned from this function which mentions that the verification state is invalid.

johnsaigle (commented):
These comments are can be resolved -- maybe we can address some of them in another PR.

johnsaigle (dismissed):
I meant to Approve with the previous review, but a commit got merged in between so I wasn't able to.

node/pkg/common/chainlock.go
evan-gray: agree, it is currently SVM-only, but I'm not sure I follow with the memory leak... the messages still get cleaned, this just caused them to get cleaned at a shorter interval and not get re-observed.

evan-gray: the comment wasn't exactly correct, it's not for "skipping cleanup" it's for failing after 5 minutes if consensus isn't reached. I'll update.

johnsaigle: Got it, I was thinking that the delete step might not trigger at all for Fogo which is why I was thinking it could leak.

mdulin2 (dismissed):

node/pkg/processor/delegate_obs.go
evan-gray: I think we can leave this as is for now. agree it's more a design choice about how canonical guardians should handle chains they don't know about (e.g. should they have to take a release before they start signing messages) and I'd rather leave that decision to a specific PR on the topic.

node/pkg/processor/observation.go
```diff
@@ -891,15 +923,22 @@ func (p *Processor) checkForDelegateQuorum(ctx context.Context, mp *node_common.
...
-	// (chain, emitter, sequence); see pkg/accountant/submit_obs.go TransferKey).
-	// Picking a stable value just makes this canonical's logs, database row,
-	// and submitted observation payload consistent rather than reflecting
-	// whichever delegate's observation happened to push the bucket over quorum.
-	// Real divergence is surfaced separately by the warning log in
-	// handleCanonicalDelegateObservation.
+	// Pick a deterministic TxID for the canonical's downstream MP. TxID is
+	// not part of the VAA signing digest, so any value here produces the same
+	// VAA — but TxID IS part of the global-accountant contract's per-entry
+	// match within a pending transfer (see the TxID field godoc on
```
johnsaigle, line 929: In general you can use the [doclink syntax](https://go.dev/doc/comment#doclinks) e.g. `[common.MessagePublication]` in a comment to create a direct godoc-supported link in code comments. This also allows it to work as hover text in editors.

johnsaigle (approved):

node/pkg/processor/observation.go
evan-gray: ooh! noted for next time, thanks 😊

mdulin2 (approved):

PR #4806, opened 2026-05-12
ci: Install JS dependencies from lockfiles

johnsaigle:
- Replace `npm install` and similar with pinned versions and/or installation via lockfiles
- Add https://github.com/asymmetric-research/locked-in to lint for future violations of this policy

See also CONTRIBUTING.md on the JS package manager policy and motivations.

### DIFF
--- a/ethereum/Dockerfile
+++ b/ethereum/Dockerfile
@@ -9,7 +9,8 @@ FROM --platform=linux/amd64 node:22.16-bullseye-slim@sha256:550b434f7edc3a187586
 RUN apt-get update && apt-get -y install \
   git python make curl netcat vim
 
-RUN npm i typescript -g
+# keep in sync with sdk/js/package.json
+RUN npm i typescript@4.3.5 -g
 RUN apt-get -y install jq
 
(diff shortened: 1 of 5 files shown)


ethereum/Dockerfile
```diff
@@ -9,7 +9,8 @@ FROM --platform=linux/amd64 node:22.16-bullseye-slim@sha256:550b434f7edc3a187586
 RUN apt-get update && apt-get -y install \
   git python make curl netcat vim
 
-RUN npm i typescript -g
+# keep in sync with sdk/js/package.json
```
djb15, line 12: Worth adding the inverse comments to the places we want to keep this in sync with? i.e. pointing at each other

johnsaigle: It's not possible to add comments to JSON files. I'm unsure about `.nvmrc` but I think the Dockerfile is probably more visible compared to the hidden config file anyway

djb15: Ah fair point 👍

djb15 (approved):

mdulin2 (approved):

evan-gray (approved):
seems reasonable to me

PR #4809, opened 2026-05-13
feat(whitepapers): token bridge pauser

evan-gray:
This PR splits the Token Bridge specific parts from #4767 

- Design #4809
- Guardian #4810
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802

### DIFF
--- a/whitepapers/0003_token_bridge.md
+++ b/whitepapers/0003_token_bridge.md
@@ -33,6 +33,7 @@ standards-compliant token between chains, creating unique wrapped representation
 * Use a universal token representation that is compatible with most VM data types.
 * Allow domain-specific payload to be transferred along the token, enabling
   tight integration with smart contracts on the target chain.
+* Allow emergency pausing of token bridge operations.
 
 ## Non-Goals
 
@@ -54,13 +55,15 @@ respective Transfer message that is posted to Wormhole, or burns a wrapped token
 
 For inbound transfers they can consume, verify and process Wormhole messages containing a token bridge payload.
 
-There will be five different payloads:
+There will be seven different payloads:
 
 * `Transfer` - Will trigger the release of locked tokens or minting of wrapped tokens.
 * `TransferWithPayload` - Will trigger the release of locked tokens or minting of wrapped tokens, with additional domain-specific payload.
 * `AssetMeta` - Attests asset metadata (required before the first transfer).
 * `RegisterChain` - Register the token bridge contract (emitter address) for a foreign chain.
 * `UpgradeContract` - Upgrade the contract.
+* `RecoverChainId` - Recover the contract's `chainId` and `evmChainId` after a chain fork (EVM-only).
+* `SetPauserAddresses` - Set the addresses authorized to pause and unpause the token bridge.
 
 Since anyone can use Wormhole to publish messages that match the payload format of the token bridge, an authorization
 payload needs to be implemented. This is done using an `(emitter_chain, emitter_address)` tuple. Every endpoint of the
@@ -101,6 +104,14 @@ The metadata of a token can be attested by calling `attestToken` on its respecti
 wormhole network using the details. A token is identified by the tuple `(chain_id, chain_address)` and metadata should
 be mapped to this identifier. A wrapped asset may only ever be created once for a given identifier and not updated.
 
+### Pausing
+
+The token bridge supports an emergency pause for use during an active exploit. While paused, every entry point reverts except for governance handlers and `unpause`. Two roles control the pause state: a `pauser` may call `pause` to set `paused` to `true`, and an `unpauser` may call `unpause` to set it back to `false`. Both roles are configured per chain via a `SetPauserAddresses` governance message.
+
+The `pauser` and `unpauser` are kept as separate roles to allow for asymmetric authority between the two assigned addresses - for example, a 2/3 multisig empowered to pause while Wormhole governance is retained for unpause.
+
+Either role may be left unset. A zero-length value or an all-zero address (of the target chain's native address size) is treated as the role being unassigned, and the corresponding entry point reverts. This allows governance to disable `pause` or `unpause` without removing the entry point - for example, leaving `pauser` unassigned on chains where pause authority is not yet desired, or zeroing a key suspected of compromise without first provisioning its replacement. If `unpauser` is unassigned while the bridge is paused, recovery requires Wormhole governance to first assign a non-zero `unpauser` via `SetPauserAddresses`.
+
 ### Handling of token amounts and decimals
 
 Due to constraints on some supported chains, all token amounts passed through the token bridge are truncated to a maximum of 8 decimals.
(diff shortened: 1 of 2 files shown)


whitepapers/0003_token_bridge.md
```diff
@@ -101,6 +104,14 @@ The metadata of a token can be attested by calling `attestToken` on its respecti
 wormhole network using the details. A token is identified by the tuple `(chain_id, chain_address)` and metadata should
 be mapped to this identifier. A wrapped asset may only ever be created once for a given identifier and not updated.
 
+### Pausing
+
+The token bridge supports an emergency pause for use during an active exploit. While paused, every entry point reverts except for governance handlers and `unpause`. Two roles control the pause state: a `pauser` may call `pause` to set `paused` to `true`, and an `unpauser` may call `unpause` to set it back to `false`. Both roles are configured per chain via a `SetPauserAddresses` governance message.
```
johnsaigle:
```suggestion
The token bridge supports an emergency pause for use during an active exploit. While paused, every entry point reverts except for governance handlers, `pause`, and `unpause`. Two roles control the pause state: a `pauser` may call `pause` to set `paused` to `true`, and an `unpauser` may call `unpause` to set it back to `false`. Both roles are configured per chain via a `SetPauserAddresses` governance message.
```

whitepapers/0003_token_bridge.md
```diff
@@ -101,6 +104,14 @@ The metadata of a token can be attested by calling `attestToken` on its respecti
...
+
+The `pauser` and `unpauser` are kept as separate roles to allow for asymmetric authority between the two assigned addresses - for example, a 2/3 multisig empowered to pause while Wormhole governance is retained for unpause.
+
+Either role may be left unset. A zero-length value or an all-zero address (of the target chain's native address size) is treated as the role being unassigned, and the corresponding entry point reverts. This allows governance to disable `pause` or `unpause` without removing the entry point - for example, leaving `pauser` unassigned on chains where pause authority is not yet desired, or zeroing a key suspected of compromise without first provisioning its replacement. If `unpauser` is unassigned while the bridge is paused, recovery requires Wormhole governance to first assign a non-zero `unpauser` via `SetPauserAddresses`.
```
johnsaigle: "the corresponding entry point reverts" Both endpoints do signature checks on the sender which should be sufficient to gate the feature. Did you want to explicitly revert in this case also?

whitepapers/0003_token_bridge.md
```diff
@@ -234,6 +253,53 @@ ChainId uint16
...
+// Module Identifier  ("TokenBridge" left-padded)
+Module [32]byte
+// Governance Action ID (4 for SetPauserAddresses)
+Action uint8 = 4
+// Target Chain (Where the governance action should be applied)
+ChainId uint16
+
+// Packet
+// Length of the pauser address. Must equal the target chain's native
+// address size (e.g. 20 on EVM, 32 on Solana), or 0 to leave the role
```
johnsaigle, line 288: So for clarity we are going to do length-based checks per-runtime on-chain.

A typical size is either:
* `32 + 1 + 2 + 1 + 20 + 1 + 20` (Module, Action, ChainID, Pauser Len, Pauser Addr, Unpauser Len, Unpauser Addr) -- for EVM

* `32 + 1 + 2 + 1 + 32 + 1 + 32` (Module, Action, ChainID, Pauser Len, Pauser Addr, Unpauser Len, Unpauser Addr) -- for Solana

Other possitibilites are e.g.: 
* `32 + 1 + 2 + 0 + 0 + 1 + 32` -- Disable pauser, set unpauser -- for Solana
* `32 + 1 + 2 + 1 + 20 + 0 + 0` -- Enable pauser, disable unpauser -- for EVM
* `32 + 1 + 2 + 0 + 0 + 0 + 0` -- Disable pause feature -- for EVM


Is that what you had in mind?

whitepapers/0003_token_bridge.md
```diff
@@ -253,6 +319,27 @@ The name and symbol fields of the Transfer payload are not guaranteed to be vali
 Implementations might truncate longer strings at the 32 byte mark, which may result in invalid UTF8 bytes at the end.
 Thus, any client wishing to present these as strings must validate them first, potentially dropping the garbage at the end.
 
+### Backwards compatibility of the pause check
```
johnsaigle, line 324: This might be a good place to clarify whether we also want the ability to rollback the pauser change. The EVM and Solana implementations so far are one-way in that they overwrite certain storage slots or account data and change their sizes. This is backward compatible in the sense that it shouldn't break existing deployments, but it can be a breaking change for integrators, SDKs, off-chain code, etc. Also, there is no way to undo the changes -- that's not strictly relevant for backward compatibility, I suppose, but it seems worth noting somewhere. Maybe under a "Deployment Considerations" section?

whitepapers/0003_token_bridge.md
```diff
@@ -101,6 +104,14 @@ The metadata of a token can be attested by calling `attestToken` on its respecti
 wormhole network using the details. A token is identified by the tuple `(chain_id, chain_address)` and metadata should
 be mapped to this identifier. A wrapped asset may only ever be created once for a given identifier and not updated.
 
+### Pausing
+
+The token bridge supports an emergency pause for use during an active exploit. While paused, every entry point reverts except for governance handlers and `unpause`. Two roles control the pause state: a `pauser` may call `pause` to set `paused` to `true`, and an `unpauser` may call `unpause` to set it back to `false`. Both roles are configured per chain via a `SetPauserAddresses` governance message.
+
+The `pauser` and `unpauser` are kept as separate roles to allow for asymmetric authority between the two assigned addresses - for example, a 2/3 multisig empowered to pause while Wormhole governance is retained for unpause.
```
johnsaigle: Let's add a note here mentioning that they SHOULD be different roles, but that this is intentionally won't be enforced on-chain or in the Guardian VAA generation process.

whitepapers/0003_token_bridge.md
evan-gray: updated

whitepapers/0003_token_bridge.md
evan-gray: I did want to explicitly revert, updated.

whitepapers/0003_token_bridge.md
evan-gray: I don't think all that math is necessary, but essentially yes. updated

whitepapers/0003_token_bridge.md
evan-gray: added a comment to this effect, technically anything can be rolled back via an upgrade 🤷

whitepapers/0003_token_bridge.md
evan-gray: updated

whitepapers/0003_token_bridge.md
johnsaigle: Got it, it was more for my own understanding of the intent here. I agree it's not necessary to in-line in the whitepaper -- but if it's already done, all good!

johnsaigle (approved):

whitepapers/0003_token_bridge.md
```diff
@@ -234,6 +253,55 @@ ChainId uint16
...
+// Target Chain (Where the governance action should be applied)
+ChainId uint16
+
+// Packet
+// Length of the pauser address. Must equal the target chain's native
+// address size (e.g. 20 on EVM, 32 on Solana), or 0 to leave the role
+// unassigned. The receiver rejects any other length. An all-zero
+// address of the native size is equivalent to a zero length and is
+// also treated as unassigned.
+PauserLen uint8
```
bemic, line 292: There is no `PauserLen` and `UnpauserLen` in the draft implementation of the [BridgeGovernance.sol](https://github.com/wormholelabs-xyz/wormhole/blob/529849c4e6f25e84910c2c1998293ae5e3619570/ethereum/contracts/bridge/BridgeGovernance.sol#L78).

evan-gray: correct, it has yet to be updated.

bemic (approved):

PR #4810, opened 2026-05-14
feat(node): token bridge set pauser gov

evan-gray:
Implemented the node portion of #4809 and only that, as opposed to #4769

- Design #4809
- Guardian #4810
- EVM WTT Pauser Role #4801
- SVM WTT Pauser Role #4802


node/pkg/adminrpc/adminserver.go
```diff
@@ -310,6 +310,54 @@ func accountantModifyBalance(req *nodev1.AccountantModifyBalance, timestamp time
...
+		return nil, fmt.Errorf("failed to serialize governance body: %w", err)
+	}
+
+	return vaa.CreateGovernanceVAA(timestamp, nonce, sequence, guardianSetIndex, body), nil
+}
+
+// decodePauserAddress decodes a hex pauser/unpauser address string. An empty
+// string decodes to an empty slice, which the serializer encodes as a
+// zero-length address - i.e. the role being unassigned.
+func decodePauserAddress(field, s string) ([]byte, error) {
```
johnsaigle, line 350: Wanna use sdk's StringToAddress instead? https://github.com/wormhole-foundation/wormhole/blob/4028572892483c79146c14df44993e46401a9b54/sdk/vaa/structs.go#L824-L848

The only difference is that it limits the length to 32 where the implementation here allows for 255. That seems better though IMO, and if we need addresses longer than 32 then there's a good case for updating the SDK to support that.

node/cmd/guardiand/admintemplate.go
```diff
@@ -574,6 +593,81 @@ func runTokenBridgeUpgradeContractTemplate(cmd *cobra.Command, args []string) {
...
+
+	b, err := prototext.MarshalOptions{Multiline: true}.Marshal(m)
+	if err != nil {
+		log.Fatal("failed to marshal request: ", err)
+	}
+	fmt.Print(string(b))
+}
+
+// parsePauserAddressHex parses a pauser/unpauser address argument and returns
+// the lower-case hex encoding used on the wire (no 0x prefix). An empty input
```
johnsaigle: I don't know if this necessarily returns the lower-case hex encoding .

```
go doc hex.DecodeString
package hex // import "encoding/hex"

func DecodeString(s string) ([]byte, error)
    DecodeString returns the bytes represented by the hexadecimal string s.

    DecodeString expects that src contains only hexadecimal characters and that
    src has even length. If the input is malformed, DecodeString returns the
    bytes decoded before the error.
```

I don't think it matters for this context though so this comment might be a misleading by how precise it is.

node/pkg/adminrpc/adminserver.go
johnsaigle: I get that the implementation of these parsing/decoding functions is to allow setting addresses of variable length, but somewhere along the line we should assert that the addresses have a length of either:
* 0 (unset)
* 20 (EVM-like)
* 32 (most other chains)
* ... any other fixed types we're aware of

Otherwise we open the opportunity to fat-finger one of these addresses by 1 character and only discover that on-chain or when the Guardians are unable to produce the same governance VAA.

It probably makes sense to add this check to `Serialize()` since everything will go through there eventually.

node/pkg/adminrpc/prototext_test.go
```diff
@@ -625,6 +625,160 @@ var adminCommandTest = []adminCommandTestEntry{
...
+				messages: {
+					sequence: 1
+					nonce: 2
+					bridge_set_pauser_addresses: {
+						module: "TokenBridge"
+						target_chain_id: 2
+						unpauser: "Hello, World!"
+					}
+				}`,
+	},
```
johnsaigle, line 780: We should add failing checks here for addresses of odd-length or other unsupported values like 18/22 and 31/33 (close to EVM or SVM length, but not quite).
Also we could add a test at the boundary of length 255 (happy path, max length) and 256 (uint8 overflow) if you want to keep support for very long address inputs.

sdk/vaa/payloads.go
```diff
@@ -412,6 +430,21 @@ func (r BodyTokenBridgeUpgradeContract) Serialize() ([]byte, error) {
 	return serializeBridgeGovernanceVaa(r.Module, ActionUpgradeTokenBridge, r.TargetChainID, r.NewContract[:])
 }
 
+func (r BodyTokenBridgeSetPauserAddresses) Serialize() ([]byte, error) {
+	if len(r.Pauser) > math.MaxUint8 {
+		return nil, fmt.Errorf("pauser too long: %d bytes (max %d)", len(r.Pauser), math.MaxUint8)
```
johnsaigle, line 438: The max should probably be 32 here unless we're aware of chains that need support for longer addresses.

sdk/vaa/payloads.go
```diff
@@ -412,6 +430,21 @@ func (r BodyTokenBridgeUpgradeContract) Serialize() ([]byte, error) {
 	return serializeBridgeGovernanceVaa(r.Module, ActionUpgradeTokenBridge, r.TargetChainID, r.NewContract[:])
 }
 
+func (r BodyTokenBridgeSetPauserAddresses) Serialize() ([]byte, error) {
```
johnsaigle, line 433: We could assert `r.Module == TokenBridge` here

node/pkg/adminrpc/adminserver.go
johnsaigle: Alternatively you could delegate all of this into Serialize() and skip the validation in the adminserver since Serialize() is checked and called here anyway using similar error messages.

evan-gray: not sure I'm convinced this should change, at face value, mostly because I'm not sure what future chains may necessarily have fixed size addresses, such as Near and Cardano

node/pkg/adminrpc/prototext_test.go
evan-gray: won't `hex.DecodeString` already reject odd-length hex?

sdk/vaa/payloads.go
evan-gray: near uses longer keys (2-64 bytes). not sure if there will be others in the future, but I don't see a reason to arbitrarily limit this when we know there may be others we don't know about.

node/cmd/guardiand/admintemplate.go
evan-gray: updated the comment

sdk/vaa/payloads.go
evan-gray: updated.

node/pkg/adminrpc/prototext_test.go
johnsaigle: Yes, oops. With your other comment about not enforcing fixed-lengths, this is resolved.

johnsaigle (approved):

panoel (approved):

PR #4812, opened 2026-05-15
node(sdk): Clarify uses and limitations of vaa.Address

johnsaigle:
Adds comments explaining how better to use the SDK's Address type and adds cosmetic refactors to some ambiguous callers.

I'm mainly adding this as a follow-up to several PR reviews where I erroneously recommended that all on-chain addresses be validated against the vaa.Address type. I wasn't aware of how this type ought to be used in several cases, so this should document the correct usage going forward.


djb15 (approved):

bemic (approved):

SEJeff (approved):
For vaa/structs.go ✅

PR #4814, opened 2026-05-15
feat(node): aptos reobserve with endpoint

evan-gray:


mdulin2:
I made some changes to the Aptos watcher to add event verification: https://github.com/wormhole-foundation/wormhole/pull/4838

I think both of these rewrites are fine but we need to come to a solution on how we want this to be structured.

mdulin2:
Can we rebase? I made some changes to some of this code, as mentioned above: https://github.com/wormhole-foundation/wormhole/pull/4838

It shouldn't require too many changes though.

mdulin2 (approved):

PR #4816, opened 2026-05-19
feat(node): multi uri for manager signers

evan-gray:
This PR modifies the manager service to accept an array of signers. This is useful both for testnet testing of a 5/7 multisig under one guardian and for mainnet rotation of running a prior set's key alongside a new key.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -1968,11 +1968,11 @@ func runNode(cmd *cobra.Command, args []string) {
 		}
 		// Ensure the manager signer is not the same as the guardian signer in non-devnet environments
 		if env != common.UnsafeDevNet && uri == *guardianSignerUri {
-			logger.Fatal("dogecoinManagerSignerUri must be different from guardianSignerUri in non-devnet environments")
+			logger.Fatal("dogecoinManagerSignerUris must be different from guardianSignerUri in non-devnet environments")
 		}
 		dogecoinSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, uri, env == common.UnsafeDevNet, "manager-dogecoin")
 		if err != nil {
-			logger.Fatal("failed to create dogecoin manager signer", zap.String("uri", uri), zap.Error(err))
+			logger.Fatal("failed to create dogecoin manager signer", zap.Error(err))
 		}
 		managerSigners[vaa.ChainIDDogecoin] = append(managerSigners[vaa.ChainIDDogecoin], dogecoinSigner)
 
@@ -1990,11 +1990,11 @@ func runNode(cmd *cobra.Command, args []string) {
 			continue
 		}
 		if env != common.UnsafeDevNet && uri == *guardianSignerUri {
-			logger.Fatal("xrplManagerSignerUri must be different from guardianSignerUri in non-devnet environments")
+			logger.Fatal("xrplManagerSignerUris must be different from guardianSignerUri in non-devnet environments")
 		}
 		xrplSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, uri, env == common.UnsafeDevNet, "manager-xrpl")
 		if err != nil {
-			logger.Fatal("failed to create XRPL manager signer", zap.String("uri", uri), zap.Error(err))
+			logger.Fatal("failed to create XRPL manager signer", zap.Error(err))
 		}
 		managerSigners[vaa.ChainIDXRPL] = append(managerSigners[vaa.ChainIDXRPL], xrplSigner)
 
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -311,9 +311,9 @@ var (
 	featureFlags  []string
 	notaryEnabled *bool
 
-	managerServiceEnabled    *bool
-	dogecoinManagerSignerUri *string
-	xrplManagerSignerUri     *string
+	managerServiceEnabled     *bool
+	dogecoinManagerSignerUris []string
+	xrplManagerSignerUris     []string
 )
 
 func init() {
@@ -560,8 +560,8 @@ func init() {
 	notaryEnabled = NodeCmd.Flags().Bool("notaryEnabled", false, "Run the notary")
 
 	managerServiceEnabled = NodeCmd.Flags().Bool("managerServiceEnabled", false, "Run the manager service")
-	dogecoinManagerSignerUri = NodeCmd.Flags().String("dogecoinManagerSignerUri", "", "Dogecoin manager signer URI")
-	xrplManagerSignerUri = NodeCmd.Flags().String("xrplManagerSignerUri", "", "XRPL manager signer URI")
+	NodeCmd.Flags().StringSliceVarP(&dogecoinManagerSignerUris, "dogecoinManagerSignerUris", "", []string{}, "Dogecoin manager signer URI(s)")
+	NodeCmd.Flags().StringSliceVarP(&xrplManagerSignerUris, "xrplManagerSignerUris", "", []string{}, "XRPL manager signer URI(s)")
 }
 
 var (
@@ -1961,17 +1961,20 @@ func runNode(cmd *cobra.Command, args []string) {
 	}
 
 	// Initialize manager signers map
-	managerSigners := make(map[vaa.ChainID]guardiansigner.GuardianSigner)
-	if *dogecoinManagerSignerUri != "" {
+	managerSigners := make(map[vaa.ChainID][]guardiansigner.GuardianSigner)
+	for _, uri := range dogecoinManagerSignerUris {
+		if uri == "" {
+			continue
+		}
 		// Ensure the manager signer is not the same as the guardian signer in non-devnet environments
-		if env != common.UnsafeDevNet && *dogecoinManagerSignerUri == *guardianSignerUri {
+		if env != common.UnsafeDevNet && uri == *guardianSignerUri {
 			logger.Fatal("dogecoinManagerSignerUri must be different from guardianSignerUri in non-devnet environments")
 		}
-		dogecoinSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, *dogecoinManagerSignerUri, env == common.UnsafeDevNet, "manager-dogecoin")
+		dogecoinSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, uri, env == common.UnsafeDevNet, "manager-dogecoin")
 		if err != nil {
-			logger.Fatal("failed to create dogecoin manager signer", zap.Error(err))
+			logger.Fatal("failed to create dogecoin manager signer", zap.String("uri", uri), zap.Error(err))
 		}
-		managerSigners[vaa.ChainIDDogecoin] = dogecoinSigner
+		managerSigners[vaa.ChainIDDogecoin] = append(managerSigners[vaa.ChainIDDogecoin], dogecoinSigner)
 
 		// Log the 33-byte compressed public key for use in P2SH multisig
 		pubKey := dogecoinSigner.PublicKey(rootCtx)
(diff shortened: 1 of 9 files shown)


bemic (commented):
During the Manager Set rotation, managers might end up signing a transaction with a different ManagerSetIndex, because each manager independently reads this value from the contract. Maybe it was discussed already, in that case feel free to ignore. 

We can add check like this in `storeSignature`
```
if aggTx != nil {
    if aggTx.ManagerSetIndex != sig.ManagerSetIndex {
        error ...
     }
}
```

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
+	// LocalSignerIndex is the index into the local signers slice for this chain
+	LocalSignerIndex int
+}
+
+// FindSignerByPubKey checks if a compressed public key is in this manager set.
+// Returns the manager set index and true if found, or 0 and false otherwise.
+func (c *ManagerSetConfig) FindSignerByPubKey(compressedPubKey []byte) (uint8, bool) {
+	for i, pk := range c.PublicKeys {
+		if bytes.Equal(compressedPubKey, pk) {
+			return uint8(i), true // #nosec G115 -- i < N which is uint8
```
johnsaigle: I think this loop could be re-written to tie `i` and `N` directly and also avoid suppressing the lint:

```
	for i := range c.N {
		if bytes.Equal(compressedPubKey, c.PublicKeys[i]) {
			return i, true
		}
	}
	return 0, false
```

Since `c.N` is uint8, `i` should be as well.

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
-	// IsSigner indicates whether this node is part of the manager set
-	IsSigner bool
-	// SignerIndex is this node's index within the manager set (0-based)
-	// Only valid if IsSigner is true
-	SignerIndex uint8
+}
+
+// SignerMatch represents a local signer that is part of a manager set.
+type SignerMatch struct {
+	// ManagerSetIndex is this signer's index within the manager set (0-based)
```
johnsaigle:
```suggestion
	// ManagerSetIndex is this signer's index within the manager set
```
I think the precision here is a bit confusing given that most things are 0-based, and the other `Index` just below this does not have the same annotation

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
+	ManagerSetIndex uint8
+	// LocalSignerIndex is the index into the local signers slice for this chain
+	LocalSignerIndex int
```
johnsaigle: Anyway to write this so that this is also a `uint8`? Mixing styles here is a little strange, and it would be better to tighten up `int` so we can validate it. (I assume this value shouldn't actually be negative or greater than the size of the signers slice, for example)

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
+// Returns a SignerMatch for each local signer found in the set.
+func (c *ManagerSetConfig) FindSigners(ctx context.Context, signers []guardiansigner.GuardianSigner) []SignerMatch {
+	var matches []SignerMatch
+	for localIdx, signer := range signers {
+		signerPubKey := signer.PublicKey(ctx)
+		signerCompressed := compressPublicKey(&signerPubKey)
+		for i, pk := range c.PublicKeys {
+			if bytes.Equal(signerCompressed, pk) {
+				matches = append(matches, SignerMatch{
+					ManagerSetIndex:  uint8(i), // #nosec G115 -- i < N which is uint8
```
johnsaigle: Similar here, you could try this:

```go
func (c *ManagerSetConfig) FindSigners(ctx context.Context, signers []guardiansigner.GuardianSigner) []SignerMatch {
	var matches []SignerMatch
	for localIdx, signer := range signers {
		signerCompressed := compressPublicKey(&signer.PublicKey(ctx))
		for i := range c.N {
			if bytes.Equal(signerCompressed, c.PublicKeys[i]) {
				matches = append(matches, SignerMatch{
					ManagerSetIndex:  i,
					LocalSignerIndex: localIdx,
				})
				break
			}
		}
	}
	return matches
}
```

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
-	IsSigner bool
-	// SignerIndex is this node's index within the manager set (0-based)
-	// Only valid if IsSigner is true
-	SignerIndex uint8
+}
+
+// SignerMatch represents a local signer that is part of a manager set.
+type SignerMatch struct {
+	// ManagerSetIndex is this signer's index within the manager set (0-based)
+	ManagerSetIndex uint8
```
johnsaigle:
```suggestion
	SignerIndex uint8
```
The other ManagerSetIndex values used in this file seem to use uint32 and refer to governance-related values, whereas this seems to be the location of a particular signer within a manager set (and thus bound by `N`, I think?)

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
+	// LocalSignerIndex is the index into the local signers slice for this chain
+	LocalSignerIndex int
+}
+
+// FindSignerByPubKey checks if a compressed public key is in this manager set.
+// Returns the manager set index and true if found, or 0 and false otherwise.
```
johnsaigle:
```suggestion
// Returns the signer index and true if found, or 0 and false otherwise.
```

This looks like it returns the location of a signer within a given manager config's slice, not the uint32 manager set index of which the signer is a member

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,45 @@ type ManagerSetConfig struct {
...
+		if bytes.Equal(compressedPubKey, pk) {
+			return uint8(i), true // #nosec G115 -- i < N which is uint8
+		}
+	}
+	return 0, false
+}
+
+// FindSigners checks which of the given signers' public keys are in this manager set.
+// Returns a SignerMatch for each local signer found in the set.
+func (c *ManagerSetConfig) FindSigners(ctx context.Context, signers []guardiansigner.GuardianSigner) []SignerMatch {
```
johnsaigle: I wonder if it's worth building a cached representation of each manager set config that is indexed by public key instead of by index.

On calls to `GetManagerSet()` you'd build the index so that you could do lookups by uint32 (MangerSetConfig) into hex-encoded pubkeys which are mapped to indexes. (Or use a new custom struct like `SignerPosition`)

That way you could get O(1) lookups instead of needing to do O(n) byte comparisons to retrieve the indices

node/pkg/manager/manager.go
```diff
@@ -710,51 +763,58 @@ func (c *ManagerService) handleXRPLBurnTicket(v *vaa.VAA) {
...
 	}
 
-	// Sign the XRPL AccountSet (burn) transaction
-	sig, err := c.signXRPLBurnTicketTransaction(v, payload, signer)
-	if err != nil {
-		c.logger.Error("failed to sign XRPL burn ticket transaction",
+	chainPubKeys := c.signerPubKeys[vaa.ChainIDXRPL]
+
+	// Sign with each signer configured for XRPL
+	for i, signer := range chainSigners {
```
johnsaigle, line 765: `chainPubKeys[i]` will panic if `len(chainSigners) > len(chainPubKeys`.

Can we add a check or at least a comment somewhere regarding the length of `chainPubKeys` vs. `chainSigners`? I get that they are initialized to be of equal length and that this should always be true. However, there's nothing guaranteeing this in the future, and we should be aware of this during refactoring.

node/pkg/manager/manager.go
johnsaigle, line 1128: I think this part could be written better ~~using `slices.SortFunc` and  `cmp.Compare`~~ with `slices.Sort`:

```go
chainIDs := make([]vaa.ChainID, 0, len(c.signerPubKeys))
for chainID := range c.signerPubKeys {
    chainIDs = append(chainIDs, chainID)
}
slices.Sort(chainIDs)
var parts []string
for _, id := range chainIDs {
    pubKeys := c.signerPubKeys[id]
    for _, pubKey := range pubKeys {
        parts = append(parts, fmt.Sprintf("%d/%s", id, hex.EncodeToString(pubKey)))
    }
}
```

You could consider using `%s` instead of or in addition to `%d` for ChainID to get the human-readable name (implicitly via `ChainID.String()`)

johnsaigle (commented):
I'm not very familiar with the logic of the manager feature so I've made some suggestions for clarity and performance only.

I'm happy to return as a codeowner review to approve the `cmd/` and `node/` changes which appear safe.

node/cmd/guardiand/node.go
```diff
@@ -311,9 +311,9 @@ var (
 	featureFlags  []string
 	notaryEnabled *bool
 
-	managerServiceEnabled    *bool
-	dogecoinManagerSignerUri *string
-	xrplManagerSignerUri     *string
+	managerServiceEnabled     *bool
+	dogecoinManagerSignerUris []string
```
johnsaigle, line 315: I think the Tiltfile needs to be updated for these values, e.g. https://github.com/wormhole-foundation/wormhole/blob/aeda92ed866d03a32c65a7209107a21374e45a86/Tiltfile#L413

node/cmd/guardiand/node.go
```diff
@@ -1961,17 +1961,20 @@ func runNode(cmd *cobra.Command, args []string) {
...
 		// Ensure the manager signer is not the same as the guardian signer in non-devnet environments
-		if env != common.UnsafeDevNet && *dogecoinManagerSignerUri == *guardianSignerUri {
+		if env != common.UnsafeDevNet && uri == *guardianSignerUri {
 			logger.Fatal("dogecoinManagerSignerUri must be different from guardianSignerUri in non-devnet environments")
 		}
-		dogecoinSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, *dogecoinManagerSignerUri, env == common.UnsafeDevNet, "manager-dogecoin")
+		dogecoinSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, uri, env == common.UnsafeDevNet, "manager-dogecoin")
 		if err != nil {
-			logger.Fatal("failed to create dogecoin manager signer", zap.Error(err))
+			logger.Fatal("failed to create dogecoin manager signer", zap.String("uri", uri), zap.Error(err))
```
johnsaigle: Are these values at all sensitive? Maybe we should avoid logging them.

node/cmd/guardiand/node.go
```diff
@@ -1982,15 +1985,18 @@ func runNode(cmd *cobra.Command, args []string) {
...
+			continue
+		}
+		if env != common.UnsafeDevNet && uri == *guardianSignerUri {
 			logger.Fatal("xrplManagerSignerUri must be different from guardianSignerUri in non-devnet environments")
 		}
-		xrplSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, *xrplManagerSignerUri, env == common.UnsafeDevNet, "manager-xrpl")
+		xrplSigner, err := guardiansigner.NewGuardianSignerFromUriWithPurpose(rootCtx, uri, env == common.UnsafeDevNet, "manager-xrpl")
 		if err != nil {
-			logger.Fatal("failed to create XRPL manager signer", zap.Error(err))
+			logger.Fatal("failed to create XRPL manager signer", zap.String("uri", uri), zap.Error(err))
```
johnsaigle: Ditto as above.

node/cmd/guardiand/node.go
johnsaigle: Do these need to be enabled in Tilt btw? I'm surprised the job completed, unless these simply aren't invoked

node/pkg/manager/manager.go
evan-gray: this referenced its index in the configured signers, nothing related to the length of the signers in any one signing set. but this code is now removed anyway.

node/pkg/manager/manager.go
evan-gray: removed

node/pkg/manager/manager.go
evan-gray: added a comment

node/cmd/guardiand/node.go
evan-gray: they're not actively used in CI, but I updated them.

node/cmd/guardiand/node.go
evan-gray: not sure, but agree the `guardianSignerUri` doesn't appear to be logged, so I'll remove this.

node/cmd/guardiand/node.go
evan-gray: removed.

node/pkg/manager/manager.go
```diff
@@ -64,11 +64,16 @@ type ManagerSetConfig struct {
...
-	// Only valid if IsSigner is true
-	SignerIndex uint8
+	// pubKeyIndex maps each compressed public key (as a string) to its position
+	// within PublicKeys for O(1) lookups by FindSignerByPubKey.
+	pubKeyIndex map[string]uint8
+}
+
+// FindSignerByPubKey checks if a compressed public key is in this manager set.
+// Returns the signer index and true if found, or 0 and false otherwise.
+func (c *ManagerSetConfig) FindSignerByPubKey(compressedPubKey []byte) (uint8, bool) {
```
johnsaigle, line 74: nit: The callers seem to treat the `false` case as an error so using a signature like `uint8, error` is probably a bit nicer compared to returning an error on false. This would better capture the idea like "we should only be doing lookups on known keys", if that's the intention.

If you want this to stay flexible or to have a map-lookup sort of feel to it though, then feel free to ignore.

evan-gray: I feel the current approach is more idiomatic. If anything, I'd sooner drop the helper altogether than make it return an error, because the caller would just end up checking and making its own error anyway, but my preference would be to leave it as is.

bemic (dismissed):

johnsaigle (approved):
Reviewed the diff but I am not familiar with the details of the manager feature.

panoel (approved):

djb15 (approved):
Approving Tiltfile

bemic (approved):
I went over the code one more time, and it looks correct to me.

PR #4818, opened 2026-05-20
sdk: Add function to convert on-chain times to Go timestamp

johnsaigle:
Adds TimeFromUnix() to convert on-chain values to the VAA's wire format type.

We can use this function when building message publications. Currently, callers parse this value manually or else truncate the timestamp value by converting it to int64 (which also generates a gosec warning). We can replace this pattern with this function which should make the code a little easier to read and makes it so callers don't need to think much about handling the time value.

It can also allow us to remove the lint exceptions rather than continuing to cover them up. This also has the fringe benefit of making Wormhole safe past 2107...!

### DIFF
--- a/sdk/vaa/structs_test.go
+++ b/sdk/vaa/structs_test.go
@@ -15,6 +15,8 @@ import (
 	"testing"
 	"time"
 
+	stdmath "math"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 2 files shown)


sdk/vaa/structs_test.go
```diff
@@ -984,6 +986,61 @@ func TestChainIDFromNumber(t *testing.T) {
...
 
+func TestTimeFromUnix(t *testing.T) {
+	type testCase[N number] struct {
+		name    string
+		input   N
+		wantErr bool
+		errMsg  string
+	}
+	// Use int64 to represent all error conditions (overflow, negative).
+	tests := []testCase[int64]{
```
djb15, line 997: Suggest adding at least one "normal" timestamp test as well with real timestamps we'd get from on-chain

djb15 (commented):
Looks good otherwise

sdk/vaa/structs_test.go
johnsaigle: done, ptal 🙏🏻

djb15 (approved):

mdulin2 (approved):

sdk/vaa/structs_test.go
```diff
@@ -984,6 +985,70 @@ func TestChainIDFromNumber(t *testing.T) {
...
+		},
+		{
+			name:    "negative timestamp",
+			input:   int64(-1),
+			wantErr: true,
+			errMsg:  "timestamp cannot be negative",
+		},
+	}
+
+	for _, tt := range tests {
```
SEJeff, line 1032: Minor testify nit. I am not sure these all need to be require. If you use require here instead of assert, every single failed assertion will fail the test. So if there are 3 invalid assertions, you'll have to fix them, run the tests, see the second one, fix it, run the tests, see the third one, and fix it.

If you use assert for most of these (unless you need to require them all!), you'll see all 3 errors in a single failed test run.

[This blog post](https://blog.8-p.info/en/2020/10/14/assert-vs-require/) has a good overview of why with some more specifics.

SEJeff (approved):

sdk/vaa/structs_test.go
johnsaigle: Ah thanks I didn't know that

PR #4819, opened 2026-05-20
node: reorder struct fields to reduce GC pointer scan area

johnsaigle:
Used the fieldalignment linter to reduce the memory required for structs used in message publications. https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment

Basically it's possible to declare the fields of a struct in an optimized order in Go such that it reduces the memory requirements. The rough process is to place pointer-type fields first, then declare variables in descending order
of size. The `fieldalignment` struct has some false positives as it reports theoretical minimums, but using
it iteratively can result in smaller structs.

---

## Details:

Total per-message GC scan savings: 56 B (-25%)
Saved for every observed message — the two major per-message structs are the MessagePublication and its tracking state:
```
  struct                  | before  | after   | delta
  ------------------------|---------|---------|------
  MessagePublication (GC) |  112 B  |   72 B  | -40 B
  state             (GC)  |  168 B  |  152 B  | -16 B
  ------------------------|---------|---------|------
  combined per message    |  280 B  |  224 B  | -56 B
  ```
 
How (MessagePublication, 128 B total struct):
  Before: pointer fields spread across offsets [0..112]
          0:   TxID (24, ptr)    48: EmitterAddr (32)    88: Timestamp (24, ptr)
          24:  Payload (24, ptr)   80: Sequence (8)
          GC scanned 112 B despite only 72 B of actual pointer fields
  After:  pointer fields contiguous at top [0..72]
          0:  TxID (24, ptr)   24: Payload (24, ptr)   48: Timestamp (24, ptr)
          72: EmitterAddr (32) ... non-pointer fields follow
          GC scanned exactly 72 B = the 3 pointer fields
How (state, 168 B total struct):
  Pushed non-pointer retryCtr (uint, 8 B) below the last pointer field
  instead of sitting between pointer-containing fields.
Other savings in processor:
  Processor struct:          288 -> 280 B total   (-8)
  DelegatedGuardianConfig:   resolved
  delegateState:             resolved
  DelegatedGuardianChainConfig: 40 -> 32 ptr B   (-8)

## Test instructions:
 
 1. Write fieldalignment-only config:
 ```sh
  $ cat > /tmp/fa.yml <<'EOF'
  version: "2"
  linters:
    enable: [govet]
    settings:
      govet:
        enable: [fieldalignment]
  EOF
  ```
 
 2. On main:
 ```sh
  $ cd node && golangci-lint run --config=/tmp/fa.yml ./pkg/common/ ./pkg/processor/
 ```
 3. On this branch, same command — compare MessagePublication:  (112->72) and state (168->152) improvements.

### DIFF
--- a/node/pkg/common/chainlock.go
+++ b/node/pkg/common/chainlock.go
@@ -162,6 +162,12 @@ func (v VerificationState) String() string {
 	}
 }
 
+// MessagePublication represents an observation of a message being published on Wormhole.
+// It is the result of a direct on-chain observation of a message being published on a supported chain,
+// or the result of a gossiped observation of a message being published on the p2p network.
+// MessagePublication is the precursor to a [vaa.VAA], containing the same data but before signature verification.
+// It also has several metadata fields that are not part of the VAA signing digest and therefore MUST NOT be relied upon
+// for consensus. However, metadata fields SHOULD be consistent across all guardians for a given message.
 type MessagePublication struct {
 	// TxID is the chain-native transaction identifier where the wormhole
 	// message was emitted. TxID is per-guardian metadata: it is NOT hashed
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: node/pkg/common/chainlock.go [hunk 1]
```diff
@@ -191,21 +197,22 @@ type MessagePublication struct {
 	// whose buckets reach quorum at different observation subsets can still
 	// pick different majorities; recovery for that case is the audit /
 	// missing_observations reobs flow rather than the delegate-quorum path.
-	TxID      []byte
```
mdulin2, line 194: This shouldn't affect database serialization/deserialization, from what I can tell.

mdulin2 (approved):
All seems good to me. Fields are still there, just in a different order. Ordering shouldn't matter in the db case because of the custom marshaller or in the p2p case because it's using different protobuf types for transmission.

djb15 (commented):
Memory usage of message publications doesn't seems like a particular bottleneck for Guardians; at least I'm not aware of throughput being that high that we need to shave off a couple of bytes?

Had no idea this was a thing so that's fun, but I suggest we don't make changes to the critical path at this time unless you have evidence this is an important change? I also don't think this will affect the serialisation/deserialisation but the marginal gain probably isn't worth the risk of being wrong

johnsaigle:
OK, I am fine to close this if it's considered not worth the refactoring risk. We can re-open it if we learn that memory usage becomes an issue that that changing the layout here might improve it.

PR #4822, opened 2026-05-20
node: Validate and sanitize RPC URLs

johnsaigle:
- Checks watcher's URLs against a deny-list of public RPC URLs. Logs an error when a public RPC URL is found for a delegated guardian. This should prevent problems where a delegate is not correct pointed at a full node.
- Sanitizes URLs in logs so that path parts are not included. (These can sometimes contain API keys or other sensitive strings).
- Normalizes logged field names across watchers for the 'watcher started' log

### DIFF
--- a/node/pkg/common/url_verification.go
+++ b/node/pkg/common/url_verification.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"errors"
 	"net"
 	"net/url"
 	"strings"
@@ -39,4 +40,71 @@ func ValidateURL(urlStr string, validSchemes []string) bool {
 		}
 	}
 	return false
+}
+
+// SafeURLForLogging returns only the host and port for a URL-like string.
+// It intentionally omits userinfo, path, query, and fragment because those may contain credentials.
+func SafeURLForLogging(urlStr string) string {
+	if urlStr == "" {
+		return ""
+	}
+
+	parsedURL, err := url.Parse(urlStr)
+	if err != nil || parsedURL.Host == "" {
+		// Schemeless host:port strings parse as scheme:opaque. Parse them again
+		// as network-path references so URL.Hostname can extract the host.
+		parsedURL, err = url.Parse("//" + urlStr)
+	}
+	if err != nil {
+		return "<invalid-url>"
+	}
+
+	if host := parsedURL.Host; host != "" {
+		return host
+	}
+
+	return "<invalid-url>"
+}
+
+// SafeErrorForLogging returns an error string with a raw URL replaced by its safe logging form.
+func SafeErrorForLogging(err error, urlStr string) string {
+	if err == nil {
+		return ""
+	}
+	if urlStr == "" {
+		return err.Error()
+	}
+
+	safeURL := SafeURLForLogging(urlStr)
+	errStr := err.Error()
+
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) && urlErr.URL != "" {
+		// url.Error carries the URL separately. Sanitize that exact value in the
+		// full error string so wrapped errors keep their additional context.
+		errStr = strings.ReplaceAll(errStr, urlErr.URL, SafeURLForLogging(urlErr.URL))
+	}
+
+	errStr = strings.ReplaceAll(errStr, urlStr, safeURL)
+
+	parsedURL, parseErr := url.Parse(urlStr)
+	if parseErr != nil {
+		return errStr
+	}
+
+	// Some libraries report URLs through url.URL.String(), which can escape or otherwise
+	// canonicalize the original string.
+	canonicalURL := parsedURL.String()
+	errStr = strings.ReplaceAll(errStr, canonicalURL, safeURL)
+
+	if parsedURL.User != nil {
+		if _, hasPassword := parsedURL.User.Password(); hasPassword {
+			// net/http redacts userinfo passwords as "***" in *url.Error while leaving
+			// path and query values intact, so the raw URL no longer matches exactly.
+			redactedURL := strings.Replace(canonicalURL, parsedURL.User.String()+"@", parsedURL.User.Username()+":***@", 1)
+			errStr = strings.ReplaceAll(errStr, redactedURL, safeURL)
+		}
+	}
+
+	return errStr
 }
--- a/node/pkg/processor/processor.go
+++ b/node/pkg/processor/processor.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"net"
 	"slices"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -25,6 +27,7 @@ import (
 	"github.com/certusone/wormhole/node/pkg/gwrelayer"
 	gossipv1 "github.com/certusone/wormhole/node/pkg/proto/gossip/v1"
 	"github.com/certusone/wormhole/node/pkg/supervisor"
+	"github.com/certusone/wormhole/node/pkg/watchers"
 	"github.com/wormhole-foundation/wormhole/sdk/vaa"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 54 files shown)


node/pkg/processor/processor.go
```diff
@@ -284,6 +289,101 @@ var (
...
+	"publicnode.com":  {},
+	"llamarpc.com":    {},
+	"1rpc.io":         {},
+	"nodereal.io":     {},
+	"bnbchain.org":    {},
+	"pocket.network":  {},
+	"monad.xyz":       {},
+	"monadinfra.com":  {},
+	"hyperliquid.xyz": {},
+	"t.conduit.xyz":   {},
```
SEJeff, line 311: drpc.org, tenderly.co, stakely, tatum.io might be worth adding.

node/pkg/processor/processor.go
```diff
@@ -284,6 +289,101 @@ var (
...
+// checkPublicRpcEndpoints checks all chains in the DGS for public RPC URLs.
+// It logs an error once per chain per run if a public RPC endpoint is detected.
+// This is a safety measure: delegating guardians that use public RPC endpoints
+// risk producing incorrect VAAs, which can cause consensus failures.
+func (p *Processor) checkPublicRpcEndpoints() {
+	if !p.delegatedGuardiansEnabled {
+		return
+	}
+	if p.dgc == nil {
+		p.logger.Warn("public RPC endpoint check skipped: delegated guardian config is nil")
```
SEJeff:
```suggestion
		p.logger.Debug("public RPC endpoint check skipped: delegated guardian config is nil")
```

Not actionable, so nothing to warn about.

node/pkg/common/url_verification.go
```diff
@@ -40,3 +40,38 @@ func ValidateURL(urlStr string, validSchemes []string) bool {
 	}
 	return false
 }
+
+// SafeURLForLogging returns only the hostname for a URL-like string.
+// It intentionally omits userinfo, path, query, and fragment because those may contain credentials.
```
djb15, line 46: Why also strip the path? I don't think that would realistically contain any secrets plus would be helpful to know the actual path delimited endpoint in some scenarios?

node/pkg/watchers/watchers.go
```diff
@@ -35,3 +37,37 @@ var (
...
+	rpcURLs[chainID] = append(rpcURLs[chainID], rpcURL)
+}
+
+// RPCURLs returns the RPC URLs registered for a chain.
+func RPCURLs(chainID vaa.ChainID) []string {
+	rpcURLsMu.RLock()
+	defer rpcURLsMu.RUnlock()
+	chainRPCURLs := rpcURLs[chainID]
+	if len(chainRPCURLs) == 0 {
+		return nil
```
djb15, line 68: Just wondering about Go convention, returning `nil` here actually returns an empty array right? Is it better to `return nil` or just `return`?

johnsaigle: No, `nil` will return `nil`. Note also that the [the zero-value for slice](https://go.dev/ref/spec#The_zero_value) is `nil` (and not an allocated slice of 0 elements). So, in this case, it's only a stylistic difference and the original code should be basically to the example in this comment.

In practice, callers can be safe either way by using `len()`, because `len(nil) == 0` so you're safe either way. `nil` would be the expectation for getting an "empty slice" result, though.

For the other part of your comment: the bare keyword `return` only works if you used a named ['result parameters'](https://go.dev/ref/spec#Return_statements) in the signature, e.g. in this case if we did:

```go
// RPCURLs returns the RPC URLs registered for a chain.
func RPCURLs(chainID vaa.ChainID) (urls []string) {
	rpcURLsMu.RLock()
	defer rpcURLsMu.RUnlock()
	chainRPCURLs := rpcURLs[chainID]
	if len(chainRPCURLs) == 0 {
		return
	}
	urls = make([]string, len(chainRPCURLs))
	copy(urls, chainRPCURLs)
	return
}
```

(`urls` gets implied if we just use `return`)

However, both of these implementations are logically equivalent. The only difference is style.

node/pkg/common/url_verification.go
johnsaigle: IMO we need it. e.g. for Alchemy, they give you a URL like this: `wss://eth-mainnet.g.alchemy.com/v2/W_...<secret here>...`, and that would be part of the path. I don't know why they do it like that, but we have to work around it.

djb15: Ah good point, thanks

node/pkg/watchers/watchers.go
djb15: > Note also that the [the zero-value for slice](https://go.dev/ref/spec#The_zero_value) is nil (and not an allocated slice of 0 elements)

TIL, thanks!

djb15 (dismissed):

johnsaigle:
@djb15 can you please re-review? I had to rebase and I also squashed the commits. Otherwise everything should be the same.

djb15 (dismissed):
Verified the rebase didn't change anything drastic

node/pkg/common/url_verification.go
```diff
@@ -40,3 +41,70 @@ func ValidateURL(urlStr string, validSchemes []string) bool {
...
+func SafeURLForLogging(urlStr string) string {
+	if urlStr == "" {
+		return ""
+	}
+
+	parsedURL, err := url.Parse(urlStr)
+	if err != nil || parsedURL.Host == "" {
+		// Schemeless host:port strings parse as scheme:opaque. Parse them again
+		// as network-path references so URL.Hostname can extract the host.
+		parsedURL, err = url.Parse("//" + urlStr)
```
mdulin2, line 56: Should we keep the scheme? I don't see the harm and it would make debugging easier.

node/pkg/processor/processor.go
```diff
@@ -284,6 +290,109 @@ var (
...
+			continue
+		}
+
+		rpcURLs := watchers.RPCURLs(chainID)
+		if len(rpcURLs) == 0 {
+			continue
+		}
+
+		for _, rpcURL := range rpcURLs {
+			hostname := common.SafeURLForLogging(rpcURL)
```
mdulin2, line 375: The function is being used for things other than logging. Maybe worth considering a different name?

node/pkg/processor/processor.go
```diff
@@ -284,6 +290,109 @@ var (
...
+	}
+
+	return "", false
+}
+
+// checkPublicRpcEndpoints checks all chains in the DGS for public RPC URLs.
+// It logs an error once per chain per run if a public RPC endpoint is detected.
+// This is a safety measure: delegating guardians that use public RPC endpoints
+// risk producing incorrect VAAs, which can cause consensus failures.
+func (p *Processor) checkPublicRpcEndpoints() {
```
mdulin2, line 343: I *believe* this is still safe for the `ReobserveWithEndpoint` calls. It's worth a double check though.

I think that preventing guardians from using public RPCs is a good call.

node/pkg/processor/processor.go
```diff
@@ -509,6 +619,8 @@ func (p *Processor) Run(ctx context.Context) error {
 
 			if err := p.dgc.Set(chains); err != nil {
 				p.logger.Error("delegate guardian config update failed", zap.Error(err))
+			} else {
+				p.checkPublicRpcEndpoints()
```
mdulin2, line 623: Should this just *log* an error or just it exit loudly?

node/pkg/common/url_verification.go
mdulin2: Is there any reason that we would WANT to log the path or query? If it's important enough, I could see a world where we only block/strip known patterns or use a regex to strip out specific portions of the path?

node/pkg/processor/processor.go
johnsaigle: If it exists here then I think it triggers a processor restart and crash loop. I think for the moment it's better to log and from there we can contract Guardians to figure out what's up

node/pkg/common/url_verification.go
johnsaigle: I'm not aware of a reason right now. I think we can go ahead with this change for now and do more advanced parsing later if we need it. It seems like a bit of a headache to try to think of every possible provider and analyze their APIs to satisfy a regex or identify known-bad patterns.

node/pkg/processor/processor.go
johnsaigle: What do you mean exactly? I think it's always used to format a string for a log or else format a string that goes into an error message that then gets logged.

PR #4831, opened 2026-05-21
ci: Improve performance of go lint and formatting

johnsaigle:
- Use `git ls-files` to retrieve check-in Go files for linting. This should be faster than using `find` to traverse all the subdirs and look up OS information.
- Allow parallel golangci-lint runs so node/ and sdk/ run at the same time

### DIFF
--- a/node/Makefile
+++ b/node/Makefile
@@ -1,6 +1,6 @@
 .PHONY: lint
-lint:
-	golangci-lint run -c ../.golangci.yml ./...
+lint: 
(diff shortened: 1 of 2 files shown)


djb15 (dismissed):

SEJeff (approved):
nice!

djb15 (approved):

johnsaigle:
Moving to draft mode for the moment so that we can merge in #4841 first. I'll come back to this afterward.

PR #4838, opened 2026-05-26
Aptos Watcher Tests + event type check

mdulin2:
The PR does the following: 
* Separates the core processing functionality for the Aptos watcher into separate functions. This is to make the code more testable. 
* Adds an event type verification to the message handling. This required a small change to the watcher configuration. This should be the ONLY functionality change to the watcher from the refactor.
* Rewrote the JSON modification framework in the test suite to be more modular.

### DIFF
--- a/node/pkg/watchers/aptos/config.go
+++ b/node/pkg/watchers/aptos/config.go
@@ -26,7 +26,7 @@ func (wc *WatcherConfig) GetChainID() vaa.ChainID {
 	return wc.ChainID
 }
 
-//nolint:unparam // Reobserver return is always nil; Aptos uses the legacy obsvReqC channel rather than the Reobserver interface.
+//nolint:unparam // error is always nil here but the return type is required to satisfy the interface.
 func (wc *WatcherConfig) Create(
 	msgC chan<- *common.MessagePublication,
 	obsvReqC <-chan *gossipv1.ObservationRequest,
--- a/node/pkg/watchers/aptos/config.go
+++ b/node/pkg/watchers/aptos/config.go
@@ -26,7 +26,7 @@ func (wc *WatcherConfig) GetChainID() vaa.ChainID {
 	return wc.ChainID
 }
 
-//nolint:unparam // error is always nil here but the return type is required to satisfy the interface.
+//nolint:unparam // Reobserver return is always nil; Aptos uses the legacy obsvReqC channel rather than the Reobserver interface.
 func (wc *WatcherConfig) Create(
 	msgC chan<- *common.MessagePublication,
 	obsvReqC <-chan *gossipv1.ObservationRequest,
@@ -35,5 +35,9 @@ func (wc *WatcherConfig) Create(
 	_ chan<- *common.GuardianSet,
 	_ common.Environment,
 ) (supervisor.Runnable, interfaces.Reobserver, error) {
-	return NewWatcher(wc.ChainID, wc.NetworkID, wc.Rpc, wc.Account, wc.Handle, msgC, obsvReqC).Run, nil, nil
+	w, err := NewWatcher(wc.ChainID, wc.NetworkID, wc.Rpc, wc.Account, wc.Handle, msgC, obsvReqC)
+	if err != nil {
+		return nil, nil, err
+	}
+	return w.Run, nil, nil
 }
--- a/node/pkg/watchers/aptos/watcher.go
+++ b/node/pkg/watchers/aptos/watcher.go
@@ -209,7 +209,7 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			events := gjson.ParseBytes(eventsJson)
 
-			if err := e.processPollingBatch(logger, events, &nextSequence); err != nil {
+			if err = e.processPollingBatch(logger, events, &nextSequence); err != nil {
 				return err
 			}
 
--- a/node/pkg/watchers/aptos/watcher.go
+++ b/node/pkg/watchers/aptos/watcher.go
@@ -30,11 +30,10 @@ type (
 		chainID   vaa.ChainID
 		networkID string
 
-		aptosRPC     string
-		aptosAccount string
-		aptosHandle  string
-		// aptosEventType is derived from aptosHandle.
-		aptosEventType string
+		aptosRPC       string
+		aptosAccount   string // Wormhole contract address
+		aptosHandle    string // Event to subscribe to Aptos RPC
+		aptosEventType string // aptosEventType is derived from aptosHandle.
 
 		msgC          chan<- *common.MessagePublication
 		obsvReqC      <-chan *gossipv1.ObservationRequest
@@ -83,10 +82,18 @@ func NewWatcher(
 	if !strings.HasSuffix(aptosHandle, "Handle") {
 		return nil, fmt.Errorf("aptosHandle %q does not end with 'Handle'", aptosHandle)
 	}
-	// Validate the configured contract address
-	if _, ok := parseAptosAddr(aptosAccount); !ok {
-		return nil, fmt.Errorf("aptosAccount %q is not a valid Aptos address", aptosAccount)
+	// Validate the handle's structure (<address>::<module>::<struct>) and its embedded address.
+	if err := validateAptosHandle(aptosHandle); err != nil {
+		return nil, err
 	}
+
+	// Validate the configured contract address. parseAptosAddr requires a valid, lowercase Aptos
+	// address so that verifyEventType can match it against the (always lowercased) event fields with
+	// an exact comparison. Fail fast on misconfiguration.
+	if _, ok := parseAptosAddr(aptosAccount); !ok {
+		return nil, fmt.Errorf("aptosAccount %q is not a valid lowercase Aptos address", aptosAccount)
+	}
+
 	return &Watcher{
 		chainID:        chainID,
 		networkID:      string(networkID),
@@ -187,7 +194,7 @@ func (e *Watcher) Run(ctx context.Context) error {
 
 			outcomes := gjson.ParseBytes(body)
 
-			e.processReobsBatch(logger, outcomes, nativeSeq)
+			e.processReobservationBatch(logger, outcomes, nativeSeq)
 
 		case <-timer.C:
 			s := ""
@@ -300,6 +307,7 @@ func (e *Watcher) processPollingBatch(logger *zap.Logger, events gjson.Result, n
 		// Consume this slot regardless of outcome so a skipped event is never re-fetched.
 		*nextSequence = eventSeq + 1
 
+		// SECURITY: Aptos event type validation
 		if err := e.verifyEventType(event); err != nil {
 			logger.Error("aptos event failed verification",
 				zap.Error(err),
@@ -332,7 +340,7 @@ func (e *Watcher) processPollingBatch(logger *zap.Logger, events gjson.Result, n
 
 // processReobsBatch handles the response to a reobservation lookup. The query
 // uses limit=1 so outcomes is expected to be a zero- or one-element array.
-func (e *Watcher) processReobsBatch(logger *zap.Logger, outcomes gjson.Result, nativeSeq uint64) {
+func (e *Watcher) processReobservationBatch(logger *zap.Logger, outcomes gjson.Result, nativeSeq uint64) {
 	for _, aptosEvent := range outcomes.Array() {
 		newSeq := aptosEvent.Get("sequence_number")
 		if !newSeq.Exists() {
@@ -344,6 +352,7 @@ func (e *Watcher) processReobsBatch(logger *zap.Logger, outcomes gjson.Result, n
 			break
 		}
 
+		// SECURITY: Aptos event type validation
 		if err := e.verifyEventType(aptosEvent); err != nil {
 			logger.Error("aptos event failed verification",
 				zap.Error(err),
@@ -369,28 +378,37 @@ func (e *Watcher) processReobsBatch(logger *zap.Logger, outcomes gjson.Result, n
 	}
 }
 
-// normalize0x strips an optional leading "0x" prefix so a configured value without it still
+// stripHexadecimalPrefix strips an optional leading "0x" prefix so a configured value without it still
 // matches the chain's 0x-prefixed representation.
-func normalize0x(s string) string {
+func stripHexadecimalPrefix(s string) string {
 	return strings.TrimPrefix(s, "0x")
 }
 
-// aptosAddrEqual reports whether two Aptos account addresses are equal. Each is decoded to its
-// canonical 32-byte form, so the "0x"/"0X" prefix, letter casing, and leading-zero padding are
-// all ignored. Returns false if either side is not a valid address.
-func aptosAddrEqual(actualAddr, expectedAddr string) bool {
-	actual, actualOK := parseAptosAddr(actualAddr)
-	expected, expectedOK := parseAptosAddr(expectedAddr)
-	return actualOK && expectedOK && actual == expected
+// validateAptosHandle checks that the event handle has the canonical
+// "<address>::<module>::<struct>" shape and that its address segment is a valid, lowercase Aptos
+// address. Splitting on "::" must yield exactly three segments; the explicit length check guards the
+// parts[0] access below from an out-of-bounds read.
+func validateAptosHandle(handle string) error {
+	parts := strings.Split(handle, "::")
+	if len(parts) != 3 {
+		return fmt.Errorf("aptosHandle %q must have the form <address>::<module>::<struct>", handle)
+	}
+	if _, ok := parseAptosAddr(parts[0]); !ok {
+		return fmt.Errorf("aptosHandle address segment %q is not a valid lowercase Aptos address", parts[0])
+	}
+	return nil
 }
 
-// parseAptosAddr decodes a hex Aptos address into its canonical 32-byte form.
+// parseAptosAddr decodes a hex Aptos address into its canonical 32-byte form. The address must be
+// lowercase with an optional lowercase "0x" prefix; uppercase hex (or a "0X" prefix) is rejected so
+// that a configured address compares equal to the always-lowercased values returned by the Aptos API.
 func parseAptosAddr(addr string) ([32]byte, bool) {
 	var out [32]byte
-	if len(addr) >= 2 && addr[0] == '0' && (addr[1] == 'x' || addr[1] == 'X') {
-		addr = addr[2:]
-	}
+	addr = stripHexadecimalPrefix(addr)
 	if len(addr) == 0 || len(addr) > 64 {
+		return out, false
+	}
+	if addr != strings.ToLower(addr) {
 		return out, false
 	}
 	decoded, err := hex.DecodeString(addr)
@@ -409,9 +427,9 @@ func (e *Watcher) verifyEventType(event gjson.Result) error {
 		return fmt.Errorf("event missing 'type' field")
 	}
 
-	// The type must match exactly, modulo the optional leading "0x" prefix. Casing is significant:
+	// The type must match exactly.
 	// Aptos returns the canonical type string, and Move module/struct names are case-sensitive.
-	if normalize0x(t.String()) != normalize0x(e.aptosEventType) {
+	if stripHexadecimalPrefix(t.String()) != stripHexadecimalPrefix(e.aptosEventType) {
 		return fmt.Errorf("event type mismatch: got %q, want %q", t.String(), e.aptosEventType)
 	}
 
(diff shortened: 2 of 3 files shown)


mdulin2:
I ran the reobservation flow on the production watcher for a while. It was picking up all of the events that were identifiable on Wormholesan.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +268,104 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+func (e *Watcher) processPollingBatch(logger *zap.Logger, events gjson.Result, nextSequence *uint64) error {
+	// the endpoint returns an array of events, ordered by sequence id (ASC)
+	for _, event := range events.Array() {
+		eventSequence := event.Get("sequence_number")
+		if !eventSequence.Exists() {
+			continue
+		}
+		eventSeq := eventSequence.Uint()
+
+		if err := e.verifyEventType(event); err != nil {
```
mdulin2, line 327: Only real change to functionality. 

Should this CONTINUE or throw an error that crashes the watcher. I'm fine with either but thought that taking down the watcher on bad event consumption may be a wise move.

mdulin2:
This broke the tilt CI. I'll figure out why and fix that.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -63,17 +65,21 @@ func NewWatcher(
...
-		networkID:     string(networkID),
-		aptosRPC:      aptosRPC,
-		aptosAccount:  aptosAccount,
-		aptosHandle:   aptosHandle,
-		msgC:          msgC,
-		obsvReqC:      obsvReqC,
-		readinessSync: common.MustConvertChainIdToReadinessSyncing(chainID),
+) (*Watcher, error) {
+	if !strings.HasSuffix(aptosHandle, "Handle") {
+		return nil, fmt.Errorf("aptosHandle %q does not end with 'Handle'", aptosHandle)
```
djb15, line 88: It would be nice if this is commented more explicitly that this only works because of the on-chain naming convention

node/pkg/watchers/aptos/watcher.go
```diff
@@ -163,23 +169,8 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-
-				}
-
-				data := chunk.Get("data")
-				if !data.Exists() {
-					break
-				}
-				e.observeData(logger, data, nativeSeq, true)
+			if err := e.processReobsBatch(logger, outcomes, nativeSeq); err != nil {
+				return err
```
djb15: Is returning the correct action here? I believe this will restart the watcher. I realise the defence in depth check should never fail, but if it does is it better to log loudly or crash loop the watcher?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -218,30 +209,8 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-				// find the next sequence that comes after the array
-				nextSequence = eventSeq + 1
-
-				data := event.Get("data")
-				if !data.Exists() {
-					continue
-				}
-				e.observeData(logger, data, eventSeq, false)
+			if err = e.processPollingBatch(logger, events, &nextSequence); err != nil {
+				return err
```
djb15: Same question here

node/pkg/watchers/aptos/watcher.go
djb15: Right, I didn't look down far enough to see this comment 😅 I'm leaning towards continue as the watcher will just restart automatically so we end up getting crash loops which seems more annoying

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +268,116 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		return fmt.Errorf("event type mismatch: got %q, want %q", t.String(), e.aptosEventType)
+	}
+
+	// The GUID identifies the on-chain EventHandle that emitted the event;
+	// account_address must equal the configured core bridge account.
+	guidAddr := event.Get("guid.account_address")
+	if !guidAddr.Exists() {
+		return fmt.Errorf("event missing 'guid.account_address' field")
+	}
+	if !strings.EqualFold(normalize0x(guidAddr.String()), normalize0x(e.aptosAccount)) {
```
djb15: Is the use of `EqualFold` intentional here?

node/pkg/watchers/aptos/watcher_test.go
```diff
@@ -219,3 +226,511 @@ func TestObserveDataFields(t *testing.T) {
...
+				"sequence_number": "171377",
+				"data": {
+					"sender": "1",
+					"sequence": "171377",
+					"nonce": "0",
+					"payload": "0xdeadbeef",
+					"consistency_level": "0",
+					"timestamp": "1700000000"
+				}
+			}`,
```
djb15: Can you build a helper function for building these json payloads? Should make the test suite a bit more concise

node/pkg/watchers/aptos/watcher.go
mdulin2: You can go either way imo. I decided to go the safer route and just ended it. If you think a loud log is better, I can do that instead, though.

node/pkg/watchers/aptos/watcher.go
mdulin2: Response above.

node/pkg/watchers/aptos/watcher.go
mdulin2: Yes. The addresses are case-insensitive, so this should work fine. Obviously, we want this to work at startup so I added this.

node/pkg/watchers/aptos/watcher.go
mdulin2: Yeah, fair point. I don't see us changing this in the future but I can add a comment to it.

node/pkg/watchers/aptos/watcher.go
mdulin2: Cool. I changed the code to just log the error and `continue` instead of throwing an error.

node/pkg/watchers/aptos/watcher_test.go
mdulin2: Ahh, you're right. I rewrote one section of the tests to do that but never went back to this one. Good call out.

mdulin2:
```
2 files changed, 109 insertions(+), 189 deletions(-)
```

node/pkg/watchers/aptos/watcher.go
mdulin2: Errors and continues now. I removed the error from the `process*` functions because this was the only error being returned from it anyway.

node/pkg/watchers/aptos/watcher.go
mdulin2: See comment above that was added.

node/pkg/watchers/aptos/watcher_test.go
mdulin2: Change made. It now uses a mutation function to modify the JSON in the way that is needed for the test.

node/pkg/watchers/aptos/watcher.go
mdulin2: I removed the error handling here and just logged the verification issue.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -66,6 +66,20 @@ func NewWatcher(
...
+		The Aptos smart contracts have two items to consider:
+		- WormholeMessageHandle - searchable event
+		- WormholeMessage - decoded event data
+
+		During event validation, the code checks that the event is a 'state::WormholeMessage' type.
+		This check prevents bugs where the RPC returns an event of the wrong type to the watcher.
+		To do this, the text 'Handle' is removed from the `aptosHandle` to derive
+		the name of the event. This is done in order to avoid passing another parameter to the watcher.
+
+		This is only be convention this optimization can be done. If the Aptos smart contracts or
```
djb15: Not sure exactly what you're trying to say in this sentence? Thanks for adding though

djb15 (dismissed):
Minor comment on the code comment, but otherwise this looks good

node/pkg/watchers/aptos/watcher.go
mdulin2: I made a change to this in the latest PR. I just do a bytewise comparison on the address instead of dealing with strings. That should make it more consistent.

node/pkg/watchers/aptos/watcher.go
mdulin2: I updated the comment to remove the typo. Any help here?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +282,152 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+	actual, actualOK := parseAptosAddr(actualAddr)
+	expected, expectedOK := parseAptosAddr(expectedAddr)
+	return actualOK && expectedOK && actual == expected
+}
+
+// parseAptosAddr decodes a hex Aptos address into its canonical 32-byte form.
+func parseAptosAddr(addr string) ([32]byte, bool) {
+	var out [32]byte
+	if len(addr) >= 2 && addr[0] == '0' && (addr[1] == 'x' || addr[1] == 'X') {
+		addr = addr[2:]
```
pleasew8t: This might read slightly better:

```suggestion
	if strings.HasPrefix(addr, "0x") || strings.HasPrefix(addr, "0X") {
		addr = addr[2:]
```

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +282,152 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		}
+
+		// Validates and publishes the message to the processor
+		e.observeData(logger, data, eventSeq, false)
+	}
+}
+
+// processReobsBatch handles the response to a reobservation lookup. The query
+// uses limit=1 so outcomes is expected to be a zero- or one-element array.
+func (e *Watcher) processReobsBatch(logger *zap.Logger, outcomes gjson.Result, nativeSeq uint64) {
```
pleasew8t:
```suggestion
func (e *Watcher) processReobservationBatch(logger *zap.Logger, outcomes gjson.Result, nativeSeq uint64) {
```

node/pkg/watchers/aptos/watcher.go
pleasew8t: I actually wonder if we don't just do:

```go
addr = strings.TrimPrefix(addr, "0x")
addr = strings.TrimPrefix(addr, "0X")
```

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +282,152 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		}
+
+		// Validates and publishes the message to the processor
+		e.observeData(logger, data, nativeSeq, true)
+	}
+}
+
+// normalize0x strips an optional leading "0x" prefix so a configured value without it still
+// matches the chain's 0x-prefixed representation.
+func normalize0x(s string) string {
```
pleasew8t: Is there a reason `parseAptosAddr` cares about lower- and uppercase `x`, but `normalize0x` does not?

pleasew8t: One suggestion is to make this function strip both (call it, `stripHexadecimalPrefix()`), and reuse in `aptosAddrEqual`.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +282,152 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+		return fmt.Errorf("event type mismatch: got %q, want %q", t.String(), e.aptosEventType)
+	}
+
+	// The GUID identifies the on-chain EventHandle that emitted the event;
+	// account_address must equal the configured core bridge account.
+	guidAddr := event.Get("guid.account_address")
+	if !guidAddr.Exists() {
+		return fmt.Errorf("event missing 'guid.account_address' field")
+	}
+	if !aptosAddrEqual(guidAddr.String(), e.aptosAccount) {
```
pleasew8t:
```suggestion
	}

	if !aptosAddrEqual(guidAddr.String(), e.aptosAccount) {
```

pleasew8t (commented):
First pass looks good. Left a few suggestions.

djb15 (dismissed):
Happy from my side. Will review any changes requested by the other reviewers

node/pkg/watchers/aptos/watcher.go
mdulin2: Yeah, good shout out. That will make this simpler. I'll trim the prefix.

node/pkg/watchers/aptos/watcher.go
mdulin2: No reason. I rewrote this to that the `0x` normalization only allows `0x`.

mdulin2:
@pleasew8t I made some changes based on your feedback. Feel free to review it again.

djb15 (dismissed):
Changes in the latest commit look good

bemic (commented):
I went over the code and my one comment would be regarding the 0x/0X, but it is already covered in Jason's comment. 
The second comment, apparently Aptos addresses with stripped zeros can be used. Example from Aptos docs is `0x1`. However I don't really know how rare it is.

mdulin2:
@bemic I saw this too but didn't decide to dig further. It appears that the `type` and the `guid` get truncated. For instance, check this out: [7730cd28ee1cdc9e999336cbc430f99e7c44397c0aa77516f6f23a78559bb5](https://fullnode.mainnet.aptoslabs.com/v1/accounts/0x007730cd28ee1cdc9e999336cbc430f99e7c44397c0aa77516f6f23a78559bb5/resources)

The type is returned as `0x7730cd28ee1cdc9e999336cbc430f99e7c44397c0aa77516f6f23a78559bb5::pool::ThalaSwap` even though the real type would have the zeros. I'll fix the code to trim leading zeros on the comparison.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -411,6 +425,11 @@ func parseAptosAddr(addr string) ([32]byte, bool) {
 	if addr != strings.ToLower(addr) {
 		return out, false
 	}
+	// The Aptos API returns special addresses in short form on a 4-bit nibble boundary
```
johnsaigle, line 428: Is the `4-bit` part relevant here? Should we do extra validation to make sure that the special addresses are multiple of this length?

node/pkg/watchers/aptos/watcher.go
```diff
@@ -427,10 +461,11 @@ func (e *Watcher) verifyEventType(event gjson.Result) error {
...
 	}
 
-	// The type must match exactly.
-	// Aptos returns the canonical type string, and Move module/struct names are case-sensitive.
-	if stripHexadecimalPrefix(t.String()) != stripHexadecimalPrefix(e.aptosEventType) {
-		return fmt.Errorf("event type mismatch: got %q, want %q", t.String(), e.aptosEventType)
+	// Compare the address segment by canonical bytes, since the API may strip leading zeros.
+	// Move module/struct names are case-sensitive and must match exactly.
+	addr, module, structName, ok := parseAptosEventType(t.String())
+	if !ok || addr != e.eventTypeAddr || module != e.eventTypeModule || structName != e.eventTypeStruct {
```
johnsaigle, line 467: suggestion: Comparisons could be done here with `bytes.Compare` rather than `!=`. This would be a bit more direct and avoid some of the specificity around whether the strings are lower-case or not

node/pkg/watchers/aptos/watcher.go
```diff
@@ -88,22 +93,33 @@ func NewWatcher(
 	}
 
 	// Validate the configured contract address. parseAptosAddr requires a valid, lowercase Aptos
-	// address so that verifyEventType can match it against the (always lowercased) event fields with
-	// an exact comparison. Fail fast on misconfiguration.
-	if _, ok := parseAptosAddr(aptosAccount); !ok {
+	// address so that verifyEventType can match it against the (always lowercased) event fields.
+	// Fail fast on misconfiguration. The canonical bytes are cached for later comparison.
+	accountAddr, ok := parseAptosAddr(aptosAccount)
```
johnsaigle, line 98: (Just sharing info, no need to change code)
In the future IMO using `error` as a return value is more natural here than `ok`. Typically `ok` (bool return type) gets used for lookups in maps. In that case boolean makes a bit more sense conceptually over error, because a key missing in a map isn't an error, it's just not there.

In the function and the parse event function though, it's arguably a problem if these pass validation: it means the RPC gave the wrong data or the watcher has a parsing problem.

Given that this function can fail for a few reasons (wrong length, not lowercase, ...), `error` allows you to specify _what_ went wrong via returning a new error with a specific string (or even a custom error type if you get fancy). That also has benefits for testing, fuzzing and so on down the line.

node/pkg/watchers/aptos/watcher_test.go
```diff
@@ -345,6 +351,43 @@ func TestVerifyEventType(t *testing.T) {
 	}
 }
 
+// TestVerifyEventTypeShortFormAddress confirms the byte comparison treats the Aptos API's
+// short-form addresses (leading zeros stripped) as equal to the zero-padded configured form.
+// The watcher is configured with the full 64-char address; the event uses the stripped form.
+func TestVerifyEventTypeShortFormAddress(t *testing.T) {
```
johnsaigle, line 357: worth adding a few negative tests here? I'm unsure what happens if the input is neither size 1 nor 65. What happens for other odd-length values, or even-length values that are in the range of 0-64? As someone less familiar with Aptos, I'm not sure what the watcher should do if the RPC returns these for whatever reason.

node/pkg/watchers/aptos/watcher.go
mdulin2: I don't think we need extra validation. What I mean is that it'll remove all leading zeros. For instance, `0x0022` would be normalized to `0x22`, and `0x0122` to `0x122` .

node/pkg/watchers/aptos/watcher.go
mdulin2: There are three checks here: 
- `addr != e.eventTypeAddr` is a [32]byte already.
- `module != e.eventTypeModule` needs to be case sensitive.
- `structName != e.eventTypeStruct` needs to be case sensitive.

Given this, I don't think anything needs to change here. Thanks for the feedback on that.

node/pkg/watchers/aptos/watcher.go
mdulin2: Hmmm, yeah. I'll try to use that pattern in the future for the error reason. Thanks for pointing out.

node/pkg/watchers/aptos/watcher_test.go
mdulin2: Sure! I added a few more tests for this parser.

node/pkg/watchers/aptos/watcher.go
johnsaigle: What I mean is that if you do a byte-by-byte comparison with `bytes.Compare`, that captures the length as well as the idea of casing, because you you're checking the bytes directly.

node/pkg/watchers/aptos/config.go
```diff
@@ -26,7 +26,7 @@ func (wc *WatcherConfig) GetChainID() vaa.ChainID {
 	return wc.ChainID
 }
 
-//nolint:unparam // error is always nil here but the return type is required to satisfy the interface.
+//nolint:unparam // Reobserver return is always nil; Aptos uses the legacy obsvReqC channel rather than the Reobserver interface.
```
johnsaigle: This comment refers to the [WatcherConfig interface](https://github.com/wormhole-foundation/wormhole/blob/5406b6a35f06d481931ad887c9e95afa98154dad/node/pkg/watchers/watchers.go#L21
): this is a method receiver for that type. I think the wording above adds some confusion compared to the original comment. 

If you want to modify it maybe we could say "... is required to satisfy the [WatcherConfig] interface."

mdulin2: Reverted comment.

node/pkg/watchers/aptos/watcher.go
```diff
@@ -299,6 +305,183 @@ func (e *Watcher) retrievePayload(s string) ([]byte, error) {
...
+			continue
+		}
+
+		// Validates and publishes the message to the processor
+		e.observeData(logger, data, eventSeq, false)
+	}
+}
+
+// processReobsBatch handles the response to a reobservation lookup. The query
+// uses limit=1 so outcomes is expected to be a zero- or one-element array.
```
johnsaigle, line 357: Worth returning early and/or logging if outcomes has a length that is greater than 1? I can see that the length == 0 case would just cause the loop to not execute.

johnsaigle (approved):

bemic (approved):

douglasgalico (approved):

PR #4839, opened 2026-05-26
node: Accountant audit improvements

djb15:
This PR includes the following changes to the Accountant:

- Increases the max batch size of observation submissions from 10 to 100. This is close to the maximum calldata that the contract can support. To support this, there is an increase in the gas limit to 5m.
- Increase the batch time from 100ms to 2s. Block time on Wormchain is ~6 seconds so there is little benefit to batching earlier, especially considering observation submission wait for block inclusion.
- The Accountant audit flow has been modified to instead call `query_all_pending_transfers` (rather than `query_missing_observations`), as the former supports pagination. This guarantees that the audit flow will always run and there is no risk of exceeding query gas limits.
- New tests are added to verify the changes perform as expected. The changes have also been tested on devnet and appear to work as expected.
- Submitting to the Accountant `subChan` now has a blocking variant. When submitting to the channel from the audit flow we will happily wait for room to open up in the channel, vs from the processor we continue to try and move on if there is a failure. 
- Manual reobservation requests will trigger a retry on the Accountant if it is stuck in the pending state of an audit. This ensures that specific transactions can be posted to the contract despite being in an audit queue.

Other avenues explored:

- I explored increase the number of messages per transaction to increase the effective batch size about 100. However submissions then use too much gas and we run into block gas constraints. With the limit of 5m per batch we could still have 95m gas per block (considering 19 guardians all submitting at the same time). We can't increase this any further.
- I explored moving away from waiting for block inclusion to submitting transactions asynchronously. This could allow more transactions to land in each block from each guardian. However we still run into the same block gas constraints and the logic becomes more complicated as we need to track account sequence numbers manually (they are only incremented after inclusion of a transaction in a block) and retries are more challenging.
- I explored modifying the accountant smart contract do add pagination for `query_missing_observations`, however this would require a smart contract upgrade. I elected to instead stick to making only Guardian changes for simplicity and ease of release.

### DIFF
--- a/node/cmd/guardiand/node.go
+++ b/node/cmd/guardiand/node.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/certusone/wormhole/node/pkg/accountant"
 	"github.com/certusone/wormhole/node/pkg/guardiansigner"
 	"github.com/certusone/wormhole/node/pkg/watchers"
 	"github.com/certusone/wormhole/node/pkg/watchers/ibc"
@@ -129,11 +130,12 @@ var (
 	ibcBlockHeightURL *string
 	ibcContract       *string
 
-	accountantContract      *string
-	accountantWS            *string
-	accountantCheckEnabled  *bool
-	accountantKeyPath       *string
-	accountantKeyPassPhrase *string
+	accountantContract                   *string
+	accountantWS                         *string
+	accountantCheckEnabled               *bool
+	accountantKeyPath                    *string
+	accountantKeyPassPhrase              *string
+	accountantSubmitObservationBatchSize *int
 
 	accountantNttContract      *string
 	accountantNttKeyPath       *string
@@ -404,6 +406,7 @@ func init() {
 	accountantKeyPath = NodeCmd.Flags().String("accountantKeyPath", "", "path to accountant private key for signing transactions")
 	accountantKeyPassPhrase = NodeCmd.Flags().String("accountantKeyPassPhrase", "", "pass phrase used to unarmor the accountant key file")
 	accountantCheckEnabled = NodeCmd.Flags().Bool("accountantCheckEnabled", false, "Should accountant be enforced on transfers")
+	accountantSubmitObservationBatchSize = NodeCmd.Flags().Int("accountantSubmitObservationBatchSize", accountant.DefaultSubmitObservationBatchSize, "maximum number of observations to submit to the accountant contract in one transaction")
 
 	accountantNttContract = NodeCmd.Flags().String("accountantNttContract", "", "Address of the NTT accountant smart contract on wormchain")
 	accountantNttKeyPath = NodeCmd.Flags().String("accountantNttKeyPath", "", "path to NTT accountant private key for signing transactions")
@@ -2055,7 +2058,7 @@ func runNode(cmd *cobra.Command, args []string) {
 	guardianOptions := []*node.GuardianOption{
 		node.GuardianOptionDatabase(db),
 		node.GuardianOptionWatchers(watcherConfigs, ibcWatcherConfig),
-		node.GuardianOptionAccountant(*accountantWS, *accountantContract, *accountantCheckEnabled, accountantWormchainConn, *accountantNttContract, accountantNttWormchainConn),
+		node.GuardianOptionAccountant(*accountantWS, *accountantContract, *accountantCheckEnabled, accountantWormchainConn, *accountantNttContract, accountantNttWormchainConn, *accountantSubmitObservationBatchSize),
 		node.GuardianOptionGovernor(*chainGovernorEnabled, *governorFlowCancelEnabled, *coinGeckoApiKey),
 		node.GuardianOptionNotary(*notaryEnabled),
 		node.GuardianOptionManagerService(*managerServiceEnabled, managerSigners, *ethRPC),
--- a/node/pkg/accountant/accountant.go
+++ b/node/pkg/accountant/accountant.go
@@ -490,13 +490,13 @@ func (acct *Accountant) submitToChannel(ctx context.Context, pe *pendingEntry, s
 			pe.state.submitPending = false
 			pe.stateLock.Unlock()
 		case <-ctx.Done():
-			acct.logger.Debug(fmt.Sprintf("context cancelled while submitting to %s channel", tag), zap.String("msgId", pe.msgId))
+			acct.logger.Warn(fmt.Sprintf("context cancelled while submitting to %s channel", tag), zap.String("msgId", pe.msgId))
 			pe.stateLock.Lock()
 			pe.state.submitPending = false
 			pe.stateLock.Unlock()
 		}
 	} else {
-		// Non-blocking write (existing behavior for processor)
+		// Non-blocking write
 		select {
 		case subChan <- pe.msg:
 			acct.logger.Debug(fmt.Sprintf("submitted observation to channel for %s", tag), zap.String("msgId", pe.msgId))
--- a/node/pkg/accountant/accountant.go
+++ b/node/pkg/accountant/accountant.go
@@ -30,7 +30,7 @@ import (
 
 // MsgChannelCapacity specifies the capacity of the message channel used to publish messages released from the accountant.
 // This channel should not back up, but if it does, the accountant will start dropping messages, which would require reobservations.
-const MsgChannelCapacity = 5 * batchSize
+const MsgChannelCapacity = 5 * DefaultSubmitObservationBatchSize
 
 type (
 	AccountantWormchainConn interface {
@@ -75,23 +75,24 @@ type (
 
 // Accountant is the object that manages the interface to the wormchain accountant smart contract.
 type Accountant struct {
-	ctx                  context.Context
-	logger               *zap.Logger
-	db                   guardianDB.AccountantDB
-	obsvReqWriteC        chan<- *gossipv1.ObservationRequest
-	contract             string
-	wsUrl                string
-	wormchainConn        AccountantWormchainConn
-	enforceFlag          bool
-	guardianSigner       guardiansigner.GuardianSigner
-	gst                  *common.GuardianSetState
-	guardianAddr         ethCommon.Address
-	msgChan              chan<- *common.MessagePublication
-	tokenBridges         validEmitters
-	pendingTransfersLock sync.Mutex
-	pendingTransfers     map[string]*pendingEntry // Key is the message ID (emitterChain/emitterAddr/seqNo)
-	subChan              chan *common.MessagePublication
-	env                  common.Environment
+	ctx                        context.Context
+	logger                     *zap.Logger
+	db                         guardianDB.AccountantDB
+	obsvReqWriteC              chan<- *gossipv1.ObservationRequest
+	contract                   string
+	wsUrl                      string
+	wormchainConn              AccountantWormchainConn
+	enforceFlag                bool
+	guardianSigner             guardiansigner.GuardianSigner
+	gst                        *common.GuardianSetState
+	guardianAddr               ethCommon.Address
+	msgChan                    chan<- *common.MessagePublication
+	tokenBridges               validEmitters
+	pendingTransfersLock       sync.Mutex
+	pendingTransfers           map[string]*pendingEntry // Key is the message ID (emitterChain/emitterAddr/seqNo)
+	subChan                    chan *common.MessagePublication
+	submitObservationBatchSize int
+	env                        common.Environment
 
 	nttContract       string
 	nttWormchainConn  AccountantWormchainConn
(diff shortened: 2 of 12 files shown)


djb15:
@claude

node/pkg/accountant/audit.go
```diff
@@ -206,102 +248,132 @@ func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchai
...
+	// observed and verified through the normal message processing pipeline. Transfers returned by
+	// the contract's all_pending_transfers query are untrusted external data. We must ONLY resubmit
+	// observations for transfers present in knownPendingTransferMap. For transfers the contract
+	// reports that we do NOT have locally, we request a reobservation from the watcher, which
+	// re-verifies the transaction on-chain before it enters the signing pipeline. Never construct
+	// a signable observation directly from contract response data.
+	for _, pt := range pendingTransfers {
+		for _, data := range pt.Data {
+			// Skip if we've already signed this
+			if hasGuardianSigned(data.Signatures, guardianIndex) {
```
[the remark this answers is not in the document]
djb15: This check was deliberately removed, I have removed the comment

node/pkg/accountant/audit.go
```diff
@@ -323,37 +395,85 @@ func (acct *Accountant) handleMissingObservation(mo MissingObservation) {
...
+
+		// Set cursor for next page
+		lastKey := page[len(page)-1].Key
+		startAfter = &lastKey
 	}
 
-	guardianIndex, found := gs.KeyIndex(acct.guardianAddr)
-	if !found {
-		return nil, fmt.Errorf("failed to get guardian index")
+	return allPending, nil
```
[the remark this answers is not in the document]
djb15: Happy to leave this as is, it's deliberate

node/pkg/accountant/audit.go
```diff
@@ -95,22 +97,48 @@ type (
...
 // makeAuditKey creates an audit map key from a pending observation entry.
 func (pe *pendingEntry) makeAuditKey() string {
 	return fmt.Sprintf("%d-%s", pe.msg.EmitterChain, strings.TrimPrefix(pe.msg.TxIDString(), "0x"))
 }
 
+// hasGuardianSigned checks if a guardian has signed based on the signatures bitmask.
+// The signatures field is a u128 represented as a decimal string where each bit
+// corresponds to a guardian index.
+func hasGuardianSigned(signatures string, guardianIndex int) bool {
+	sigInt, ok := new(big.Int).SetString(signatures, 10) //nolint:mnd // Base 10 because the signatures are encoded as a decimal string
```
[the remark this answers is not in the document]
djb15: Not going to resolve this one

node/pkg/accountant/audit.go
```diff
@@ -206,102 +247,132 @@ func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchai
...
-					auditErrors.Inc()
-					acct.logger.Error("contract reported pending observation as missing, resubmitted it", zap.String("msgID", pe.msgId))
-				} else {
-					acct.logger.Info("contract reported pending observation as missing but it is queued up to be submitted, skipping it", zap.String("msgID", pe.msgId))
-				}
+	acct.logger.Info("audit queried pending transfers",
+		zap.Int("totalPending", len(pendingTransfers)),
+		zap.String("contract", contract))
+
+	// SECURITY: knownPendingTransferMap contains only transfers that this guardian has personally
```
johnsaigle: I think these sentences can be reworded slightly given that the Guardian may be a canonical Guardian. In that case it's not really correct to say that it 'has personally observed and verified'

johnsaigle (dismissed):
One small wording suggestion, feel free to ignore.

johnsaigle (dismissed):

node/pkg/accountant/audit.go
```diff
@@ -121,21 +148,21 @@ func (acct *Accountant) audit(ctx context.Context) error {
...
+func (acct *Accountant) runAudit(ctx context.Context) {
+	knownPendingTransferMap := acct.createAuditMap(false)
+	acct.logger.Debug("in AuditPendingTransfers: starting base audit", zap.Int("numPending", numPendingEntries(knownPendingTransferMap)))
+	acct.performAudit(ctx, knownPendingTransferMap, acct.wormchainConn, acct.contract)
 	acct.logger.Debug("in AuditPendingTransfers: finished base audit")
 
-	tmpMap = acct.createAuditMap(true)
-	acct.logger.Debug("in AuditPendingTransfers: starting ntt audit", zap.Int("numPending", numPendingEntries(tmpMap)))
-	acct.performAudit(tmpMap, acct.nttWormchainConn, acct.nttContract)
+	knownPendingTransferMap = acct.createAuditMap(true)
```
pleasew8t: nit: rename to `knownPendingNttTransferMap` (or similar)

node/pkg/accountant/audit.go
```diff
@@ -206,102 +247,131 @@ func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchai
...
+				continue
 			}
 
-			delete(tmpMap, key)
-		} else {
-			acct.handleMissingObservation(mo)
-		}
-	}
+			// We haven't signed - build key to check our local map
+			key := fmt.Sprintf("%d-%s", data.EmitterChain,
```
pleasew8t, line 272: Could this key construction be moved to its own function? Seeing as `pendingEntry` also has a `makeAuditKey` function, we could point that function and this code to a single function that produces a key.

node/pkg/accountant/accountant.go
```diff
@@ -436,37 +444,68 @@ func (acct *Accountant) loadPendingTransfers() error {
...
+		case <-time.After(timeout):
+			channelSubmitTimeouts.Inc()
+			acct.logger.Warn(fmt.Sprintf("timeout submitting observation to %s channel, will retry next audit", tag),
+				zap.String("msgId", pe.msgId),
+				zap.Duration("timeout", timeout))
+			pe.stateLock.Lock()
+			pe.state.submitPending = false
+			pe.stateLock.Unlock()
+		case <-ctx.Done():
+			acct.logger.Debug(fmt.Sprintf("context cancelled while submitting to %s channel", tag), zap.String("msgId", pe.msgId))
```
pleasew8t: Since the accountant's `performAudit()` doesn't really care about the result of this function, does it make sense to promote this to a warning or error? Since this could be one of the reasons a submission fails, we would want to know about it for investigating later on.

node/pkg/accountant/audit.go
```diff
@@ -206,102 +247,131 @@ func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchai
...
+				})
 			}
-			return
 		}
+	}
 
-		for _, pe := range pendingTransfers {
-			// There should be no nil entries, but we'll skip to be safe
+	if len(knownPendingTransferMap) == 0 {
+		acct.logger.Debug("exiting performAudit")
```
pleasew8t:
```suggestion
		acct.logger.Debug("exiting performAudit with no known pending transfers left")
```

pleasew8t (dismissed):

node/pkg/accountant/accountant.go
```diff
@@ -436,37 +444,68 @@ func (acct *Accountant) loadPendingTransfers() error {
...
-		pe.state.submitPending = false
+// submitToChannel submits an observation to the specified channel. If blocking is false and the channel is full,
+// it marks the transfer as no longer pending so it will be resubmitted by the audit. If blocking is true, it will
+// block until the channel has space, a timeout occurs, or the context is cancelled.
+func (acct *Accountant) submitToChannel(ctx context.Context, pe *pendingEntry, subChan chan *common.MessagePublication, tag string, blocking bool, timeout time.Duration) {
+	if blocking {
+		select {
+		case subChan <- pe.msg:
+			acct.logger.Debug(fmt.Sprintf("submitted observation to channel for %s", tag), zap.String("msgId", pe.msgId))
+		case <-time.After(timeout):
```
mdulin2, line 495: If the timeout is zero (aka the non-blocking path), I think you should just send it regularly without the timeout. This is because if two parts of a `select` are fired simultaneously, it will choose one of them at random; this will lead to A LOT (50% ish) of failed sends that you'd expect to succeed. https://medium.com/@pedram.esmaeeli/a-pattern-for-overcoming-non-determinism-of-golang-select-statement-139dbe93db98

mdulin2: Here's a test that you can use to demonstrate this is true:
```golang

package main

import (
	"fmt"
	"time"
)

func main() {
	sent := 0
	timedOut := 0

	for i := 0; i < 10000; i++ {
		ch := make(chan string, 1)

		select {
		case ch <- "hello":
			sent++
		case <-time.After(0):
			timedOut++
		}
	}

	fmt.Printf("sent:     %d\n", sent)
	fmt.Printf("timeout:  %d\n", timedOut)
}

```
Output is non-deterministic too: 

```
sent:     4950
timeout:  5050
```

node/pkg/accountant/accountant.go
```diff
@@ -436,37 +444,68 @@ func (acct *Accountant) loadPendingTransfers() error {
...
 // marked as "submit pending", this function returns false without doing anything. Otherwise it returns true. The return value can
-// be used to avoid unnecessary error logging. If writing to the channel would block, this function returns without doing anything,
-// assuming the pending transfer will be handled on the next audit interval. This function grabs the state lock.
-func (acct *Accountant) submitObservation(pe *pendingEntry) bool {
+// be used to avoid unnecessary error logging. If blocking is false and writing to the channel would block, this function returns
+// without doing anything, assuming the pending transfer will be handled on the next audit interval. If blocking is true, it will
+// block until the channel has space, a timeout occurs, or the context is cancelled. This function grabs the state lock.
+func (acct *Accountant) submitObservation(ctx context.Context, pe *pendingEntry, blocking bool) bool {
 	pe.stateLock.Lock()
-	defer pe.stateLock.Unlock()
```
mdulin2, line 443: Nice optimization! Makes it faster, as long as we don't deadlock it.

node/pkg/accountant/accountant.go
```diff
@@ -436,37 +444,68 @@ func (acct *Accountant) loadPendingTransfers() error {
...
+			pe.state.submitPending = false
+			pe.stateLock.Unlock()
+		case <-ctx.Done():
+			acct.logger.Debug(fmt.Sprintf("context cancelled while submitting to %s channel", tag), zap.String("msgId", pe.msgId))
+			pe.stateLock.Lock()
+			pe.state.submitPending = false
+			pe.stateLock.Unlock()
+		}
+	} else {
+		// Non-blocking write (existing behavior for processor)
```
mdulin2: Weird comment that I don't see the purpose of. Seems like something that Claude would write based on discussion with it.

node/pkg/accountant/accountant.go
mdulin2: Upon checking usage of this function, the `blocking=true` and `duration=0` doesn't seem to happen. So, probably not a huge deal then.

node/pkg/accountant/audit.go
mdulin2, line 248: Is the purpose of this loop just to loop errors if we don't know the status? It seems weird that we're iterating through these and just logging here.

node/pkg/accountant/audit.go
```diff
@@ -323,37 +393,85 @@ func (acct *Accountant) handleMissingObservation(mo MissingObservation) {
...
+	ctx context.Context,
+	logger *zap.Logger,
+	qc queryConn,
+	contract string,
+	pageSize int,
+) ([]PendingTransfer, error) {
+	var allPending []PendingTransfer
+	var startAfter *TransferKey
+
+	for {
```
mdulin2, line 425: How large is this right now? I imagine there are lots of pending transfers from over the years that never got processed?

node/pkg/accountant/audit.go
```diff
@@ -206,102 +247,131 @@ func (acct *Accountant) performAudit(tmpMap map[string][]*pendingEntry, wormchai
...
-					auditErrors.Inc()
-					acct.logger.Error("contract reported pending observation as missing, resubmitted it", zap.String("msgID", pe.msgId))
-				} else {
-					acct.logger.Info("contract reported pending observation as missing but it is queued up to be submitted, skipping it", zap.String("msgID", pe.msgId))
-				}
+	acct.logger.Info("audit queried pending transfers",
+		zap.Int("totalPending", len(pendingTransfers)),
+		zap.String("contract", contract))
+
+	// SECURITY: knownPendingTransferMap contains only transfers that are verified through the normal message processing pipeline. Transfers returned by
```
mdulin2, line 258: Good comment. Injecting VAAs at this stage could be catastrophic.

node/pkg/accountant/submit_obs.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	batchSize    = 10
-	batchTimeout = 100 * time.Millisecond
+	batchSize    = 100             // Observations per batch (limited by wasm contract input size of 64KB)
```
bemic: I think we can potentially hit 64kb limit with this batch size. We have custom payloads.
But I don't think it is a blocker. It would require most of the observation to have large payload.

node/pkg/accountant/audit.go
```diff
@@ -323,37 +393,85 @@ func (acct *Accountant) handleMissingObservation(mo MissingObservation) {
...
+
+	for {
+		page, err := queryAllPendingTransfersPage(ctx, logger, qc, contract, startAfter, pageSize)
+		if err != nil {
+			return nil, err
+		}
+
+		allPending = append(allPending, page...)
+
+		if len(page) < pageSize {
```
bemic, line 433: Most likely non-blocker comment.
I wonder if this can be a problem causing non-ending loop. Do we want any max cap here?

node/pkg/accountant/accountant.go
djb15: Good to know! But yeah that's not a realistic use case for this, the timeout will always be non-zero so I will leave this as is

node/pkg/accountant/submit_obs.go
djb15: Thanks Lukas, this is a good callout. I've double checked and I agree I don't think this is likely but it's feasibly possible. It's also possible now with the batch size of 10, but of course this change makes it 10x easier.

Do you think it's worth adding a command line flag to let the guardians change the batch size, but default to 100? In the case 100 is too large and we're having issues, it's a simple command line arg change vs a whole new deployment?

node/pkg/accountant/audit.go
djb15: Yep, it will list out all the message ids we're having problems with querying that may be helpful for debugging purposes

node/pkg/accountant/audit.go
djb15: I'm going to leave this as is, I thought about doing that before but because we're operating on different data structures it's not as simple as just calling the existing function. And because we only do this manually once I was ok with leaving it inline

node/pkg/accountant/audit.go
djb15: As far as I'm aware the accountant part of the dashboard is the authoritative source of all the messages stuck in the accountant; if I'm correct then of course this isn't very many and trends towards 0 when chain connections are working properly

node/pkg/accountant/submit_obs.go
bemic: Yes I think it's a good idea to give guardians the ability to change the number. It can potentially save us some troubles in case we have a set of observations with big payloads and it is repeatedly hitting this limit.

node/pkg/accountant/audit.go
djb15: Each page is of size 500, so we'd need probably tens to hundreds of thousands of pending transfers for this to be an issue which I think is practically infeasible

mdulin2 (approved):

bemic (approved):

node/cmd/guardiand/node.go
```diff
@@ -129,11 +130,12 @@ var (
...
-	accountantWS            *string
-	accountantCheckEnabled  *bool
-	accountantKeyPath       *string
-	accountantKeyPassPhrase *string
+	accountantContract                   *string
+	accountantWS                         *string
+	accountantCheckEnabled               *bool
+	accountantKeyPath                    *string
+	accountantKeyPassPhrase              *string
+	accountantSubmitObservationBatchSize *int
```
johnsaigle, line 138: nit: could probably be a smaller unsigned type value. We probably don't want to support the full int range. The checks below for `<= 0` on a value called "size" also feels off; sizes shouldn't ever be negative.

node/pkg/accountant/accountant_test.go
```diff
@@ -102,10 +102,19 @@ type AuditMockWormchainConn struct {
 	allPendingTransfersErr  error
 	batchTransferStatusResp []byte
 	batchTransferStatusErr  error
+
+	queryLock         sync.Mutex
```
johnsaigle, line 106: In the future, consider using [RWMutex](https://pkg.go.dev/sync#RWMutex) instead, as it allows you to do multiple reads concurrently. So when you query for length below, you could have any number of independent readers safely do that. The regular Mutex is all-or-nothing.

Anyway it doesn't really matter for that test, but generally there's usually a good argument for using the RWMutex instead.

johnsaigle (approved):

PR #4841, opened 2026-05-27
Custom ci lints

mdulin2:
The PR is introducing the capability for custom linters in Wormhole: 
* Adds a new folder on the top level named `linters`
* Adds the `channelcheck` linter but doesn't enable it.
* Replaces the standard golangci-lint binary with a build of our linter for both local testing and in CI. This new flow will allow us to add custom linters to the Wormhole repository in the future.
* Adds linter runs for the linter modules themselves, the node directory, and the SDK. Wormchain historically hasn't been linted and it would likely require a lot of work to do this. So, not making this change currently. 

`channelcheck` is NOT enabled in this PR to limit the number of files being touched. This will be enabled in a separate follow-up PR. There are comments labeled `Note on channel capacity` that will be converted either into `nolint` comments or into code that satisfies the linter.


scripts/lint.sh
```diff
@@ -54,26 +58,57 @@ format(){
...
+            exit 1
+            ;;
+    esac
+
+    dir="$WORMHOLE_GOLANGCI_LINT_CACHE/$WORMHOLE_GOLANGCI_LINT_VERSION"
+    bin="$dir/wormhole-golangci-lint"
+
+    if [ ! -x "$bin" ]; then
+        mkdir -p "$dir"
+        url="https://github.com/asymmetric-research/wormhole-custom-ci/releases/download/${WORMHOLE_GOLANGCI_LINT_VERSION}/wormhole-golangci-lint_${WORMHOLE_GOLANGCI_LINT_VERSION}_${asset}"
```
johnsaigle: You might be able to get this to work with `go install github.com/asymmetric-research/wormhole-custom-ci@<version>` instead, might be a bit simpler.

If you do that then the binary could live in the user's `GOPATH` and you can use Go's built-in caching and other toolchain advantages instead of creating one ad-hoc here

mdulin2: golangci-lint itself is go-installable, but the custom-built variant intentionally isn't because its `main` lives upstream. We'd have to add some extra wiring to support this, which I don't think is worth the effort.

johnsaigle:
Adding a note here to cross-reference this PR with two other open PRs which introduce changes to golangci-lint this PR will likely conflict with.

https://github.com/wormhole-foundation/wormhole/pull/4701#issuecomment-4576517471
https://github.com/wormhole-foundation/wormhole/pull/4765

.github/CODEOWNERS
```diff
@@ -39,6 +39,7 @@
 /wormchain/devnet/ @evan-gray
 /wormchain/devnet/txverifier @djb15 @johnsaigle @mdulin2 @pleasew8t
 /wormchain/ts-sdk/ @evan-gray @kev1n-peters @panoel
+/linters/ @djb15 @johnsaigle @mdulin2 @pleasew8t
```
johnsaigle: Can you add @bemic?

linters/rules/channelcheck/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/certusone/wormhole/linters/rules/channelcheck
+
+go 1.25.0
```
johnsaigle: suggestion: use the same go version as the one in `node/go.mod`

scripts/lint.sh
```diff
@@ -54,26 +65,79 @@ format(){
...
-    cd "$ROOT"/node
-    golangci-lint run --timeout=10m $GOLANGCI_LINT_ARGS ./...
 
-    cd "${ROOT}/sdk"
-    golangci-lint run --timeout=10m $GOLANGCI_LINT_ARGS ./...
+    # === Go linting (custom wormhole-golangci-lint)
+    local LINT_BIN
+    LINT_BIN="$(ensure_wormhole_golangci_lint)"
+
+    # Lint each Go module root. linters/ is the custom linter's own code; its
```
johnsaigle: We're not linting `wormchain/` here so this comment is a little misleading.

scripts/lint.sh
```diff
@@ -54,26 +65,79 @@ format(){
     fi
 }
 
+# Content hash of the linters/ source, used to decide whether a rebuild is needed.
+# Hashes every file under linters/ (regardless of git tracking) except the
+# bin/ and build/ build outputs, so any source edit triggers exactly one rebuild.
+# Runs find from inside LINTERS_DIR so paths are relative: sha256sum embeds the
+# path, so an absolute one would make the hash checkout-path dependent.
+linters_source_hash() {
```
johnsaigle: My understanding is that you don't need any of this caching logic. Go should handle this normally when you build the custom golangci-lint binary. You can read more about this by running `go help environment` and looking at the `GOCACHE` and `GOMODCACHE` variables. Basically, Go is smart enough to figure out when a binary needs to be rebuilt so there's no need to add custom caching logic on top of what Go does normally.

linters/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/certusone/wormhole/linters
```
johnsaigle:
```suggestion
module github.com/wormhole-foundation/wormhole/linters
```

suggestion: Since this is a new package, I don't think we need compatibility with the old `certusone` namespace. I don't think it has an impact either way but we may as well use the real path that exist on github.com instead of the fake name.

linters/rules/channelcheck/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/certusone/wormhole/linters/rules/channelcheck
```
johnsaigle: suggestion: Since this is a new package, I don't think we need compatibility with the old certusone namespace. I don't think it has an impact either way but we may as well use the real path that exist on github.com instead of the fake name.

linters/.gitignore
```diff
@@ -0,0 +1,11 @@
...
+/bin/
+
+# Stray root-level builds, e.g. from a bare `go build ./...` (the Makefile
+# targets put these under bin/).
+/wormhole-lint
+/wormhole-golangci-lint
+
+# Cloned + patched upstream sources (revive, golangci-lint) used to build
+# bin/wormhole-golangci-lint. Recreated by `make build-golangci-lint`.
+/build/
```
johnsaigle: Doesn't this use `bin` now? Also, the reference to `revive` can be removed

linters/rules/channelcheck/channelcheck.go
```diff
@@ -0,0 +1,488 @@
...
+	"go/types"
+
+	"github.com/golangci/plugin-module-register/register"
+	"golang.org/x/tools/go/analysis"
+)
+
+// ChannelCheckPlugin is the golangci-lint module-plugin entry point. Its
+// configuration lives in the package-level settings var (populated by New),
+// which is the single source of truth — shared with the standalone
+// cmd/wormhole-lint multichecker, which populates the same var via flags.
```
johnsaigle, line 25: This part looks out of date, I think? I don't think we're using the multichecker approach.

linters/rules/channelcheck/channelcheck.go
```diff
@@ -0,0 +1,488 @@
...
+					break
+				}
+
+				// If the SendStmt was NOT found within a Select clause, then add a linter error.
+				tokenID := n.Pos()
+				if _, ok := seenPositions[tokenID]; !ok {
+					if name, named := sendChanName(n); named && settings.ignoreChannelNames[name] {
+						return true
+					}
+					pass.Reportf(tokenID, "Blocking send. Add timer, ticker, or default case: %q", render(pass.Fset, n))
```
johnsaigle, line 215: suggestion: The `render` output can get sort of noisy, e.g.

```
node/pkg/adminrpc/adminserver.go:1061:3: Blocking send. Add timer, ticker, or default case: "s.injectC <- &common.MessagePublication{\n\tTxID:\t\t\tethcommon.Hash{}.Bytes(),\n\tTimestamp:\t\tv.Timestamp,\n\tNonce:\t\t\tv.Nonce,\n\tSequence:\t\tv.Sequence,\n\tConsistencyLevel:\tv.ConsistencyLevel,\n\tEmitterChain:\t\tv.EmitterChain,\n\tEmitterAddress:\t\tv.EmitterAddress,\n\tPayload:\t\tv.Payload,\n\tIsReobservation:\tfalse,\n\tUnreliable:\t\tfalse,\n}" (channelcheck)
		s.injectC <- &common.MessagePublication{ // Note on channel capacity: Only blocks this command
```

Up to you but I think the line number is sufficient.

scripts/lint.sh
```diff
@@ -54,26 +65,79 @@ format(){
...
+    # rules/<name>/ are separate modules (mirroring `make -C linters test`), so
+    # lint each one too. Run in a subshell so a failure still halts under
+    # `set -e` without leaking the working directory between modules.
+    lint_module() {
+        ( cd "$1" && "$LINT_BIN" run --timeout=10m $GOLANGCI_LINT_ARGS ./... )
+    }
+
+    lint_module "$ROOT/node"
+    lint_module "$ROOT/sdk"
+    lint_module "$ROOT/linters"
```
johnsaigle, line 104: Also, are you sure you want to lint the `linters/` directory? We use a pretty strict config in Wormhole mostly because the security is crucial, but this internal tool doesn't need to be as strict. It's not a bad idea necessarily but it does mean that if we add more lints in the future that we'll need to update a lot more files.

On the other hand, if you're adding self-linting of the channelcheck, you could also consider adding unit tests to CI to make sure that changes to the tool don't break anything going forward.

linters/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+test:
+	go test -v ./...
+	@for mod in $$(find rules -name go.mod); do \
+	  dir=$$(dirname $$mod); \
+	  echo "==> testing $$dir"; \
+	  (cd $$dir && go test -v ./...) || exit $$?; \
+	done
+
+vet:
```
johnsaigle: Since right now you have included `linters/` as a target for golangci-lint, you don't need this target as `golangci-lint` has `govet` enabled there already.

If you want a lint target here I think you can copy `node/Makefile`'s `lint` target to get accurate coverage for this directory.

linters/Makefile
```diff
@@ -0,0 +1,36 @@
+GOLANGCI_LINT_VERSION ?= v2.12.2
```
johnsaigle: Maybe we can use a git commit hash here instead of a version number? That way we are protected incase the version is malleable.

So this line would look something more like `GOLANGCI_LINT_HASH ?= 123abc... # v2.12.2` and the `go install` below would use `go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$GOLANGCI_LINT_HASH`

johnsaigle (commented):
The linter works and I tested the new lint make targets from each of the subdirectories and via the Docker method.

I left some comments mostly around deleting a custom caching system which I don't believe is necessary given Go's built-in caching mechanism.

I also left several notes marked "suggestion" that can be ignored/resolved without comment. 👍🏻

.github/CODEOWNERS
mdulin2: Yeah, totally. Please don't hate me, Lukas.

linters/rules/channelcheck/channelcheck.go
mdulin2: This was large because it was a raw struct being used for the send. I like the verbosity of it for reading linter errors in bulk so I think I'll keep it.

scripts/lint.sh
mdulin2: We don't have to. I thought the more things are linted, the better. Otherwise, it will become outdated over time. What do you think? I can go either way on it.

scripts/lint.sh
mdulin2: Interesting! I hadn't considered relying on the built-in Go caching for this. 

I somewhat made this for the remote CI runs. After thinking about this comment some more, the real bottleneck isn't the linter runs, it's the tilt runs. So, we're probably not optimizing for the correct thing there. I'll remove the caching aspect of the linter.

johnsaigle: Yeah exactly, CI is set-up to run Tilt in a parallel job from the lints so they're independent.

scripts/lint.sh
johnsaigle: Sure we can leave it as-is. Thoughts on adding its tests to CI?

linters/rules/channelcheck/channelcheck.go
mdulin2: It uses `multichecker` so we can compile multiple lints into a single `wormhole-lint` binary. So, I think this is correct. I don't necessarily think we need this build option, but it's really nice for testing lints when you first create them.

johnsaigle (dismissed):
Ran the tests on the latest commit with and without channelcheck enabled.

scripts/lint.sh
mdulin2: It definitely should be. I hadn't considered that the test suite wasn't running automatically. I'll add that.

mdulin2: Added to the `test-coverage` makefile command. Does anything else need to be done for the coverage aspect of this?

djb15 (dismissed):
I didn't review the linter itself in a tonne of detail but I think the changes to the existing code look good, seems to be pretty unobtrusive and it also doesn't seem to change the developer workflow at all which is great.

The linter itself doesn't appear to do anything dangerous and the test cases cover the channel use cases we have in the monorepo so far.

scripts/lint.sh
johnsaigle: The gap I was suggesting is that the unit tests for the new custom linter don't run in CI at all. It's true that they are also excluded from the test _coverage_ metrics but I don't think we need to add them there as those are designed to check the coverage of the node code itself. (I suppose it doesn't hurt though)

So, more specifically, I think you should add a new job in `.github/workflows/build.yml` at the same level as e.g. `node-tests`, only it should be called something like `custom-golangci-tests` and its `run` step should be to run `go test` or similar within the `linters/` directory. That way, CI will fail if the unit tests for the custom linters fails. This should be run in a separate step run from the node tests (i.e. not bundled directly into `make`) since we're just checking the custom tool here, not the `node/` code, so their CI tests can be isolated.

mdulin2: Okay, sure. I can change that.

mdulin2:
I messed up the git history' commits were duplicated for some reason. Changes are much cleaner now.

scripts/lint.sh
mdulin2: - Removed the linter tests from the coverage
- Created a new job to run specifically for the linter tests.

johnsaigle:
@mdulin2 looks like you need to rebase after your EVM watcher PR got merged

johnsaigle (dismissed):

linters/rules/channelcheck/README.md
```diff
@@ -0,0 +1,24 @@
...
+Many of these will lead to false positives or situations where we *want* a blocking channel send. In these cases, `nolint:channelcheck` is easy to add (assuming this is integrated directly with golangci-lint). Regardless, having this issue pointed out automatically is a good way to fix bugs.
+
+## Configuration
+
+Each option can be set two ways: as a golangci-lint module setting (under
+`settings.custom.channelcheck.settings:`, using the **Setting** name) or as a
+standalone analyzer flag (when running the `wormhole-lint` binary, using the
+**Flag** name).
+
+| Setting                   | Flag         | Type       | Default | Description                                                                                     |
```
bemic, line 19: Should we include `CheckEmptyDefault` in the table?

linters/README.md
```diff
@@ -0,0 +1,125 @@
...
+          blocking: true
+          unbuffered: false
+          bufferMax: 0
+          emptyDefault: false
+          ignoreChannelsByName: [errC]
+```
+
+`channelcheck` settings:
+
+| Setting                | Type       | Default | Description                                                                                                                              |
```
bemic, line 67: Might be better to use actual setting names (`CheckBlockingSends`) instead of flags (`blocking`)?

djb15 (dismissed):

linters/rules/channelcheck/README.md
bemic: @mdulin2 feel free to mark my 2 comments as resolved and ignore, they are non blockers.

SEJeff (commented):
Why not use golangci-lint? I am unclear on why we want this vs using the industry standard linting tool and plugging any custom linters we want into it.

johnsaigle:
@SEJeff This is how you do that. https://golangci-lint.run/docs/plugins/module-plugins/

I agree it's a weird system that they make you recompile a custom version of golangci-lint but that appears to be the way this gets done. We also took a look at using revive instead but it has basically the same set-up, so we went this way.

mdulin2:
Any thoughts from folks on this? I'd love to get this merged so that we can turn on the channel linter in the repo, as well as add other custom linters.

johnsaigle:
@mdulin2 I think we're good on our side, you just need codeowner review now

.gitignore
```diff
@@ -33,4 +33,4 @@ sui/examples/wrapped_coin
 /coverage-check
 # Only used for internal testing
 *.testnet.yaml
-coverage.txt
+coverage.txt
```
SEJeff: I scoff at you @mdulin2 for removing a trailing `\n`. You should fix your editor going forward, but resolve this comment for this PR.

linters/rules/channelcheck/channelcheck_test.go
SEJeff, line 1: These could be massively simplified with table driven tests.

linters/rules/channelcheck/channelcheck.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+
+	for _, file := range pass.Files {
+		var seenPositions = make(map[token.Pos]bool)
+
+		ast.Inspect(file, func(node ast.Node) bool {
+			switch n := node.(type) {
+			// Fails open by design. Will
```
SEJeff: Unfinished comment?

SEJeff (dismissed):
All of my comments are minor nits you can close if you don't want to do them for this PR. Nice!

johnsaigle (approved):

djb15 (approved):

SEJeff (approved):

PR #4843, opened 2026-05-28
feat: sui wtt pauser

douglasgalico:


sdk/js/src/utils/consts.ts
```diff
@@ -831,7 +831,7 @@ const DEVNET = {
   sui: {
     core: "0xea31c369d1f873d87d37f313ec37f1ee20a0b8136f06e3d3521330ee467312a4", // wormhole module State object ID
     token_bridge:
-      "0xbab50719a0e35350677b75bddc25f20bdc7a8c4788c7a0746b9796956380a2d2", // token_bridge module State object ID
+      "0xfa434e835120d31598abceee7208e5236aead674aa4e216b5ca761454a7351a7", // token_bridge module State object ID
```
douglasgalico: as bytecode of state has changed

sui/token_bridge/sources/state.move
```diff
@@ -371,8 +471,16 @@ module token_bridge::state {
     ///
     /// NOTE: Please keep this method as public(friend) because we never want
     /// to expose this method as a public method.
-    public(friend) fun migrate__v__0_2_0(_self: &mut State) {
-        // Intentionally do nothing.
+    public(friend) fun migrate__v__0_3_0(self: &mut State) {
+        // Initialize pause dynamic fields with defaults.
+        // paused = false, pauser = @0x0, unpauser = @0x0
+        // Guarded with exists_ check for resilience against partial migration
```
kcsongor: partial migration is not possible, and the `migrate` handler ensures it can only execute once. no harm in keeping this check here, but it's redundant!

kcsongor:
I think we should consider introducing pauser/unpauser capabilities, and guard the relevant functions behind ownership of these objects. The issue with the current method is that checking `tx::sender` means the pauser/unpauser can only be an EOA (and not something like the governance program or another smart contract).

Unfortunately it's not possible for the admin to then claw back a capability object from its current holder, so this approach would come with some extra complexity in the form of a table that holds which capability object is the active one (which would look a lot like the current implementation which stores the address of the pauser/unpauser. What I'm proposing would instead store the id of the pauser/unpauser cap objects, and check those instead of tx::sender)

sui/token_bridge/sources/governance/set_pauser_addresses.move
```diff
@@ -71,30 +93,65 @@ module token_bridge::set_pauser_addresses {
...
-        sui::event::emit(PauserAddressesSet { pauser, unpauser });
+    /// Mint a `PauserCap` for `owner` and record its id as active. A zero owner
+    /// unassigns the role (records @0x0, mints nothing). Returns the recorded id.
+    fun assign_pauser(
+        token_bridge_state: &mut State,
+        latest_only: &state::LatestOnly,
+        owner: address,
+        ctx: &mut TxContext
+    ): address {
+        if (owner == @0x0) {
```
kcsongor: I think it would be cleaner if we used `option<address>` throughout rather than using `0x0` as a special address

sui/token_bridge/sources/state.move
```diff
@@ -172,7 +172,8 @@ module token_bridge::state {
         }
     }
 
-    /// Returns the configured pauser address. Returns `@0x0` if unset.
+    /// Returns the designated pauser capability's object id (as an `address`).
+    /// Returns `@0x0` if unset. See `token_bridge::pause`.
     public fun pauser(self: &State): address {
```
kcsongor: `pauser` and `unpauser` should be represented as `ID` instead of `address` as it's semantically more accurate

sui/token_bridge/sources/governance/set_pauser_addresses.move
kcsongor: or better yet, `option<ID>` per my other comment

kcsongor (dismissed):

kcsongor (dismissed):

sui/token_bridge/sources/pause.move
```diff
@@ -0,0 +1,350 @@
...
+    const E_NOT_EXPIRED: u64 = 7;
+
+    /// Temporary-pause duration: 5 days, in milliseconds (Sui Clock is ms).
+    const PAUSE_DURATION_MS: u64 = 432_000_000;
+    /// Maximum representable timestamp; used by `freeze_bridge`.
+    const MAX_TIMESTAMP_MS: u64 = 0xFFFFFFFFFFFFFFFF;
+
+    /// Capability whose holder may `pause` the bridge, while its id is the
+    /// active pauser. Minted and transferred by governance.
+    struct PauserCap has key, store {
```
mdulin2, line 96: `store` allows for `public_transfer()` calls to transfer the object to other users, and be `shared`. If we removed `store`, this would no longer be possible. 

However, this would make it so that a contract couldn't own the object. 

If contracts need to be able to own the object for something like Unpauser via governance, then I'd leave as is. Otherwise, I'd remove the `store` portion of this.

sui/token_bridge/sources/test/pause_tests.move
```diff
@@ -0,0 +1,965 @@
+// SPDX-License-Identifier: Apache 2
+
+#[test_only]
+module token_bridge::pause_tests {
```
mdulin2, line 4: Good tests! Could you add a test that calls one of the guarded functions while it's paused to ensure that it reverts?

mdulin2 (dismissed):

sdk/js/src/utils/consts.ts
mdulin2: I don't think that the object ids should change with a contract upgrade.

douglasgalico: afaik we don't upgrade in devnet? But just redeploy it from tilt - so it needs a new one

sui/token_bridge/sources/state.move
```diff
@@ -371,8 +536,24 @@ module token_bridge::state {
     ///
     /// NOTE: Please keep this method as public(friend) because we never want
     /// to expose this method as a public method.
-    public(friend) fun migrate__v__0_2_0(_self: &mut State) {
-        // Intentionally do nothing.
+    public(friend) fun migrate__v__0_3_0(self: &mut State) {
```
mdulin2, line 539: Do we need to add these changes to the initialization path for usage on Tilt? or, do we run the upgrade process incrementally? Just want to make sure that Tilt matches the real state as much as possible when testing.

mdulin2 (approved):

kcsongor (approved):

kev1n-peters (approved):

PR #4846, opened 2026-06-01
node(sui): Deserialize txID as string, not *string

johnsaigle:
The function `getEvents` is missing a nil check on `datum.ID.TxDigest`.
However, it doesn't look like this field is actually nil-able upstream:
https://docs.sui.io/sui-api-ref#suix_queryevents

So, it makes more sense to deserialize this value as a string (not
`*string`) which allows us to avoid nil checks. Where nil checks already
existed, this replaces them with length == 0 checks.

### DIFF
--- a/node/pkg/watchers/sui/watcher.go
+++ b/node/pkg/watchers/sui/watcher.go
@@ -87,7 +87,7 @@ type (
 
 	SuiResult struct {
 		ID struct {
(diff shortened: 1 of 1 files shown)


djb15:
Arguably many more of these fields don't need to be pointers according to the docs. Anyhow this is soon going to be superseded by the gRPC changes so suggest we just close this

PR #4847, opened 2026-06-02
Add test-fast for Go unit testing

johnsaigle:
Speed up local human + agent unit test loops by making a new testing
target. CI is untouched so it still runs the full test suite.

- Add `test-fast` to get a much faster unit test loop in Go code. This
  passes `-short` to go test, skips fuzzing, and omits `-race`.
- Gate slow integration-style tests, stress tests, and tests relying on
  sleeps behind Go's testing.Short() configuration.
- Add t.Parallel() to certain tests where appropriate (Go tests run in
- Shortens fuzz time for a specific test since it hasn't had any issues
  since merged

LLM summary from diffing the two test times:
> make test-fast gives a fast local path: 18.43s vs main make test at 89.15s, about 79% faster.

PR #4850, opened 2026-06-05
node: Migrate Sui watcher to use `suiclient` package

djb15:
In this PR we update the Sui watcher and txverifier to call into the new `suiclient` package that was added in #4804. The most significant change is that we now BCS decode the returned on-chain data, instead of decoding the JSON data that was returned by the JSON-RPC. Although the new `gRPC` API does support formatting BCS return data as JSON, we elect to use the BCS return data to avoid theoretically vulnerabilities in that translation layer.

The [watcher_live_test.go](https://github.com/wormhole-foundation/wormhole/pull/4850/changes#diff-8e73bb58efd137d13fb2ecfc8785f68d88279e89a0adcd351012bebee4ec4217) file was added in this PR and it validates that the message data parsed by the new logic exactly matches the data that is returned by the soon to be deprecated JSON-RPC.

A flag has been removed from the txverifier as the new subscription based client is forwards looking. The backwards polling on startup based on the latest block is no longer supported but as far as I can tell this isn't critical functionality.

The `worm` cli tool should also be updated to utilise the new Sui gRPC interface, but that is left for a future PR as it's an intrusive change that requires the updating of numerous dependencies.

### DIFF
--- a/node/pkg/suiclient/suiclient.go
+++ b/node/pkg/suiclient/suiclient.go
@@ -106,7 +106,7 @@ type SuiObjectChange struct {
 // Callers MUST nil-check every field they read. The Get* methods do not
 // enforce per-field presence — that responsibility belongs to the caller.
 type SuiTransaction struct {
-	Digest        *string
+	TxDigest      *string
 	Events        []SuiEvent
 	Checkpoint    *uint64
 	Timestamp     *time.Time
--- a/node/pkg/suiclient/suiclient.go
+++ b/node/pkg/suiclient/suiclient.go
@@ -23,6 +23,7 @@ const (
 	TransactionFieldCheckpoint     = "checkpoint"
 	TransactionFieldTimestamp      = "timestamp"
 	TransactionFieldChangedObjects = "effects.changed_objects"
+	TransactionFieldStatus         = "effects.status"
 )
 
 const (
@@ -180,6 +181,10 @@ type SuiClient interface {
 	// SuiTransaction; see the TransactionField* constants. At least one field
 	// is required.
 	//
+	// The execution status is always fetched in addition to the requested
+	// fields, and an error is returned unless the transaction executed
+	// successfully — callers never see data from failed transactions.
+	//
 	// Fields not requested (and fields requested but missing from the upstream
 	// response) come back nil/empty. Callers MUST nil-check every field they
 	// read — this method does not enforce per-field presence.
@@ -187,7 +192,7 @@ type SuiClient interface {
 
 	// Subscribe to transaction events of type `eventType`. Each matching event is delivered
 	// on the channel as a SuiTransactionEvent, pairing it with the digest of the transaction
-	// that emitted it.
+	// that emitted it. Events from transactions that did not execute successfully are dropped.
 	SubscribeToTransactionEvent(ctx context.Context, eventType string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error)
 	SubscribeToTransactionEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error)
 
--- a/node/pkg/suiclient/suiclient.go
+++ b/node/pkg/suiclient/suiclient.go
@@ -82,9 +82,9 @@ type SuiTransactionEvent struct {
 	Event    SuiEvent
 }
 
-// SuiObjectChange holds the subset of a transaction effect's changed-object
-// information needed to look up an object's state before and after a
-// transaction. It is populated from effects.changed_objects; request
+// SuiObjectChange holds basic metadata of an object that changed during the
+// execution of a transaction, such as it's ID, type and versions before and
+// after transaction execution. It is populated from effects.changed_objects; request
 // TransactionFieldChangedObjects to have it populated on SuiTransaction.
 //
 // Callers MUST nil-check every field they read.
--- a/node/pkg/suiclient/suiclient.go
+++ b/node/pkg/suiclient/suiclient.go
@@ -18,10 +18,11 @@ const (
 // any string, so callers needing a sub-message path (e.g. "contents.value")
 // can pass the literal directly.
 const (
-	TransactionFieldDigest     = "digest"
-	TransactionFieldEvents     = "events"
-	TransactionFieldCheckpoint = "checkpoint"
-	TransactionFieldTimestamp  = "timestamp"
+	TransactionFieldDigest         = "digest"
+	TransactionFieldEvents         = "events"
+	TransactionFieldCheckpoint     = "checkpoint"
+	TransactionFieldTimestamp      = "timestamp"
+	TransactionFieldChangedObjects = "effects.changed_objects"
 )
 
 const (
@@ -70,6 +71,33 @@ type SuiEvent struct {
 	BcsBytes          []byte
 }
 
+// SuiTransactionEvent pairs an event with the digest of the transaction that emitted it.
+// It is produced by the subscription, where events are delivered individually and would
+// otherwise lose their transaction context — the bare gRPC Event carries no digest, so the
+// digest is taken from the parent ExecutedTransaction. Events obtained via GetTransaction do
+// not need this wrapper, since the caller already knows the digest.
+type SuiTransactionEvent struct {
+	TxDigest string
+	Event    SuiEvent
+}
+
+// SuiObjectChange holds the subset of a transaction effect's changed-object
+// information needed to look up an object's state before and after a
+// transaction. It is populated from effects.changed_objects; request
+// TransactionFieldChangedObjects to have it populated on SuiTransaction.
+//
+// Callers MUST nil-check every field they read.
+type SuiObjectChange struct {
+	ObjectID   *string
+	ObjectType *string
+	// InputVersion is the object's version before this transaction executed
+	// (i.e. the previous version).
+	InputVersion *uint64
+	// OutputVersion is the object's version after this transaction executed
+	// (i.e. the current version).
+	OutputVersion *uint64
+}
+
 // SuiTransaction holds the flat-primitive fields of an executed transaction.
 // A field is populated only when the caller requested it via the field mask
 // AND the upstream node returned it; otherwise the field is nil/empty.
@@ -77,10 +105,11 @@ type SuiEvent struct {
 // Callers MUST nil-check every field they read. The Get* methods do not
 // enforce per-field presence — that responsibility belongs to the caller.
 type SuiTransaction struct {
-	Digest     *string
-	Events     []SuiEvent
-	Checkpoint *uint64
-	Timestamp  *time.Time
+	Digest        *string
+	Events        []SuiEvent
+	Checkpoint    *uint64
+	Timestamp     *time.Time
+	ObjectChanges []SuiObjectChange
 }
 
 // SuiCheckpoint holds the flat-primitive fields of a checkpoint. A field is
@@ -156,9 +185,11 @@ type SuiClient interface {
 	// read — this method does not enforce per-field presence.
 	GetTransaction(ctx context.Context, digest string, fields []string) (SuiTransaction, error)
 
-	// Subscribe to transaction events of type `eventType`
-	SubscribeToTransactionEvent(ctx context.Context, eventType string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error)
-	SubscribeToTransactionEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error)
+	// Subscribe to transaction events of type `eventType`. Each matching event is delivered
+	// on the channel as a SuiTransactionEvent, pairing it with the digest of the transaction
+	// that emitted it.
+	SubscribeToTransactionEvent(ctx context.Context, eventType string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error)
+	SubscribeToTransactionEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error)
 
 	// Close the client
 	Close() error
--- a/node/pkg/suiclient/suigrpc.go
+++ b/node/pkg/suiclient/suigrpc.go
@@ -168,7 +168,7 @@ func (s *SuiGrpcClient) createCheckpointStream(ctx context.Context, fields []str
 	return stream, err
 }
 
-func (s *SuiGrpcClient) SubscribeToTransactionEvent(ctx context.Context, event string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
+func (s *SuiGrpcClient) SubscribeToTransactionEvent(ctx context.Context, event string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error) {
 	eventTypes := []string{
 		event,
 	}
@@ -176,7 +176,7 @@ func (s *SuiGrpcClient) SubscribeToTransactionEvent(ctx context.Context, event s
 	return s.SubscribeToTransactionEvents(ctx, eventTypes, eventWriteChannel)
 }
 
-func (s *SuiGrpcClient) SubscribeToTransactionEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiEvent) (SuiSubscription, error) {
+func (s *SuiGrpcClient) SubscribeToTransactionEvents(ctx context.Context, eventTypes []string, eventWriteChannel chan<- SuiTransactionEvent) (SuiSubscription, error) {
 	if len(eventTypes) == 0 {
 		return SuiSubscription{}, fmt.Errorf("sui gRPC SubscribeToTransactionEvents requires at least one event type")
 	}
@@ -184,8 +184,10 @@ func (s *SuiGrpcClient) SubscribeToTransactionEvents(ctx context.Context, eventT
 		return SuiSubscription{}, fmt.Errorf("sui gRPC SubscribeToTransactionEvents requires a non-nil eventWriteChannel")
 	}
 
-	// This stream is only concerned with transaction events
+	// This stream is concerned with transaction events plus the digest of the transaction
+	// that emitted them, which is paired with each event as a SuiTransactionEvent below.
 	fields := []string{
+		"transactions.digest",
 		"transactions.events",
 	}
 
@@ -273,10 +275,17 @@ func (s *SuiGrpcClient) SubscribeToTransactionEvents(ctx context.Context, eventT
 						continue
 					}
 
+					// Pair the event with the digest of the transaction that emitted it; the
+					// bare gRPC Event carries no digest of its own.
+					txDigest := ""
+					if tx.Digest != nil {
+						txDigest = *tx.Digest
+					}
+
 					// Writing to eventWriteChannel could be blocking if there are no readers. Use a select to include
 					// a simultaneous check to bail out if the context is cancelled.
 					select {
-					case eventWriteChannel <- *suiEvent:
+					case eventWriteChannel <- SuiTransactionEvent{TxDigest: txDigest, Event: *suiEvent}:
 					case <-ctx.Done():
 						return
 					}
@@ -298,9 +307,11 @@ func (s *SuiGrpcClient) Close() error {
 	return nil
 }
 
-// Create a new SuiClient, with the gRPC service as implementation. Additional gRPC dial options
-// (e.g. for custom transport credentials, interceptors, or per-RPC metadata via interceptors) may
-// be supplied; they are appended after the defaults so callers can override them.
+// Create a new SuiClient, with the gRPC service as implementation. The default transport is TLS.
+// Additional gRPC dial options (e.g. for custom transport credentials, interceptors, or per-RPC
+// metadata via interceptors) may be supplied; they are appended after the defaults so callers can
+// override them. In particular, passing grpc.WithTransportCredentials(insecure.NewCredentials())
+// overrides the default TLS transport with a plaintext one, as needed for a local dev-mode node.
 func NewSuiGrpcClient(rpcURL string, logger *zap.Logger, extraOpts ...grpc.DialOption) (SuiClient, error) {
 	if logger == nil {
 		logger = zap.NewNop()
@@ -403,6 +414,20 @@ func grpcExecutedTransactionToSuiTransaction(grpcTransaction *pb.ExecutedTransac
 			if suiEventPtr := grpcEventToSuiEvent(event); suiEventPtr != nil {
 				out.Events = append(out.Events, *suiEventPtr)
 			}
+		}
+	}
+
+	if grpcTransaction.Effects != nil {
+		for _, changedObject := range grpcTransaction.Effects.ChangedObjects {
+			if changedObject == nil {
+				continue
+			}
+			out.ObjectChanges = append(out.ObjectChanges, SuiObjectChange{
+				ObjectID:      changedObject.ObjectId,
+				ObjectType:    changedObject.ObjectType,
+				InputVersion:  changedObject.InputVersion,
+				OutputVersion: changedObject.OutputVersion,
+			})
 		}
 	}
 
--- a/node/pkg/suiclient/suigrpc.go
+++ b/node/pkg/suiclient/suigrpc.go
@@ -130,12 +130,23 @@ func (s *SuiGrpcClient) GetLatestCheckpoint(ctx context.Context, fields []string
 // Replaces `sui_getTransactionBlock`.
 // Docs: https://www.quicknode.com/docs/sui/sui-grpc/ledger/get-transaction
 //
+// The execution status is always fetched in addition to the requested fields, and an error is
+// returned unless the transaction carries an explicit successful status. Wormhole only ever
+// cares about successful transactions, so enforcing this here guarantees callers never act on
+// data from a failed (or status-less) transaction.
+//
 // Returned-field nil-checking is the caller's responsibility — any field
 // not requested OR not returned by the upstream node comes back nil/empty,
 // and this method does not enforce per-field presence.
 func (s *SuiGrpcClient) GetTransaction(ctx context.Context, digest string, fields []string) (SuiTransaction, error) {
 	if len(fields) == 0 {
 		return SuiTransaction{}, fmt.Errorf("sui gRPC GetTransaction requires at least one field for digest=%s", digest)
+	}
+
+	if !slices.Contains(fields, TransactionFieldStatus) {
+		// Clone before appending so the caller's slice is never mutated through a shared
+		// backing array.
+		fields = append(slices.Clone(fields), TransactionFieldStatus)
 	}
 
 	getTransactionRequest := pb.GetTransactionRequest{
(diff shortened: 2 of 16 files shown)


djb15:
@claude

node/pkg/suiclient/suigrpc.go
```diff
@@ -273,10 +275,17 @@ func (s *SuiGrpcClient) SubscribeToTransactionEvents(ctx context.Context, eventT
 						continue
 					}
 
+					// Pair the event with the digest of the transaction that emitted it; the
+					// bare gRPC Event carries no digest of its own.
+					txDigest := ""
+					if tx.Digest != nil {
+						txDigest = *tx.Digest
```
johnsaigle, line 320: Should `tx.Digest` have a nil check first?

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
 		}
 
-		// If any of these event parameters are nil, skip the event
-		if wormholeMessage.Sender == nil || wormholeMessage.Sequence == nil || event.Type == nil {
+		// BCS-decode the event contents into a WormholeMessage. This is done explicitly to avoid any
+		// unnecessary error logging for events that can't be deserialized into `WormholeMessage`
+		// instances. If the contents cannot be decoded, the event is simply skipped.
+		wormholeMessage, err := suiclient.DecodeBcs[WormholeMessage](event.BcsBytes)
+		if err != nil {
+			logger.Debug("Failed to BCS-decode WormholeMessage event", zap.Error(err))
```
johnsaigle: Is this worth a Warn or Error level? We won't see Debug in production

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
+		// Parse the wormhole message. vaa.IsTransfer can be omitted, since this is done
+		// inside `DecodeTransferPayloadHdr` already.
+		hdr, err := vaa.DecodeTransferPayloadHdr(wormholeMessage.Payload)
 
-			// The sender address is prefixed with "0x", but the message ID format does not include that prefix.
-			senderWithout0x := strings.TrimPrefix(*wormholeMessage.Sender, "0x")
+		// If there is an error decoding the payload, skip the event. One reason for a potential
+		// failure in decoding is that an attestation of a token was requested.
+		if err != nil {
+			logger.Debug("Error decoding payload", zap.Error(err))
```
johnsaigle: Same here, is Debug sufficient? I think this log could be omitted if the main issue is that an attestation is missing, but if there are other cases then it might be good to promote this to a higher level.

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+//   - A Move `Balance<T>`/`Supply<T>` wraps a single `u64` and serializes as a bare u64.
+//   - `wormhole::bytes32::Bytes32 { data: vector<u8> }` is a length-prefixed vector, NOT a
+//     fixed array, so it decodes into a Go []byte (the length is always 32).
+//   - A Move `UID`/`ID`/`address` serializes as a fixed 32-byte value (no length prefix).
+//   - `token_registry::Key<C>` is an empty Move struct, which Sui serializes as a single
+//     `dummy_field: bool` (1 byte).
+
+// wormhole::bytes32::Bytes32 { data: vector<u8> }
+type suiBytes32 struct {
+	Data []byte
```
johnsaigle, line 29:
```suggestion
	Data [32]byte
```

Based on the above comments, it looks like we should fix the length here like what's done for other types.

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+}
+
+//	token_bridge::wrapped_asset::ForeignInfo<C> {
+//		token_chain: u16,
+//		token_address: ExternalAddress,
+//		native_decimals: u8,
+//		symbol: String,                // serialized as vector<u8>
+//	}
+type suiForeignInfo struct {
+	TokenChain     uint16
```
johnsaigle: Any reason to prefer `unit16` here to `vaa.ChainID`?

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+// BCS notes:
+//   - A Move `Balance<T>`/`Supply<T>` wraps a single `u64` and serializes as a bare u64.
+//   - `wormhole::bytes32::Bytes32 { data: vector<u8> }` is a length-prefixed vector, NOT a
+//     fixed array, so it decodes into a Go []byte (the length is always 32).
+//   - A Move `UID`/`ID`/`address` serializes as a fixed 32-byte value (no length prefix).
+//   - `token_registry::Key<C>` is an empty Move struct, which Sui serializes as a single
+//     `dummy_field: bool` (1 byte).
+
+// wormhole::bytes32::Bytes32 { data: vector<u8> }
+type suiBytes32 struct {
```
johnsaigle, line 28:
```suggestion
type Bytes32 struct {
```

nit: This is in a file `suitypes` that works only with Sui so prefixing each type with `sui` isn't necessary

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
-		scanPackage1 := matches[1]
-		scanCoinType1 := matches[2]
-		scanPackage2 := matches[3]
-		scanAssetType := matches[4]
-		scanCoinType2 := matches[5]
-
-		// Ensure that the asset type is wrapped or native
-		if scanAssetType != "wrapped_asset::WrappedAsset" && scanAssetType != "native_asset::NativeAsset" {
-			return false
+// The following structs mirror the on-chain Move types emitted/stored by the Sui core
```
johnsaigle, line 14: nit: These comments could be moved to package-level comments on the first line to be more compatible with go tooling e.g. `go doc`

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+// and token bridges. Their field order and types must match the Move definitions exactly
+// so that BCS decoding of the gRPC-provided bytes succeeds. The layouts were validated
+// against live mainnet objects.
+//
+// BCS notes:
+//   - A Move `Balance<T>`/`Supply<T>` wraps a single `u64` and serializes as a bare u64.
```
johnsaigle, line 20: nit: If you move these comments so they directly above the types that they're commenting on, then they'll show up correctly in e.g. hover text dev's editors when they want to know more about the types.

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+	Decimals    uint8
+}
+
+//	0x2::dynamic_field::Field<Name, Value> {
+//		id: UID,        // 32-byte address
+//		name: Name,     // token_registry::Key<C>, an empty struct -> 1-byte dummy bool
+//		value: Value,
+//	}
+type suiNativeAssetField struct {
+	ID    [32]byte
```
johnsaigle: Why not use `suiBytes32` in these other types?

node/pkg/txverifier/sui_test.go
```diff
@@ -43,127 +50,247 @@ type ResultTestCase struct {
...
+	}
+	return uint16(v)
+}
+
+// parseByteCSV parses a comma-separated list of decimal byte values (e.g. "0,1,255") into a []byte.
+func parseByteCSV(s string) []byte {
+	parts := strings.Split(s, ",")
+	out := make([]byte, len(parts))
+	for i, p := range parts {
+		v, err := strconv.Atoi(strings.TrimSpace(p))
```
johnsaigle: Better to use `strconv.ParseUint(..., 10, 8)` here so that you know that: the maximum value will be 255 and that negative values are rejected.

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+//	}
+type suiWrappedAsset struct {
+	Info        suiForeignInfo
+	TreasuryCap suiTreasuryCap
+	Decimals    uint8
+}
+
+//	0x2::dynamic_field::Field<Name, Value> {
+//		id: UID,        // 32-byte address
+//		name: Name,     // token_registry::Key<C>, an empty struct -> 1-byte dummy bool
```
johnsaigle, line 82: nit: Maybe this kind of empty structs could be called `DummyName` or `NameKeyDummy` or something for clarity. It's weird to see a variable called `Name` with the `bool` type.

node/pkg/watchers/sui/watcher_live_test.go
```diff
@@ -0,0 +1,252 @@
+package sui
```
johnsaigle, line 3: Could you add `//go:build integration` to the top? This will skip these tests when running normal unit tests.

If we do this more consistently, we can skip slower tests during quick dev loops but continue to run these in CI.

Right now the normal go unit test suite runs in 2-3 minutes and this test alone will basically 2x that time.

See alsol: https://github.com/wormhole-foundation/wormhole/pull/4847

node/pkg/watchers/sui/watcher_live_test.go
```diff
@@ -0,0 +1,252 @@
...
+//
+// Optional environment variables:
+//
+//	SUI_WATCHER_RPC         gRPC endpoint host:port      (default fullnode.mainnet.sui.io:443)
+//	SUI_WATCHER_JSONRPC     JSON-RPC URL for cross-check (default https://fullnode.mainnet.sui.io)
+//	SUI_WATCHER_EVENT_TYPE  core bridge WormholeMessage  (default mainnet core bridge type)
+//	SUI_WATCHER_SECONDS     how long to observe          (default 120)
+//	SUI_WATCHER_TXVERIFIER  set to any value to enable the transfer verifier
+func TestLiveSuiWatcher(t *testing.T) {
+	if os.Getenv("SUI_WATCHER_LIVE") == "" {
```
johnsaigle: Using something like the build tag mentioned above is a bit nicer IMO than using custom env vars to gate tests.

If you don't want to use that for whatever reason, consider instead using ` testing.Short()`. This way dev can choose "run just the fast tests" for speed or "run the full tests" for completeness. 

If someone just wants to run this test, they can use the integration tag and the path instead of a custom env var, e.g. `go test -tags integration ./pkg/watchers/sui -run TestLiveSuiWatcher`

node/pkg/suiclient/suigrpc.go
mdulin2: This is a nil check, correct? `if tx.Digest != nil`

node/pkg/watchers/sui/watcher.go
```diff
@@ -400,54 +292,69 @@ func (e *Watcher) Run(ctx context.Context) error {
...
-						if err != nil {
-							logger.Error("inspectBody Error", zap.Error(err))
-							continue
-						}
-						if event.checkpoint > e.latestProcessedCheckpoint {
-							e.latestProcessedCheckpoint = event.checkpoint
-						}
-					}
+
+				if err := e.processEvent(ctx, logger, txEvent.Event, txEvent.TxDigest, false); err != nil {
```
mdulin2, line 356: I like the rewrite to put this in a separate function. It makes this MUCH easier to test going forward.

node/pkg/suiclient/suiclient.go
```diff
@@ -18,10 +18,11 @@ const (
...
 const (
-	TransactionFieldDigest     = "digest"
-	TransactionFieldEvents     = "events"
-	TransactionFieldCheckpoint = "checkpoint"
-	TransactionFieldTimestamp  = "timestamp"
+	TransactionFieldDigest         = "digest"
+	TransactionFieldEvents         = "events"
+	TransactionFieldCheckpoint     = "checkpoint"
+	TransactionFieldTimestamp      = "timestamp"
+	TransactionFieldChangedObjects = "effects.changed_objects"
```
mdulin2, line 25: Could we add a check for `transactions.effects.status` on the sui client side? This way, the status is guaranteed to be valid once it reaches the watcher code. It should add a tiny more about of data, and a small `if` statement. So, I think it's worth the additional security controls.

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
+		// again after decoding the object contents.
+		if !validateSuiAssetType(objectType, s.suiTokenBridgePackageId) {
 			continue
 		}
 
-		// Get the previous version of the object. This makes a call to the Sui API.
-		resp, err := s.suiApiConnection.TryMultiGetPastObjects(ctx, objectChange.ObjectId, objectChange.Version, objectChange.PreviousVersion)
+		// Fetch the object at the version after this transaction executed (the current version)
+		// and at the version before it executed (the previous version). These calls go to the Sui API.
+		currentObject, err := s.client.GetObjectAtVersion(ctx, *objectChange.ObjectID, objectChange.OutputVersion, []string{suiclient.ObjectFieldContents})
```
mdulin2, line 164: Any chance we could add caching for these calls? At the very least, the previous object version has a good chance of already being there from a previous usage of the tx verifier.

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
 			continue
 		}
 
-		// Get the change in balance
-		balanceChange, err := resp.GetBalanceChange()
-		if err != nil {
-			logger.Error("Error getting balance difference", zap.Error(err))
+		// The decimals, token address, and token chain are immutable properties of an asset and
+		// must not change across versions. A mismatch indicates malformed or unexpected data.
+		if currentInfo.decimals != previousInfo.decimals {
```
mdulin2, line 196: Good sanity check to add.

mdulin2 (changes requested):

mdulin2:
The changes look pretty sane to me. The changes keep the core functionality but just slide in the new code. I'm glad the txverifier was super well-tested; it made this a little less scary to modify.

I'm running a scanner that cross-references the txs on Wormholescan with the newer generated ones from the watcher, to ensure there are no discrepancies. If this passes, I think we'll be in the clear.

mdulin2:
I got an error on `5dj299xhuM7q6wmB4srNQLGrw3m96HDUU3Exe7oQaJ7G` not being found by the gRPC path, but is valid on Wormholescan, and the JSON-RPC path is to be deprecated. This new watcher may change some Sui node requirements if you expect reobservations to work.

node/pkg/suiclient/suigrpc.go
johnsaigle: Yeah idk what I was thinking 🫠

node/pkg/txverifier/sui.go
djb15: Elected to remove

node/pkg/txverifier/suitypes.go
djb15: The on-chain type is technically an unbounded byte array but there is a length check to enforce that it's actually 32 bytes exactly. I was 50:50 about which way to take this but I will fix the length here

node/pkg/txverifier/suitypes.go
djb15: I personally like the explicit disambiguation between the runtimes in the txverifier so will keep these as they are

node/pkg/txverifier/suitypes.go
djb15: Ah turns out I need to keep this as is for the deserialisation because the on-chain type is prefixed by a length which is part of the deserialisation step.

node/pkg/txverifier/suitypes.go
djb15: Basically the same reason as above. The data is serialised as 32 bytes here but 33 bytes in the former case

node/pkg/txverifier/suitypes.go
djb15: It kind of applies to a bunch of these types so happy where it is

node/pkg/watchers/sui/watcher_live_test.go
djb15: Thanks, I didn't know about the `//go:build integration` pattern. I've made that change and also made it so the test skips if the RPC isn't provided. The general idea for this test wasn't to run in CI given the external dependencies so I think the left over environment variables are acceptable?

node/pkg/suiclient/suiclient.go
djb15: Good shout, added

node/pkg/txverifier/sui.go
djb15: Agree caching would be nice. I'd suggest that happens in a future PR though, the previous implementation didn't have caching either so I'd prefer to avoid the scope creep

djb15:
> I got an error on `5dj299xhuM7q6wmB4srNQLGrw3m96HDUU3Exe7oQaJ7G` not being found by the gRPC path, but is valid on Wormholescan, and the JSON-RPC path is to be deprecated. This new watcher may change some Sui node requirements if you expect reobservations to work.

Max and I had a brief chat about this earlier, it seems the new gRPC endpoints only hold 15-30 days of history by default, so any reobservation requests beyond that time frame will fail unless pointed at an archival node. I believe 15-30 days is sufficient but would appreciate thoughts from other core contributors

node/pkg/suiclient/suiclient.go
```diff
@@ -70,17 +72,45 @@ type SuiEvent struct {
...
+// digest is taken from the parent ExecutedTransaction. Events obtained via GetTransaction do
+// not need this wrapper, since the caller already knows the digest.
+type SuiTransactionEvent struct {
+	TxDigest string
+	Event    SuiEvent
+}
+
+// SuiObjectChange holds the subset of a transaction effect's changed-object
+// information needed to look up an object's state before and after a
+// transaction. It is populated from effects.changed_objects; request
```
pleasew8t:
```suggestion
// SuiObjectChange holds basic metadata of an object that changed during the
// execution of a transaction, such as it's ID, type and versions before and
// after transaction execution. It is populated from effects.changed_objects; request
```

node/pkg/suiclient/suigrpc.go
```diff
@@ -153,9 +164,23 @@ func (s *SuiGrpcClient) GetTransaction(ctx context.Context, digest string, field
 		return SuiTransaction{}, fmt.Errorf("sui gRPC GetTransaction returned nil top-level properties for digest=%s fields=%v", digest, fields)
 	}
 
+	if !transactionSucceeded(resp.Transaction) {
```
pleasew8t, line 167: I think this is a check that should be delegated to the caller, which would be consistent with how other watchers do this, and with how we are treating the gPRC client as more of a generic library to use the gRPC service (like with the field masks).

node/pkg/suiclient/suiclient.go
```diff
@@ -70,17 +72,45 @@ type SuiEvent struct {
 	BcsBytes          []byte
 }
 
+// SuiTransactionEvent pairs an event with the digest of the transaction that emitted it.
+// It is produced by the subscription, where events are delivered individually and would
+// otherwise lose their transaction context — the bare gRPC Event carries no digest, so the
+// digest is taken from the parent ExecutedTransaction. Events obtained via GetTransaction do
+// not need this wrapper, since the caller already knows the digest.
+type SuiTransactionEvent struct {
+	TxDigest string
```
pleasew8t, line 81: I think it's worth trying to maintain consistency with "TxDigest" (here) vs "Digest" (SuiTransaction). It can get quite confusing when we use different names for the same thing.

node/pkg/suiclient/suigrpc.go
```diff
@@ -168,25 +193,30 @@ func (s *SuiGrpcClient) createCheckpointStream(ctx context.Context, fields []str
...
 		return SuiSubscription{}, fmt.Errorf("sui gRPC SubscribeToTransactionEvents requires a non-nil eventWriteChannel")
 	}
 
-	// This stream is only concerned with transaction events
+	// This stream is concerned with transaction events plus the digest of the transaction
+	// that emitted them, which is paired with each event as a SuiTransactionEvent below.
+	// The execution status is fetched as well so that events from failed transactions can
+	// be dropped.
 	fields := []string{
+		"transactions.digest",
```
pleasew8t, line 217: For consistency, could these strings be defined as constants in the package?

node/pkg/watchers/sui/watcher.go
```diff
@@ -232,93 +126,62 @@ func NewWatcher(
...
 
-	// Now that we know this is a wormhole event, we can unmarshal the specifics.
-	var fields FieldsData
-	err := json.Unmarshal(*body.Fields, &fields)
+	msg, err := suiclient.DecodeBcs[txverifier.WormholeMessage](event.BcsBytes)
 	if err != nil {
-		logger.Error("failed to unmarshal FieldsData", zap.String("SuiResult.Fields", string(*body.Fields)), zap.Error(err))
 		p2p.DefaultRegistry.AddErrorCount(vaa.ChainIDSui, 1)
-		return fmt.Errorf("inspectBody failed to unmarshal FieldsData: %w", err)
+		return fmt.Errorf("processEvent failed to BCS-decode WormholeMessage: %w", err)
```
pleasew8t:
```suggestion
		return fmt.Errorf("processEvent failed to decode WormholeMessage BCS bytes: %w", err)
```

node/pkg/watchers/sui/watcher.go
```diff
@@ -40,129 +50,16 @@ type (
...
-		maximumBatchSize          int
-		descendingOrder           bool
-		loopDelay                 time.Duration
-		queryEventsCmd            string
-		postTimeout               time.Duration
+		msgChan       chan<- *common.MessagePublication
+		obsvReqC      <-chan *gossipv1.ObservationRequest
+		readinessSync readiness.Component
+
+		suiClient suiclient.SuiClient
```
pleasew8t, line 58: Suggested comment to add:

```
Note: suiclient.SuiClient is an interface. A `nil` check is therefore fine for checking initialization.
```

node/pkg/watchers/sui/watcher.go
```diff
@@ -460,11 +367,14 @@ func (e *Watcher) Run(ctx context.Context) error {
...
 			case <-timer.C:
-				height, err := e.getLatestCheckpointSN(ctx, logger)
+				checkpoint, err := client.GetLatestCheckpoint(ctx, []string{suiclient.CheckpointFieldSequenceNumber})
 				if err != nil {
 					logger.Error("Failed to get latest checkpoint sequence number", zap.Error(err))
+				} else if checkpoint.SequenceNumber == nil {
+					logger.Error("latest checkpoint response missing sequence number")
 				} else {
-					currentSuiHeight.Set(float64(height))
+					height := int64(*checkpoint.SequenceNumber) // #nosec G115 -- Sui checkpoint sequence numbers will not exceed int64 for the foreseeable future
```
pleasew8t:
```suggestion
					height := int64(*checkpoint.SequenceNumber) // #nosec G115 -- Sui checkpoint sequence numbers will not exceed math.MaxInt64 for the foreseeable future
```

node/pkg/watchers/sui/watcher.go
```diff
@@ -400,54 +292,69 @@ func (e *Watcher) Run(ctx context.Context) error {
 		zap.Bool("unsafeDevMode", e.unsafeDevMode),
 	)
 
-	// Get the node version for troubleshooting
-	e.logVersion(ctx, logger)
+	// Use an injected client (e.g. from tests) if present, otherwise create one for the
```
pleasew8t, line 296: I like the idea of an injected client. Nice and clean.

node/pkg/txverifier/suitypes.go
```diff
@@ -1,451 +1,208 @@
...
+
+// suiDynamicFieldTypeRegex matches a token-registry dynamic field object type of the form
+//
+//	0x...2::dynamic_field::Field<<pkg>::token_registry::Key<<coin>>,<pkg>::<asset>::<Asset><<coin>>>
+//
+// The framework address may be abbreviated (0x2) or fully padded (0x000...0002), and the
+// comma separating the two type arguments may or may not be followed by whitespace; the gRPC
+// API uses the padded address and no space, whereas the legacy JSON-RPC API used the short
+// form and a trailing space.
+var suiDynamicFieldTypeRegex = regexp.MustCompile(`^0x0*2::dynamic_field::Field<([^:]+)::token_registry::Key<([^>]+)>,\s*([^:]+)::([^<]+)<([^>]+)>>$`)
```
pleasew8t, line 128: Long live the regex!

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
+		// If there is an error decoding the payload, skip the event. One reason for a potential
+		// failure in decoding is that an attestation of a token was requested.
+		if err != nil {
+			continue
+		}
 
-			msgIDStr := fmt.Sprintf("%d/%s/%s", vaa.ChainIDSui, senderWithout0x, *wormholeMessage.Sequence)
-			assetKey := fmt.Sprintf(KEY_FORMAT, hdr.OriginAddress.String(), hdr.OriginChain)
+		// The sender address is prefixed with "0x", but the message ID format does not include that prefix.
+		senderWithout0x := strings.TrimPrefix(sender, "0x")
```
pleasew8t: Is there a reason for not just reusing wormholeMessage.Sender in this case? I'm almost wondering, if the gRPC API does not return the "0x" prefix, if we shouldn't just remove it from the transfer verifier's initialization?

node/pkg/txverifier/sui.go
```diff
@@ -48,172 +48,181 @@ type SuiTransferVerifier struct {
...
+				zap.Uint64("version", *objectChange.OutputVersion),
 				zap.Error(err))
 			continue
 		}
 
-		decimals, err := resp.GetDecimals()
+		previousObject, err := s.client.GetObjectAtVersion(ctx, *objectChange.ObjectID, objectChange.InputVersion, []string{suiclient.ObjectFieldContents})
 		if err != nil {
-			logger.Error("Error getting decimals", zap.Error(err))
+			logger.Error("Error getting previous object version",
```
pleasew8t:
```suggestion
			logger.Error("Error getting object at previous version",
```

pleasew8t (commented):
I really like this. Especially how the PR removes lots of code from both the transfer verifier and the watcher, which where both doing their own thing getting Sui data. Good job!

One thing to note: I ran the standalone transfer verifier a few times today, and the subscription error'd out. The errors weren't related to this code. From some searching, I think it's because I was running against the public full node that maybe disconnects subscriptions after a period of time - potentially because nothing is being sent over the subscription (likely, given low Wormhole traffic from Sui). So it's just worth ensuring the go routine in the watcher does actually restart after dying because subscription error.

node/pkg/suiclient/suigrpc.go
johnsaigle: I disagree, I'd much rather guarantee that the transaction succeeded than to make this the responsibility of every call site. I feel like that would open up very real security issues in the long-term, and the potential inconsistency here is justified.

From my understanding, the only reason the other watchers do this is because libraries like e.g. solana-go have (IMO) a weakness where they surface failed transactions in their APIs without informing the dev. (A better pattern for that might be making a safe `GetTransaction` and another function `GetTransactionUnchecked` that warns about the risks.)

We could add a code comment to note that failed tx's are skipped by this function.

djb15: There's a comment above as to the justification:

> // The execution status is always fetched in addition to the requested fields, and an error is
// returned unless the transaction carries an explicit successful status. Wormhole only ever
// cares about successful transactions, so enforcing this here guarantees callers never act on
// data from a failed (or status-less) transaction.

I understand the wish to make it a generic library, but I can't think of a reason we wouldn't want this check in the context of Wormhole

node/pkg/suiclient/suigrpc.go
djb15: These strings are only ever used once, so I'm going to leave them as they are. If reuse happens down the line the follow up PR can make them constants

node/pkg/txverifier/sui.go
djb15: No good point, I've slightly rewritten that to avoid extra trim call. But yeah I think it could make sense to remove the `0x` prefix from the initialisation but consider that scope creep on this PR

node/pkg/watchers/sui/watcher.go
```diff
@@ -232,93 +138,62 @@ func NewWatcher(
 			suiTokenBridgeEmitter = "0x" + hex.EncodeToString(sdk.KnownTokenbridgeEmitters[vaa.ChainIDSui])
 		}
 
-		// Create the Sui Api connection, and query the state object to get the token bridge address and emitter.
-		suiApiConnection := txverifier.NewSuiApiConnection(suiRPC)
+		// Create the Sui gRPC client used by the transfer verifier to query transactions and objects.
+		suiClient, err := suiclient.NewSuiGrpcClient(suiRPC, nil, suiGrpcDialOpts(unsafeDevMode)...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create Sui gRPC client for transfer verifier: %w", err)
+		}
```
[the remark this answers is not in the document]
djb15: This finding isn't really a problem because that path is only ever hit once on startup, it's not in the runnable restart logic. So you only leak a connection if you were to restart the whole Guardian which would terminate all the go processes anyway

bengtlofgren (approved):
Lgtm 💯 🚀

mdulin2 (approved):

pleasew8t (approved):

evan-gray (approved):
approval of devnet changes as codeowner.

PR #4852, opened 2026-06-10
node: close EVM watcher connector on supervisor restart

bengtlofgren:
## Summary

Fixes a guardian EVM watcher leak. The `Connector` had no `Close()`, so on supervisor restart the previous connector's `*rpc.Client` (and its dispatch goroutine) was never released, and the pending-observation map — held on the watcher value, which persists across restarts — was reused with stale entries.

## Change

- Add `Close()` to the `Connector` interface and `EthereumBaseConnector`.
- `Run()` defers `Close()` and re-initialises the pending-observation map on entry.

## Test

`TestEthereumBaseConnector_CloseReleasesGoroutines` dials and closes 25 connectors against an in-process WebSocket endpoint and asserts no goroutine leak via `goleak`. The `go.mod` change only promotes `goleak` and `gorilla/websocket` from indirect to direct — no new modules.

## Notes

Scoped to the EVM connector fix only. The Sui data-pump backoff / RPC-error-surfacing change has been dropped from this PR; it will be handled as part of the upcoming Sui SDK migration.

### DIFF
--- a/node/pkg/watchers/evm/connectors/ethereum.go
+++ b/node/pkg/watchers/evm/connectors/ethereum.go
@@ -145,11 +145,9 @@ func (e *EthereumBaseConnector) SubscribeNewHead(ctx context.Context, ch chan<- 
 	return e.client.SubscribeNewHead(ctx, ch)
 }
 
-// Close releases the JSON-RPC client, signalling its dispatch/read/write
-// goroutines to exit. Safe to call multiple times.
-func (e *EthereumBaseConnector) Close() error {
+// Close releases the JSON-RPC client and its goroutines. Safe to call multiple times.
+func (e *EthereumBaseConnector) Close() {
 	if e.rawClient != nil {
 		e.rawClient.Close()
 	}
-	return nil
 }
--- a/node/pkg/watchers/evm/connectors/ethereum.go
+++ b/node/pkg/watchers/evm/connectors/ethereum.go
@@ -144,3 +144,12 @@ func (e *EthereumBaseConnector) Client() *ethClient.Client {
 func (e *EthereumBaseConnector) SubscribeNewHead(ctx context.Context, ch chan<- *ethTypes.Header) (ethereum.Subscription, error) {
 	return e.client.SubscribeNewHead(ctx, ch)
 }
+
+// Close releases the JSON-RPC client, signalling its dispatch/read/write
+// goroutines to exit. Safe to call multiple times.
+func (e *EthereumBaseConnector) Close() error {
+	if e.rawClient != nil {
+		e.rawClient.Close()
+	}
+	return nil
+}
--- a/node/pkg/watchers/evm/watcher.go
+++ b/node/pkg/watchers/evm/watcher.go
@@ -289,6 +289,12 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 		return fmt.Errorf("failed to verify evm chain id: %w", verifyErr)
 	}
 
+	// Reset Run-scoped state: the watcher value persists across supervisor
+	// restarts, so a map left full by a prior Run accumulates without this.
+	w.pendingMu.Lock()
+	w.pending = make(map[pendingKey]*pendingMessage)
+	w.pendingMu.Unlock()
+
 	// Connect to the node using the appropriate type of connector.
 	{
 		var finalizedPollingSupported, safePollingSupported bool
@@ -300,6 +306,15 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
 			return fmt.Errorf(`failed to create connection to url "%s": %w`, w.url, err)
 		}
+
+		// Release the rpc.Client when Run returns so a supervisor restart does
+		// not strand the prior client's dispatch goroutines. Captured by value.
+		ethConnToClose := w.ethConn
+		defer func() {
+			if ethConnToClose != nil {
+				_ = ethConnToClose.Close()
+			}
+		}()
 
 		// Log the connector details for troubleshooting purposes.
 		if finalizedPollingSupported {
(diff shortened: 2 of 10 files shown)


node/pkg/watchers/evm/watcher.go
```diff
@@ -301,6 +307,15 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			return fmt.Errorf(`failed to create connection to url "%s": %w`, w.url, err)
 		}
 
+		// Release the rpc.Client when Run returns so a supervisor restart does
+		// not strand the prior client's dispatch goroutines. Captured by value.
+		ethConnToClose := w.ethConn
+		defer func() {
+			if ethConnToClose != nil {
+				_ = ethConnToClose.Close()
```
johnsaigle: As noted above, this always returns nil 
```suggestion
				ethConnToClose.Close()
```

node/pkg/watchers/evm/connectors/ethereum.go
```diff
@@ -144,3 +144,12 @@ func (e *EthereumBaseConnector) Client() *ethClient.Client {
...
 	return e.client.SubscribeNewHead(ctx, ch)
 }
+
+// Close releases the JSON-RPC client, signalling its dispatch/read/write
+// goroutines to exit. Safe to call multiple times.
+func (e *EthereumBaseConnector) Close() error {
+	if e.rawClient != nil {
+		e.rawClient.Close()
+	}
+	return nil
```
johnsaigle: It looks like this method always returns `nil` so I think you can change the signature to not have a return value.

node/pkg/watchers/evm/watcher.go
```diff
@@ -301,6 +307,15 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			return fmt.Errorf(`failed to create connection to url "%s": %w`, w.url, err)
 		}
 
+		// Release the rpc.Client when Run returns so a supervisor restart does
+		// not strand the prior client's dispatch goroutines. Captured by value.
+		ethConnToClose := w.ethConn
+		defer func() {
```
johnsaigle: Should this be a more general `Cleanup()` function rather than in-lined? That would provide you a single place to cleanup the pending message cache and the eth connection.

node/pkg/watchers/evm/watcher.go
```diff
@@ -289,6 +289,12 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 		return fmt.Errorf("failed to verify evm chain id: %w", verifyErr)
 	}
 
+	// Reset Run-scoped state: the watcher value persists across supervisor
+	// restarts, so a map left full by a prior Run accumulates without this.
+	w.pendingMu.Lock()
+	w.pending = make(map[pendingKey]*pendingMessage)
```
johnsaigle: I understand the desire to avoid a memory leak here, however this is only cleaning up some of the watcher's state. The CCQ caches and `txverifier` also have cached data that will persist across restarts, from my understanding.

I think it makes sense to attempt to clear all of these or none of them. If you really want to only clear `w.pending` then it would be helpful to note explicitly that the other data in the watcher might still leak.

node/pkg/watchers/evm/watcher.go
```diff
@@ -301,6 +307,15 @@ func (w *Watcher) Run(parentCtx context.Context) error {
 			return fmt.Errorf(`failed to create connection to url "%s": %w`, w.url, err)
 		}
 
+		// Release the rpc.Client when Run returns so a supervisor restart does
+		// not strand the prior client's dispatch goroutines. Captured by value.
```
johnsaigle: Why 'captured by value'? The type isn't using a pointer so I'm unclear why this is being specified here.

node/pkg/watchers/evm/connectors/ethereum.go
```diff
@@ -144,3 +144,12 @@ func (e *EthereumBaseConnector) Client() *ethClient.Client {
 func (e *EthereumBaseConnector) SubscribeNewHead(ctx context.Context, ch chan<- *ethTypes.Header) (ethereum.Subscription, error) {
 	return e.client.SubscribeNewHead(ctx, ch)
 }
+
+// Close releases the JSON-RPC client, signalling its dispatch/read/write
+// goroutines to exit. Safe to call multiple times.
+func (e *EthereumBaseConnector) Close() error {
```
johnsaigle: Could you add a close call to the reobservation path in `handleReobservationRequest` also?

johnsaigle (commented):
While you're here, you may want to look into testing `rpc.DialContext` as well. GPT is telling me the following. Take it or leave it -- also feel free to capture in a GitHub Issue and resolve separately. I just wanted to note it here in case the memory leaks are interfering with your work. If so, you could look into DialContext as well to help get unstuck.

```
Why Context Cancel Is Not Enough
In go-ethereum v1.10.21, rpc.DialContext says the context only controls initial connection establishment. It does not close the resulting client.
For non-HTTP clients, initClient starts dispatch/read goroutines:
if !isHTTP {
    go c.dispatch(conn)
}
dispatch starts the read loop:
go c.read(codec)
The intended shutdown path is Client.Close():
func (c *Client) Close() {
    if c.isHTTP {
        return
    }
    select {
    case c.close <- struct{}{}:
        <-c.didClose
    case <-c.didClose:
    }
}
So defer cancel() only cancels the timeout context. It does not signal c.close, does not wait on c.didClose, and does not terminate those goroutines.
Proper Fix
Close immediately after a successful dial:
c, err := rpc.DialContext(timeout, url)
if err != nil {
    return fmt.Errorf("failed to connect to endpoint: %w", err)
}
defer c.Close()
Same pattern in getFinality:
c, err := rpc.DialContext(timeout, w.url)
if err != nil {
    return false, false, fmt.Errorf("failed to connect to endpoint: %w", err)
}
defer c.Close()
That ensures every return path after successful dial closes the temporary client, including error returns from the RPC call or response parsing.
```

node/pkg/watchers/evm/watcher.go
evan-gray: I don't think we want to clear `w.pending` as that would cause dropped messages!

johnsaigle (approved):
My comments have been addressed.

djb15 (approved):

PR #4855, opened 2026-06-11
feat(whitepapers): update pauser with temporary pause

evan-gray:
This PR updates the WTT Pauser design to include a `freeze` role and change the `pause` role only have the authority to temporarily pause (for 5 days in this initial design)


whitepapers/0003_token_bridge.md
```diff
@@ -106,11 +106,19 @@ be mapped to this identifier. A wrapped asset may only ever be created once for
...
+- A `freeze` authority may call `freeze` to set `paused` to `true` and set `pauseExpiry` to the maximum representable timestamp, pausing the bridge for the maximum amount of time. A frozen bridge will not become permissionlessly unpausable in practice and can only be lifted by the `unpauser`. `freeze` is the higher-trust counterpart to the temporary, self-expiring `pauser`.
+- An `unpauser` may call `unpause` to set `paused` back to `false` and `pauseExpiry` to `0` at any time, regardless of `pauseExpiry`. This is the privileged path to lift a pause (including a `freeze`) before it would otherwise expire.
+
+Additionally, a permissionless `unpauseExpired` entry point allows anyone to set `paused` to `false` once `block.timestamp >= pauseExpiry`. This bounds a `pauser`-initiated pause to `PAUSE_DURATION` without requiring the `unpauser` to act, while ensuring the pause never lapses prematurely: the boolean `paused` remains authoritative for every other entry point's check, so a pause is only lifted by an explicit `unpause` or `unpauseExpired` call - never silently by the passage of time.
```
johnsaigle:
```suggestion
Additionally, a permissionless `unpauseExpired` entry point allows anyone to set `paused` to `false` once `block.timestamp >= pauseExpiry`. This bounds a `pauser`-initiated pause to `PAUSE_DURATION` without requiring the `unpauser` to act, while ensuring the pause never lapses prematurely: the boolean `paused` remains authoritative, so a pause is only lifted by an explicit `unpause` or `unpauseExpired` call - never silently by the passage of time.
```

nit: Not _every_ other entrypoint, just the non-governance, non-pause ones.

whitepapers/0003_token_bridge.md
```diff
@@ -106,11 +106,19 @@ be mapped to this identifier. A wrapped asset may only ever be created once for
...
+
+Any role may be left unset. A zero-length value or an all-zero address (of the target chain's native address size) is treated as the role being unassigned. When a role is unassigned, the corresponding entry point MUST revert before comparing the caller against the configured role. Implementations MUST NOT treat an all-zero address as an authorized caller. This allows governance to disable `pause`, `freeze`, or `unpause` without removing the entry point - for example, leaving `pauser` unassigned on chains where pause authority is not yet desired, or zeroing a key suspected of compromise without first provisioning its replacement. The permissionless `unpauseExpired` entry point has no associated role and is always callable. If `unpauser` is unassigned while the bridge is paused, recovery before `pauseExpiry` requires Wormhole governance to first assign a non-zero `unpauser` via `SetPauserAddresses`; otherwise the pause lifts permissionlessly once `pauseExpiry` has passed.
```
johnsaigle:
```suggestion
Any role may be left unset. A zero-length value or an all-zero address (of the target chain's native address size) is treated as the role being unassigned. When a role is unassigned, the corresponding entry point MUST revert before comparing the caller against the configured role. Implementations MUST NOT treat an all-zero address as an authorized caller. This allows governance to disable `pause`, `freeze`, or `unpause` without removing the entry point - for example, leaving `pauser` unassigned on chains where pause authority is not yet desired, or zeroing a key suspected of compromise without first provisioning its replacement. The permissionless `unpauseExpired` entry point has no associated role and is always callable. If `unpauser` is unassigned while the bridge is paused, recovery before `pauseExpiry` requires Wormhole governance to first assign a non-zero `unpauser` via `SetPauserAddresses`; otherwise the pause can be permissionlessly lifted once `pauseExpiry` has passed.
```

nit: More explicit that a call to unpause/unpauseExpired is required; it doesn't lift on its own

whitepapers/0003_token_bridge.md
```diff
@@ -106,11 +106,19 @@ be mapped to this identifier. A wrapped asset may only ever be created once for
...
+- A `pauser` may call `pause` to set `paused` to `true` and set `pauseExpiry` to `block.timestamp + PAUSE_DURATION`, where `PAUSE_DURATION` is initially a hard-coded constant of 5 days. `pause` may be called repeatedly; each call pushes `pauseExpiry` to 5 days from the current time, so the bridge stays paused for as long as the `pauser` keeps re-pausing. A `pause` call MUST NOT reduce a `pauseExpiry` already further in the future (e.g. one set by `freeze`) - a lower-trust `pauser` cannot curtail a `freeze`. This requirement is why unpausing MUST reset `pauseExpiry` to `0`: a stale expiry left over from a prior `freeze` would otherwise block a later `pause`.
```
kcsongor: alternatively unpause could also set `pauseExpiry` to `block.timestamp` (current), that way there would be on-chain evidence of the last unpause

whitepapers/0003_token_bridge.md
```diff
@@ -154,13 +164,18 @@ a `TransferWithPayload`. Amount in the tokens native decimals. `payload` is an a
 
 `submitRecoverChainId(Message recoverChainId)` - Execute a `RecoverChainId` governance message. Only callable on a forked chain (EVM-only).
 
-`pause()` - Set `paused` to `true`. Callable only by the `pauser`; reverts when `pauser` is unassigned.
+`pause()` - Set `paused` to `true` and `pauseExpiry` to `block.timestamp + PAUSE_DURATION` (5 days), never reducing a `pauseExpiry` already further in the future. Callable only by the `pauser`; reverts when `pauser` is unassigned.
```
johnsaigle: You could mention that freeze, unpause, and unpauseExpired are idempotent, but pause is not. The main benefit would be guiding reviews and test generation, so it may not be necessary long-term.

whitepapers/0003_token_bridge.md
kcsongor: I'm curious why require this explicit step and not just do it implicitly when the pause time expires?

whitepapers/0003_token_bridge.md
```diff
@@ -290,8 +305,14 @@ ChainId uint16
...
 PauserLen uint8
-// Address authorized to pause the bridge
+// Address authorized to temporarily pause the bridge (for PAUSE_DURATION)
 Pauser [PauserLen]uint8
+// Length of the freeze address. Must equal the target chain's native
+// address size, or 0 to leave the role unassigned. An all-zero address
+// of the native size is also treated as unassigned.
+FreezeLen uint8
+// Address authorized to pause the bridge for the maximum duration
+Freeze [FreezeLen]uint8
```
johnsaigle: nit: Freezer instead? I like either pause/freeze/unpause or pauser/freezer/unpauser but mixing them feels off

whitepapers/0003_token_bridge.md
```diff
@@ -154,13 +164,18 @@ a `TransferWithPayload`. Amount in the tokens native decimals. `payload` is an a
 
 `submitRecoverChainId(Message recoverChainId)` - Execute a `RecoverChainId` governance message. Only callable on a forked chain (EVM-only).
 
-`pause()` - Set `paused` to `true`. Callable only by the `pauser`; reverts when `pauser` is unassigned.
+`pause()` - Set `paused` to `true` and `pauseExpiry` to `block.timestamp + PAUSE_DURATION` (5 days), never reducing a `pauseExpiry` already further in the future. Callable only by the `pauser`; reverts when `pauser` is unassigned.
+
+`freeze()` - Set `paused` to `true` and `pauseExpiry` to the maximum representable timestamp. Callable only by the `freeze` authority; reverts when `freeze` is unassigned.
```
johnsaigle:
```suggestion
`freeze()` - Set `paused` to `true` and `pauseExpiry` to the maximum representable timestamp. Callable only by the `freeze` role; reverts when `freeze` is unassigned.
```
Reads as a little Solana-specific and overloads conceptually with a token mint's freeze authority

whitepapers/0003_token_bridge.md
kcsongor: I see this is explained below. I'm not sure the overhead is significant (no extra storage slot reads if we structure things right), but I guess it's fine

whitepapers/0003_token_bridge.md
```diff
@@ -106,11 +106,19 @@ be mapped to this identifier. A wrapped asset may only ever be created once for
...
+- A `pauser` may call `pause` to set `paused` to `true` and set `pauseExpiry` to `block.timestamp + PAUSE_DURATION`, where `PAUSE_DURATION` is initially a hard-coded constant of 5 days. `pause` may be called repeatedly; each call pushes `pauseExpiry` to 5 days from the current time, so the bridge stays paused for as long as the `pauser` keeps re-pausing. A `pause` call MUST NOT reduce a `pauseExpiry` already further in the future (e.g. one set by `freeze`) - a lower-trust `pauser` cannot curtail a `freeze`. This requirement is why unpausing MUST reset `pauseExpiry` to `0`: a stale expiry left over from a prior `freeze` would otherwise block a later `pause`.
+- A `freeze` authority may call `freeze` to set `paused` to `true` and set `pauseExpiry` to the maximum representable timestamp, pausing the bridge for the maximum amount of time. A frozen bridge will not become permissionlessly unpausable in practice and can only be lifted by the `unpauser`. `freeze` is the higher-trust counterpart to the temporary, self-expiring `pauser`.
```
johnsaigle:
```suggestion
- A `freeze` role may call `freeze` to set `paused` to `true` and set `pauseExpiry` to the maximum representable timestamp, pausing the bridge for the maximum amount of time. A frozen bridge will not become permissionlessly unpausable in practice and can only be lifted by the `unpauser`. `freeze` is the higher-trust counterpart to the temporary, self-expiring `pauser`. A `freeze` call from the `freeze` role always succeeds: if the bridge is unpaused, `freeze` causes a pause; if paused, a `freeze` effectively extends an existing `pause` to the maximum duration. Successive calls to `freeze()` are a no-op.
```

- `authority` reads as a little Solana-specific and overloads conceptually with a token mint's freeze authority
- makes explicit how freeze works as an override for any existing paused state

whitepapers/0003_token_bridge.md
```diff
@@ -335,9 +359,12 @@ However, deploying pause support updates existing contract state. For example, t
 
 Rather than a single boolean `paused` state, the bridge could expose finer-grained pauses (e.g. inbound vs. outbound) so that, for example, redemptions can continue while new deposits are blocked. We are not taking this approach now in favor of a simple boolean; granular pause can be added later without breaking the pause/unpause governance defined here.
 
-### Temporary pause
+### Pause state as a pure timestamp
```
johnsaigle, line 362: We could add a section here for "Unpausing is always permissioned" to explain why we have the permissionless `unpauseExpired`, if you like.

johnsaigle (commented):
Largely looks good, left some nits that you can feel free to apply or dismiss as you like.

whitepapers/0003_token_bridge.md
evan-gray: because the explicit step allows an event to be emitted, which seems preferable.

whitepapers/0003_token_bridge.md
evan-gray: good point, I could go for that

whitepapers/0003_token_bridge.md
evan-gray: 🥶

mdulin2 (dismissed):

bemic (dismissed):

whitepapers/0003_token_bridge.md
evan-gray: perhaps this comment will be enough to say because this approach was recommended instead 😬 I'm not sure I have additional color to add aside from there's a permissionless unpause because some thought it best to have a permissionless unpause (and a separate `freeze` authority to less temporarily pause)

whitepapers/0003_token_bridge.md
```diff
@@ -154,13 +164,18 @@ a `TransferWithPayload`. Amount in the tokens native decimals. `payload` is an a
...
 
-`pause()` - Set `paused` to `true`. Callable only by the `pauser`; reverts when `pauser` is unassigned.
+`pause()` - Set `paused` to `true` and `pauseExpiry` to `block.timestamp + PAUSE_DURATION` (5 days), never reducing a `pauseExpiry` already further in the future. Callable only by the `pauser`; reverts when `pauser` is unassigned. Not idempotent: each call pushes `pauseExpiry` forward.
+
+`freeze()` - Set `paused` to `true` and `pauseExpiry` to the maximum representable timestamp. Callable only by the `freezer`; reverts when `freezer` is unassigned. Idempotent.
 
-`unpause()` - Set `paused` to `false`. Callable only by the `unpauser`; reverts when `unpauser` is unassigned.
+`unpause()` - Set `paused` to `false` and `pauseExpiry` to `block.timestamp`. Callable only by the `unpauser`; reverts when `unpauser` is unassigned.
+
+`unpauseExpired()` - Set `paused` to `false` and `pauseExpiry` to `block.timestamp`. Permissionless; reverts unless `block.timestamp >= pauseExpiry`.
```
bemic: Should we add "reverts when `paused != true`" ?
It can be called when the bridge was never paused and emitting misleading events

johnsaigle (dismissed):

whitepapers/0003_token_bridge.md
johnsaigle: ➕ Yeah it would be annoying if someone overwrote the timestamp.

mdulin2 (approved):

bemic (approved):

PR #4859, opened 2026-06-12
cleanup: remove CGO

johnsaigle:
CGO was originally required in order to support Celo. However, their
architecture has changed substantially and so we shouldn't need CGO
support anymore.

See also `56cf37a7ecee3de2d8eae0456c4360f3023922e9` for prior work.

### DIFF
--- a/Dockerfile.proto
+++ b/Dockerfile.proto
@@ -7,7 +7,7 @@ COPY tools/build.sh /app/tools/
 COPY tools/go.* /app/tools/
 
 RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
-	cd /app/tools && CGO_ENABLED=0 ./build.sh
+	cd /app/tools && ./build.sh
 
 # syntax=docker.io/docker/dockerfile:1.3@sha256:42399d4635eddd7a9b8a24be879d2f9a930d0ed040a61324cfdf59ef1357b3b2
 FROM docker.io/golang:1.25.10-bookworm@sha256:154bd7001b6eb339e88c964442c0ad6ed5e53f09844cc818a41ce4ecb3ce3b43 AS go-build
--- a/Dockerfile.proto
+++ b/Dockerfile.proto
@@ -6,8 +6,12 @@ RUN mkdir /app
 COPY tools/build.sh /app/tools/
 COPY tools/go.* /app/tools/
 
+# Build these tools as static binaries. They are compiled in this Debian-based
+# image but later copied into Alpine (below in the `node-build` step), which 
+# does not provide Debian's glibc dynamic loader. Without CGO_ENABLED=0, tools 
+# such as buf can be linked against glibc and fail in Alpine with a misleading "not found" error.
 RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
-	cd /app/tools && ./build.sh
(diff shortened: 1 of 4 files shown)


mdulin2:
If this builds, I think it's fine. It appears to build with the `cgo` things removed.

mdulin2:
I think this is still required. By default (on my system at least), `CGO_ENABLED` is set to `1`. So, when you remove the environment variable, you're just using the system default. 

If I explicitly set the env variable to 0, then I get the following error when building:

```
make node
cd tools && ./build.sh
rm -rf bridge
rm -rf node/pkg/proto
tools/bin/buf generate
cd node && CGO_ENABLED=0 go build -ldflags "-X github.com/certusone/wormhole/node/pkg/version.version=v2.63.0-32-g48258bc67-dirty" \
          -mod=readonly -o ../build/bin/guardiand \
          github.com/certusone/wormhole/node
# github.com/ethereum/go-ethereum/crypto
../../../../go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.21/crypto/signature_nocgo.go:85:14: assignment mismatch: 2 variables but btc_ecdsa.SignCompact returns 1 value
# github.com/CosmWasm/wasmvm/internal/api
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:195:2: undefined: GasMeter
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:196:20: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:200:11: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:201:11: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:205:28: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:212:38: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:216:32: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:230:42: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:329:7: undefined: KVStore
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mock_failure.go:15:27: undefined: GoAPI
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:329:7: too many errors
```

johnsaigle:
Max is right, CGO is still required based on the above.

In theory, I think this could be resolved if we were to upgrade go-ethereum + it's btc library dependency. This would allow us to avoid CGO in that location. The CosmWasm aspect could still use CGO but it might allow us to get rid of the CGO requirement in the rest of the repo.

However, this PR was intended as a quick cleanup with the assumption that disabling this flag would be clean. It's not a priority right not to refactor just for the purpose of getting rid of CGO.

PR #4861, opened 2026-06-16
cleanup: remove old encryption package

johnsaigle:


mdulin2:
I saw this before; this is just dead code. If we ever needed it in the future, we could just look at the git history. Good change!

mdulin2 (approved):

bemic (approved):

PR #4867, opened 2026-06-17
docs: Add guidelines for writing tests

johnsaigle:
This update captures some unspecified practices that in practice we're already doing. The goal here is to explain some of the practices to new devs in a shared doc instead of repeatedly during PR comment threads, as well as to guide LLMs to write tests more in-line with our desired style.

See also @SEJeff 's comment here: https://github.com/wormhole-foundation/wormhole/pull/4818#discussion_r3421615568

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -245,6 +245,8 @@ For Go specifically:
   - So, require should be used carefully for setting up e.g. a mock or a test database that is required for the test to run. Asserts should be used for checking actual properties and business logic.
 - Use `t.Parallel()` for tests that can safely be run in parallel.
 - Use `testing.Short()` to gate  tests that are more integration-like or that otherwise are slow.
+- Use `t.Helper()` for any utility/helper functions that are not themselves explicitly tests that you do not want to see in the error traceback.
+
 
 ### Unit tests anti-patterns
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -230,6 +230,27 @@ Places to find out more about existing test coverage and how to run those tests:
 
 The best place to understand how we invoke these tests via GitHub Actions on every commit can be found via `./.github/workflows/*.yml` and the best place to observe the results of these builds can be found via [https://github.com/wormhole-foundation/wormhole/actions](https://github.com/wormhole-foundation/wormhole/actions).
 
+### Writing Tests
+
+Tests should be written to be extensible and to cover as much of the expected behavior as possible, especially failure cases.
+Tests added to the repo are expected to prove their error cases, not just the happy path.
+
+Where possible, real example data should be used as the test input rather than hand-crafted data. For example,
+a real RPC response with on-chain can be saved in-line and tested against a parser.
+
+For Go specifically:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -230,6 +230,27 @@ Places to find out more about existing test coverage and how to run those tests:
...
+Where possible, real example data should be used as the test input rather than hand-crafted data. For example,
+a real RPC response with on-chain can be saved in-line and tested against a parser.
+
+For Go specifically:
+- Use [table-driven tests](https://go.dev/wiki/TableDrivenTests#example-of-a-table-driven-test) to test multiple inputs in an extensible way.
+- Be intentional when using require vs assert. 
+  - Require halts the test and hides all subsequent logging, whereas multiple failing asserts will be logged.
+  - So, require should be used carefully for setting up e.g. a mock or a test database that is required for the test to run. Asserts should be used for checking actual properties and business logic.
+- Use `t.Parallel()` for tests that can safely be run in parallel.
+- Use `testing.Short()` to gate  tests that are more integration-like or that otherwise are slow.
```
SEJeff, line 247:
```suggestion
- Use `testing.Short()` to gate  tests that are more integration-like or that otherwise are slow.
- Use `t.Helper()` for any utility/helper functions that are not themselves explicitly tests that you do not want to see in the error traceback.

```

SEJeff (dismissed):

SEJeff (dismissed):

djb15 (approved):

CONTRIBUTING.md
```diff
@@ -230,6 +230,29 @@ Places to find out more about existing test coverage and how to run those tests:
...
+- Use `t.Helper()` for any utility/helper functions that are not themselves explicitly tests that you do not want to see in the error traceback.
+
+
+### Unit tests anti-patterns
+
+- Hard-coded sleeps
+- Live, external network calls
```
pleasew8t, line 254: I wonder if it isn't worth expanding on this point, though. The "live" tests we built into the Sui client and watcher are a really nice alternative to writing a standalone utility to see the magic happen. So I wouldn't consider it an anti-pattern.

bemic (approved):

CONTRIBUTING.md
pleasew8t: A unit test shouldn't be making these live external network calls, I agree. But maybe a short section on writing "live" tests could be useful. Similarly, it's maybe worth including a bit about fuzz tests where appropriate? I don't think we need full blown guidance on fuzz testing, but it could talk about basic fuzzing and how, in some cases, the corpus could act much like the table-driven testing you mention earlier.

johnsaigle: I do feel that it's correct to say that live calls during a unit test is an anti-pattern. I wouldn't consider these "live" tests to be unit tests; they're integration tests. 

I think we could separately add a section on Integration Testing and Fuzzing if we want but I think both could be quite involved especially if the idea is to establish a pattern around how we do this -- currently we have a mix of TS-based tests in Tilt, pure Go tests gated beyond flags under `go test`, and standalone CLI utilities (i.e. transfer verifier) so I'm not sure we have even an informal standard around this yet.

Given that this PR has the necessary approvals, I'll merge it now but if you want to write a follow-up that adds new sections for other types of testing, I'm happy to review it.

PR #4873, opened 2026-06-18
feat(evm): wtt temporary pause + freeze role

douglasgalico:


ethereum/contracts/bridge/Bridge.sol
```diff
@@ -95,36 +115,99 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
     ///      VAA — letting whoever holds the pauser key shut the bridge down on the fork
     ///      immediately, before chain-id recovery completes.
     function pause() external {
-        address p = pauser();
-        if (p == address(0)) revert NotPauser();
-        if (msg.sender != p) revert NotPauser();
+        _requireRole(pauser(), NotPauser.selector);
+        uint64 newExpiry = uint64(block.timestamp) + PAUSE_DURATION;
+        // Never reduce an expiry already further out (e.g. one set by `freeze`).
+        if (newExpiry > pauseExpiry()) {
```
mdulin2: Should we revert in this case instead? I feel like succeeding here gives a false sense of success to the caller; they didn't do anything on their `pause()` call if the current expiry is larger than the new one.

ethereum/contracts/bridge/BridgeGovernance.sol
```diff
@@ -72,15 +72,22 @@ contract BridgeGovernance is BridgeGetters, BridgeSetters, ERC1967Upgrade {
...
+        address indexed freezer,
+        address indexed unpauser
+    );
+
+    /// @notice Set the pauser, freezer, and unpauser addresses via a `SetPauserAddresses`
+    ///         (action 4) governance VAA.
     /// @dev Payload layout:
     ///        module(32) | action(1)=4 | chainId(2)
     ///      | pauserLen(1) | pauser[pauserLen]
+    ///      | freezerLen(1) | freezer[freezerLen]
```
mdulin2, line 90: Do we need to update the go sdk, or node software for this change?

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -95,36 +115,99 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+    uint64 constant PAUSE_DURATION = 5 days;
+
+    /// @dev Authorize `msg.sender` against a configured role address, reverting `err` (a 4-byte
+    ///      custom-error selector) if the role is unassigned (all-zero) or the caller does not
+    ///      match. The unassigned check is first so an all-zero role is never authorized. Shared by
+    ///      pause/freeze/unpause and reverts via assembly so the three callsites collapse to one
+    ///      body — keeping `BridgeImplementation` under the EIP-170 limit.
+    function _requireRole(address role, bytes4 err) internal view {
+        if (role == address(0) || msg.sender != role) {
+            assembly {
```
mdulin2, line 136: We don't call pause and unpause very often. So, is the assembly optimization necessary? It seems like we're adding complexity without a hard gas requirement.

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -95,36 +115,99 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+    /// @dev The pauser is configured via the `SetPauserAddresses` (action 4) governance VAA and may
+    ///      be left unassigned; when unassigned this entry point reverts before comparing
+    ///      `msg.sender`, so an all-zero `pauser` is never authorized.
     /// @dev Intentionally does not check `isFork()`. On a forked chain the pre-fork pauser can
     ///      still pause the bridge without first waiting for a `submitRecoverChainId` governance
     ///      VAA — letting whoever holds the pauser key shut the bridge down on the fork
     ///      immediately, before chain-id recovery completes.
     function pause() external {
-        address p = pauser();
-        if (p == address(0)) revert NotPauser();
```
mdulin2, line 108: This check feels redundant because of the `msg.sender` check directly after it. I think it's fine to remove.

ethereum/contracts/bridge/Bridge.sol
mdulin2: You can't pass an error to call as the second parameter. So, I see three options: 
- Keep as is. 
- Use a generic error message here with a simple `revert pauserError()`
- Perform role check here, return a boolean, and revert at the callsite. 

None of these are ideal imo.

mdulin2 (changes requested):

ethereum/contracts/bridge/Bridge.sol
douglasgalico: Not related to gas optimization, it's the bytecode size limit which BridgeImplementation is close to it. Doing that I prevent inlining the logix 3x while keeping distinct per role errors. I'd keep as it is to preserve the distinct errors and keeping under the size limit

ethereum/contracts/bridge/Bridge.sol
douglasgalico: That's by design and also happens in svm/sui (whitepaper 0003).  This branch is only hit when the bridge is already frozen (the only expiry > now + 5 days). The guard make sure the spec rule that a pauser can't reduce a freeze's expiry, then reverting would let a pauser detect/weaken a freeze

ethereum/contracts/bridge/BridgeGovernance.sol
douglasgalico: Yes! That's in a separated PR

ethereum/contracts/bridge/Bridge.sol
douglasgalico: Good catch, this was already refactored. Now _requireRole combine both checks and follows the whitepaper recomendation that unassigned role MUST revert before comparing msg.sender (like an allzero is never authorized), so we short circuit this zero check first

ethereum/contracts/bridge/Bridge.sol
johnsaigle: > reverting would let a pauser detect/weaken a freeze

Can you explain this part? I'm unclear on 'detect' here because the bridge being frozen is not a secret. 

I think that Max is suggesting that, if the `newExpiry < pauseExpiry()` we should revert instead of calling `setPaused()` and emitting an event below this block.

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -95,36 +115,99 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
+        emit Paused(msg.sender, pauseExpiry());
     }
 
-    /// @notice Unpause the bridge. Only callable by the configured unpauser. The unpauser is
-    ///         configured via the `SetPauserAddresses` (action 4) governance VAA and may be left
-    ///         unassigned; when unassigned this entry point reverts before comparing `msg.sender`,
-    ///         so an all-zero `unpauser` is never authorized. Note that `unpause` is intentionally
-    ///         exempt from the `notPaused` modifier so it remains callable while paused.
+    /// @notice Freeze the bridge for the maximum duration. Only callable by the configured freezer.
+    ///         Sets `paused` and `pauseExpiry` to the maximum timestamp. The higher-trust
```
johnsaigle:
```suggestion
    ///         Sets `paused` to `true` and `pauseExpiry` to the maximum timestamp. The higher-trust
```

suggestion: this wording is a bit more clear IMO

ethereum/contracts/bridge/Bridge.sol
```diff
@@ -32,12 +32,23 @@ contract Bridge is BridgeGovernance, ReentrancyGuard {
...
-    /// @notice Emitted when the bridge is unpaused.
+    /// @notice Emitted when the bridge is temporarily paused via `pause`.
+    /// @param by The pauser that paused the bridge.
+    /// @param pauseExpiry The timestamp (unix seconds) at which the pause becomes permissionlessly
+    ///        liftable via `unpauseExpired`.
+    event Paused(address indexed by, uint256 pauseExpiry);
+
+    /// @notice Emitted when the bridge is frozen via `freeze`.
+    /// @param by The freezer that froze the bridge.
+    /// @param pauseExpiry The pause expiry, set to the maximum timestamp by `freeze`.
```
johnsaigle, line 43: Since `pauseExpiry` is always a constant here, and we're basically just using a max value to mean "pause forever", I'm not sure there's much value in emitting the pauseExpiry value

johnsaigle (dismissed):
Left some non-blocking suggestions. Otherwise, looks good.

ethereum/contracts/bridge/Bridge.sol
mdulin2: Yeah, exactly. If this is intentional, I'm happy to move on though.

mdulin2 (dismissed):

ethereum/contracts/bridge/Bridge.sol
douglasgalico: True, ignore the detect part, bad working and indeed frozen isnt a secret. For the else-revert I'd keep it, the whitepaper says the only revert for pause() is unassigned pauser ,so the noop is on purpose. The intent is to make sure it's paused, if it's frozen state then that's already true, so succeeding is fine (being idempotent as freeze() ). And the event emits the real pauseExpiry() not the newExpiry (so it not misleading) - Same applies for SUI/SVM

evan-gray: I intended the MUST NOT here to mean revert. https://github.com/wormhole-foundation/wormhole/blob/5a21b21d54e5aa8e4779914beb86158749c23a85/whitepapers/0003_token_bridge.md?plain=1#L113

mdulin2 (approved):

johnsaigle (approved):

kcsongor (approved):

PR #4874, opened 2026-06-18
feat: add freezer role and timed pause/unpause-expired to Solana wtt

douglasgalico:


solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
-        /* paused */ true,
-    )?;
+    let config_info = accs.config.info();
+    require_role(config_info, accs.pauser.key, Role::Pauser)?;
+
+    let new_expiry = accs.clock.unix_timestamp.saturating_add(PAUSE_DURATION);
+    {
+        let mut data = config_info.data.borrow_mut();
+        // Never reduce an expiry already further out (e.g. one set by `freeze`).
+        if new_expiry > pause_expiry(&data) {
```
mdulin2: Same comment as EVM PR. I'd expect this to fail.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+
+#[derive(FromAccounts)]
+pub struct Freeze<'b> {
+    /// Caller must equal the configured freezer stored in the Config tail.
+    pub freezer: Signer<AccountInfo<'b>>,
+
+    pub config: Mut<ConfigAccount<'b, { AccountState::Initialized }>>,
+
+    /// Event authority PDA for Anchor CPI event signing.
+    pub event_authority: Info<'b>,
```
mdulin2, line 114: The `emit_event_cpi` has the `event_authority` PDA verification. So, this seems safe.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+
+    /// This program (for self-CPI).
+    pub self_program: Info<'b>,
```
mdulin2, line 117: The `emit_event_cpi` effectively checks whether this is correct. I appreciate the extra check here; better safe than sorry.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -96,36 +198,108 @@ pub fn unpause(ctx: &ExecutionContext, accs: &mut Unpause, _data: UnpauseData) -
...
+pub fn unpause_expired(
+    ctx: &ExecutionContext,
+    accs: &mut UnpauseExpired,
+    _data: UnpauseExpiredData,
 ) -> Result<()> {
-    // Reject if the Config account hasn't been migrated yet (legacy 32-byte layout).
+    let config_info = accs.config.info();
+
+    // Migration check: a legacy (un-extended) Config can never be paused, but guard explicitly.
     if config_info.data_len() < CONFIG_WITH_PAUSER_LEN {
```
mdulin2, line 237: Is there a reason this exists HERE but not on the other functions?

mdulin2 (changes requested):

solana/modules/token_bridge/program/src/api/pause.rs
mdulin2: It'd probably be good to include this check on the other three functions as well.

mdulin2:
Solana tests of the account migration seem pretty solid. Thanks for testing that out!

solana/modules/token_bridge/program/src/api/pause.rs
douglasgalico: Same as evm, by design. The guard stops a lower trust pauser from reducing a freeze expiry, failing would leak/weaken the freeze, and event emits the actual expiry at the end

solana/modules/token_bridge/program/src/api/pause.rs
douglasgalico: Yep! They actually happen inside `require_role` for pause/freeze/unpause. unpause_expired is permissionless so it never calls require_role, that's why we need a copy of the migration guard. So all four paths cover it

solana/modules/token_bridge/program/src/api/governance.rs
```diff
@@ -214,10 +214,11 @@ pub fn set_pauser_addresses(
     verify_governance(&accs.vaa)?;
     claim::consume(ctx, accs.payer.key, &mut accs.claim, &accs.vaa)?;
 
-    // One-time migration: legacy 32-byte Config accounts grow to 97 bytes the first time this
-    // VAA is processed. Subsequent VAAs reuse the already-extended layout. The `paused` flag
-    // is preserved across updates — rotating the pauser/unpauser keys does not implicitly
-    // unpause the bridge.
+    // One-time migration: the Config account is created by the bridge `initialize` at its
+    // original 32-byte size; the first time this VAA is processed it grows straight to the full
+    // CONFIG_WITH_PAUSER_LEN (137) bytes. Subsequent VAAs reuse the already-extended layout. The
```
johnsaigle: suggestion
```suggestion
    // [`CONFIG_WITH_PAUSER_LEN`] bytes. Subsequent VAAs reuse the already-extended layout. The
```

suggestion: intra-doc links are more clear than in-lining constants that might become stale. This generally applies to a lot of the code in this PR but I'll avoid leaving additional comments on this topic.
https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

solana/modules/token_bridge/program/src/api/pause.rs
johnsaigle: I think the same confusion is happening here as the other PR. The failure would happen in an `else` branch here. If the new_expiry < pause_expiry, we should revert rather than call `write_paused` and emit an event.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
         &accs.event_authority,
         &PAUSED_EVENT_DISCRIMINATOR,
-        &accs.pauser.key.to_bytes(),
+        &payload,
+    )
+}
+
+#[derive(FromAccounts)]
+pub struct Freeze<'b> {
+    /// Caller must equal the configured freezer stored in the Config tail.
```
johnsaigle:
```suggestion
    /// Caller must equal the freezer stored in Config.
```

suggestion: not sure that being precise about the storage layout is helpful here

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+
+    /// This program (for self-CPI).
+    pub self_program: Info<'b>,
+}
+
+#[derive(BorshDeserialize, BorshSerialize, Default)]
+pub struct FreezeData {}
+
+/// Freeze the bridge for the maximum duration. Only callable by the configured freezer. Sets
+/// `paused` and `pause_expiry` to the maximum timestamp. The higher-trust counterpart to the
```
johnsaigle:
```suggestion
/// `paused` to `true` and `pause_expiry` to the maximum timestamp. The higher-trust counterpart to the
```

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+    pub event_authority: Info<'b>,
+
+    /// This program (for self-CPI).
+    pub self_program: Info<'b>,
+}
+
+#[derive(BorshDeserialize, BorshSerialize, Default)]
+pub struct FreezeData {}
+
+/// Freeze the bridge for the maximum duration. Only callable by the configured freezer. Sets
```
johnsaigle:
```suggestion
/// Freeze the bridge. Only callable by the configured freezer. Sets
```

suggestion: The constant is mentioned a few times later in this comment block

solana/modules/token_bridge/program/src/api/pause.rs
johnsaigle: suggestion: a bit more clear IMO

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+/// `paused` and `pause_expiry` to the maximum timestamp. The higher-trust counterpart to the
+/// temporary, self-expiring `pause`: a frozen bridge will not become permissionlessly unpausable
```
johnsaigle:
```suggestion
/// temporary `pause`: a frozen bridge will not become permissionlessly unpausable
```

suggestion: redundant with "temporary"

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -43,21 +69,82 @@ pub struct Pause<'b> {
...
+    let config_info = accs.config.info();
+    require_role(config_info, accs.freezer.key, Role::Freezer)?;
+
+    {
+        let mut data = config_info.data.borrow_mut();
+        write_pause_expiry(&mut data, i64::MAX);
+        write_paused(&mut data, true);
+    }
+
+    if accs.self_program.key != ctx.program_id {
```
johnsaigle: suggestion: it's a bit nicer to do error handling at the top of the function, then do the happy path state changes later, instead of mixing the writes with the errors. In practice it doesn't matter because transactions in solana are all-or-nothing

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -96,36 +198,108 @@ pub fn unpause(ctx: &ExecutionContext, accs: &mut Unpause, _data: UnpauseData) -
...
+    accs: &mut UnpauseExpired,
+    _data: UnpauseExpiredData,
 ) -> Result<()> {
-    // Reject if the Config account hasn't been migrated yet (legacy 32-byte layout).
+    let config_info = accs.config.info();
+
+    // Migration check: a legacy (un-extended) Config can never be paused, but guard explicitly.
     if config_info.data_len() < CONFIG_WITH_PAUSER_LEN {
-        return Err(PauserNotConfigured.into());
+        return Err(NotPaused.into());
```
johnsaigle: I think this should stay as the previous error `PauserNotConfigured` since it would only trigger for legacy accounts

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -96,36 +198,108 @@ pub fn unpause(ctx: &ExecutionContext, accs: &mut Unpause, _data: UnpauseData) -
...
-        if &configured != signer {
-            return Err(InvalidPauser.into());
+        if accs.clock.unix_timestamp < pause_expiry(&data) {
+            return Err(NotExpired.into());
         }
+        write_pause_expiry(&mut data, accs.clock.unix_timestamp);
+        write_paused(&mut data, false);
+    }
+
+    if accs.self_program.key != ctx.program_id {
```
johnsaigle: suggestion: same comment as the other file. It's a bit nicer to group error handling together at the beginning of the program rather than mix it in with potential state changes.

solana/modules/token_bridge/program/src/api/pause.rs
```diff
@@ -96,36 +198,108 @@ pub fn unpause(ctx: &ExecutionContext, accs: &mut Unpause, _data: UnpauseData) -
...
-    write_paused(&mut data, paused);
+/// Verify `signer` is the configured holder of `role`. Reverts before comparing the caller if the
+/// role is unassigned (zero pubkey) or the Config account has not been migrated — so an all-zero
+/// role is never authorized (whitepapers/0003_token_bridge.md Pausing). Reads only; writes happen
+/// in the caller after the borrow is released.
+fn require_role(config_info: &AccountInfo, signer: &Pubkey, role: Role) -> Result<()> {
+    // Reject if the Config account hasn't been migrated yet (legacy layout → role unassigned).
+    if config_info.data_len() < CONFIG_WITH_PAUSER_LEN {
+        return match role {
+            Role::Freezer => Err(NotFreezer.into()),
```
johnsaigle: I think `PauserNotConfigured` is more appropriate here and below

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -110,6 +116,15 @@ pub enum TokenBridgeError {
     /// `self_program` account passed by the caller does not match this program's `program_id`.
     /// Only used to defensively guard the self-CPI that emits Anchor-style events.
     InvalidSelfProgram,
+    // NOTE: new variants are APPENDED below — `From<TokenBridgeError>` maps each variant to its
+    // numeric position (`t as u64`), so inserting in the middle would renumber existing error
+    // codes. Always append.
+    /// Caller of `freeze` is not the configured freezer (or the freezer role is unassigned).
+    NotFreezer,
```
johnsaigle:
```suggestion
    InvalidFreezer,
```

This would be more consistent with the existing error codes. I think that "freezer role is unassigned" should be covered instead by `PauserNotConfigured`. You could consider renaming that error to `PausingNotConfigured` as it usually means that the bridge is in legacy mode and hasn't upgraded yet. Alternatively, maybe there's an argument for an error code `RoleNotConfigured` which could be used for all 3 roles.

As written now, I think the errors are inconsistent.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -1606,23 +1768,27 @@ async fn pause_is_idempotent() {
     .unwrap();
     assert!(read_paused(&fetch_config_data(&mut context).await));
 
-    // Second pause on an already-paused bridge: succeeds, state unchanged.
+    // Second pause on an already-paused bridge: succeeds, stays paused.
```
johnsaigle: Maybe this should also assert that pauseExpiry has increased

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -1606,23 +1768,27 @@ async fn pause_is_idempotent() {
...
-    .expect("pause on already-paused bridge must be a no-op success");
+    .expect("pause on already-paused bridge must succeed");
     assert!(read_paused(&fetch_config_data(&mut context).await));
 }
 
 #[tokio::test]
-async fn unpause_is_idempotent() {
+async fn unpause_reverts_when_not_paused() {
+    // Per whitepaper 0003, `unpause` is NOT a no-op on an unpaused bridge — it reverts with
+    // `NotPaused`. (This is the behavior change from the pre-expiry design, matching EVM/SUI:
```
johnsaigle: suggestion: Delete the `pre-expiry design` part. LLMs have a tendency to comment code as if they're changelogs, but this isn't relevant to add in the source code given the pre-expiry design was never deployed.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -1654,23 +1820,501 @@ async fn set_pauser_addresses_is_idempotent() {
...
+    let second = fetch_config_data(&mut context).await;
+    assert_eq!(
+        second, first,
+        "a repeated freeze must not change Config bytes",
+    );
+}
+
+#[tokio::test]
+async fn pause_does_not_curtail_freeze() {
+    // A `pause` must never reduce a `pause_expiry` that is already further in the future. After a
```
johnsaigle, line 2013: There's an argument for making this case return an error in the contract rather than a no-op + event emission. If that is changed, this test should get updated too.

solana/modules/token_bridge/program/tests/integration.rs
```diff
@@ -1654,23 +1820,501 @@ async fn set_pauser_addresses_is_idempotent() {
...
+    .unwrap();
+    common::unpause_expired(
+        &mut context.client,
+        context.token_bridge,
+        &stranger,
+        &context.payer,
+    )
+    .await
+    .expect_err("unpause_expired before expiry must revert with NotExpired");
+    assert!(
```
johnsaigle, line 2183: The pauseExpiry should probably be asserted too.

solana/modules/token_bridge/program/src/types.rs
```diff
@@ -85,24 +86,30 @@ impl Owned for Config {
...
+/// Tail layout (137-byte total Config account):
+///   bytes  32 ..  33 : `paused`       (u8, exactly 0 or 1 — see `write_paused` / `paused()`)
+///   bytes  33 ..  65 : `pauser`       (Pubkey)
+///   bytes  65 ..  97 : `unpauser`     (Pubkey)
+///   bytes  97 .. 129 : `freezer`      (Pubkey)
+///   bytes 129 .. 137 : `pause_expiry` (i64 LE, unix seconds — matches Clock.unix_timestamp)
 ///
-/// See the "Pausing" section of whitepapers/0003_token_bridge.md.
+/// `freezer` and `pause_expiry` are APPENDED after the original pauser/unpauser tail so the
+/// existing offsets are preserved; the first `SetPauserAddresses` realloc's straight to the full
```
johnsaigle: I don't think backward compatibility is important here since we've not deployed this yet. Ideally we'd use the same storage layouts on the wire formats and across the different implementations -- IIRC the EVM implementation's ordering is pauser/freezer/unpauser. It's not a big problem if the layouts differ in practice, but I also don't see a good reason why they are different currently.

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -136,4 +151,8 @@ solitaire! {
     SetPauserAddresses => set_pauser_addresses,
     Pause => pause,
     Unpause => unpause,
+    // Appended after the existing instructions so their dispatch indices (the on-chain
```
johnsaigle: You can delete this. We haven't deployed this so backward compatibility isn't important. Also this type of comment is better as a changelog/PR description rather than a code comment.

johnsaigle (commented):
Core logic looks good. I left some comments that would help improve the code quality and clarity of the comments+errors going forward, plus some suggestions.

mdulin2 (dismissed):

solana/modules/token_bridge/program/src/api/pause.rs
douglasgalico: `pause`, `unpause` are also eligible to that - so applying there too. Thanks!

solana/modules/token_bridge/program/src/lib.rs
```diff
@@ -110,6 +116,19 @@ pub enum TokenBridgeError {
     /// `self_program` account passed by the caller does not match this program's `program_id`.
     /// Only used to defensively guard the self-CPI that emits Anchor-style events.
     InvalidSelfProgram,
+    // NOTE: new variants are APPENDED below — `From<TokenBridgeError>` maps each variant to its
```
johnsaigle, line 119: suggestion: shouldn't this go on top of `TokenBridgeError` above rather than the middle of the list?

johnsaigle (approved):

mdulin2 (approved):

kcsongor (approved):

PR #4875, opened 2026-06-18
Enable Channel Check Linter

mdulin2:
This is part two of integrating the channel linter into the repository. Part one added the custom linter flow, and the channel linter; this enables the channel linter. 

In general, three types of changes were made: 
* Made the channel non-blocking by using the `WriteToChannelWithoutBlocking` helper. Done on `errC` and metric channel writes mostly. 
* Added `ctx.Done()` to the `select` for smoother shutdown. 
* Add `nolint:channelcheck` with an explanation on why the current setup is correct. 

It should be noted that `msgC` is made as an exception to the `channelcheck` linter because it should always be a blocking write. With the addition of the `nolint` linter, we can't add `nolint:channelcheck` with a comment there. So, I left the original `Note on channel capacity` comment as is.


johnsaigle:
@mdulin2 looks like this needs a rebase

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -257,7 +257,7 @@ func (w *Watcher) Run(ctx context.Context) error {
 
 				if msg != nil {
 					msg.IsReobservation = true
-					w.msgChan <- msg
+					w.msgChan <- msg // Note on channel capacity: We never want to drop messages.
```
johnsaigle, line 260: Should the changes to this file be `nolint` instead of the "Note" style?

node/pkg/watchers/near/poll.go
```diff
@@ -50,7 +51,7 @@ func (e *Watcher) recursivelyReadFinalizedBlocks(logger *zap.Logger, ctx context
 	// we want to avoid going too far back because that would increase the likelihood of error somewhere in the recursion stack.
 	// If we go back too far, we just report the error and terminate early.
 	if recursionDepth > maxFallBehindBlocks {
-		e.eventChan <- EVENT_NEAR_WATCHER_TOO_FAR_BEHIND // Note on channel capacity: Only pauses this watcher
+		common.WriteToChannelWithoutBlocking(e.eventChan, EVENT_NEAR_WATCHER_TOO_FAR_BEHIND, "near_event_chan") // Non-blocking: eventChan is buffered (cap 10) and metrics-only, so dropping an event is preferable to stalling the watcher.
```
johnsaigle, line 54: suggestion: Avoid putting constants directly in documentation in case they change later
```suggestion
		common.WriteToChannelWithoutBlocking(e.eventChan, EVENT_NEAR_WATCHER_TOO_FAR_BEHIND, "near_event_chan") // Non-blocking: eventChan is buffered and metrics-only, so dropping an event is preferable to stalling the watcher.
```

Same with the `cap 10` comment in the other Near file

johnsaigle (commented):
The changes to the channel handling seem sensible. It's also nice to see more detail getting added to the nolint comments 👍🏻 

I'm happy to approve after the Near comments are updated.

EDIT: Looks like a rebase is required too.

PR #4878, opened 2026-06-19
Add testutils pkg; use TimeFromUnix when building message publications

johnsaigle:
Use the SDK's new [TimeFromUnix](https://github.com/wormhole-foundation/wormhole/blob/2771cd01da6e83ccf3ab3991357af0afb17cfc0b/sdk/vaa/structs.go#L215) timestamp validation function at all call sites, including in test code. 

This normalizes timestamp validation to avoid niche overflow scenarios as well as divergences between various types used to represent timestamps in the Go code. It adds an error-handling step but it also makes it so that authors don't need to know the implementation details of how a timestamp is represented. It also allows us to avoid linter warnings by resolving them correctly instead of muting them at every site where a Message Publication is built.

In order to make tests a bit more ergonomic, a
helper test function was added in a new testutils/ package. Going
forward, we can move shared test helpers into this package rather than
re-implementing functions that are needed by many tests (e.g. building
and validating Message Publications).


node/pkg/watchers/sui/watcher.go
```diff
@@ -188,10 +188,14 @@ func (e *Watcher) processEvent(ctx context.Context, logger *zap.Logger, event su
 	}
 
 	txHashEthFormat := eth_common.BytesToHash(txHashBytes)
+	timestamp, err := vaa.TimeFromUnix(msg.Timestamp)
+	if err != nil {
+		return fmt.Errorf("processEvent failed to parse timestamp: %w", err)
+	}
```
djb15, line 195: Just a quick Q to aid in review, what's the logic you used for error handling here and elsewhere in this PR? Is the justification for returning an error that anything with a `nosec G115` must have been trusted before, so an error should never be returned? And if it ever were, that's better than creating a message publication for a timestamp that isn't valid?

johnsaigle: A lot of the context for this is in the function comment added here: https://github.com/wormhole-foundation/wormhole/blob/2771cd01da6e83ccf3ab3991357af0afb17cfc0b/sdk/vaa/structs.go#L215

Essentially the `serializeBody` function is the source of truth and it hard truncates any Go timestamp value (int64) to a uint32. 

With that in mind, values for Timestamp that are outside of `[0, max uint32]` are incorrect. Two Message Publications that differ only by timestamps `0` and `maxUint32 + 1` would map to the same VAA because of the downcasting in the serialization logic.

Regardless of the nosec comments, the goal of the PR is to make sure that the range of the timestamp field is bound to a safe domain.

If you want a more cautious patch, we could change this to logging at the error level instead?

node/pkg/watchers/algorand/watcher.go
```diff
@@ -179,9 +179,15 @@ func lookAtTxn(e *Watcher, t types.SignedTxnInBlock, b types.Block, logger *zap.
 	}
 
 	for _, obs := range observations {
+		timestamp, err := vaa.TimeFromUnix(b.TimeStamp)
+		if err != nil {
+			logger.Error("invalid block timestamp", zap.Error(err), zap.Int64("timestamp", b.TimeStamp))
```
djb15: I know not all of the existing error messages are great, but it would be nice if we could add more information about a transaction here to aid in debugging if this ever occured

node/pkg/watchers/aptos/watcher.go
```diff
@@ -369,11 +369,16 @@ func (e *Watcher) observeData(logger *zap.Logger, data gjson.Result, nativeSeq u
 		logger.Error("consistency level is larger than expected MaxUint8")
 		return
 	}
+	timestamp, err := vaa.TimeFromUnix(ts.Uint())
+	if err != nil {
+		logger.Error("invalid timestamp", zap.Error(err), zap.Uint64("timestamp", ts.Uint()))
```
djb15: Same here

node/pkg/watchers/evm/by_transaction.go
```diff
@@ -92,10 +91,14 @@ func MessageEventsForTransaction(
 		if err != nil {
 			return nil, 0, nil, fmt.Errorf("failed to parse log: %w", err)
 		}
+		timestamp, err := vaa.TimeFromUnix(blockTime)
+		if err != nil {
+			return nil, 0, nil, fmt.Errorf("invalid block timestamp: %w", err)
```
djb15: Same here

node/pkg/watchers/near/tx_processing.go
```diff
@@ -235,10 +234,14 @@ func (e *Watcher) processWormholeLog(logger *zap.Logger, _ context.Context, job
 
 	// SECURITY the timestamp of an observation is the timestamp of the block in which the wormhole core receipt has been finalized.
 	ts := outcomeBlockHeader.Timestamp
+	timestamp, err := vaa.TimeFromUnix(ts)
+	if err != nil {
+		return fmt.Errorf("invalid block timestamp: %w", err)
```
djb15: Same here

node/pkg/watchers/solana/shim.go
```diff
@@ -357,10 +357,14 @@ func (s *SolanaWatcher) shimProcessRest(
 		}
 		return nil
 	}
+	timestamp, err := vaa.TimeFromUnix(messageEvent.Timestamp)
+	if err != nil {
+		return fmt.Errorf("invalid shim message timestamp: %w", err)
```
djb15: Same here

node/pkg/watchers/sui/watcher.go
djb15: Thanks John. No I'm happy with the changes, I'd rather be more defensive here

node/pkg/watchers/sui/watcher.go
```diff
@@ -188,10 +188,14 @@ func (e *Watcher) processEvent(ctx context.Context, logger *zap.Logger, event su
 	}
 
 	txHashEthFormat := eth_common.BytesToHash(txHashBytes)
+	timestamp, err := vaa.TimeFromUnix(msg.Timestamp)
+	if err != nil {
+		return fmt.Errorf("processEvent failed to parse timestamp: %w", err)
```
djb15: Same here

node/pkg/watchers/evm/by_transaction.go
johnsaigle: Sounds good, I'll rework these so that they're more actionable

johnsaigle:
@djb15 could you re-review?

djb15 (approved):
Thanks for the changes to the logs

johnsaigle:
In draft until I resolve the dictionary conflict

PR #4880, opened 2026-06-22
feat(CORE-2970): worm-cli sui migration

bengtlofgren:
# worm CLI: migrate Sui support to `@mysten/sui` v2 gRPC

## Linear Ticket

**Ticket**: CORE-2970

## Summary

Migrates the `worm` CLI's Sui support from `@mysten/sui` v1 (`SuiClient`, JSON-RPC) to v2.19
(`SuiGrpcClient`, the gRPC fullnode API). The Sui RPC surface changed substantially between
majors — method names, response envelopes, the `json` representation of Move values, BCS, and
transaction building — so every Sui code path (provider, object/coin reads, registrations,
deploy, init, attest/createWrapped, transfer, governance submit) was updated. Adds a network
safety guard and hermetic unit tests, and resolves the ESM-only nature of the v2 package in the
TypeScript/Jest/esbuild toolchain.

## Major Changes

- **SDK bump and provider.** `@mysten/sui` `^1.45.0` → exact `2.19.0`; lockfile updated.
  `getProviderForChain`/`getProvider` now construct `new SuiGrpcClient({ network, baseUrl })`
  instead of `new SuiClient({ url })`. `ChainProvider<"Sui">` is now `SuiGrpcClient`.
- **gRPC response-shape adapters.** The gRPC `json` representation flattens Move structs (no
  nested `.fields`), collapses UID `id.id` to a plain `id`, and encodes `vector<u8>` as base64.
  `getObjectFields`, `getOriginalAssetSui` (wrapped/native branches), `getTokenCoinType`,
  `getPackageId`, `getOriginalPackageId`, and the registrations reader were rewritten against
  this shape.
- **Transaction execution envelope.** `executeTransactionBlock` now unwraps the gRPC
  `Transaction | FailedTransaction` result, reads `status.success`, and flattens
  `effects.changedObjects` together with the `objectTypes` map into the CLI's normalized
  `SuiTransactionResult`.
- **BCS and transaction building.** Adds `CoinTypeKeyBcs`/`RegistryKeyBcs` schemas for
  dynamic-field key lookups; switches VAA arguments to typed `tx.pure("vector<u8>", ...)` and
  publishing to `tx.publish`. Pagination is rewritten as explicit `listOwnedObjects` /
  `listDynamicFields` / `listCoins` cursor loops.
- **Deploy path.** Devnet uses `sui client test-publish --publish-unpublished-deps` with a
  `fromCliJson` normalizer that maps the legacy CLI `objectChanges` shape onto
  `SuiTransactionResult`; persistent networks use an SDK `tx.publish` flow.
- **Network safety guard.** `getSuiNetwork` is now exhaustive (explicit `Devnet`→`localnet`,
  throws on unknown). `assertSuiNetwork` verifies the endpoint's chain identifier matches the
  selected `--network` (mainnet/testnet) and is invoked inside `executeTransactionBlock`, so any
  `--network`/`--rpc` mismatch is refused before signing. The signer carries its `network`.
- **Toolchain for an ESM-only dependency.** `tsconfig.json` adds `paths` overrides mapping the
  v2 subpath types (`@mysten/sui/{grpc,graphql,transactions,...}`) to their `.d.mts` files; a new
  `tsconfig.build.json` clears `paths` for esbuild (which resolves the package `exports` map
  natively); the build script passes `--tsconfig=tsconfig.build.json`; the Jest transform now
  handles `.mjs`/`.mts` and transpiles `@mysten`/`@noble`/`@scure` ESM.
- **Tests.** Adds hermetic unit tests: `sui-utils.test.ts` (object filters, published-package
  extraction, address/type normalization+validation, and a `CoinTypeKeyBcs` wire-byte assertion
  against on-chain ground truth) and `smoke.test.ts` (a toolchain canary that loads the v2
  subpath modules).

## Files Modified

- `clients/js/src/chains/sui/utils.ts` — provider, signer, `executeTransactionBlock` envelope,
  pagination helpers, `getSuiNetwork`/`assertSuiNetwork` guard.
- `clients/js/src/sdk/sui.ts` — `getOriginalAssetSui`/`getForeignAssetSui`/`getTokenCoinType`,
  `getObjectFields`, BCS schemas, type/address validators.
- `clients/js/src/chains/sui/{publish,submit,transfer,registrations,buildCoin,log}.ts` — deploy
  normalization, governance submit/createWrapped, transfer PTB, registration listing, coin build.
- `clients/js/src/cmds/sui/{deploy,init,setup,publishMessage,utils}.ts` — command wiring to the
  new client and result shape.
- `clients/js/src/chains/generic/{provider,getWrappedAssetAddress}.ts`,
  `clients/js/src/chains/generic/getOriginalAsset.ts` — `SuiGrpcClient` provider type and
  `Chain` lookups.
- `clients/js/{tsconfig.json,tsconfig.build.json,jest.config.json,package.json,package-lock.json}`
  — ESM resolution, build/test config, dependency pin.
- `clients/js/src/chains/sui/__tests__/{sui-utils,smoke}.test.ts` — new unit tests.

## Testing

- Type-check (`tsc --noEmit`) and build (`esbuild`) clean; unit suites green.
- Read-only decoders validated live against the public Sui gRPC fullnode on **testnet and
  mainnet** (object fields, wrapped/native original-asset decode, coin-type forward/reverse
  round-trip, registration listing, error paths).
- Write path validated against a **local Sui node**: `executeTransactionBlock` envelope mapping
  on a real signed transaction; `worm sui deploy` end-to-end through real `sui client
  test-publish` output; the `WormholeMessage`-style event decode confirmed (gRPC encodes the
  `vector<u8>` event fields as base64); and the network guard confirmed to refuse a mismatched
  `--network`/endpoint before signing.
- The bundled `worm` artifact runs read-only Sui subcommands against testnet, confirming the
  ESM-only dependency loads at runtime.

## Additional Notes

- `@mysten/sui` v2 is ESM-only with `engines.node >= 22`; the `tsconfig.json` `paths` overrides
  are pinned to the exact `2.19.0` `dist` layout. A follow-up to `moduleResolution: "bundler"`
  would let `tsc` read the `exports` map natively and remove the overrides.
- The public Sui fullnodes serve gRPC-web on the existing `:443` endpoints, so the default RPCs
  in `networks.ts` are unchanged.
- Known minor issue (diagnostic command only): `queryRegistrationsSui` keys results by chain
  name via `chainIdToChain`, which returns `undefined` for chain ids newer than the installed
  SDK, collapsing all such rows under a single `undefined` key. Falling back to the numeric chain
  id as the key would preserve them.


clients/js/src/chains/sui/transfer.ts
```diff
@@ -29,27 +33,100 @@ export async function transferSui(
...
+  const nonce = Math.floor(Math.random() * 0xffffffff);
+
+  const [transferTicket, dust] = tx.moveCall({
+    target: `${tokenBridgePackageId}::transfer_tokens::prepare_transfer`,
+    arguments: [
+      assetInfo,
+      transferCoin,
+      tx.pure("u16", recipientChainId),
+      tx.pure("vector<u8>", [...recipient]),
+      tx.pure("u64", BigInt(0)),
```
douglasgalico, line 110: Even that we know the fee is zero, the right would be reading from core bridge state and passing the fee here.

douglasgalico (changes requested):

douglasgalico (dismissed):

clients/js/src/chains/sui/publish.ts
```diff
@@ -59,68 +60,62 @@ export const publishPackage = async (
...
-        )}`
-      );
-    }
-
-    // Find the main package (not dependencies) - it's typically the last one published
-    const mainPackage = publishedChanges[publishedChanges.length - 1];
-    console.log(`Published package ID: ${mainPackage.packageId}`);
-
-    return result;
+    output = execSync(cmd, { encoding: "utf-8" });
```
johnsaigle: Can you try to refactor this using `execFileSync` here unless there's a specific reason to invoke a shell? Using that method instead is more resistant to command injection.

clients/js/src/chains/sui/publish.ts
```diff
@@ -59,68 +60,62 @@ export const publishPackage = async (
...
+ * `objectChanges` payload.
+ */
+const publishPackageTestPublish = async (
+  signer: SuiSigner,
+  packagePath: string
+): Promise<SuiTransactionResult> => {
+  // `--build-env testnet` selects testnet dependency pins (localnet is not a
+  // pinned environment); `--publish-unpublished-deps` publishes dependencies
+  // that are not yet on-chain for this network.
+  const cmd = `sui client test-publish ${packagePath} --publish-unpublished-deps --build-env testnet --json 2>&1`;
```
johnsaigle: Can you add a check that `packagePath` is a directory first using a variant of `fs`'s realpath functions? This would make sure that this is actually a directory on the system before invoking the command.

johnsaigle: Does using `2>&1` make sense here? It looks like you're merging the stdout with stderr and then working around that using the first-brace parsing above. If you leave these split, can you just grab the pure JSON from one of those output sources?

clients/js/src/chains/sui/publish.ts
```diff
@@ -59,68 +60,62 @@ export const publishPackage = async (
...
   }
+  return publishPackageSDK(signer, network, packagePath);
 };
 
 /**
- * Use `sui client test-publish` for ephemeral/local deployments.
- * This handles dependencies automatically and doesn't require Published.toml manipulation.
- * Note: Uses the locally configured Sui CLI signer, not a programmatic signer.
+ * Extract the transaction digest from `sui client test-publish --json` output.
+ * The CLI prepends human-readable build lines before the JSON object, so the
```
johnsaigle: Are you sure there's nothing like `--quiet` that would return only the JSON? It would be very unusual for such a tool to require all consumers to manually parse out text from JSON

clients/js/src/chains/sui/transfer.ts
```diff
@@ -29,27 +34,103 @@ export async function transferSui(
...
+
+  const [feeCoin] = tx.splitCoins(tx.gas, [tx.pure("u64", messageFee)]);
+
+  const [assetInfo] = tx.moveCall({
+    target: `${tokenBridgePackageId}::state::verified_asset`,
+    arguments: [tx.object(token_bridge)],
+    typeArguments: [coinType],
+  });
+
+  // Random 32-bit transfer nonce.
```
johnsaigle:
```suggestion
  // Pseudo-random 32-bit transfer nonce.
```
Note that `Math.random()` is not cryptographically random and can be predicted

clients/js/src/chains/sui/transfer.ts
```diff
@@ -29,27 +34,103 @@ export async function transferSui(
...
+  const nonce = Math.floor(Math.random() * 0xffffffff);
```
johnsaigle: Can you explain why this step is needed?

clients/js/src/cmds/sui/publishMessage.ts
```diff
@@ -97,13 +97,16 @@ export const addPublishMessageCommands: YargsAddCommandsFn = (
 
       logTransactionDigest(res);
       logTransactionSender(res);
+      // In the gRPC event JSON, `payload` (vector<u8>) is base64 while
+      // `sender` (ID) is a 0x-prefixed hex string.
+      const parsed = event.json as any;
```
johnsaigle, line 102: Worth checking the results of `parsed` here? Otherwise the tx might succeed but this part could decode wrong or give an error

clients/js/src/cmds/sui/setup.ts
```diff
@@ -233,9 +233,9 @@ export const addSetupCommands: YargsAddCommandsFn = (y: typeof yargs) =>
   );
 
 const getEmitterCapObjectId = async (
-  client: SuiClient,
+  client: SuiGrpcClient,
   tokenBridgeStateObjectId: string
-): Promise<string> => {
+): Promise<string | undefined> => {
```
johnsaigle: Maybe better to throw an error here if the emitter cap id is undefined?

clients/js/src/sdk/sui.ts
```diff
@@ -17,75 +35,158 @@ export async function getForeignAssetSui(
...
+  }
+
+  const obj = await client.getObject({
+    objectId: fieldId,
+    include: { json: true },
+  });
+  const type = obj.object.type;
+  const value = (obj.object.json as any)?.value;
+
+  if (type.includes("wrapped_asset::WrappedAsset<")) {
```
johnsaigle: Any way to use Sui's functions to do this? My LLM is suggesting the following might help:

```
isValidStructTag,
  isValidSuiAddress,
  normalizeStructTag,
  normalizeSuiAddress,
  parseStructTag,
```

clients/js/src/sdk/sui.ts
```diff
@@ -17,75 +35,158 @@ export async function getForeignAssetSui(
...
+
+  const state = await client.getObject({
+    objectId: tokenBridgeStateObjectId,
+    include: { json: true },
+  });
+  const registryId = (state.object.json as any)?.token_registry?.id;
+  if (!registryId) {
+    throw new Error("Unable to fetch token registry object ID");
+  }
+  const originalPackageId = state.object.type.split("::")[0];
```
johnsaigle: Are you able to use the built-in Sui SDK to do this parsing instead?

clients/js/src/sdk/sui.ts
```diff
@@ -17,75 +35,158 @@ export async function getForeignAssetSui(
...
 
-  const keyType = getTableKeyType(coinTypes?.type);
-  if (!keyType) {
-    throw new Error("Unable to get key type");
-  }
+  // The dynamic-field key type is declared in the *original* package that
+  // defines `token_registry`, which is the package embedded in the state
+  // object's type (Move type origins are stable across upgrades), not the
+  // upgraded package returned by `getPackageId`.
+  const originalPackageId = state.object.type.split("::")[0];
```
johnsaigle: Same here, can we use Sui's parsing instead of a manual split?

clients/js/src/sdk/sui.ts
```diff
@@ -17,75 +35,158 @@ export async function getForeignAssetSui(
...
+      parentId: registryId,
+      name: {
+        type: `${originalPackageId}::token_registry::Key<${coinType}>`,
+        bcs: getRegistryKeyBcs().serialize({ dummy_field: false }).toBytes(),
+      },
+    });
+    fieldId = res.dynamicField.fieldId;
+  } catch {
+    throw new Error(
+      `Token of type ${coinType} has not been registered with the token bridge`
```
johnsaigle: I'm not sure this error message is correct for all instances of `getDynamicField` errors. It's probably better to surface and log the actual error here instead of assume that it's a registration issue.

clients/js/src/chains/sui/publish.ts
mdulin2: Would also solve a potential command injection issue here? Can't have a packagePath that's also a command injection payload.

clients/js/src/chains/sui/transfer.ts
```diff
@@ -29,27 +34,103 @@ export async function transferSui(
...
+  const tx = new Transaction();
+
+  const [transferCoin] = (() => {
+    if (coinType === SUI_TYPE_ARG) {
+      return tx.splitCoins(tx.gas, [tx.pure("u64", amt)]);
+    }
+    const [primaryCoinId, ...mergeCoinIds] = coinObjectIds;
+    const primaryCoinInput = tx.object(primaryCoinId);
+    if (mergeCoinIds.length) {
+      tx.mergeCoins(
```
mdulin2, line 84: If a large number of dust coins were sent here, could this cause `mergeCoins()` to perform too many operations and fail? This could be done by an outside attacker for almost nothing.

Maybe it's worth A) sorting from largest to smallest and B) only merging as many as it takes to get the amount required for the transfer.

clients/js/src/sdk/sui.ts
mdulin2: Agreed, that seems better than this.

mdulin2 (changes requested):

clients/js/src/chains/sui/utils.ts
```diff
@@ -15,93 +15,241 @@ const UPGRADE_CAP_TYPE = "0x2::package::UpgradeCap";
...
+// include-gated members (`effects`, `objectTypes`, ...) are present on the type.
+type GrpcTransaction = SuiClientTypes.Transaction<typeof TX_RESULT_INCLUDE>;
+
+/**
+ * Flatten a gRPC transaction envelope into the CLI's `SuiTransactionResult`.
+ * The gRPC transport exposes mutations as `effects.changedObjects` (with
+ * `outputState`/`idOperation` enums) plus a separate `objectId -> type` map;
+ * this reassembles the two into the single shape the rest of the CLI consumes.
+ */
+export const toSuiTransactionResult = (
```
johnsaigle, line 164: Because of how a `resFailedTransaction` gets converted by this function, the callers need to do `assertSuccess` on a `SuiTransactionResult` or else they'll continue even after transactions fail.

At least these callers are missing asserts:
- transferSui: transfer.ts:142-143
- Guardian-set governance submission: submit.ts:90-92
- Register-chain governance submission: submit.ts:221-223

It would be good to review all the callers of this function to make sure that they're correctly handling failed transactions.

clients/js/src/chains/sui/transfer.ts
mdulin2: Seems like a fix was done. Going to resolve this comment.

mdulin2 (dismissed):
LGTM!

johnsaigle:
Note that CI is failing because of how the output is being handled:

```
          sui │ [sui-contracts] Error: Unexpected non-JSON output from test-publish: INCLUDING DEPENDENCY MoveStdlib
          sui │ [sui-contracts] INCLUDING DEPENDENCY Sui
          sui │ [sui-contracts] BUILDING wormhole
```

IMO the way to approach this is the following:
* Output _only_ JSON on STDOUT if using a flag like `--json`
* Change your debug messages to print to STDERR
* Avoid any bash commands like that redirect STDERR to STDOUT as I mentioned earlier (https://github.com/wormhole-foundation/wormhole/pull/4880#discussion_r3590054862)

clients/js/src/chains/sui/publish.ts
```diff
@@ -59,68 +88,89 @@ export const publishPackage = async (
...
   }
+  return publishPackageSDK(signer, network, packagePath);
 };
 
 /**
- * Use `sui client test-publish` for ephemeral/local deployments.
- * This handles dependencies automatically and doesn't require Published.toml manipulation.
- * Note: Uses the locally configured Sui CLI signer, not a programmatic signer.
+ * Extract the transaction digest from `sui client test-publish -q --json`
+ * output. Despite `-q --json`, the Sui CLI still prints Move build progress
```
johnsaigle: FYI I _think_ this is resolved a newer version of the Sui CLI tools, and https://github.com/wormhole-foundation/wormhole/pull/4922/ will bump this version. 

It's not a major deal, but I think it would be a bit nicer to avoid this custom parsing if indeed the logs are about to disappear anyway.

Do you mind testing with the new version being used in that PR to see if it fixes the output? If so, we could remove this parsing code and wait until the new version is applied upstream before merging this PR. If the logs are still there anyway, we can include this workaround.

kev1n-peters (approved):

douglasgalico (approved):

kcsongor (approved):

panoel (approved):

djb15 (approved):
Approving the Tiltfile change

PR #4885, opened 2026-06-29
Literal Eval to Eval (Algorand)

mdulin2:
The Python `eval()` is a known bad sink for code execution bugs. Although there's no security impact with this currently, we're proactively changing the `eval()` to `literal_eval()` to prevent this from ever becoming an issue.

### DIFF
--- a/algorand/admin.py
+++ b/algorand/admin.py
@@ -2,19 +2,17 @@
 
 import os
 from os.path import exists
-from time import time, sleep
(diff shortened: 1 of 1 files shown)


djb15 (approved):

bemic (approved):

mdulin2:
The Sui build process seems to be broken. This only touched the Algorand tests, so I hope we can just ignore that test failure and merge this.

PR #4888, opened 2026-07-01
Update 0g governor daily limit to 50k from 10k

aadam-10:


bemic (approved):
Approving based on Dirk's agreement on Slack

evan-gray (approved):

PR #4890, opened 2026-07-03
node: Add ability to pass custom headers to Sui gRPC

djb15:
Moving to draft for the time being. Guardians should be able to work around this, we can consider reopening soon if this continues to be a blocker.

### DIFF
--- a/node/pkg/suiclient/grpc_headers.go
+++ b/node/pkg/suiclient/grpc_headers.go
@@ -1,0 +1,100 @@
+package suiclient
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// GrpcHeaderDialOptions converts key=value header specs from command line args
+// into gRPC client interceptors that attach the headers as outgoing metadata
+func GrpcHeaderDialOptions(headerSpecs []string) ([]grpc.DialOption, error) {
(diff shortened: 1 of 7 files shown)


node/pkg/suiclient/grpc_headers.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	headers = headers.Copy()
+	return []grpc.DialOption{
+		grpc.WithChainUnaryInterceptor(grpcHeaderUnaryInterceptor(headers)),
+		grpc.WithChainStreamInterceptor(grpcHeaderStreamInterceptor(headers)),
+	}, nil
+}
+
+func parseGrpcHeaderSpecs(headerSpecs []string) (metadata.MD, error) {
```
johnsaigle, line 30: suggestion: I think some of this logic could be replaced by using cobra's built-in parsing.

This is an example that parses key-value pairs:

```go
var labels map[string]string

cmd.Flags().StringToStringVar(
    &labels,
    "label",
    nil,
    "key-value labels, e.g. --label env=prod,team=infra",
)
```

It might make more sense to put this in a the `cmd/` code as well since this is related to CLI input parsing rather than Sui client code.

djb15:
I'm going to close this as clearly the Guardians have been able to connect to their Sui nodes in the meantime, and as discussed internally it's preferable not to make this change. This PR can always be re-opened if required.

PR #4892, opened 2026-07-03
node: Consolidate borsch library usage

djb15:
I spent some time looking through the existing tests and adding a couple of new ones to check whether we're hitting all the edge cases we care about. Not a place we want to be making a mistake, so I'm happy to abandon these changes if the risks outweigh the upside

### DIFF
--- a/node/pkg/watchers/solana/client_test.go
+++ b/node/pkg/watchers/solana/client_test.go
@@ -218,10 +218,6 @@ func TestParseMessagePublicationAccount(t *testing.T) {
 	)
 
 	const (
-		// Define error string for testing. This is returned by the borsh-go library when it fails to
-		// deserialize a struct. In this case, we're relying on the library to fail when
-		// the message account data can't be deserialized into [MessagePublicationAccount].
-		errStringBorsh         = "failed to read required bytes"
 		errStringParseTooShort = "message account data is too short"
 	)
 
@@ -230,6 +226,7 @@ func TestParseMessagePublicationAccount(t *testing.T) {
 		// Named input parameters for target function.
 		messageAccountData func(t *testing.T) MessageAccountData
 		want               *MessagePublicationAccount
+		wantErr            bool
 		errStr             string
 	}{
 		{
@@ -272,6 +269,25 @@ func TestParseMessagePublicationAccount(t *testing.T) {
 			errStr: "",
 		},
 		{
+			name: "success -- trailing bytes are ignored",
+			messageAccountData: func(t *testing.T) MessageAccountData {
+				withTrailingBytes := append([]byte{}, validMessageAccountDataReliable...)
+				withTrailingBytes = append(withTrailingBytes, 0xff, 0xee, 0xdd)
+				return mustNewMessageAccountData(t, withTrailingBytes)
+			},
+			want: &MessagePublicationAccount{
+				VaaVersion:       0,
+				ConsistencyLevel: 32,
+				SubmissionTime:   1770212672,
+				Nonce:            0,
+				Sequence:         1367797,
+				EmitterChain:     1,
+				EmitterAddress:   emitterAddrReliable,
+				Payload:          payload,
+			},
+			errStr: "",
+		},
+		{
 			name: "success -- reliable message with non-zero vaa time",
 			messageAccountData: func(t *testing.T) MessageAccountData {
 				acct := testMessagePublicationAccount(payload, 32)
@@ -304,8 +320,9 @@ func TestParseMessagePublicationAccount(t *testing.T) {
 			messageAccountData: func(t *testing.T) MessageAccountData {
 				return MessageAccountData{}
 			},
-			want:   &MessagePublicationAccount{},
-			errStr: errStringParseTooShort,
+			want:    &MessagePublicationAccount{},
+			wantErr: true,
+			errStr:  errStringParseTooShort,
 		},
 		{
 			name: "failure -- data too short",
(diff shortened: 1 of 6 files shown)


node/pkg/watchers/solana/client_test.go
```diff
@@ -267,7 +264,29 @@ func TestParseMessagePublicationAccount(t *testing.T) {
 				Sequence:         3458072,
 				EmitterChain:     1,
 				EmitterAddress:   emitterAddrUnreliable,
-				Payload:          nil, // borsh deserialization results in this being nil rather than an empty slice
+				Payload:          nil, // Borsh deserialization results in this being nil rather than an empty slice
+			},
+			errStr: "",
+		},
+		{
+			name: "success -- trailing bytes are ignored",
```
johnsaigle: This is a change in behaviour, right? Previously I believe this returned an error, and that's probably safer

johnsaigle: If gagliardetto doesn't do this then I think that's arguably a bug in that library

johnsaigle: Got this from an LLM FWIW:

---

_Borsh rejects trailing bytes. The reference implementation (borsh-rs) treats any unconsumed bytes after deserialization as an error.
The high-level entry points enforce this:
- from_slice / try_from_slice — after T::deserialize, check if !v_mut.is_empty() and return Err(InvalidData, "Not all bytes read") (borsh-rs/borsh/src/de/mod.rs).
- from_reader / try_from_reader — after deserializing, attempt to read 1 more byte; success (non-EOF) triggers the same "Not all bytes read" error.
The low-level deserialize / deserialize_reader methods themselves do not check — they only consume what the type needs. So the "no trailing bytes" rule is enforced at the API boundary, not per-field. The written spec (the README pseudocode) doesn't call it out explicitly, but the reference implementation's behavior is the de facto standard, and the error constant ERROR_NOT_ALL_BYTES_READ makes the stance unambiguous: the entire input must be consumed._

djb15: Yes definitely a change of behaviour, although I couldn't think of a way to abuse this, and I also couldn't think of a neat way to keep the behaviour without just creating a wrapper function with the extra trailing bytes validation step. Any strong opinions (including abandoning this PR)?

johnsaigle: I think the PR is good and I do think it would be better to make just such a wrapper function! I'd prefer that the Guardian follows the expected behaviour of Borsh even if it's not exploitable at the moment. It'll keep things easier to reason about IMO, plus set us up to be more compatible with other libraries.

It's also possible that the solana-go behaviour has been patched in a newer version, so if we upgrade that library in the future, that might address it also.

I think it'd be best if there was a code comment on top of it where we link the Borsh standard. 

This would be easy to do in a follow-up PR instead, but it would mean that we'd have to revert these changes to the test behaviour later.

bemic (commented):
I haven't really looked into "shim" code. Therefore, consider this comment more as an datapoint that you can take into account when making the decision. 

I diff tested `near/borsh-go v0.3.0` vs `gagliardetto/binary v0.7.7` across all four affected struct layouts (`PostMessageData`, `MessagePublicationAccount`, `ShimPostMessageData`, `ShimMessageEventData`) using  more than 2.5mil canonical, edge case and malformed inputs. No difference in accepted inputs or successfully decoded values, and no panic.

Only difference spot by gpt5.6:

> The meaningful differences are that gagliardetto/binary decodes directly into the destination, so failed decoding may leave partial data and reused structs may retain an old slice when a zero-length payload is decoded. It also returns []byte slices that alias the input buffer, whereas near/borsh-go copied them; error messages and wrapping differ as well.
> 
> None of these differences currently breaks the reviewed flows because they use fresh destinations, discard values on error, and do not mutate decoded payloads.

johnsaigle:
@djb15 needs rebase

node/pkg/watchers/solana/client_test.go
djb15: I updated this to enforce that there are never trailing bytes when borsch decoding now in the latest commit

node/pkg/watchers/solana/shim.go
```diff
@@ -116,7 +115,7 @@ func shimParsePostMessage(shimPostMessageDiscriminator []byte, buf []byte) (*Shi
 	}
 
 	data := new(ShimPostMessageData)
-	if err := borsh.Deserialize(data, buf[len(shimPostMessageDiscriminator):]); err != nil {
+	if err := decodeBorsh(data, buf[len(shimPostMessageDiscriminator):]); err != nil {
```
johnsaigle: Hey I followed up on the Borsh trailing bytes thing. It looks like Rust actually does it in two different ways, allowing trailing bytes in some cases and giving an error in other cases.

Also, the [Shim post_message function on-chain](https://github.com/wormhole-foundation/wormhole/blob/e155608982af3a6e0813dafc54ecba3933559e3c/svm/wormhole-core-shims/crates/shim/src/post_message.rs#L157-L181) explicitly says that trailing bytes are ignored.

So my understanding right now is that the strict checks that you added based on my earlier review in `decodeBorsh` are generally the right way to go, especially for decoding single values. However, for decoding shim post message data _specifically_, we should allow trailing bytes so that we don't accidentally give an error here. 

I believe the change now would just be to replace this line with a call to `Decode` with no `HasRemaining` check, and then to update the tests accordingly so we explicitly allow trailing bytes in a payload.

Let me know if this makes sense, we can talk about other approaches if you want. Sorry for the bad suggestion earlier -- I wasn't aware that the shim code was so permissive.

djb15: Appreciate you reviewing this change carefully, good catch! Claude seems to agree here, I've made the requested change. Decided to use a new wrapper function rather than adding an argument to the existing function to make it extra clear what's happening at the call site.

Also have some mainly LLM generated comments on the test cases, but I thought it would be helpful for future us to understand why we have the differing decoding cases.

djb15: Will deal with the rebase in a bit

johnsaigle: Awesome, yes the changes look good and I think the comments are super helpful. 👍🏻 I can approve after the rebase

johnsaigle (approved):

bemic (approved):
I found no code-level blocker.

PR #4893, opened 2026-07-06
Xrpl ntt accountant

douglasgalico:
Integrates XRPL into the NTT global accountant. The XRPL watcher now recognizes NTT transfer emitters and synthesizes transceiver registration messages, so XRPL custody accounts can be tracked by the ntt-global-accountant

### DIFF
--- a/node/pkg/accountant/ntt_config.go
+++ b/node/pkg/accountant/ntt_config.go
@@ -48,6 +48,9 @@ func nttGetEmitters(env common.Environment) (validEmitters, validEmitters, error
 			{chainId: vaa.ChainIDArbitrumSepolia, addr: "000000000000000000000000fA42603152E4f133F5F3DA610CDa91dF5821d8bc"},
 			{chainId: vaa.ChainIDBaseSepolia, addr: "000000000000000000000000149987472333cD48ac6D28293A338a1EEa6Be7EE"},
 			{chainId: vaa.ChainIDOptimismSepolia, addr: "000000000000000000000000eCF0496DE01e9Aa4ADB50ae56dB550f52003bdB7"},
+			// XRPL NTT custody emitters — one per (custody account, token), keyed on keccak256("ntt" + manager32 + sourceToken32). Log-only until the XREG registration pipeline is live.
+			{chainId: vaa.ChainIDXRPL, addr: "264a5b2d2d0d0722a05f3ebe0403aee658729456cb50b8c5d524036bd0c06aaa", logOnly: true}, // rpQdxE7TGjyhFWQeVrnxYo4j8dqwj9BTLH (XRP)
+			{chainId: vaa.ChainIDXRPL, addr: "31155b0ad6a8e59f4411d0276a915b90f36de48163a5ddc1886b59e4525cac24", logOnly: true}, // rN6CtQsYcAFBRciP9ktiPRv2R6FZ5sk5Fm (IOU "TS1")
 		}
 		arEmitterConfig = sdk.KnownTestnetAutomaticRelayerEmitters
 	} else {
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -100,6 +100,8 @@ uint32 nonce
 []byte payload
 ```
 
+<!-- cspell:ignore XREG vaas -->
+
 ### Registration (Initialize Transceiver & Peer Registration)
 
 Because XRPL has no smart contracts, the `WormholeTransceiverInfo` (hub init) and
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -100,7 +100,7 @@ uint32 nonce
 []byte payload
 ```
 
-<!-- cspell:ignore XREG vaas -->
+<!-- cspell:ignore XREG XREL vaas -->
 
 ### Registration (Initialize Transceiver & Peer Registration)
 
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -100,13 +100,70 @@ uint32 nonce
 []byte payload
 ```
 
-### Initialize Transceiver
+### Registration (Initialize Transceiver & Peer Registration)
 
-> 🚧 TODO
+Because XRPL has no smart contracts, the `WormholeTransceiverInfo` (hub init) and
+`WormholeTransceiverRegistration` (peer) messages cannot be emitted natively.
+Instead, the Solana Sequencer emits an `XrplRegistration` (prefix `XREG`) message
+when a custody account is initialized (`initialize`) or a peer is registered
+(`register_peer`), and the executor relays it back to XRPL as a Payment to the
+Core Account. The watcher then re-keys and synthesizes the canonical transceiver
+message under the correct NTT transceiver emitter.
 
-### Transceiver (Peer) Registration
+Upon a Payment to the Core Account, the guardian watcher MUST
 
-> 🚧 TODO
+- Verify that the transaction is `validated`
+- Verify that the transaction result is `tesSUCCESS`
+- Verify that the transaction type is `Payment`
+- Verify that the **first** memo (index 0) has a MemoFormat of
+  `application/x-ntt-registration`, hex-encoded
+- Verify that the MemoData begins with the `XREG` prefix
+
+The MemoData carries the raw `XREG` bytes:
+
+```go
+[4]byte  prefix = "XREG"
+uint8    kind              // 0 = Hub, 1 = Peer
+[20]byte manager           // XRPL custody account id
+[]byte   token_id          // wire form: XRP 1B / IOU 41B / MPT 25B
+// Hub tail (kind=0):
+uint8    token_decimals
+// Peer tail (kind=1):
+uint16   peer_chain        // big-endian
+[32]byte peer_address
+```
+
+The watcher derives the NTT transceiver emitter
+`keccak256("ntt" + manager32 + token32)` — the **same** emitter used for
+`NativeTokenTransfer` messages from this manager+token pair — and emits the VAA
+under it, so the accountant keys hub/peer state on the same emitter it sees
+transfers from.
+
+#### Initialize Transceiver (kind = Hub)
+
+Synthesize a `WormholeTransceiverInfo` payload (70 bytes):
+
+```go
+[4]byte  prefix = 0x9c23bd3b
+[32]byte manager_address    // manager32 (left-padded XRPL account)
+uint8    manager_mode = 1   // locking
+[32]byte token_address      // token32 derived from token_id
+uint8    token_decimals
+```
+
+#### Transceiver (Peer) Registration (kind = Peer)
+
+Synthesize a `WormholeTransceiverRegistration` payload (38 bytes):
+
+```go
+[4]byte  prefix = 0x18fc67c2
+uint16   chain_id            // big-endian, the peer chain
+[32]byte transceiver_address // the peer address
+```
+
+An operator submits the resulting VAAs to the NTT global accountant via
+`submit_vaas` to establish the hub/peer registration before transfers are
+enforced.
 
 ### NativeTokenTransfer TransceiverMessage
 
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -106,10 +106,12 @@ uint32 nonce
 
 Because XRPL has no smart contracts, the [`WormholeTransceiverInfo`](https://github.com/wormhole-foundation/native-token-transfers/blob/main/docs/Transceiver.md#initialize-transceiver) (hub init) and
 [`WormholeTransceiverRegistration`](https://github.com/wormhole-foundation/native-token-transfers/blob/main/docs/Transceiver.md#transceiver-peer-registration) (peer) messages cannot be emitted natively.
-Instead, the Solana Sequencer emits an `XrplRegistration` (prefix `XREG`) message
-when a custody account is initialized (`initialize`) or a peer is registered
-(`register_peer`), and the executor relays it back to XRPL as a Payment to the
-Core Account. The watcher then re-keys and synthesizes the canonical transceiver
+Instead, when a custody account is initialized (`initialize`) or a peer is
+registered (`register_peer`), the Solana Sequencer emits the registration as an
+`XrplRelease` (XREL) carrying the `XrplRegistration` (XREG) bytes in its first
+memo. The delegated manager set signs and submits that XREL as a multisig Payment
+to the Core Account (there is no standalone XREG message and the executor is not
+involved). The watcher then re-keys and synthesizes the canonical transceiver
 message under the correct NTT transceiver emitter.
 
 Upon a Payment to the Core Account, the guardian watcher MUST
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -104,8 +104,8 @@ uint32 nonce
 
 ### Registration (Initialize Transceiver & Peer Registration)
 
-Because XRPL has no smart contracts, the `WormholeTransceiverInfo` (hub init) and
-`WormholeTransceiverRegistration` (peer) messages cannot be emitted natively.
+Because XRPL has no smart contracts, the [`WormholeTransceiverInfo`](https://github.com/wormhole-foundation/native-token-transfers/blob/main/docs/Transceiver.md#initialize-transceiver) (hub init) and
+[`WormholeTransceiverRegistration`](https://github.com/wormhole-foundation/native-token-transfers/blob/main/docs/Transceiver.md#transceiver-peer-registration) (peer) messages cannot be emitted natively.
 Instead, the Solana Sequencer emits an `XrplRegistration` (prefix `XREG`) message
 when a custody account is initialized (`initialize`) or a peer is registered
 (`register_peer`), and the executor relays it back to XRPL as a Payment to the
@@ -127,13 +127,21 @@ The MemoData carries the raw `XREG` bytes:
 [4]byte  prefix = "XREG"
 uint8    kind              // 0 = Hub, 1 = Peer
 [20]byte manager           // XRPL custody account id
-[]byte   token_id          // wire form: XRP 1B / IOU 41B / MPT 25B
+[]byte   token_id          // wire form, per the sequencer token_id encoding (see below)
 // Hub tail (kind=0):
 uint8    token_decimals
 // Peer tail (kind=1):
 uint16   peer_chain        // big-endian
 [32]byte peer_address
 ```
+
+The `token_id` uses the same encoding as the sequencer's XREL payload:
+
+| Token Type | Byte 0 | Bytes 1+                    | Total    |
+| ---------- | ------ | --------------------------- | -------- |
+| XRP        | 0x00   | (none)                      | 1 byte   |
+| IOU        | 0x01   | currency (20) + issuer (20) | 41 bytes |
+| MPT        | 0x02   | mpt_issuance_id (24)        | 25 bytes |
 
 The watcher derives the NTT transceiver emitter
 `keccak256("ntt" + manager32 + token32)` — the **same** emitter used for
@@ -148,10 +156,18 @@ Synthesize a `WormholeTransceiverInfo` payload (70 bytes):
 ```go
 [4]byte  prefix = 0x9c23bd3b
 [32]byte manager_address    // manager32 (left-padded XRPL account)
-uint8    manager_mode = 1   // locking
+uint8    manager_mode        // 0 = Locking, 1 = Burning (see below)
 [32]byte token_address      // token32 derived from token_id
 uint8    token_decimals
 ```
+
+`manager_mode` is derived by the watcher from the XREG `manager` and `token_id`:
+an XRPL custody account is in **Burning** mode when it is itself the token's
+issuer (it can mint/burn), and **Locking** mode otherwise (it holds a token
+issued by another account). The issuer is the last 20 bytes of an IOU `token_id`,
+or the first 20 bytes of an MPT `mpt_issuance_id`; XRP has no issuer and is always
+Locking. The `ntt-global-accountant` only accepts Locking hubs, but the init
+message must attest the manager's true mode rather than assume Locking.
 
 #### Transceiver (Peer) Registration (kind = Peer)
 
--- a/node/pkg/watchers/xrpl/README.md
+++ b/node/pkg/watchers/xrpl/README.md
@@ -100,8 +100,6 @@ uint32 nonce
 []byte payload
 ```
 
-<!-- cspell:ignore XREG XREL vaas -->
-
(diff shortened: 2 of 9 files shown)


M-Picco (dismissed):

node/pkg/watchers/xrpl/parse.go
```diff
@@ -82,6 +82,108 @@ const (
...
+
+// buildTransceiverInfoPayload builds a WormholeTransceiverInfo payload:
+//
+//	prefix(4=0x9c23bd3b) + manager_address(32) + manager_mode(1) +
+//	token_address(32) + token_decimals(1)  = 70 bytes
+func buildTransceiverInfoPayload(manager32, token32 [32]byte, decimals uint8) []byte {
+	out := make([]byte, 0, transceiverInfoLen)
+	out = append(out, transceiverInfoPrefix[:]...)
+	out = append(out, manager32[:]...)
+	out = append(out, nttModeLocking)
```
evan-gray: why is this always locking? an XRPL manager could be either locking or burning depending on if it is the asset issuer.

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,72 @@ uint32 nonce
...
+  `application/x-ntt-registration`, hex-encoded
+- Verify that the MemoData begins with the `XREG` prefix
+
+The MemoData carries the raw `XREG` bytes:
+
+```go
+[4]byte  prefix = "XREG"
+uint8    kind              // 0 = Hub, 1 = Peer
+[20]byte manager           // XRPL custody account id
+[]byte   token_id          // wire form: XRP 1B / IOU 41B / MPT 25B
```
evan-gray: I assume this follows the encoding from the sequencer spec?

**token_id encoding:**

| Token Type | Byte 103 | Bytes 104+                  | Total token_id |
|------------|----------|-----------------------------|----------------|
| XRP        | 0x00     | (none)                      | 1 byte         |
| IOU        | 0x01     | currency (20) + issuer (20) | 41 bytes       |
| MPT        | 0x02     | mpt_issuance_id (24)        | 25 bytes       |

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,72 @@ uint32 nonce
...
+transfers from.
+
+#### Initialize Transceiver (kind = Hub)
+
+Synthesize a `WormholeTransceiverInfo` payload (70 bytes):
+
+```go
+[4]byte  prefix = 0x9c23bd3b
+[32]byte manager_address    // manager32 (left-padded XRPL account)
+uint8    manager_mode = 1   // locking
```
evan-gray: where is the mode determined?

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,72 @@ uint32 nonce
 []byte payload
 ```
 
-### Initialize Transceiver
+<!-- cspell:ignore XREG vaas -->
 
-> 🚧 TODO
+### Registration (Initialize Transceiver & Peer Registration)
```
evan-gray, line 103: Might be worth linking to https://github.com/wormhole-foundation/native-token-transfers/blob/main/docs/Transceiver.md in this section

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,72 @@ uint32 nonce
...
 
-### Transceiver (Peer) Registration
+Because XRPL has no smart contracts, the `WormholeTransceiverInfo` (hub init) and
+`WormholeTransceiverRegistration` (peer) messages cannot be emitted natively.
+Instead, the Solana Sequencer emits an `XrplRegistration` (prefix `XREG`) message
```
evan-gray: I might have expected the Solana Sequencer to form the payload and the core bridge to allow for a generic "emit as transceiver" functionality that accepts a `token_id` and generic payload. Though I suppose this approach prevents malicious transfer messages from being emitted directly via the core bridge so perhaps it is more secure albeit more tedious (and would need to be extended across multiple codebases should the need for another message type arise).

node/pkg/watchers/xrpl/parse.go
```diff
@@ -82,6 +82,108 @@ const (
...
+var transceiverRegPrefix = [4]byte{0x18, 0xfc, 0x67, 0xc2}  // WormholeTransceiverRegistration
+
+// XREG constants
+const (
+	xregKindHub    = 0x00 // hub sub-type (byte after prefix), mirrors sequencer XREG_KIND_HUB
+	xregKindPeer   = 0x01 // peer sub-type (byte after prefix), mirrors sequencer XREG_KIND_PEER
+	xregTokenXRP   = 0x00 // wire token_id opening byte for XRP, mirrors XrplTokenId
+	xregTokenIOU   = 0x01 // wire token_id opening byte for issued currencies
+	xregTokenMPT   = 0x02 // wire token_id opening byte for MPT
+	nttModeLocking = 0x00 // manager mode for XRPL custody accounts (locking; accountant only accepts locking)
```
evan-gray: accountant only accepts locking hubs, which is true, but these init messages should reflect whether the manager is in locking or burning mode. all XRPL managers should not have messages attesting that they are locking when they may not be.

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,72 @@ uint32 nonce
...
+when a custody account is initialized (`initialize`) or a peer is registered
+(`register_peer`), and the executor relays it back to XRPL as a Payment to the
+Core Account. The watcher then re-keys and synthesizes the canonical transceiver
+message under the correct NTT transceiver emitter.
 
-> 🚧 TODO
+Upon a Payment to the Core Account, the guardian watcher MUST
```
evan-gray, line 115: this also must verify that the message came from the matching manager? though I'm confused about how an XREG message from the sequencer results in a payment from the manager without either manager changes in this PR or a generic payment message in the Sequencer PR. I would expect the latter - that the Sequencer issues a generic payment to the core bridge containing this registration. Maybe I'm missing where that occurs, but it looks like this is simply posted as a VAA payload. 😕

node/pkg/watchers/xrpl/parse.go
```diff
@@ -352,6 +461,191 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
...
+	}
+	return data, nil
+}
+
+// registrationSenderAccountID returns the sender's 20-byte XRPL account ID when
+// the sender is a managed (guardian-controlled) account, or nil when it is not
+// managed — used to authenticate an XREG registration publish (see the SECURITY
+// note in parseRegistrationTransaction). A missing/malformed Account field is an
+// error; an unmanaged sender is a silent decline (nil, nil).
+func (p *Parser) registrationSenderAccountID(tx transaction.FlatTransaction) ([]byte, error) {
```
evan-gray, line 647: we seem to have this code multiple times now - this is very similar to `extractSender`

node/pkg/watchers/xrpl/parse.go
```diff
@@ -352,6 +461,191 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
...
+	if len(data) < xregHeaderLen+1 {
+		return nil, emitter, fmt.Errorf("xreg payload too short: %d", len(data))
+	}
+	kind := data[4]
+	var manager20 [20]byte
+	copy(manager20[:], data[xregManagerOffset:xregHeaderLen])
+
+	// manager32 = left-padded 20-byte account id (same as transfer sourceNTTManager).
+	var manager32 [32]byte
+	copy(manager32[12:], manager20[:])
```
evan-gray: getting nervous about the multiple re-implementations in this file - there's logic for this in `addressToEmitter` but they don't quite match.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -352,6 +461,191 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
...
+
+	sourceToken, consumed, err := regSourceTokenFromWire(data[xregHeaderLen:])
+	if err != nil {
+		return nil, emitter, err
+	}
+	tail := data[xregHeaderLen+consumed:]
+
+	emitter = (&Parser{}).calculateEmitterAddress(manager32, sourceToken)
```
evan-gray: not sure `calculateEmitterAddress` needs to be method of parser, maybe switch it to a free function instead or match the other helpers and make `buildRegistrationMessage` a method of Parser.

evan-gray (changes requested):
I might be misunderstanding something, but between this and the sequencer PR, I'm not clear how the XRPL manager sends the payment to the XRPL core bridge account. I'm also concerned about not being able to accurately distinguish locking and burning configurations.

node/pkg/watchers/xrpl/parse.go
johnsaigle: Note that common operations like this are often covered and well-tested by functions in the Go SDK. (under `sdk/`)

node/pkg/watchers/xrpl/README.md
douglasgalico: Changed at sequencer level! XREG is wrapped in XREL which is signed by delegated manager and captured by watcher.

node/pkg/watchers/xrpl/parse.go
douglasgalico: Added logic to detect based on token_id (IOU and MPT) by checking if Issuer is the same as the Manager/CustodyAccount

node/pkg/watchers/xrpl/parse.go
mdulin2, line 445: Missing comment for 'parseRegistrationTransaction' on the flow of this. Since we do it for the other four, I'd recommend adding it here.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -344,6 +489,13 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
 		return msg, err
 	}
 
+	// Registration (XREG) publishes go to the core account but use a distinct
+	// MemoFormat, so they must be checked before the generic core parser.
+	msg, err = p.parseRegistrationTransaction(tx)
```
bemic: If the XREG is wrapped as XREL and send as standard payment, the transaction should produce XACK.
In that case this code won't be reached because `parseXACKTransaction` will consume the tx.

The test did not catch this because the `regTxStream` function builds the transaction without the `TicketSequence`. In such case `parseXACKTransaction` returns (nil, nil) because the field is empty and `parseRegistrationTransaction` takes place. However the manager code assigns `TicketSequence: uint32(payload.TicketID)`.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -352,6 +504,193 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
...
+	return &common.MessagePublication{
+		TxID:             txHash,
+		Timestamp:        tx.Timestamp,
+		Nonce:            0,
+		Sequence:         sequence,
+		EmitterChain:     vaa.ChainIDXRPL,
+		EmitterAddress:   emitter,
+		Payload:          payload,
+		ConsistencyLevel: 0,
+		IsReobservation:  false,
```
mdulin2, line 595: Is this actually true? I would imagine on a reobservation this wouldn't be the case.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -344,6 +489,13 @@ func (p *Parser) parseTransaction(tx GenericTx) (*common.MessagePublication, err
 		return msg, err
 	}
 
+	// Registration (XREG) publishes go to the core account but use a distinct
```
mdulin2: The `XREG` is being sent through the manager, which uses tickets. This currently doesn't handle the ticket management because it's not sending an ACK back to the sequencer. This means that the ticket would effectively be stuck there until a `BURN` was triggered for the ticket. Naturally, this would fail because the sequence was indeed used up. 

Additionally, what should we do in the case that the `XREG` fails? Should the hub or peer registration accountant message be rebroadcasted?

mdulin2 (changes requested):

node/pkg/watchers/xrpl/parse.go
```diff
@@ -82,6 +82,151 @@ const (
...
+
+// xregPrefix is the 4-byte prefix for XRPL registration payloads
+var xregPrefix = [4]byte{'X', 'R', 'E', 'G'}
+
+// Prefixes for canonical NTT transceiver messages (from the NTT spec / ntt-messages)
+var transceiverInfoPrefix = [4]byte{0x9c, 0x23, 0xbd, 0x3b} // WormholeTransceiverInfo
+var transceiverRegPrefix = [4]byte{0x18, 0xfc, 0x67, 0xc2}  // WormholeTransceiverRegistration
+
+// XREG constants
+const (
```
johnsaigle: suggestion: Could we move the XRPL discriminator constants into custom typed domains instead of leaving them as untyped constants?

Right now values like `xRegKind` and `xRegTokenXRP` are easy to mix in-place because the compiler does not know they represent different concepts. A future change could accidentally compare or pass a registration kind where a token kind is expected, and it would still compile if both are just plain constants.

It would look something like this:

```go
type xRegKind string
type xRegToken string

const (
    xRegKindRegister xRegKind = "..."
    xRegKindAck      xRegKind = "..."

    xRegTokenXRP xRegToken = "..."
)
```

Then helper signatures can encode intent:

```go
func buildXRegPayload(kind xRegKind, token xRegToken) ...
```

That makes invalid mixing harder to write and easier to catch during review/compile time.

I’d also suggest putting these XRPL domain types and constants in a dedicated file, e.g. `node/pkg/watchers/xrpl/types.go`, so the parser code stays focused on transaction handling and the XRPL wire/domain vocabulary has one obvious home.

node/pkg/watchers/xrpl/README.md
```diff
@@ -100,13 +100,90 @@ uint32 nonce
 []byte payload
 ```
 
-### Initialize Transceiver
+<!-- cspell:ignore XREG XREL vaas -->
```
johnsaigle: suggestion: Consider adding these to the dictionary at `cspell-custom-words.txt`

node/pkg/accountant/ntt_config.go
```diff
@@ -48,6 +48,9 @@ func nttGetEmitters(env common.Environment) (validEmitters, validEmitters, error
 			{chainId: vaa.ChainIDArbitrumSepolia, addr: "000000000000000000000000fA42603152E4f133F5F3DA610CDa91dF5821d8bc"},
 			{chainId: vaa.ChainIDBaseSepolia, addr: "000000000000000000000000149987472333cD48ac6D28293A338a1EEa6Be7EE"},
 			{chainId: vaa.ChainIDOptimismSepolia, addr: "000000000000000000000000eCF0496DE01e9Aa4ADB50ae56dB550f52003bdB7"},
+			// XRPL NTT custody emitters — one per (custody account, token), keyed on keccak256("ntt" + manager32 + sourceToken32). Log-only until the XREG registration pipeline is live.
```
johnsaigle, line 51: Are you tracking this somewhere? Flipping these bools seems like an easy thing to forget. Is the 'registration pipeline' happening in different PR?

node/pkg/watchers/xrpl/parse.go
douglasgalico: That comes also with the other message parsers we have, and it's based on the EVM approach: the parser always set to false and what watcher's reobserv path overrides this to true at `case req := <-w.obsvReqC`

node/pkg/watchers/xrpl/parse.go
douglasgalico: Thanks for spotting it! Indeed it's a miss, parser was adjusted to output a list of transactions (the registration message for accountant and ACK) - adjusted also at watcher code in all paths

node/pkg/watchers/xrpl/parse.go
douglasgalico: Fixed the ordering of checks in parser and ACK is handled properly. For the XREG failure there was also a gap, now after checking the transaction result - and when it fails - it will pass through parseXACKTransaction, with that we can consume XACK and clear the ticket on sequencer.

node/pkg/watchers/xrpl/parse.go
douglasgalico: Yep! Moved the XREG wire consts into a dedicated types.go file. Kept them untyped to match the package other wire discriminants (tokenType*, xackTxType*, ..) since they are raw bytes read from/appended to []byte

node/pkg/accountant/ntt_config.go
douglasgalico: Yes, the pipeline is done after the sequencer is upgraded to new code and executor watching and dispatching the msgs back to XRPL after XREG on sequencer. After that we'll remove logOnly for final test

node/pkg/watchers/xrpl/parse.go
```diff
@@ -321,35 +427,275 @@ func (p *Parser) parseNttTransaction(
...
+// the core account with regMemoFormat) and synthesizes the canonical NTT
+// transceiver registration message under the NTT transceiver emitter.
+//
+// Returns (nil, nil) if this is not a registration publish.
+func (p *Parser) parseRegistrationTransaction(tx GenericTx) (*common.MessagePublication, error) {
+	if p.coreAccount == "" {
+		return nil, nil
+	}
+
+	destination, err := p.extractDestination(tx.Transaction)
```
mdulin2, line 542: This returning an error is a problem with the current ordering again. There are many transaction types where the `destination` is NOT set. Both a failed `TicketCreate` and `AccountSet` (burn ticket) will hit this error, and return, skipping the `parseXACKTransaction` logic.

mdulin2: I think we need more tests to ensure things work with the current ordering realistic RPC responses, and realistic settings. This is the second change that would have broken existing functionality in the watcher within this PR. 

From what I can tell, the reason the existing tests didn't catch this was because the `coreAccount` was never set on the tests, hitting the exit path with no error above this. In reality, the existing tests (if properly configured) should have caught this regression.

node/pkg/watchers/xrpl/watcher.go
```diff
@@ -339,28 +339,26 @@ func (w *Watcher) processTransaction(tx *streamtypes.TransactionStream) error {
...
-	xrplMessagesConfirmed.Inc()
-
-	w.logger.Info("message observed", msg.ZapFields()...)
-
 	return nil
 }
 
 // fetchAndParseTransaction fetches a specific transaction by hash for reobservation.
-func (w *Watcher) fetchAndParseTransaction(txHash []byte) (*common.MessagePublication, error) {
+func (w *Watcher) fetchAndParseTransaction(txHash []byte) ([]*common.MessagePublication, error) {
```
johnsaigle: suggestion: update the function name and comment to clarify that this now returns multiple message publications

node/pkg/watchers/xrpl/parse.go
```diff
@@ -321,35 +427,276 @@ func (p *Parser) parseNttTransaction(
...
+//     and the registration would be lost.
+//   - Core/NTT come last because they require a Destination field that TicketCreate
+//     transactions do not have.
+//
+// A registration publish is the one case that yields TWO messages: the synthesized
+// registration (for the accountant) AND the XACK that clears the ticket the
+// registration consumed on the sequencer. Every other transaction yields at most
+// one message.
+//
+// Returns an empty slice if none matched.
```
johnsaigle:
```suggestion
// Returns nil if none matched.
```
The function only returns nil or non-empty slices, not an empty slice.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -925,11 +1281,12 @@ func (p *Parser) addressToEmitter(address string) (vaa.Address, error) {
 		return vaa.Address{}, fmt.Errorf("unexpected account ID length: got %d, want %d", len(accountID), addresscodec.AccountAddressLength)
 	}
 
-	// Left-pad with zeros to create 32-byte emitter address
-	// vaa.Address is [32]byte, accountID is 20 bytes
-	// Place accountID in the last 20 bytes (indices 12-31)
-	var emitter vaa.Address
-	copy(emitter[32-addresscodec.AccountAddressLength:], accountID)
+	// Right-align the 20-byte account id in a 32-byte vaa.Address (left-padded
+	// with 12 zero bytes). Uses the SDK helper rather than hand-rolled copying.
```
johnsaigle:
```suggestion
	// with 12 zero bytes).
```
This line can be removed -- I assume this was added during an LLM revision but we want to avoid comments like this getting into the codebase

node/pkg/watchers/xrpl/parse.go
```diff
@@ -1283,7 +1640,8 @@ func (p *Parser) scaleAmount(amount uint64, fromDecimals, toDecimals uint8) (uin
 
 // calculateEmitterAddress calculates the emitter address from source NTT manager and source token.
 // emitter = keccak256("ntt" + source_ntt_manager_address + source_token)
-func (p *Parser) calculateEmitterAddress(sourceNTTManager, sourceToken [32]byte) vaa.Address {
+// It is a pure function of its inputs (no Parser state), so it is a free function.
```
johnsaigle: This line can be removed -- I assume this was added during an LLM revision but we want to avoid comments like this getting into the codebase

node/pkg/watchers/xrpl/parse.go
```diff
@@ -321,35 +427,276 @@ func (p *Parser) parseNttTransaction(
...
+//   - Core/NTT come last because they require a Destination field that TicketCreate
+//     transactions do not have.
+//
+// A registration publish is the one case that yields TWO messages: the synthesized
+// registration (for the accountant) AND the XACK that clears the ticket the
+// registration consumed on the sequencer. Every other transaction yields at most
+// one message.
+//
+// Returns an empty slice if none matched.
+func (p *Parser) parseTransaction(tx GenericTx) ([]*common.MessagePublication, error) {
```
johnsaigle, line 464: Given how complicated this function is, I think it's worth threading a `zap.Logger` through to the `Parser` type. It could just use a pointer to the logger the watcher already has.

This would allow us to add logging alongside the detection of the transaction type. It would be helpful for testing and debugging to know if this function ever returns the wrong type, for example.

node/pkg/watchers/xrpl/parse.go
```diff
@@ -321,35 +434,283 @@ func (p *Parser) parseNttTransaction(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode sender address %s: %w", account, err)
+	}
+	return accountID, nil
+}
+
+// buildRegistrationMessage parses the XREG bytes and returns the canonical
+// transceiver payload plus the NTT transceiver emitter to publish under.
+//
+// XREG layout: prefix(4) + kind(1) + manager(20) + token_id(wire 1-41) + tail
```
johnsaigle, line 665: If you're expecting a fixed number of total bytes here (or two possible fixed-lengths depending on whether it's Hub or Peer), could you add checks to make sure that exactly that number of bytes is present in the data?

Typically we try to be strict with data parsing, including checks that there are no trailing bytes following an expected payload.

PR #4894, opened 2026-07-09
node/solana: base58-decode websocket account pubkey

peaktwilight:
The Pythnet account-subscription websocket path (processAccountSubscriptionData) built the account PublicKey with solana.PublicKeyFromBytes([]byte(value.Pubkey)), but value.Pubkey is a base58-encoded string (the pubkey field of the accountSubscribe JSON notification). Casting the string to []byte reinterprets the base58 text as raw 32-byte key material, producing an incorrect PublicKey.

This key becomes the observation TxID on the account-based path, so the effect is a wrong/garbled TxID in emitted observations and logs. It does not affect the VAA digest (derived from emitter, sequence and payload), so this is a correctness/observability fix, not a consensus or security change.

Fix: decode the base58 string with solana.PublicKeyFromBase58 and surface any decode error, matching how pubkey strings are parsed elsewhere in the watcher (config.go). Adds a table-driven test using real base58 pubkeys that fails against the pre-fix code.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -1154,7 +1154,14 @@ func (s *SolanaWatcher) processAccountSubscriptionData(_ context.Context, data [
 		return nil
 	}
 
-	acc := solana.PublicKeyFromBytes([]byte(value.Pubkey))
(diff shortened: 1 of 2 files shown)


johnsaigle:

PR #4895, opened 2026-07-09
node/evm: return per-element error in batch poller

peaktwilight:
When a batched eth_getBlockByNumber call succeeds at the transport level but an individual element fails, getBlocks and getBlockRange logged the per-element error but then returned the outer err, which is nil on that path, so a failed block fetch was reported as an empty success instead of an error.

Note: geth records per-element failures on batch[idx].Error; the previously-referenced results[idx].err field is never populated, so the error branch never fired. This fix reads batch[idx].Error at both sites and removes the dead field.

Worst case previously was downstream nil/len handling of the empty result leading to a watcher restart, so this is a correctness cleanup, not a security fix. Adds a table-driven test proving the per-element error now propagates.

### DIFF
--- a/node/pkg/watchers/evm/connectors/batch_poller.go
+++ b/node/pkg/watchers/evm/connectors/batch_poller.go
@@ -34,7 +34,6 @@ type (
 
 	BatchResult struct {
 		result BlockMarshaller
-		err    error
 	}
 )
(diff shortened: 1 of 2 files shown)


johnsaigle:

PR #4907, opened 2026-07-17
ci: Compile integration tests in CI

johnsaigle:
Adds a Makefile target that compiles the Go integration tests (gated by the `-integration` flag). 

These tests currently get run manually during development, and not as a part of CI. This is fine, but recent refactoring has introduced syntax errors that we would not have seen until the next time these tests were run.

This PR ensures that the integration tests continue to compile by doing a fast, cheap build step as part of the `node-tests` CI target. If syntax errors occur, CI will fail.

This PR also adds minimal fixes for the existing syntax errors.

### DIFF
--- a/node/pkg/watchers/xrpl/parse_integration_test.go
+++ b/node/pkg/watchers/xrpl/parse_integration_test.go
@@ -47,17 +47,13 @@ func TestValidateTransactionResult_Integration(t *testing.T) {
 	// Verify the transaction is validated
 	require.True(t, txResp.Validated, "transaction should be validated")
 
-	// Create a parser and test validateTransactionResult
-	parser := NewParser("", nil, nil)
-
 	// Create a GenericTx with the transaction result from the response
 	tx := GenericTx{
-		Transaction:           txResp.TxJSON,
 		MetaTransactionResult: txResp.Meta.TransactionResult,
 	}
 
 	// Test validateTransactionResult - should succeed for a successful transaction
-	err = parser.validateTransactionResult(tx)
+	err = validateTransactionResult(tx)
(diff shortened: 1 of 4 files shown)


bemic (approved):

node/pkg/watchers/xrpl/parse_integration_test.go
```diff
@@ -68,38 +64,6 @@ func TestValidateTransactionResult_Integration(t *testing.T) {
 	t.Logf("Validated: %t", txResp.Validated)
 }
 
-// TestValidateTransactionResult_Integration_FailedTransaction tests that
-// validateTransactionResult correctly rejects a failed transaction.
-func TestValidateTransactionResult_Integration_FailedTransaction(t *testing.T) {
```
mdulin2, line 73: Why are these tests being removed?

mdulin2 (changes requested):

node/pkg/watchers/xrpl/parse_integration_test.go
johnsaigle: I mentioned in this commit: https://github.com/wormhole-foundation/wormhole/pull/4907/changes/1bcfc758e82aaa50f1a9563d4635b79a66e9f9a4

Basically there are already tests in `parse_test.go` under this package that do the same, and they're done as normal unit tests so they don't need to be gated by the integration flag. 

Let me know if you think differently

mdulin2: If they're done as other tests already, then I think it's fine to remove.

mdulin2 (approved):

djb15 (approved):

PR #4921, opened 2026-07-29
node: Increase supported sui grpc checkpoint size

djb15:
Increasing the gRPC return data limit from 4MiB to 32MiB to lower the probability of messages being skipped.


node/pkg/suiclient/suigrpc.go
```diff
@@ -24,6 +24,8 @@ var suiGrpcNilResponses = promauto.NewCounter(
 		Help: "Total number of nil checkpoint responses received from the Sui gRPC subscription stream",
 	})
 
+const suiGrpcMaxReceiveMessageSize = 32 * 1024 * 1024
```
johnsaigle, line 29: Can you add a permalink to their source code or else a reference to the Sui docs? It'd be better than needing to dig around for the PR description to find a source for the constant and why we need it.

djb15: Good shout, done

mdulin2: https://github.com/MystenLabs/sui/blob/1981116f572033f684d7f014a107718c98616cb2/crates/sui-protocol-config/src/lib.rs#L2573

mdulin2: Should this be slightly bigger than 32 MB? If the checkpoint is returned at 32 MB, then wouldn't the TOTAL request (with whatever other data) be slightly larger than 32MB? I would assume that 33MB would be safe.

mdulin2 (dismissed):

node/pkg/suiclient/suigrpc.go
djb15: It's already bigger 😄 Sui caps to 30 MiB, I've set to 32 MiB

johnsaigle: Can you mention this in the comment also? Sorry to request another commit but it'd be good to document why the limit is a little bigger than the constant we're linking too. Just trying to save future us from a headache

johnsaigle (dismissed):

bemic (approved):

pleasew8t (approved):

PR #4925, opened 2026-07-30
fix: fix worm evm info for testnet

douglasgalico:


douglasgalico:
CI to be fixed after https://github.com/wormhole-foundation/wormhole/pull/4880 migrated (due to RPC deprecation)

PR #4927, opened 2026-07-31
node(governor): Add freshness check for token updates in "hack" script

johnsaigle:
The recent token updating PR https://github.com/wormhole-foundation/wormhole/pull/4923 got closed because it was essentially a no-op. This PR makes that failing case more clear and points to the upstream data source as the potential cause of the issue.

### DIFF
--- a/node/hack/governor/src/index.ts
+++ b/node/hack/governor/src/index.ts
@@ -111,14 +111,15 @@ var depeggedUSDStablecoins = [];
 
 {
   const doc = fs.readFileSync("../../pkg/governor/generated_mainnet_tokens.go", "utf8");
-  var matches = doc.matchAll(/{Chain: (?<chain>[0-9]+).+Addr: "(?<addr>[0-9a-fA-F]+)".*Symbol: "(?<symbol>.*)", CoinGeckoId: "(?<coinGeckoId>.*)", Decimals.*Price: (?<price>.*)}.*\n/g);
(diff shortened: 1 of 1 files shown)


node/hack/governor/src/index.ts
```diff
@@ -384,6 +404,13 @@ axios
       }
     }
 
+    if (newTokensCount === 0) {
+      throw Error("The TVL response produced no tokens; refusing to overwrite the generated token list.");
+    }
```
djb15, line 409: I don't think this is a good failure case, it is perfectly valid to only remove tokens during a token list update

djb15 (commented):
I don't fully understand the purpose of this change? I think you're trying to fail loudly if nothing changes, I'm just not sure why that's helpful?

For example if I run the script twice in quick succession, it shouldn't fail the second time imo. I know this script is already a mess but it would be nice not to add more external dependencies to it 😄

johnsaigle:
@djb15 It was mostly to help us figure out if there's an issue upstream, but I don't mind if we close it if you think this will make your UX worse

djb15:
@johnsaigle it would make my UX a bit worse so I'd rather close. But thanks for taking a look, did you run this locally to get any signal on upstream?

johnsaigle:
I did run it locally and it looks like the problem is upstream (stale data)

PR #4930, opened 2026-08-03
Remove reference to Moonbeam

djb15:
Remove references to Moonbeam as the chain has shut down as of 31st July


johnsaigle (approved):
Looks good and I don't see any problems as long as tests pass.

PR #4936, opened 2026-08-05
fix(node): compact broadcast path skips SentTimestamp freshness, leaving delegate signatures replayable

SashaMIT:
## Summary

- `processSignedDelegateObservation` (direct path) rejects a `DelegateObservation` whose `SentTimestamp` is older/further ahead than `delegateObservationMaxTimeDifference` (15 min).
- The compact `SignedDelegateSignaturesBroadcast` path reconstructs each inner observation in `expandBroadcastToSignedDelegateObservations`, verifies guardian-set membership and the ECDSA signature — but **never checks `SentTimestamp`** — and feeds the result straight into `delegateObsvRecvC`, bypassing the direct path's check.
- This PR applies the same 15-minute freshness bound per inner signature in the broadcast expander, before signature recovery.

## Impact if unsolved

Any inner delegate signature ever observed on the network remains replayable forever when repackaged into a compact broadcast: a stale attestation can be re-injected at any time, and the freshness control that exists precisely to bound this replay is silently absent on one of the two ingest paths. The two paths now enforce the same bound.

## Test plan

- New `TestExpandBroadcastDropsStaleSentTimestamp`: two validly-signed inner observations (one fresh, one 2× the bound old) — only the fresh one is emitted. **Fails without the fix** (mutation-verified).
- `go test ./pkg/p2p -run TestExpandBroadcastDropsStaleSentTimestamp` passes.
- Note: the full `pkg/p2p` suite panics at pristine HEAD in this environment (`crypto/tls bug: where's my session ticket?` from quic-go v0.48.1 vs local Go — pre-existing, unrelated; libp2p network tests).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/node/pkg/p2p/p2p.go
+++ b/node/pkg/p2p/p2p.go
@@ -1706,6 +1706,16 @@ func expandBroadcastToSignedDelegateObservations(broadcast *gossipv1.DelegateSig
 			}
 		}
 
+		// Enforce the same SentTimestamp freshness bound the direct delegate-observation
+		// path enforces (processSignedDelegateObservation); without it, an inner delegate
+		// signature stays replayable indefinitely when re-broadcast in compact form.
+		if time.Until(time.Unix(sig.SentTimestamp, 0)).Abs() > delegateObservationMaxTimeDifference {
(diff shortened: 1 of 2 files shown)


djb15:

PR #4937, opened 2026-08-05
fix(node): one malformed delegated-guardian governance config can kill every guardian's EVM watcher

SashaMIT:
## Summary

- `WormholeDelegatedGuardians` (immutable) accepts any config where `threshold == 0` XOR `keys empty` — it does **not** check `threshold <= keys.length`, the quorum floor, or duplicate keys.
- Every guardian node *does* reject those configs in `NewDelegatedGuardianChainConfig` — and the EVM watcher propagated that rejection as an error, which the polling routine forwards to the fatal channel (`The watcher will exit anyway`).
- Result: a single malformed (or compromised) governance config VAA exits the entire observation pipeline for that chain on **every** guardian until a corrective governance VAA lands — delegated attestation and normal VAA observation alike.
- Fix: `buildDelegatedGuardianConfig` (extracted pure helper) skips the offending chain — fail closed, so delegated attestation for that chain stays off until a valid config arrives — while valid chains in the same config keep working. Mirrors the existing "no keys" skip pattern. A contract-side validation mirror is still worthwhile for future deployments, but the node-side resilience fixes the live liveness DoS on the already-deployed immutable contract.

## Impact if unsolved

Governance (or a governance-key compromise) can publish one config with `threshold > keys.length` and halt observation for an entire chain across the guardian set — a network-wide liveness failure triggered by a single on-chain write, on a contract that cannot be patched.

## Test plan

- New `TestBuildDelegatedGuardianConfigSkipsInvalidChains`: valid chain survives; threshold-too-high, threshold-below-quorum, duplicate-keys, and no-keys chains are all skipped without error.
- `go test ./pkg/watchers/evm` passes; package builds clean.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/node/pkg/watchers/evm/delegated_guardian_config_test.go
+++ b/node/pkg/watchers/evm/delegated_guardian_config_test.go
@@ -1,0 +1,52 @@
+package evm
+
+import (
+	"testing"
+
+	"github.com/wormhole-foundation/wormhole/sdk/vaa"
+	dgAbi "github.com/certusone/wormhole/node/pkg/watchers/evm/connectors/delegated_guardians"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+)
+
+// The on-chain WormholeDelegatedGuardians contract accepts configs the node rejects.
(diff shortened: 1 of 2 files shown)


djb15:

PR #4940, opened 2026-08-07
feat: upgrade SUI token bridge on testnet

douglasgalico:


kcsongor (dismissed):

sui/testing/scripts/upgrade-token-bridge.ts
```diff
@@ -1,105 +1,143 @@
...
-} from "@mysten/sui.js";
+/**
+ * Testnet Sui Token Bridge upgrade via governance VAA — gRPC edition.
+ *
+ * Usage:
+ *   TESTNET_WALLET_PRIVATE_KEY="suiprivkey1..." \
+ *   TESTNET_GUARDIAN_PRIVATE_KEY="<hex>" \
+ *   npx tsx scripts/upgrade-token-bridge-grpc.ts
+ *
+ *   npx tsx scripts/upgrade-token-bridge-grpc.ts --build-only
```
djb15: Comments are referencing a non-existent file

sui/testing/scripts/upgrade-token-bridge.ts
```diff
@@ -1,105 +1,143 @@
...
+ * TESTNET_WALLET_PRIVATE_KEY accepts the bech32 `suiprivkey1...` export
+ * directly, or the legacy base64 (scheme-byte-prefixed) format.
+ */
+import { SuiGrpcClient } from "@mysten/sui/grpc";
+import { Transaction } from "@mysten/sui/transactions";
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import { decodeSuiPrivateKey } from "@mysten/sui/cryptography";
+import { SUI_CLOCK_OBJECT_ID, normalizeSuiObjectId } from "@mysten/sui/utils";
+import { secp256k1 } from "@noble/curves/secp256k1";
+import { keccak_256 } from "@noble/hashes/sha3";
```
djb15: This package isn't listed as an explicit direct dependency in the `package.json` and lock file?

sui/testing/scripts/upgrade-token-bridge.ts
```diff
@@ -109,156 +147,169 @@ function buildForBytecodeAndDigest(packagePath: string) {
...
+    modules: buildOutput.modules, // base64 strings, accepted by tx.upgrade
+    dependencies: buildOutput.dependencies.map((d) => normalizeSuiObjectId(d)),
     digest: Buffer.from(buildOutput.digest),
   };
 }
 
+/// UpgradeContract governance VAA: module "TokenBridge", action 2, chain 21,
+/// payload = 32-byte build digest. Signed by the single testnet guardian
+/// (guardian set index 0).
+function makeUpgradeVaa(digest: Buffer, guardianKeyHex: string): Buffer {
```
djb15, line 164: Mind adding some more inline comments to this function to explain the structure? Tbf the original file wasn't much better but it would be nice to verify what's going on by referring to a whitepaper or explaining in plain english how we're building the payload.

djb15 (commented):
Just some small comments, generally looks good, need to give the VAA building function another review. Could you confirm the following just for my understanding:
1. This updated script was used to upgrade the testnet token bridge to test out the new pausing changes?
2. The changes are required to use the new gRPC endpoints?
3. This script will also be used for the eventual mainnet upgrade?

douglasgalico:
> Just some small comments, generally looks good, need to give the VAA building function another review. Could you confirm the following just for my understanding:
> 
> 1. This updated script was used to upgrade the testnet token bridge to test out the new pausing changes?
> 2. The changes are required to use the new gRPC endpoints?
> 3. This script will also be used for the eventual mainnet upgrade?

1. Yes, used to create the new implementation, generate the upgrade and migration actions and calls.
2. Exactly, as the previous script was relying into the deprecated RPC
3. Not without a couple of modifications: The consts for Core contract and Token Bridge state ID (now pointing to testnet), gRPC of Mainnet and support for the upgrade VAA signed by guardians (instead of we signing the testnet one with a single one)

sui/testing/package.json
```diff
@@ -6,12 +6,16 @@
   "license": "MIT",
   "dependencies": {
     "@certusone/wormhole-sdk": "^0.9.12",
+    "@mysten/sui": "^2.23.2",
     "@mysten/sui.js": "^0.32.2",
+    "@noble/curves": "^1.0.0",
+    "@noble/hashes": "^1.3.0",
```
djb15: Could you set these `noble` versions to `2.x.x` please? Looks like `mysten/sui` uses `2.x.x` so makes sense to be at least aligned on major version imo

djb15 (commented):
Sorry last Q, it seems like we're re-spinning `makeUpgradeVaa` to build the VAA object from scratch here? Not necessarily a non starter, but what's the justification for doing that here vs using existing libraries? I'm going to guess it's because we trying not to vendor anything here and it's strictly for testing so better to be more self contained?

douglasgalico:
> Sorry last Q, it seems like we're re-spinning `makeUpgradeVaa` to build the VAA object from scratch here? Not necessarily a non starter, but what's the justification for doing that here vs using existing libraries? I'm going to guess it's because we trying not to vendor anything here and it's strictly for testing so better to be more self contained?

No hard blocker at all, the old certusone mock could have worked too... but the mock has no usable TokenBridge upgrade builde, so the old script was calling `publishWormholeUpgradeContract` and doing some byte surgery to overwrite the module, action, target chain. Since the file is being rewritten, explicitly building there seemed better thank keeping the old sdk plus byte patching. Happy to switch to either if the team prefers

sui/testing/package.json
douglasgalico: adjusted

djb15 (approved):

kcsongor (approved):

Issue #4473, opened 2025-08-18
Remove remaining references to Terra, Terra2, and other deprecated chains


johnsaigle:
https://github.com/wormhole-foundation/wormhole/pull/4446 did a lot of work to remove recently deprecated chains. However, there are still code paths and helper files lingering in the repo for deprecated chains, especially Terra and Terra 2.

The task here is to go through the codebase and remove all these references, especially any branches that execute conditionally on the state of deprecated chains. Completing this task will help to make the code easier to reason about.

Non-exhaustive list:
* `scripts/contract-upgrade-governance.sh`
* `Tiltfile`
* CSV files under `deployments`
* `src/token_bridge/__tests__/terra-integration.ts`
* devnet/node.yaml
* devnet/terra-devnet.yaml
* node/cmd/guardiand/node.go
* scripts/devnet-consts.json
* sdk/js/src/nft_bridge/__tests__/utils/consts.ts
* sdk/js/src/token_bridge/__tests__/utils/consts.ts

johnsaigle:
See also #4525 . The remaining Tilt references are interfering with other work.

johnsaigle:
If Terra2 is removed from Github CI, a thoughtful review must be done in order to ensure coverage for the CosmWasm and accountant code is not reduced.

See comment here: https://github.com/wormhole-foundation/wormhole/pull/4537#discussion_r2445013533

johnsaigle:
https://github.com/wormhole-foundation/wormhole/issues/4724

johnsaigle:
https://github.com/wormhole-foundation/wormhole/pull/4792 removes Terra

PR #4941, opened 2026-08-10
chore: bump wormhole sdk to latest version

M-Picco:
* Clean up deprecated chains' code
* Move cli related logic from `build` action to its own `build-cli` action triggered when modifying its files. This should prevent from running the ci on PRs unrelated to it


clients/js/src/cmds/status.ts
```diff
@@ -47,14 +52,25 @@ export const handler = async (
...
+      continue;
+    }
     targetChainProviders.set(
-      toChainName(chainToChainId(key as Chain)),
+      toChainName(chainId as Parameters<typeof toChainName>[0]),
       new ethers.providers.JsonRpcProvider(NETWORKS[network][key as Chain].rpc)
     );
   }
 
   // TODO: Convert this over to sdkv2
```
johnsaigle: Can this be removed?

clients/js/info_tests/info-1.expected
```diff
@@ -14,41 +14,49 @@
...
     "Injective": "0x00000000000000000000000045dbea4617971d93188eda21530bc6503d153313",
     "Osmosis": "0x0000000000000000000000000000000000000000000000000000000000000000",
     "Sui": "0xccceeb29348f71bdd22ffef43a2a19c1f5b5e17c5cca5411529120182672ade5",
     "Aptos": "0x0000000000000000000000000000000000000000000000000000000000000001",
     "Arbitrum": "0x0000000000000000000000000b2402144bb366a632d14b83f244d2e0e21bd39c",
     "Optimism": "0x0000000000000000000000001d68124e65fafc907325e3edbf8c4d84499daa8b",
-    "Gnosis": "0x0000000000000000000000000000000000000000000000000000000000000000",
     "Pythnet": "0x0000000000000000000000000000000000000000000000000000000000000000",
-    "Xpla": "0x8f9cf727175353b17a5f574270e370776123d90fd74956ae4277962b4fdee24c",
     "Btc": "0x0000000000000000000000000000000000000000000000000000000000000000",
```
johnsaigle, line 31: Should Btc get removed?

clients/js/info_tests/info-1.expected
```diff
@@ -14,41 +14,49 @@
...
-    "Oasis": "0x0000000000000000000000005848c791e09901b40a9ef749f2a6735b418d7564",
     "Algorand": "0x67e93fa6c8ac5c819990aa7340c0c16b508abb1178be9b30d024b8ac25193d45",
-    "Aurora": "0x00000000000000000000000051b5123a7b0f9b2ba265f9c4c8de7d78d52f510f",
-    "Fantom": "0x0000000000000000000000007c9fc5741288cdfdd83ceb07f3ea7e22618d79d2",
-    "Karura": "0x000000000000000000000000ae9d7fe007b3327aa64a32824aaac52c42a6e624",
-    "Acala": "0x000000000000000000000000ae9d7fe007b3327aa64a32824aaac52c42a6e624",
     "Klaytn": "0x0000000000000000000000005b08ac39eaed75c0439fc750d9fe7e1f9dd0193f",
     "Celo": "0x000000000000000000000000796dff6d74f3e27060b71255fe517bfb23c93eed",
     "Near": "0x148410499d3fcda4dcfd68a1ebfcdddda16ab28326448d4aae4d2f0465cdfcb7",
     "Moonbeam": "0x000000000000000000000000b1731c586ca89a23809861c6103f0b96b3f57d92",
```
johnsaigle, line 23: I believe Moonbeam has shut down

.github/workflows/build-cli.yml
```diff
@@ -0,0 +1,28 @@
+name: Build
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+    paths:
+      - ./clients/js
```
johnsaigle:
```suggestion
      - clients/js/**
```

I'm not sure this `paths` value is correct. I think we need to do it from the repo root. It's worth doing a proper red-green test here to confirm

I suggested a different path but I'd recommend checking the correct syntax here as I'm not 100% my suggestion is right.

clients/js/src/consts/networks.ts
```diff
@@ -91,267 +82,76 @@ const Mainnet = {
...
-  },
-  Btc: {
-    rpc: undefined,
-    key: undefined,
-  },
-  Wormchain: {
-    rpc: undefined,
-    key: undefined,
-  },
   Moonbeam: {
```
johnsaigle: Moonbeam is gone

clients/js/src/cli.test.ts
```diff
@@ -163,32 +205,18 @@ describe("Info Tests", () => {
...
-                "0x0000000000000000000000007c9fc5741288cdfdd83ceb07f3ea7e22618d79d2",
-              Karura:
-                "0x000000000000000000000000ae9d7fe007b3327aa64a32824aaac52c42a6e624",
-              Acala:
-                "0x000000000000000000000000ae9d7fe007b3327aa64a32824aaac52c42a6e624",
               Klaytn:
                 "0x0000000000000000000000005b08ac39eaed75c0439fc750d9fe7e1f9dd0193f",
               Celo: "0x000000000000000000000000796dff6d74f3e27060b71255fe517bfb23c93eed",
               Near: "0x148410499d3fcda4dcfd68a1ebfcdddda16ab28326448d4aae4d2f0465cdfcb7",
               Moonbeam:
```
johnsaigle: Can be deleted

clients/js/src/cli.test.ts
```diff
@@ -432,6 +447,202 @@ describe("Generate Tests", () => {
...
+        exec(
+          `node build/main.js parse ${vaa}`,
+          (error: any, stdout: string, stderr: any) => {
+            if (error) {
+              return done(new Error(`Execution error during parse: ${error}`));
+            }
+            try {
+              const outputObject = JSON.parse(stdout);
+
+              // Can't check the signature, sequence, or digest because they
```
johnsaigle: You could consider checking that they are non-empty and within the expected range for their types?

e.g.
* Signature is a fixed, expected size (and probably a hex value?)
* Sequence is in the bounds of u64
* Digest is also a fixed, expected size (and probably a hex value?)

johnsaigle (changes requested):
* Would you be able to explain more the decision to remove the IBC relayer tests? From my understanding this is still a live path on Wormhole via Gateway/Wormchain. So, it's something we should be running in CI. I know the _standard relayer_ was deprecated in favour of the Executor, but I'm not aware that the IBC relayer is also replaced or deprecated.

* Also, we should be careful with removing the Terra2 tests. I don't have a full understanding of the impact but I remember Bruce flagged this when I tried to remove Terra2 things before. IMO it's worth reviewing this Issue and the related PRs to get a sense of the impact here. https://github.com/wormhole-foundation/wormhole/issues/4473#issuecomment-3432730045.

* Can you regenerate the package-lock.json file? It is missing the integrity hashes that we need in order to safely use `npm ci` (see here: https://github.com/wormhole-foundation/wormhole/blob/main/CONTRIBUTING.md#working-with-node-dependencies). AFAICT this can happen if you re-generate the lockfile without first deleting `node_modules/` locally.

M-Picco:
@johnsaigle Thanks for the feedback.

> Would you be able to explain more the decision to remove the IBC relayer tests?

The rationale for the skip was the Terra2 deprecation. I thought about replacing Terra2 with a different Cosmos chain but I feared the PR scope would increase too much. Forgot to mention that on the PR description.

I hadn't considered the coverage angle, which makes sense. I'll lean towards covering the worm cli/wh sdk gap and leave the Terra2 removal for later.

> IMO it's worth reviewing this Issue and the related PRs to get a sense of the impact here. https://github.com/wormhole-foundation/wormhole/issues/4473#issuecomment-3432730045

Thanks for the referenced issue too, I was not a aware of it

> Can you regenerate the package-lock.json file?

Will do, thanks for the pointer

clients/js/info_tests/info-1.expected
M-Picco: The `registrations` part of the response seems to be generated dynamically [here](https://github.com/wormholelabs-xyz/wormhole/blob/c700836452eb6db087c0b308d4e93fbc16d1f2f7/clients/js/src/cmds/info/registrations.ts#L103) based on the chains on the SDK. Looks like Moonbeam and Btc are still defined [there](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/12489addd0b33ef6426956ff9d7975836f3c7cac/core/base/src/constants/chains.ts#L29)

clients/js/info_tests/info-1.expected
M-Picco: Same as above

M-Picco:
@johnsaigle I've reverted the Terra2 removal from the test harness and added the support for Terra2 at the worm cli now that it was deprecated from the wh sdk. I've also regenerated the `package-lock.json` according to the contributing guidelines.

Thanks for the feedback and let me know if there's anything else you think is missing.

clients/js/src/cmds/submit.ts
```diff
@@ -163,8 +164,6 @@ async function executeSubmit(
       contractAddress,
       rpc
     );
-  } else if (chain === "Terra" || chain === "Terra2") {
-    await execute_terra(parsedVaa.payload, buf, network, chain);
   } else if (chain === "Solana" || chain === "Pythnet") {
```
johnsaigle: Could you add Fogo here too?

clients/js/src/cmds/transfer.ts
```diff
@@ -93,16 +95,6 @@ export const handler = async (
...
-    await transferTerra(
-      srcChain,
-      dstChain,
-      dstAddr,
-      tokenAddr,
-      amount,
-      network,
-      rpc
-    );
   } else if (srcChain === "Solana" || srcChain === "Pythnet") {
```
johnsaigle: Could you add Fogo here too?

clients/js/src/utils.ts
```diff
@@ -52,10 +148,13 @@ export function getNetwork(network: string): Network {
   throw new Error(`Unknown network: ${network}`);
 }
 
-export function chainToChain(input: string): Chain {
+export function chainToCliChain(input: string): CliChain {
   if (input.length < 2) {
     throw new Error(`Invalid chain: ${input}`);
   }
   const chainStr = input[0].toUpperCase() + input.slice(1).toLowerCase();
```
johnsaigle: I think this is probably broken for chains with mixed-case names, e.g. `HyperEVM`.

johnsaigle (changes requested):

M-Picco:
@johnsaigle Addressed your comments from last review ([diff](https://github.com/wormhole-foundation/wormhole/compare/1c87b8f200949fbcec854620b9a1ac857b2e2d34...313aa94b376538fddce9bac0b810ab118f6d59ea)). I've also noticed that some commands where failing when deprecated chains were present on the on-chain state (e.g. TB registrations), so added support for them as well. They should not be allowed for normal commands though.

douglasgalico (approved):

djb15 (approved):
Thanks for keeping the Terra2 Tilt tests around for now

M-Picco:
Closing in favor of #4953

PR #4943, opened 2026-08-13
node: Tighten Sui transaction verifier parsing

djb15:
Avoid the use of `string.Contains` in this context where it can be overly permissive. Instead of re-deriving the asset type, we instead take it from the parsing/validation step.

### DIFF
--- a/node/pkg/txverifier/suitypes.go
+++ b/node/pkg/txverifier/suitypes.go
@@ -2,6 +2,7 @@ package txverifier
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/big"
 	"regexp"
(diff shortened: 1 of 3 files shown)


johnsaigle (dismissed):
Good improvement 👍🏻

I feel like these changes could be stronger if you introduced a new custom `Asset` type (alias for `string`) and an enum that contains only the values you used for `suiNativeAssetType` and `suiWrappedAssetType`. That seems the right solution here because we don't really want `string` here, we want exactly two values and any other case is an error.

That way, you'd have stronger guarantees on the functions. `parseSuiAssetType` could return `Asset, error` and decodeSuiAssetObject could take an `Asset` as its parameter.
Right now, those are only conventions and the comments are trying to warn future devs not to mess up.

In any case, the code looks correct so I'm still good if you wanna merge as-is rather than go with the approach I suggested.

djb15:
@johnsaigle appreciate the feedback, I like the suggestion. Particularly returning an error instead of a boolean, that's a lot nicer and allowed me to the make the test cases a bit stricter on the error checking. I guess ultimately the only additional protection we have is that string variables passed into `parseSuiAssetType` will be rejected by the compiler, but string literals and `suiAssetType` of arbitrary strings will still work. Anyhow, definitely worth doing!

johnsaigle (approved):

node/pkg/txverifier/suitypes.go
```diff
@@ -127,49 +136,55 @@ type suiAssetInfo struct {
 // form and a trailing space.
 var suiDynamicFieldTypeRegex = regexp.MustCompile(`^0x0*2::dynamic_field::Field<([^:]+)::token_registry::Key<([^>]+)>,\s*([^:]+)::([^<]+)<([^>]+)>>$`)
 
-// validateSuiAssetType validates the type information of a token-registry dynamic field object.
-// The following checks are performed:
+// Errors identifying which parseSuiAssetType check rejected an object type.
+var (
+	errNotTokenRegistryField = errors.New("object type is not a token-registry dynamic field")
+	errNotAssetType          = errors.New("asset value type is not a native or wrapped token-bridge asset")
```
pleasew8t, line 142: nit: `errUnsupportedAssetType`

pleasew8t (approved):

PR #4947, opened 2026-08-17
node: Update accountant audit logic

djb15:
Observation batches are now packed against the exact serialised transaction size rather than always batching up to a maximum batch size of 100, deferring messages that don't fit to a later batch. A message too large to ever fit is submitted in its own batch, with a new metric and associated log.

### DIFF
--- a/node/pkg/accountant/submit_obs.go
+++ b/node/pkg/accountant/submit_obs.go
@@ -2,6 +2,7 @@ package accountant
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -22,7 +23,8 @@ import (
 )
 
 const (
-	DefaultSubmitObservationBatchSize = 100             // Observations per batch (limited by wasm contract input size of 64KB)
+	DefaultSubmitObservationBatchSize = 100             // Maximum observations per batch, also subject to maxSubmitObservationsMsgSize
+	maxSubmitObservationsMsgSize      = 64 * 1024       // Maximum size of the marshaled submit_observations message (the wasm contract input is limited to 64KB)
 	batchTimeout                      = 2 * time.Second // Time to collect observations before submitting
 )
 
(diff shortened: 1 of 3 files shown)


djb15:
@claude

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
 	}
 
-	acct.submitObservationsToContract(msgs, gs.Index, uint32(guardianIndex), wormchainConn, contract, prefix, tag) // #nosec G115 -- This is checked above
+	batches, oversized := packObservationBatches(msgs, acct.submitObservationBatchSize, maxSubmitObservationsMsgSize)
+	if len(batches) > 1 {
+		acct.logger.Info(fmt.Sprintf("split observations for %s into multiple batches to stay within the transaction size limit", tag), zap.Int("numMsgs", len(msgs)), zap.Int("numBatches", len(batches)))
+		batchSizeSplits.Inc()
+	}
+	for _, msg := range oversized {
+		// The message is submitted anyway (alone, so it can't take any other observations down with it) in case our size accounting is too conservative.
```
johnsaigle, line 106: Should we add a log in the case where an oversized message does _not_ fail? This would help give us a signal to change the limit.

johnsaigle: Also do they still get submitted? The loop below seems to only submit messages under `batches`, not `oversized`

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
 }
 
+// packObservationBatches partitions the messages into batches of at most maxBatchCount messages whose marshaled
+// submit_observations messages each stay within maxMsgSize. Each message is placed in the first batch with enough room for it,
+// so a large message that does not fit in the current batch is deferred to a later batch rather than failing the messages
+// around it. A message too large to fit even in a batch by itself is returned in oversized as well as being placed in its own
+// batch.
+func packObservationBatches(msgs []*common.MessagePublication, maxBatchCount int, maxMsgSize int) (batches [][]*common.MessagePublication, oversized []*common.MessagePublication) {
+	// batchSizes[i] is the size the marshaled observations array for batches[i] would have, including the enclosing brackets.
+	var batchSizes []int
```
johnsaigle: You could consider pre-allocating here to avoid allocations in the loop. Maybe the capacity could be set to `len(msgs) / maxBatchCount`?

node/pkg/accountant/submit_obs.go
```diff
@@ -22,7 +23,8 @@ import (
 )
 
 const (
-	DefaultSubmitObservationBatchSize = 100             // Observations per batch (limited by wasm contract input size of 64KB)
+	DefaultSubmitObservationBatchSize = 100             // Maximum observations per batch, also subject to maxSubmitObservationsMsgSize
```
johnsaigle, line 26: Could you change this to be `uint8` (rather than the default `int`)?

node/pkg/accountant/submit_obs.go
```diff
@@ -22,7 +23,8 @@ import (
...
+	maxSubmitObservationsMsgSize      = 64 * 1024       // Maximum size of the marshaled submit_observations message (the wasm contract input is limited to 64KB)
```
johnsaigle, line 27: Could you change this to be e.g. `uint16`?

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
 	transfersSubmitted.Add(float64(len(msgs)))
 	return nil
 }
 
+// packObservationBatches partitions the messages into batches of at most maxBatchCount messages whose marshaled
+// submit_observations messages each stay within maxMsgSize. Each message is placed in the first batch with enough room for it,
+// so a large message that does not fit in the current batch is deferred to a later batch rather than failing the messages
+// around it. A message too large to fit even in a batch by itself is returned in oversized as well as being placed in its own
+// batch.
+func packObservationBatches(msgs []*common.MessagePublication, maxBatchCount int, maxMsgSize int) (batches [][]*common.MessagePublication, oversized []*common.MessagePublication) {
```
johnsaigle, line 123:
```suggestion
func packObservationBatches(msgs []*common.MessagePublication, maxBatchCount uint8, maxMsgSize uint16) (batches [][]*common.MessagePublication, oversized []*common.MessagePublication) {
```

We can use stricter types here to make it impossible to accidentally call these with negative values, or values that would be too big to make sense.

node/pkg/accountant/submit_obs.go
johnsaigle, line 125:
```suggestion
	var batchSizes []uint16
```

node/pkg/accountant/submit_obs.go
johnsaigle, line 123: Could you add checks to explicitly handle the case where `msgs` is empty? I think we could also return early and log if either of the `max` arguments are 0. Right now these cases will either return empty values in a silent way, or else could result in all messages getting put into `oversized`.

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
+			}
+			batches = append(batches, []*common.MessagePublication{msg})
+			batchSizes = append(batchSizes, obsSize+jsonBracketsSize)
+		}
+	}
+	return batches, oversized
+}
+
+// marshaledObservationSize returns the number of bytes the observation for the message occupies in the marshaled observations array.
+func marshaledObservationSize(msg *common.MessagePublication) int {
```
johnsaigle, line 150: Might be better to use `uint16` as the return type here

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
+// Sizes of the JSON punctuation accounted for when computing how large a marshaled submit_observations message will be.
+const (
+	jsonQuotesSize   = len(`""`) // The quotes around a JSON string, such as the base64-encoded observations
+	jsonBracketsSize = len(`[]`) // The brackets around a JSON array, such as the marshaled observations
+	jsonCommaSize    = len(`,`)  // The comma between JSON array elements
+)
+
+// submitObservationsMsgSize returns the size of the marshaled submit_observations message for an observations array of
+// obsArraySize bytes, assuming worst-case sizes for the other fields.
+func submitObservationsMsgSize(obsArraySize int) int {
```
johnsaigle, line 168: Might be better to use uint16 as the return type

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
+	// The marshaled observations array appears in the message as a quoted base64 string.
+	return submitObservationsMsgOverhead + jsonQuotesSize + base64.StdEncoding.EncodedLen(obsArraySize)
+}
+
+// submitObservationsMsgOverhead is the number of bytes in a marshaled submit_observations message excluding the base64-encoded
+// observations string, computed with worst-case values for the other fields.
+var submitObservationsMsgOverhead = func() int {
```
johnsaigle, line 175: Might be better to use uint16 as the return type

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
+	}
+	bytes, err := json.Marshal(SubmitObservationsMsg{
+		Params: SubmitObservationsParams{
+			Observations:     []byte{},
+			GuardianSetIndex: math.MaxUint32,
+			Signature:        SignatureType{Index: math.MaxUint32, Signature: sig},
+		},
+	})
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal submit observations message: %v", err))
```
johnsaigle, line 188: Worth double-checking that the caller here is ultimately caught by a `recover` in the supervisor

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
+)
+
+// submitObservationsMsgSize returns the size of the marshaled submit_observations message for an observations array of
+// obsArraySize bytes, assuming worst-case sizes for the other fields.
+func submitObservationsMsgSize(obsArraySize int) int {
+	// The marshaled observations array appears in the message as a quoted base64 string.
+	return submitObservationsMsgOverhead + jsonQuotesSize + base64.StdEncoding.EncodedLen(obsArraySize)
+}
+
+// submitObservationsMsgOverhead is the number of bytes in a marshaled submit_observations message excluding the base64-encoded
```
johnsaigle, line 173: Is this value a constant ultimately? Maybe we could pre-compute the value and use unit tests to lock it in.

node/pkg/accountant/submit_obs.go
```diff
@@ -188,6 +278,20 @@ type (
 var SubmitObservationPrefix = []byte("acct_sub_obsfig_000000000000000000|")
 var NttSubmitObservationPrefix = []byte("ntt_acct_sub_obsfig_00000000000000|")
 
+// makeObservation converts a message publication into the observation that is submitted to the smart contract.
+func makeObservation(msg *common.MessagePublication) Observation {
+	return Observation{
+		TxHash:           msg.TxID,
+		Timestamp:        uint32(msg.Timestamp.Unix()), // #nosec G115 -- This conversion is safe until year 2106
```
johnsaigle, line 285: Would be nice to use the SDK's `TimeFromUnix` here instead of fighting the linter.

node/pkg/accountant/submit_obs_test.go
```diff
@@ -65,3 +69,196 @@ func TestParseObservationResponseData(t *testing.T) {
 	assert.Equal(t, expectedResult0, responses[0])
 	assert.Equal(t, expectedResult1, responses[1])
 }
+
+func makeMsgForPackingTest(t *testing.T, sequence uint64, payloadLen int) *common.MessagePublication {
+	t.Helper()
+	emitterAddr, err := vaa.StringToAddress("0x0290fb167208af455bb137780163b7b7a9a10c16")
+	require.NoError(t, err)
+	return &common.MessagePublication{
```
johnsaigle, line 77: Just for completeness it would be nice to add all of the fields here, i.e. IsReobservation, Unreliable, VerificationState

node/pkg/accountant/submit_obs_test.go
```diff
@@ -65,3 +69,196 @@ func TestParseObservationResponseData(t *testing.T) {
...
+	indices := make([]int, 0, end-start)
+	for i := start; i < end; i++ {
+		indices = append(indices, i)
+	}
+	return indices
+}
+
+// marshaledMsgSizeForBatch builds the submit_observations message for a batch the same way SubmitObservationsToContract does,
+// with worst-case values for the fields other than the observations, and returns its marshaled size.
+func marshaledMsgSizeForBatch(t *testing.T, batch []*common.MessagePublication) int {
```
johnsaigle, line 108: It might be better to not use `int` here as the return type

node/pkg/accountant/submit_obs_test.go
```diff
@@ -65,3 +69,196 @@ func TestParseObservationResponseData(t *testing.T) {
...
+func indexRange(start int, end int) []int {
+	indices := make([]int, 0, end-start)
+	for i := start; i < end; i++ {
+		indices = append(indices, i)
+	}
+	return indices
+}
+
+// marshaledMsgSizeForBatch builds the submit_observations message for a batch the same way SubmitObservationsToContract does,
+// with worst-case values for the fields other than the observations, and returns its marshaled size.
```
johnsaigle, line 107: Is this a constant value we could save instead of using a function?

node/pkg/accountant/submit_obs.go
```diff
@@ -95,11 +97,99 @@ func (acct *Accountant) handleBatch(ctx context.Context, subChan chan *common.Me
...
 
+// packObservationBatches partitions the messages into batches of at most maxBatchCount messages whose marshaled
+// submit_observations messages each stay within maxMsgSize. Each message is placed in the first batch with enough room for it,
+// so a large message that does not fit in the current batch is deferred to a later batch rather than failing the messages
+// around it. A message too large to fit even in a batch by itself is returned in oversized as well as being placed in its own
+// batch.
+func packObservationBatches(msgs []*common.MessagePublication, maxBatchCount int, maxMsgSize int) (batches [][]*common.MessagePublication, oversized []*common.MessagePublication) {
+	// batchSizes[i] is the size the marshaled observations array for batches[i] would have, including the enclosing brackets.
+	var batchSizes []int
+	for _, msg := range msgs {
```
johnsaigle, line 126: Let's add a nil check here

node/pkg/accountant/submit_obs.go
djb15: We don't use the non default `int` type for constants in most other places in the Guardian, why the ask to change that?

node/pkg/accountant/submit_obs_test.go
djb15: Vs a smaller type it's practically impossible for us to overflow an `int`. And based on the logic in the function it will never be negative so I'm not sure why we should go more defensive here?

node/pkg/accountant/submit_obs.go
djb15: I need to get those CodeQL rules going!!

PR #4950, opened 2026-08-22
node: use loaded addresses for Solana ALT resolution #4649

### LINKED ISSUE #4649: Solana Account Lookup Table (ALT) Message Dropping
## Description
The `v0` version of Solana transactions added [address lookup tables](https://solana.com/developers/guides/advanced/lookup-tables#how-to-create-an-address-lookup-table). These are accounts that store accounts that can then be referenced as a single index in the instruction instead of all the bytes of a public key.

ALTs can be *deleted*. There are [protections](https://docs.anza.xyz/proposals/versioned-transactions#security-concerns) in place to mitigate the impact on on-chain usage, such as preventing reinitialization.

In the Wormhole watcher, to fill out the accounts correctly, there's a call to `getAccountInfo()` that will get all of the addresses associated with an ALT. Then it fixes the `AccountKeys()` structure to ensure processing proceeds as expected.

If an account was stored within a deleted ALT that is being observed by the Guardian, it cannot be consumed. For many accounts, this is fine because a reobservation on that particular account could see the TX. For the `shim`, which doesn't look at the account but the full transaction information, this would prevent the message from being consumed. 

As a result, this is largely a self-denial of service. Not a must-fix, but something to consider in the future.

ayushmishra2005:
Fixes #4649.

## Summary

For versioned Solana transactions, use `TransactionMeta.LoadedAddresses` to populate lookup-table account keys when the metadata contains the complete set of loaded addresses.

This avoids fetching historical ALT accounts when the resolved addresses are already available in the transaction metadata, so transactions can still be processed after an ALT has been closed.

If `LoadedAddresses` is missing or incomplete, the watcher falls back to the existing ALT RPC lookup path.

## Changes

* Use loaded writable and readonly addresses from transaction metadata when complete.
* Preserve Solana account ordering: static keys, writable loaded addresses, then readonly loaded addresses.
* Keep the existing ALT lookup path as a fallback.
* Add coverage for deleted ALTs, missing/incomplete metadata, and loaded-address ordering.

### DIFF
--- a/node/pkg/watchers/solana/client.go
+++ b/node/pkg/watchers/solana/client.go
@@ -761,7 +761,7 @@ func (s *SolanaWatcher) processTransaction(ctx context.Context, rpcClient *rpc.C
 	}
 
 	signature := tx.Signatures[0]
-	err := s.populateLookupTableAccounts(ctx, rpcClient, tx)
+	err := s.populateLookupTableAccounts(ctx, rpcClient, tx, meta)
 	if err != nil {
 		s.logger.Error("failed to fetch lookup table accounts for transaction",
 			zap.Uint64("slot", slot),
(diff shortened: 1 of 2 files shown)


djb15:

ayushmishra2005:
Thanks @djb15 for pointing me to the contribution policy. I understand and won’t open further PRs without coordinating first. I’m actively interested in contributing to Wormhole, particularly around Solana protocol/watcher work. If there’s an appropriate issue or area where an external contributor would be useful, I’d appreciate any guidance.

djb15:
Thanks for understanding. Glad to hear you're interested in contributing. Generally we aim to avoid changing Watcher related components where possible since they're high risk areas of the codebase. One area we can improve in that's low risk is test coverage. If you're interested in Solana specifically you could consider adding test coverage there. Specifically I would suggest looking at basing any changes off of https://github.com/wormhole-foundation/wormhole/pull/4924 which makes some small refactors to the Solana watcher to aid in writing effective tests.

ayushmishra2005:
Thanks, that’s helpful.

PR #4955, opened 2026-08-25
feat(cli): add TokenBridge SetPauserAddresses and Sui ContractUpgrade…

douglasgalico:
… support


.github/workflows/build.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         with:
           node-version: "22"
       - name: Install Foundry
-        uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1.7.0
+        uses: foundry-rs/foundry-toolchain@908c540300062bd5a7e473851cdb4282204cee09 # v1.9.1
```
djb15, line 84: Ty for fixing the builds! Verified the hash matches the claimed version

clients/js/src/chains/sui/submit.ts
```diff
@@ -222,6 +283,20 @@ export const submit = async (
...
+            client,
+            network,
+            vaa,
+            coreBridgeStateObjectId,
+            tokenBridgeStateObjectId
+          );
+          setMaxGasBudgetDevnet(network, tx);
+          const res = await executeTransactionBlock(signer, tx);
+          console.log(JSON.stringify(res));
+          break;
```
djb15, line 298: Do we need to assert whether or not the transaction actually succeeded here?

clients/js/src/chains/sui/utils.ts
```diff
@@ -534,6 +537,279 @@ export const registerChain = async (
...
+      `token bridge package not found above ${__dirname}; set SUI_PACKAGE_DIR`
+    );
+  }
+
+  const buildOutput: {
+    modules: string[];
+    dependencies: string[];
+    digest: number[];
+  } = JSON.parse(
+    execSync(
```
djb15: Prefer `execFileSync` here so that we're not passing unsanitized variables directly to the shell

djb15 (commented):
I'm not that familiar with the js client so this was a pretty light review

clients/js/src/chains/sui/submit.ts
douglasgalico: That's parsed and checked at the executoTransactionBlock (and parsing the tx output). So the full result (wether succeed or not) is sent to output - As that's a final action, it's fine not asserting as that will (and should) be validated by who's operating

clients/js/src/chains/sui/utils.ts
douglasgalico: 👍

djb15 (approved):

evan-gray (approved):
checked the encoding appears to match https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0003_token_bridge.md

PR #4961, opened 2026-09-02
node: support solana transaction v1

evan-gray:


node/pkg/watchers/solana/shim.go
djb15: This comment (copied in other places too) needs updating now. The typecasting is still safe as v1 caps the number of instructions to 64

node/pkg/watchers/solana/client.go
```diff
@@ -971,7 +971,19 @@ func (s *SolanaWatcher) processTransaction(ctx context.Context, rpcClient *rpc.C
 	return
 }
 
-func (s *SolanaWatcher) processInstruction(ctx context.Context, rpcClient *rpc.Client, slot uint64, inst solana.CompiledInstruction, programIndex uint16, tx *solana.Transaction, signature solana.Signature, idx int, isReobservation bool) (bool, error) {
+// toRPCCompiledInstruction converts a top-level solana.CompiledInstruction
+// (from tx.Message.Instructions) into an rpc.CompiledInstruction, the type
+// used for inner instructions (meta.InnerInstructions[i].Instructions), so
+// both can be passed through the same instruction-processing functions.
+func toRPCCompiledInstruction(inst solana.CompiledInstruction) rpc.CompiledInstruction {
+	return rpc.CompiledInstruction{
```
djb15, line 979: Worth adding a comment here stating that we always set the stack height to 0 (the default) and that it shouldn't be used in any checks downstream?

node/pkg/watchers/solana/client.go
djb15: `PublicKeyFromBytes` got stricter on the library version bump, and now fails for any inputs that aren't 32 bytes. I think the input is still base58 encoded, so we should use `PublicKeyFromBase58` instead?

node/hack/repair_solana/repair.go
djb15: Want to bump it here too?

djb15: Also missing a bump at `node/hack/solana/account_lookup.go`

node/pkg/watchers/solana/client.go
bengtlofgren: Should we just add it explicitly in the fields as well?

djb15 (approved):

bemic (approved):

evan-gray:
[646MzxVgeVUasyjZfQ9GLZ6D1wuGGUgibeEXtb8HfQoDNdcdurnhT9zUg33df3kpcG6hfF8zCygUPJDteJbMaUpt](https://wormholescan.io/#/tx/646MzxVgeVUasyjZfQ9GLZ6D1wuGGUgibeEXtb8HfQoDNdcdurnhT9zUg33df3kpcG6hfF8zCygUPJDteJbMaUpt?network=Testnet)'s VAA was made while running commit 8834933b on the testnet guardian

